PR #8, opened 2020-06-02
Added "You're the one that I want" to list of songs

michaelclapham:
It's a real hit! It's going to help with parenting so much!


michaelclapham:
Wait but is it? This seems not like content we want to add 😢 
Also it's copyrighted music we can't add that! 😱 
Sorry pull request DENIED

PR #10, opened 2020-06-03
Added "You're the One that I Want" song

michaelclapham:
It's a great song! The app is lacking 80's hits!


src/app/feature/songs/song-lyrics/one-i-want.ts
```diff
@@ -0,0 +1,17 @@
+import { Song } from '../song.model';
+
+export const oneIWant: Song = {
+    songTitle: "You're the One That I Want",
+    audioURL: "assets/audio/one_I_want.mp3",
+    lyrics: [
+        {
```
michaelclapham, line 7: There's only one lyric here? This is nowhere near complete!

michaelclapham:
I'm pretty sure this song is copyrighted, so we can't put it in the app. Also it would be a bit of a waste of space, we have too many songs already 😢 
Closing this PR.

PR #23, opened 2020-06-08
Revert "Create the link component"

maxwellfundi:
Reverts IDEMSInternational/parenting-app-ui#18

PR #24, opened 2020-06-08
Create the think component

praise-ebe:


maxwellfundi:
@praise-ebe  I suggest you create a new component from a new branch. the changes from the old one can not be present because of the revert i did on #23

PR #40, opened 2020-06-09
Updating

barbra2020:


src/app/feature/take-a-pause/breathing-anim/breathing-anim.component.scss
```diff
@@ -16,7 +16,7 @@
 .mouth {
     z-index: 3;
     top: 80px;
-    left: 50px;
+    left: 50p
```
michaelclapham, line 19: 50p is not correct. p is not a valid unit. px or pt or em are. https://www.w3schools.com/cssref/css_units.asp

PR #71, opened 2020-10-05
Feature/offline chat service2

michaelclapham:
- Moved chat service to shared folder as it likely to be used across multiple screens

- Created Online and Offline Chat services that extend Chat Service abstract class so that switching between the two only requires a one line change in the component.

- Moving from parent to subflow and back is working but buggy.


src/app/shared/services/chat/chat-msg.model.ts
```diff
@@ -0,0 +1,26 @@
+export interface IRapidProMessage {
+    message: string;
+    message_id: string;
+    title: string;
+    type: string;
+    wasTapped: boolean; // informs whether message received in app foreground or background
+    quick_replies?: string; // string with JSON array. e.g "["English","Malay"]"
```
chrismclarke, line 7: nice to know that is the format. If you find this happening in lots of places or want to make it more explicit you could always define an alternate string type type, e.g. `type stringifiedJson = string` and then just say `quick_replies?: stringifiedJson`
But equally the comment achieves that

src/app/shared/services/chat/chat.service.ts
```diff
@@ -0,0 +1,10 @@
+import { BehaviorSubject, Observable } from 'rxjs';
+import { ChatMessage } from './chat-msg.model';
+
+export abstract class ChatService {
```
chrismclarke, line 4: Abstract classes! getting fancy... (all I have to add on that, makes sense)

src/app/shared/services/chat/offline/chat.flow.ts
```diff
@@ -0,0 +1,207 @@
+import { Observable, of, BehaviorSubject } from 'rxjs';
+import { ChatMessage, ChatResponseOption } from '../chat-msg.model';
+import { FlowStatusChange } from './offline-chat.service';
+import { RapidProFlowExport } from './rapid-pro-export.model';
+
+export interface ChatFlow {
```
chrismclarke, line 6: Exporting an interface from the class could get messy if used elsewhere (pretty easy to end up with circular dependencies), so might be better to have in the other typings/models files

src/app/shared/services/chat/offline/chat.flow.ts
```diff
@@ -0,0 +1,207 @@
...
+                }
+            }
+        });
+    }
+
+    private enterNode(node: RapidProFlowExport.Node) {
+        this.currentNode = node;
+        console.log("Entered node ", node.uuid, node);
+        for (let action of node.actions) {
+            if (action.type === "enter_flow") {
```
chrismclarke, line 41: This might be better as a switch statement so that we can include a default case to at least log/report if there are any cases not handled. Also would reduce the number of if statements which is a bit confusing (but I'm assuming required given a lot of possible cases... although you could sacrifice a bit of efficiency for readability and just calculate things like latestFlowEvent at a higher level regardless of action type). Similarly you could add getter functions for things like latestFlow that could just return if exists, e.g.  `get latestFlow(){return flowEvents.length > 0 ? flowEvents[flowEvents.length-1] : null`  and call directly `if(latestFlow){...}

src/app/shared/services/chat/offline/chat.flow.ts
```diff
@@ -0,0 +1,207 @@
...
+            }
+        }
+        if (matchingCategoryId) {
+            this.exitUsingCategoryId(node, matchingCategoryId);
+        } else {
+            console.warn("Nothing matches :(");
+        }
+    }
+
+    private useRouter(node: RapidProFlowExport.Node) {
```
chrismclarke, line 113: Are routers a rapidpro thing, or like angular router? Maybe a brief comment might be useful to explain what the different userRouter cases do

src/app/shared/services/chat/offline/chat.flow.ts
```diff
@@ -0,0 +1,207 @@
...
+        for (let arg of routerCase.arguments) {
+            matchStrings = matchStrings.concat(arg.toLowerCase().split(" "));
+        }
+        for (let matchString of matchStrings) {
+            if (incomingMsg.toLowerCase() === matchString) {
+                return true;
+            }
+        }
+    }
+
```
chrismclarke, line 173: Nice to have all these actions and matchers separated out, I was going to suggest whether it might be useful to include any 3rd party libs to help with the various manipulations (e.g. lodash/underscore/voca etc.), but looking at it I think the cases are probably pretty unique so maybe easiest to just handle manually.

If you want to make any of these oneliners regex is your friend here, but probably less easy on the eyes, e.g. to pass a case-insensitive flag and use a variable that would be `return new RegExp(incomingMsg,'i').test(matchstring));`

But honestly, if I really wanted one-liners I'd probably keep just do `if (incomingMsg.toLowerCase() === matchString) return true` and anger the lint fairies, so probably not the best use case for regex... but if ever useful elsewhere.

Another micro optimisation, is that you convert incomingMsg to lower case every time the loop runs, you could move it out to just convert once and save some precious zepto seconds there

src/app/shared/services/chat/offline/offline-chat.service.ts
```diff
@@ -0,0 +1,99 @@
...
+  private contactFields: { [field: string]: string } = {};
+  private currentFlow: RapidProOfflineFlow;
+
+  public messages$: BehaviorSubject<ChatMessage[]>;
+  private flowStatus$: BehaviorSubject<FlowStatusChange[]>;
+
+  constructor(
+    protected http: HttpClient,
+  ) {
+    this.messages$ = new BehaviorSubject([]);
```
chrismclarke, line 31: personal preference, I often just instantiate in the definition `private messages$ = new BehaviorSubject<ChatMessage[]>([]), but know lots of people who consider that sacrilege and instantiate everything within the constructor. So just to mention in case you weren't aware

chrismclarke (approved):
There's probably a bit too much going on for me to really understand how it all works without spending more time, but from what I can see looks solid enough (particularly the rxjs bits which is newer), see the one comment about memory leaks, and then anything else is mostly just random musings.

I couldn't really see why things are firing twice to start (suspect it's just something minor), but can take a deeper look another time if still causing issue later

michaelclapham:
@chrismclarke I'm still working on the offline flow processing, and I will address the majority of your comments, however I'm merging this down to master as I've merged lots of changes to the Rapid Pro Flow export definition into this branch, and changes to allow unit tests for the scripts which I'll be working on as a part of other pull requests.

I've switched the chat screen to use the new Online chat service (which just wraps the old one in a nice abstract class) so user functionality should remain the same.

chrismclarke:
All sounds good to me, I've noticed the releases coming through so nice that system is working as intended. I'm still behind on my planned dev so will update as and when I've got anything to report

PR #81, opened 2020-10-30
Reminders screen

idems-bot1:
List of reminders
Dragging around to pick when the reminder is for.
Clicking reminder to say that it is done with green tick animation
Create reminder screen.
Calendar on home screen now goes to reminders screen

The dragging around to change when the reminder is for between today, tomorrow and later is very buggy but it's updating the backend datastore which then publishes the changed reminders list back to the UI, which causes a view referesh. Also it's hard to know what date dragging something to Later means the reminder is for.


chrismclarke (approved):
Not sure if you were waiting on me to review, but looks really clean and clear to me so I'm going to go ahead and merge so that I can also make a few updates based on this.

PR #82, opened 2020-11-03
Add Survey Feature Module

chrismclarke:
## Main Changes
Adds service, components, and data to generate survey forms, and save results to the database.

Surveys can either be authored in [suvey-js](https://surveyjs.io/) or custom components integrated with angular forms. Additional handlers included for checking/validating entries, processing survey schema updates, and database bindings. 

Example surveys given in both available formats, the first a rewrite of the intro-tutorial component to the `analytics` survey (where user provides consent for analytics), using custom components and angular forms. The second an example partial rewrite of the welcome flow to use a simple survey interface. 

## Additional Changes
My local tslint/prettier seemed pretty opinionated on line wraps and quotation marks, and have overwritten a few extra files. As these changes are in-line with the current repo defaults (were appearing as lint errors) assume this is fine and unlikely to be reverted back. A few linting rules were also relaxed


michaelclapham:
Looks sensible to me. I like the use of component inputs for custom survey components.

PR #86, opened 2020-11-09
Feat/reminders

chrismclarke:
Update reminders module:

## Main Changes
- Add database bindings
- Update models and mock generation
- Add local push notification bindings
- Improve component interface
- Refactor to own module

## TODOs
- Local push notifications requires testing
- Could possibly also expose notification time (current default 9am)
- Requires testing how timezones are handled
- Add more ways to integrate 3rd party notifications
- Past and future (beyond 5 days) should be visible on click (possibly dropdown expandable section) - currently just total counter

## Known Issues
- If checking multiple issues in a row any animations already started will be restarted


michaelclapham (approved):
Code looks good but bottom add button seems to not work when running on web on device. will merge down and find a solution to this on a separate branch.

PR #98, opened 2020-11-15
add intro splash component

chrismclarke:
Hey @michaelclapham,

Following last week's call with Esmee here is an updated component to show an intro splash screen with faded sections. The technical document sent to me outlining it was as follows:
```
App name (ParentApp)+ logo 
Encouraging/inspirational slogans that fades. 
<fade in> 
Bringing up teenagers is hard 
<fade out> 
<fade in> 
You are the best parent your teen could have 
<fade out> 
<fade in> 
Join the 124 million parents who have used our resources 
<fade out> 
<fade in> Powered by Parenting for Lifelong Health PLH LOGO Logos of UNICEF and WHO <fade out>
```

I'm not sure if this data should be populated from somewhere specific and if we have any of the images/logos/names. If you have any more on this could you please add to the component, it's laid out in a fairly basic html component so should be pretty straight forward to work with (at some point in the future I'm assuming we will redefine in some sort of xlsx survey format, but for now I think we just wanted to get something in quickly)


chrismclarke:
Note - requires angular animations which is already added in another branch so should hold off until #99 merged

chrismclarke:
#99 Now open so should be good to merge after

michaelclapham:
I don't really understand why the splash screen is a survey, but it doesn't break the build. Merging

PR #114, opened 2020-11-20
Popup modal with animations to reward user

### LINKED ISSUE #107: Add reward animation component
As discussed, we need a general component that can be opened either from a model or included in any other template as part of a rewards system.

## Specific requirements:
- Accepts `id` as `@input` property for specifying the reward id
- Accepts `animate` prop which will be used to trigger animation on create
- Reads animation json from assets folder (given by reward id)
- When mounted (using either `onInit` or `ionViewDidEnter` lifecycle) should load the animation json from assets and play (if animate specified from prop)

## Nice-to-haves
- Display mode that can be used statically (possibly the last frame of the animation, or maybe a standalone image - tbc)
- Could possibly also include a text input property to also display a message (not currently requested, but could be useful)
- Provide fallback animation if the specific id isn't found

## General info
- A couple starting ids can be seen in the `AwardsList` tab at: https://docs.google.com/spreadsheets/d/1x8hluzT4IwEPapyNsknciKdv1Bc9z8WiF7ZxTCrnoPY/edit?usp=sharing
- Recommend just using whatever stock animations can reasonably be found for now.
- It's not sure exactly how we will be calling it yet, likely via a modal but that should behave mostly the same way as if included in HTML template, e.g. `<plh-reward-animation id='fireworks' animate='true'></plh-reward-animation>
- The component should sit in the `features/goals` folder, and can use any existing services or create a new one there if required (or can just handle logic from component ts)

michaelclapham:
PR Checklist

## Description

Animation modal for rewards system.

Supports Lottie or SVG + CSS based animations

Usage example:
```typescript
const modal = await this.modalController.create({
  cssClass: "anim-slide-up-modal",
  component: AnimModalComponent,
  componentProps: {
    id: id,
    title: "Congratulations!",
    autoCloseMs: 4000
  }
});
modal.present();
```

## Git Issues

Closes #107

## Screenshots/Videos


chrismclarke (approved):
That's awesome! Really nice functionality and implementation, like the subtle modal styling and added blob animations a lot - sure we can build on these in some nice ways.

Issue #122, opened 2020-11-23
Auto-populate all data from gsheets


chrismclarke:
A few things that we need to consider before implementing:

1. What's the best way to download the source sheets? Manual? Script? 
I'm guessing as we might want multiple sources as parts of different build pipelines at some point will want to use google sheets API to download an entire folder, and then convert/populate

2. What about assets?
I'm guessing we will want to allow assets to be included in same google folder, so that any assets available automatically overwrite default build assets. Will need to decide whether to handle as part of nested folder structures, or tidy up naming and go for a flatter folder structure.

3. What file format's best for outputs?
Up till now have been using JSONs, although it might actually be easier to convert to `.ts` with the json content as a default export so that we can also use linting to check if data has been defined correctly. It also means we can use data directly without awaiting HTTP imports and can implement tree-shaking.

4. What should be checked into source control? 
There are trade-offs of having all data within source control to avoid surprises of changing data, vs keeping a cleaner git history and conflict management. I'm leaning towards keeping converted data within git, as I have a feeling that future build pipelines will likely opt for a common core of data and then custom overrides (e.g. a country-specific deployment may want to only change some data but keep aligned with main updates)

5. How can we make it easier for authors to see updates in the app or different build variants (e.g. country-specific deployment)?
Short term I'm leaning towards a simple page/button in the app where a user can drop an xlsx file, automatically run the existing conversion script and updating the current data service (or whatever is storing a master copy of all data read into the app) with the new data. Open to suggestions

It would be good to get your thoughts on these points @michaelclapham, I don't think it will be on the roadmap for this upcoming release (v0.5.0) but likely shortly after.

PR #135, opened 2020-11-24
export xlsx script data direct to app

chrismclarke:
PR Checklist

## Description
Takes existing script to convert xlsx sheets to json (merging with content sheet data), and now also writes
to the main app `src/data` folder (as discussed in #122).
Adds to prettierignore to exclude from auto-formatting (easier to keep consistent and view git changes)

Note - further tweaks expected to include content changes such as marking xlsx files as 'draft' to be excluded from the app


chrismclarke:
Note, will be breaking as some of reminders module depends on older data structures, but these will be refactored shortly

PR #146, opened 2020-11-26
Store user settings in backend so that user settings can be changed remotely for user testing

michaelclapham:
PR Checklist

## Description

This PR is a proof of concept for using a MongoDB driven backend + Firebase authentication to store user settings.
Backend is messy but reasonably secure here: https://github.com/michaelclapham/firebase-mongo-server
Hosted at https://plh-demo1.idems.international/user-api/

It supports an On Behalf of (OBO) feature where OBO Admins are entitled to act on behalf of any other user who has logged in. This allows them to modify another users settings. This could be used by the user testing team to toggle options in the app between different versions we want to test.

## Screenshots/Videos

Change settings for yourself or for another logged in user.
image

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -88,7 +88,8 @@
         "serve": {
           "builder": "@angular-devkit/build-angular:dev-server",
           "options": {
-            "browserTarget": "app:build"
+            "browserTarget": "app:build",
+            "proxyConfig": "src/proxy.config.json"
           },
           "configurations": {
             "production": {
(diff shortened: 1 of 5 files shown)


chrismclarke (changes requested):
Interesting concept, although I'm still not sold on the core idea (I can't think of many use cases for wanting to modify an individual user's settings as opposed to a group/cohort, and even then it feels a bit invasive - I'd expect changes we make more to be on a db tidying level, and that way we can just sync general settings changes back down to a user profile instead of modifying ones they have set.

As this is only p.o.c and introducing a new backend db structure I'm going to say not to merge, but instead you should keep as a feature module in case we want to build on at a future date. I'd suggest naming that branch something more declarative, such as `concept/act-on-user-behalf` or similar

Issue #189, opened 2020-12-07
Right hand side module circles


michaelclapham:
![image](https://user-images.githubusercontent.com/3285072/101388161-7bac6b80-38b7-11eb-8949-a991c0997d9d.png)

PR #193, opened 2020-12-07
Slider side panel for modules

michaelclapham:
PR Checklist

## Description

Right hand slider for selecting module to start

## Git Issues

_Closes #189

## Screenshots/Videos

image


chrismclarke (approved):
Looks great! slider animation works well too. Also nice to see the effects of changing module.
My only suggestion would be to try and avoid the hardcoded colours, even if it means not perfectly matching the mockups (I might try quickly using a combination of brightness and opacity filters on the primary colour to see if I can get something reasonable). Will also be interested to see how it handles adding more modules! But for now really happy to have it in

chrismclarke:
Actually, I see some maths in there so have a good feeling!

chrismclarke:
image
Done, now working in glorious hot pink.

PR #212, opened 2020-12-10
Generate Rapid Pro flow UUIDs deterministically

michaelclapham:
PR Checklist

## Description

Generates Rapid Pro flow export UUID's (universally unique identifiers) in a deterministic way.
This means that when the spreadsheet to flowchart code is run on the same sheets the output will not be changed.

The UUID's now follow a format of

```
`uuid_${flowName}_${entityType}_${counter}
```

For example the first node in the flow example_main would have the id
uuid_example_main_node_0


chrismclarke (approved):
Sounds like exactly what we want, I feel like this will also really help with debugging more generally as well as keeping track of data updates. nice to have the added typings also

PR #224, opened 2020-12-11
WIP: Show error when flow name not found

michaelclapham:
PR Checklist

## Description

Shows an alert when something tries to load a flow name that does not exist


chrismclarke:
The implementation makes sense, although it would be good to get checks into the parser so that authors are aware without having to wait for a fail. The parser scripts already have a check from flows exist that are referenced from task flows, it would be useful if we could also have a check so that flows referenced from other flows also exist

chrismclarke:
Happy for you to merge this once conflicts resolved and create another issue for the checks

chrismclarke (changes requested):
Happy for you to merge this once conflicts resolved and create another issue for the checks as mentioned in comment above

michaelclapham:
Closing as chat no longer used

PR #229, opened 2020-12-12
Open first app opening flow if first time on home screen

michaelclapham:
PR Checklist

## Description

Opens initial flow (first_app_opening) if first time visiting home screen.
This uses same code that was on old home screen page


chrismclarke (commented):
I really don't like this, if we ever change the home screen page then we need to remember to copy this across. The app component would be a better place for required initialisation logic. Also as we need to implement a system for tracking triggers and events this week I expect this code will be removed in a couple days, so no point merging now and confusing people if it is working in this way only for a couple days.

chrismclarke:
Not required, will be handled through xlsx sheet definitions instead

Issue #230, opened 2020-12-12
Crash in one flow causes the next flow not to launch


michaelclapham:
Crash in one flow causes the next flow not to launch

chrismclarke:
Recommend adding some sort of init/reset function that can be used when first initialised but also after any flow completion (where completion is for the full flow stack in case of linked flows). Ideally this should all be handled within the service, but as a fallback we could also call in chat page lifecycle events to make sure things have been setup/torn-down appropriately between navigation

michaelclapham:
Closing as chat no longer being used

PR #231, opened 2020-12-12
Stop a flow crash from breaking next flow open.

michaelclapham:
PR Checklist

## Description

Before this fix a crash in one flow, would prevent the next flow from starting.

This was because the exception in the subscription to flow events caused the subscription for future events to unsubscribe.

## Git Issues

_Closes #230


chrismclarke (commented):
Would it not just be better to re-initialise all the variables or even the service itself when a new set of flows begins? (e.g. triggered from the chat page ngOnInit itself)? That way a broken flow can't persist errors within the service.

chrismclarke (approved):
Will merge as a short term fix, but still feels a bit fragile (see previous comment)

Issue #159, opened 2020-11-30
Handle end of flows


michaelclapham:
## Update 2020-12-12
There are a few different ways a flow can end which should be handled in specific ways:
1. Links to another flow (currently handled correctly)
2. Links to a specific content page (need to fix handling, ideally presenting a button like `Take me to tips` if a tips page.
3. No links (needs fixing, ideally presenting a button like `Return to Content` to take the user back to the previous page (if exists, otherwise home). 

I didn't see this button at all in some cases when testing, so added a message to say 'end of content'. Once this is resolved that message should be resolved also also.

## Original Issue
At the end of a conversation flow when I click on the "Take me to homescreen" button, it does not work.

![image](https://user-images.githubusercontent.com/3285072/100594874-1f16e280-32f2-11eb-8d47-2f3d8d89c731.png)

When I click on the button I get an error in the console
![image](https://user-images.githubusercontent.com/3285072/100594997-440b5580-32f2-11eb-890b-809a1868bb01.png)

This is also causing an issue where it warns you about losing progress incorrectly when you try to exit the chat screen at the end of a flow.

PR #262, opened 2020-12-16
Fix end of flow errors

michaelclapham:
- Deal with flow end from router node
- Clear flows stack when starting new flow
- Fix flow continuation

PR Checklist

## Description

Fixes exceptions happenning when a flow should have eneded but didn't, causing it to try to enter a null node.

## Git Issues

_Closes #159


chrismclarke (approved):
looks like better handling all round, thanks for this

PR #325, opened 2021-03-02
Feature: Slider Template Component

egor-mir:
PR Checklist

## Description

Develope Slider Template Component

Slider Parameters:

- value - number or null. If null - don't display slider picker circle.
- help - text or null. If null - don't display help icon.
- min - number - minimum value of slider. Default - 0
- max - number - maximum value of slider. Default - 7
- step - number - slider value accuracy. Default - 1
- title - text or null - slider title text.

## Git Issues

_Closes ##320

## Screenshots/Videos


smborio:
Slider Parameters:

help - text or null. If null - don't display help icon.
min_value - number - minimum value of slider. Default - 0
max_value - number - maximum value of slider. Default - 7
min_text - string - reference text for minimum value (left aligned with label for minimum value)
max_text - string - reference text for maximum value (right aligned with label for maximum value)
unit_text - string - reference to display values (centered under display values for the slider)
step - number - slider value accuracy. Default - 1
title - text or null - slider title text
default_value - null

Note:
_value is a variable that is set through the slider, not a parameter. There should be 3 states. Nothing done (question skipped for example) it stays as null. if the user selects the no answer circle on the right of the slider then we want to assign the value "na". If the user selects a number we want to assign this number.

PR #343, opened 2021-03-12
Support @fields in templates

michaelclapham:
PR Checklist

## Description

Replace @fields.var_name in templates using the contact field service used for chatbot implementation.

Work in progress as we need to support a way of populating these values from content code.


chrismclarke:
makes sense to me

PR #349, opened 2021-03-14
bugfix so that template names are used correctly from excel authoring

smborio:
PR Checklist

## Description
Adding templatename as a property of template and distinguished between when template.name and template.templatename are used to ensure correct implementation of excel authoring (when giving a name to a nested template)


smborio:
Second commit is a temporary fix to enable the nav_groups to work for the demo

PR #373, opened 2021-03-19
Global and field variable implementation

michaelclapham:
- Change globals implementation
- Added global data file
- Content update
- Allow @field as well as @fields
- Content sync
- Add set_field action

PR Checklist

## Description

@global constants are now set in sheets with flow_type global.
Sheets with flow_type global can also be used to set initial field values using row type set_field
@field or @fields can be used for getting field values.
set_field action working.
set_field row type working.


chrismclarke (commented):
All in all looks pretty nice and tidy, although just to check have you tested it with the example templates? I'd be surprised if working correctly due to the code:
```
if(this.getField(row.name)===undefined
```
as I thought local storage returns null if a key is not set? At least that's what the official typing says:

>  /**
     * Returns the current value associated with the given key, or null if the given key does not exist in the list associated with the object.
     */
    getItem(key: string): string | null;

PR #387, opened 2021-03-23
Hidden JS evaulation

michaelclapham:
PR Checklist

## Description

_What this PR does_

## Git Issues

_Closes #_

## Screenshots/Videos


chrismclarke:
really like this implementation, good idea to start with just the hidden evaluation.
My own question is if we need the exceptions for "true" and "false", and instead just process everything and then convert back to a boolstring after? Because currently it looks like our evaluator function could be returning either a string or boolean depending, which might lead to some inconsistencies

PR #379, opened 2021-03-21
Fix/local variables

chrismclarke:
PR Checklist

- [ ] - Latest `master` branch merged
- [ ] - PR title descriptive (can be used in release notes)

## Description

- add `sheetVariables` to template to define variables which should be made available to all other rows in the same excel sheet
- add recursive check for dynamic expressions
- rename `set_local` row type to `declare_variable`
- tidy template row filter types
- add fallbacks to getField and getGlobal expressions
- make paragraph text break on new-line characters

## TODO / Feedback required
- possibly rename `declare_variable` to something to reflect a) it is accessed via `@local` property and b) it will not be overwritten (it is a constant referred to in the sheet)
- possibly re-interpret if in fact it shouldn't be a constant (should the variables set in `declare_variable` be overwritten by parent?

## Git Issues
Should check if helps with #366 or #360

## Screenshots/Videos
https://docs.google.com/spreadsheets/d/16fNOe2W8inFxSIkZSKDa_lvQv68xFEKy5_LtNSjZyDY/edit?usp=sharing

image

https://plh-teens-app1--pr379-fix-local-variables-1l66imd2.web.app/template/debug_nesting_variables

image

PR #420, opened 2021-03-27
Update system for parsing parameter lists

chrismclarke:
PR Checklist

## Description

- Convert paramater_list to key-value pairs in parser (to avoid needing to do in frontend code)
- Update frontend paramater_list type parsers (string, boolean, number) to reflect key-value pairs instead of array
- Update parser to convert rows with names containing `_list` to array (e.g. `@local.button_list`)
- Create new radio group example with local variable list to allow list in parameter list
- Refactor identified components where new parameter format breaking

## Known issues and todo
- Radio button components cannot be tested properly as require parsing of recursive local variables (depends on refactor of #379)
- Would be good to test some of the other components that use parameter lists to check methods still behave as intended.
 
@michaelclapham - could you take a quick look at some of the components you have worked on? I tried to keep the same-named util functions to just work with objects instead of arrays, however I did not understand the `getStringParamFromTemplateRowValueList` which seemed identical to `getStringParamFromTemplateRow`, and not aware if there's other places parameters will now be parsed as a list (or should be)

## Git Issues

_Closes #_

## Screenshots/Videos

### DIFF
--- a/scripts/src/plh-data-convert/parsers/default/default.parser.ts
+++ b/scripts/src/plh-data-convert/parsers/default/default.parser.ts
@@ -2,6 +2,7 @@ import chalk from "chalk";
 import * as fs from "fs-extra";
 import { FlowTypes } from "../../../../types";
 import { AbstractParser } from "../abstract.parser";
+import { parsePLHListString } from "../utils";
 // When running this parser assumes there is a 'type' column
 type IRowData = { type: string; name?: string };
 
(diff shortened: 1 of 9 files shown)


smborio:
@chrismclarke The list is also used in [combo box](https://docs.google.com/spreadsheets/d/1ZnBQ_hJnbnJ0SwY4uA9rCjonHRaGK0n0GUdg1EB-HLc/edit#gid=456120628&range=F4), see this example, which is not reading the list properly.

PR #500, opened 2021-04-06
fix nested popups

chrismclarke:
PR Checklist

## Description

Update popups to associate with top-level template instead of parent (so they trigger correctly when called from within nested structures)

## Git Issues

_Closes #_

## Screenshots/Videos


michaelclapham:
If I go to https://plh-teens-app1--pr500-fix-nested-popups-a4246rvw.web.app/template/debug_pop_ups_override and then:
1. Click on button 3 (it opens a popup)
2. Click on button 1, (it opens another template)
3. Click on completed (it goes back to the previous template)
4. Now none of the buttons open the pop-up any more

If at step 3 you click uncompleted, it goes back to the previous template, but the popup that should be open is not open.

PR #503, opened 2021-04-06
Refactor/db structures

chrismclarke:
PR Checklist

## Description
@michaelclapham - I've already discussed all these aspects with David so from a content side should be good, just wanted to make you aware of the changes before merging. Of possible note is the change to the contact field service, which keeps the local storage system implementation but provides an additional db write (for historic tracking). I noticed the same code was in 2 places but thought best to keep this way for now until we know better what will happen with the chat system.

- Add a common `flow_events` table for tracking data from flow interactions
- Add a common `data_events` table for tracking data from specific data updates (e.g. setting contact fields)
- Add authoring opt-out for data capturing (`db_ignore_events`)
- Add db writes for template emit completed actions
- Set contact fields to provide an additional write to the database 
- remove duplicate template service file

## Git Issues

_Closes #_

## Screenshots/Videos


michaelclapham (approved):
All looks sensible to me!

Issue #513, opened 2021-04-08
Move template list button in header


egor-mir:
Add new ion-button to header, on click navigate to /template

PR #514, opened 2021-04-08
feat: add ion-button to header.

egor-mir:
PR Checklist

## Description
Add new ion-button to header, on click navigate to /template
_What this PR does_

## Git Issues
#513 
_Closes #_

## Screenshots/Videos
image


chrismclarke:
I'm not sure if it is something already discussed with Michael, but I think probably we want to keep the yarn lock files checked into the repo as it ensures everyone uses the same versions of dependencies.

Here's an old blog post explaining reasons: https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

Once we have released and things are more settled I think we will want to migrate to yarn v2 which will have a slightly different format

PR #566, opened 2021-04-14
Fix intro.js to support Angular navigation

michaelclapham:
- Workaround for intro.js to support ng routes
- Tour can highlight template comp by name

PR Checklist

## Description

Workaround to deal with the fact that intro.js does not support single page applications.

When changing step, don't actually change step until navigation is done, then use workaround to select elements that are only just shown on page.

Also adds support for highlighting elements in a tour by their name in a template page.
_

## Git Issues

_Closes #_

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/template-component.ts
+++ b/src/app/shared/components/template/template-component.ts
@@ -16,8 +16,6 @@ import { TEMPLATE_COMPONENT_MAPPING } from "./components";
 import { FlowTypes, ITemplateRowProps } from "./models";
 import { TemplateContainerComponent } from "./template-container.component";
 
-export const NAME_CLASS_PREFIX = "tc-name-";
-
 /*********************************************************************
  *  Directive used as part of loading dynamic flow component elemnt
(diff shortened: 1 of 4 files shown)


src/app/shared/components/template/template-component.ts
```diff
@@ -42,7 +44,7 @@ export class TmplCompHostDirective {
   selector: "plh-template-component",
   template: `
     <div
-      class="plh-tmpl-comp"
+      class="plh-tmpl-comp {{ rowNameClass }}"
```
chrismclarke, line 47: would it not just be easier to use data attributes like below, and specify
```
[attr.data-rowname]="row.name"
```
That would also mean we don't need any of the other lines of code in this file, keeping it more concise

chrismclarke (changes requested):
would the change I suggested work? would be nice to limit the amount of custom code going into this general component (happy with whatever else goes into the tour service)

PR #572, opened 2021-04-15
Change styles in TmplParentPointBoxComponent

egor-mir:
PR Checklist

## Description
Css changes in TmplParentPointBox Component:
Add scale factor and flex property for display two items per row.
Template [Link](https://plh-teens-app1--pr572-feat-parent-point-bo-l37d6rfp.web.app/template/feature_parent_point_box)
_What this PR does_

## Git Issues

_Closes #_

## Screenshots/Videos
image


michaelclapham (approved):

src/app/shared/components/template/components/points-item/points-item.component.scss
```diff
@@ -1,10 +1,10 @@
 .wrapper-point {
```
michaelclapham, line 1: .wrapper-point {
    --scale-factor--point: 1;
}
@media screen and (max-width: 400px) {
  .wrapper-point {
    --scale-factor--point: 0.227vw; // I think this is the right number 100 vw / 440
  }
}

michaelclapham (changes requested):
Is there a way we could set --scale-factor--point using a CSS media query? It seems like it would be more reliable and efficient that way.

For example
.wrapper-point {
--scale-factor--point: 1;
}
@media screen and (max-width: 400px) {
.wrapper-point {
--scale-factor--point: 0.227vw; // I think this is the right number 100 vw / 440
}
}

Not sure if the media query logic is exactly right, but you get the idea.

src/app/shared/components/template/components/points-item/points-item.component.scss
egor-mir: But now it is dynamically and calculates based on template-container width. This is more universal solution from my point of view

michaelclapham: I don't think that's true, but merging anyway

smborio (approved):
problems sorted, looking good

PR #575, opened 2021-04-15
Fix popup height

michaelclapham:
PR Checklist

## Description

Instead of always being all, only be as tall as popup content.

Also corrects centering of popup


smborio (approved):
You haven't specified a template to test with so I'll have a quick look at the preview and approve as the changes seem sensible

PR #678, opened 2021-04-29
Add support for data exported by subtype

chrismclarke:
PR Checklist

## Description
Adds support so that providing a `flow_subtype` column in a flow will separate out data exports in the app. This makes it easier to debug and work with data in the app, and also will make it easier to exclude specific flows from future builds. It also will make it easier to track changed files on github.

The changes are purely backend/scripts, as the final export made available to the app is formatted in the same way as previously (instead of one file providing an export of all data inline, a new index file collates from multiple files before providing the same singular export)

**Note** - This PR does not include updated data (to avoid merge conflicts), so after merge the `plh-sync-data` script should be run to reformat the data to the new folder structures

## Future TODOs
- would be good to add an optional `flow_model` column if we also want to provide more type-checking for specific flow sub_types (ensuring the sheets column matches expected)

## Git Issues

_Closes #_

## Screenshots/Videos
_Sheet column_
image

_Data output folder_
image


smborio (approved):
approved in meeting but needs testing.

PR #735, opened 2021-05-13
Fix/asset urls

chrismclarke:
PR Checklist

## Description
- Add common method to correct urls for asset images
- Refactor components to use common method instead of individual per-component methods

## TODO
Finish replacing methods in rest of components
image

## Git Issues

_Closes #_

## Screenshots/Videos
image


smborio (approved):
quick functional test passed

PR #741, opened 2021-05-16
Campaigns Feature - core updates

chrismclarke:
PR Checklist

## Description

Provide necessary infrastructure changes to support campaign system implementation. Including:

- updating data parsers and services to correctly collate all campaign data
- adding data bindings to link @campaigns in templates to the campaign list data
- making all template processing asynchronous so that campaign lookups can also evaluate conditions at runtime (previously all @data, @fields, @global, @local etc. were synchronous, but now as we directly need to query the database for the first time fields were set as opposed to just current values that means we need things to be more flexible/powerful)
- adding script updates to allow processing actions from a called variable
- script updates to process the evaluation lists more neatly (and allow self @row notation)
- refactoring a lot of legacy code part-implemented for the first and second iterations of the reminders system
- creating structures to support testing debugging of work-in-progress implementation
- generally trying to maintain some degree of code tidiness and slow move towards more generalised data structures

## TODO (required for current implementation)

- add debug templates
- add 'sent' handling and evaluation
- test and review other condition evaluation requirements
- add better campaigns debug page (?)

## TODO (longer term)
- Make `@campaigns` an implementation of a more generalised `@data` or `@query` lookup
- Unify template `condition` and data_list `condition_list` evaluation methods
- Add type-checking to data lists for known types (e.g. campaign list)
- add way to expose data_list type from within campaign module folder
- add way to expose evaluation processing (i.e. @campaign) from within campaign module folder

## Git Issues

_Closes #_

## Screenshots/Videos
Example/testing **home_screen_campaigns** template (to be removed/refactored)
image

image

image


smborio (approved):
Quick functional test done. Will follow up with issues, if any

PR #784, opened 2021-05-29
Fix/debug actions

chrismclarke:
PR Checklist

## Updates (2021-06-07)

- Added `update_action_list` row type that can be used to set a templates actions from within the template itself
- Added `nav_resume` action trigger which is triggered after popup close or back navigation
- Added `example_update_actions` template to demo 
- Updated home screen to use these to emit a `force_reload` to fix the campaign reload

https://user-images.githubusercontent.com/10515065/121130002-9ddd6280-c7e2-11eb-9798-b6e614a2c4cb.mp4

https://user-images.githubusercontent.com/10515065/121130203-ed239300-c7e2-11eb-98d3-f2e34f9d464e.mp4

image

## Updates (2021-06-05)
- The initial issue of action list inheritance should now be fixed, although the authoring updates mentioned in multiple places are still required. I've now manually updated the action lists for `w_self_care_reward.think_temp` and `w_self_care_recognise.think_temp` so at least the first workshop runs smoothly

- A related issue where text_box and text_area fields weren't updating correctly was identified and a fix applied (similar issue to before where it wasn't always waiting for the input value to be saved before emitting change). This was further masked but a more general angular change detection issue (we relied on components updating their own value, but this would not detect changes if their value was updated as a side-effect of another component updating a common field).

- `set_variable` rows have been added to the list of things recalculated on every re-render for cases where the variable is specified dynamically. This hopefully makes sense from an authoring perspective, anything hardcoded as a fixed value will only ever be set once (and can be overwritten from inside the template), anything dynamically assigned will always update (so updating the value from within the template will not work).

- Discovered related issue for home screen campaigns (not updating correctly). The campaign is successfully re-evaluated and the row updated, but angular does not correctly detect the change as the row is within a display group (the display group cannot detect the change that has happened within its child rows array). Essentially all components that render templates would need more fine-tuned change detection applied. This is not ideal, so likely will require larger overhaul to fix (where each component subscribes to their value directly from the container, instead of the child display component).

- The `force_rerender` emit has been renamed to `force_reload` and all current usages have been changed to use the `force_reprocess` method instead (which does not fully re-initialise, just recalculate existing)

--------

## Known Issues
- ~~w_selfcare_stepper appears to be marking radio set_fields as undefined (need to check if that is correct or not, likely pre-existing issue)~~ (not caused by pr)
- All workshop templates need to be checked to see if they are emitting completed/uncompleted correctly, with additional actions added where missing (author updates)
- Home page does not update campaign icon on return due to angular change detection issue (won't fix in this pr)

## Description
This PR aims to address bugs seen whilst debugging nested action overrides. A few main issues addressed include:

1. The debug template itself had multiple errors which led to misleading diagnosis
2. The default handling of emit actions were to propagate to parent unless otherwise listed (this was a conscious choice at an earlier time, but assume that is now legacy and updated to require explict re-emit from parent).
3. Refactor required better separate out the code to make for easier debugging and future dev
4. Possible minor issues detected in deep merge of nested dynamic lists

Sufficient changes and refactoring made that made it pretty impossible to keep meaningful comparison, so should be considered a rewrite of the core methods. These are now separated out enough that it should be possible to make better sense of the methods from within their standalone files (only need to refer `template-row.service`, refactored template-actions left unchanged). As such extensive testing is recommended.

Main changes:

- Remove default action handling that passed actions to parent if no other handlers specified (see inline comment)
- Refactor and merge separated template merge methods
- Separate out actions and row processing code to own services
- Update row processing methods
- Update testing mock methods
- Improved logging for emit methods to make it easier to review action behaviour at runtime

## Git Issues

_Closes #_

## Screenshots/Videos

Deep grandchild element rerendering

https://user-images.githubusercontent.com/10515065/120725538-08ad3780-c48b-11eb-83f3-145a7a02527d.mp4


src/app/shared/components/template/services/template-action.service.ts
```diff
@@ -0,0 +1,229 @@
...
+            const actionsForEmittedEvent = row.action_list.filter((a) => a.trigger === emit_value);
+            log("Excuting actions matching event ", { emit_value, actionsForEmittedEvent });
+            // process in parallel, do not return/await
+
+            // TODO - check if should be parent row or current row
+            parent.templateActionService.handleActions(actionsForEmittedEvent, row);
+          } else {
+            log("No action list for row ", row, "on template name ", name);
+            // NOTE - CC 2021-08-05 - Remove default action passing to parent
+            // parent.handleActions([action], row);
```
chrismclarke, line 149: This is the main change made to parent action handling (commenting out default behaviour to pass to parent).

I'm not sure exactly why we had it before, so possible to have knock-on effects

src/app/shared/components/template/services/template-row.service.ts
```diff
@@ -0,0 +1,241 @@
...
+  /** Hashmap of all rows keyed by nested row name (e.g. contentBox1.row1.title)  */
+  public templateRowMap: Map<string, FlowTypes.TemplateRow> = new Map();
+  public processedRows: FlowTypes.TemplateRow[];
+  public renderedRows: FlowTypes.TemplateRow[]; // rows processed and filtered by condition
+  constructor(public container: TemplateContainerComponent) {}
+
+  /***************************************************************************************
+   *  Row Initialisation
+   **************************************************************************************/
+
```
chrismclarke, line 30: These are the rewritten methods for handling row intialisation and processing

src/app/shared/components/template/services/template-row.service.ts
```diff
@@ -0,0 +1,241 @@
...
+        const { name, value } = override as any;
+        const type = "set_variable";
+        const rowEntry: FlowTypes.TemplateRow = { name, value, type, _nested_name: name };
+        processedRows.unshift(rowEntry);
+      });
+    }
+    return processedRows;
+  }
+
+  /***************************************************************************************
```
chrismclarke, line 114: These are the updated methods for handling row processing.

src/app/shared/components/template/utils/template-utils.ts
```diff
@@ -1,27 +1,35 @@
 import { FlowTypes } from "src/app/shared/model";
 import { arrayToHashmap } from "src/app/shared/utils";
 
-/**
```
chrismclarke, line 4: These are the modified methods that perform template deep merging, the changes are not intuitive as the 2 main methods were essentially flipped to allow merge with the template row processing methods (per TODO item below)

chrismclarke:
`nav_group` issue now fixed (adding default emit actions), but will still see some issues in nav group page changes in cases where emit conditions not explicitly authored, e.g.

**Incorrect syntax** - would previously still emit as a default but now no longer does
image

**Correct syntax** - specificy emit (exists on some but not all)
image

@smborio - content will require update to make consistent

smborio (approved):
Merging to test for undesired consequences

PR #797, opened 2021-06-02
Update android permissions

chrismclarke:
PR Checklist

## TODO
Review any further changes required

## Description

Following issue where user device does not support GPS and so cannot be installed

- remove gps permission 
- better document required permissions

## Git Issues

_Closes #_

## Screenshots/Videos

Capacitor plugin docs (individual plugins will reference any android permissions required)
https://capacitorjs.com/docs/v2/apis

E.g. unsupported device - **android.hardware.location.gps** 

image


chrismclarke:
OK, I've now commented out all permissions as making a permission optional turns out to be more effort than we would want to put in right now (essentially you need one line of code to enable the permission, then multiple lines for every 'feature' that relates to it (e.g. the camera alone would need around 10 extra lines to specify fully optional: https://developer.android.com/guide/topics/manifest/uses-feature-element#camera-hw-features)

But in the current app format I don't think we actually need any, and as we make more features enabled we can improve on.

smborio (approved):
looks good, great notes, thanks!

PR #809, opened 2021-06-05
Refactor - Dynamic variables and calculation statements

chrismclarke:
PR Checklist

## Breaking Changes
Due to improvements to how variables are processed, some of the dynamic statements that relied first on a string replacement and then javascript evaluation no longer work (as they are processed as a first pass in javascript). Specifically:

- Mathematical statements on field values no longer parse correctly. As all fields are retrieved as string values, the first pass will evaluate as a string instead of a number. 
**Before**: `@fields.some_value+1`    (now would simply append a `1` onto the end of the number)
**After**:  `+@fields.some_value+1` or `@calc(parseInt(@fields.some_value)+1)`
(the `+` forces javascript to evaluate as a number)

- Dynamic variables should no longer be quoted as strings for comparison.
**Before**: `"@local.radio_group"=="happy"`     (now would compare the first word with a double set of quotation marks)
**After**: `@local.radio_group=="happy"`

## Update 2021-06-10
- Much larger overhaul given to allow calling predefined functions from calculations and rewrite dynamic parser to be cleaner and address a number of identified bugs with the new implementation
- Add new functions to pick a random item from a list and to lookup an answer_list item text from name
- Create example calc sheet to demo core functionality
- Lots of local testing of different dynamic expression types

https://user-images.githubusercontent.com/10515065/121605241-eb371b00-ca00-11eb-8ae5-30607462ed07.mp4

image
https://docs.google.com/spreadsheets/d/15TccctjzepSPPv9xrLp46fhYldNN0wrho1zA5voTYcg/edit?usp=sharing

## Update 2021-06-07
- Following call mostly good to go, just want to add an example of picking a random item from a list (possibly with use of helper method)

## Description

- Provide support for template statements including `@calc(....)` which will evaluate the calculation in JavaScript
- Create example template

## TODO
The PR should be in reasonable shape, just need to wait on #784 for re-rendering to work in example and confirm naming and general utiltiy.

## Git Issues

_Closes #_

## Screenshots/Videos
image

image


smborio (approved):
Discussed in meeting

PR #826, opened 2021-06-10
content reviewing

AndriiHubarievW:
PR Checklist

## Description
created new component 'reviewing-form.component'. Which is placed in app.component in order to divide screen into two 
 parts. The 'reviewing-form.component' listening for two events of the app 'contextmenu' - launching custom context menu 
 and 'click' - it helps to close presented context menu. 

## Future TODOs
- Landing page is a comment box for the whole page with a comment to right click on a specific component to comment about it
- if the selected item is a text of some kind, the right click should have the options for suggest change or make comment
- in non-text object, the right click only allows to make comment
- split screen in 'half' where the bottom has a title 'Existing comments about this screen' and a line saying no comments submitted yet
-  If it is a text, the selected item should only show the value, not all the information is currently showing. If it's not a text it shouldn't give any info (this won't make sense to people)

## Git Issues

_Closes #_

## Screenshots/Videos


chrismclarke (approved):
Hi @AndriiHubarievW 

@smborio mentioned this will be your last week contributing so I just wanted to say a huge thanks for all the work you've done, and wish you the best for your future plans.

I've also included below a few bits of specific feedback for this PR, although it isn't critical that you implement any of them (overall the code is good and any minor updates I can just make at a later point)

Feedback:

(+) The overall code separation is really good. I know some people will be opinionated to split angular projects into separate html/scss/js files for everything, but I think smaller features (like the `ExistingCommentsComponent` work well as a single file (in the style more like react) - It's good to realise that these things are all just down to preference

(+) Functions have clear names and are split well (really happy you use state things like `dismissContextMenu` instead of just `dismiss`

(?) Variable naming conventions - I think we are a bit inconsistent in the project whether to use `camel_case` or `snakeCase` - but I think sometimes mixing can be a bit confusing. E.g. `(open_suggest)="openSuggest". I might be tempted to make the action something like `handleOpenSuggest` or `toggleOpenSuggest` to be slightly clearer that it is an event emitter and not a variable or function

(+) Code style - it's really good that you've managed to implement the correct code style (via prettier/eslint), as this saves a lot of headaches if people keep reverting other's changes because they prefer single quotation marks over double quotation marks, for example.

(?) CSS Size notation - We have a bit of a mix of `%`, `px`, `em` and `rem`, which is probably a bit confusing. I don't have strong opinions myself and think there are probably times when each have their place but I think mixing %, rem and px in the same class (e.g. `plh-template-component:first-child` ) could lead to some inconsistent behaviour. There are loads of good articles on https://css-tricks.com/ with suggestions for when to use each

(+) Handling or rxjs - Subjects, BehaviourSubjects, etc. are always a bit of a tricky area, so really nice to see that you have managed to use successfully (including using async pipe binding). One thing I learnt the hard way when starting to work with subscriptions, is the potential for memory leaks when used too often, or without methods to unsubscribe. Here's a good article that covers some of the key points: https://blog.bitsrc.io/6-ways-to-unsubscribe-from-observables-in-angular-ab912819a78f

(?) Code comments - The code is clean and functions named well so there isn't much need for comments, however there are a few places that stand out as potentially confusing where a short comment might be beneficial (e.g. in the `getTargetRow` function we have `e["path"].find((c) => c.localName === "plh-template-component")?.__ngContext__[21];` - so I would like to understand why we are doing this (one line comment would do either in the function or before the function).

(?) Screenshots / videos - I find these are usually really handy to have for reviewers (even when we also have the preview build). So I'd recommend including a quick screenshot or video when you make a new feature PR. Ideally you want this to be as easy for you as possible - I'm on a windows pc and the in-built print screen lets me select the area to include (if you open the snip & sketch app there is an option to make this the default behaviour). This copies to my clipboard and I paste straight into GitHub (no saving files required) - I expect there are similar on mac/linux. As for videos, I use the free `screencastify` plugin for google chrome to record short videos of features - again this takes me very very little time to do.

But overall been really great to work with you, I've definitely been impressed with how quickly you've managed to integrate into the rather complex project and contribute in meaningful ways (particularly as I haven't been able to give much support myself).

If you ever need any extra endorsements or references in the future just let me know!

chrismclarke:
Oh, and as a final note, if you're trying to decide what areas you want to work on developing more in the future, a couple links that I think are really good for understanding more of the big picture (to transition junior->intermediate->senior) are:

https://roadmap.sh/frontend
(a visual roadmap of things to learn about for frontend development. They have similar ones for backend and devops)

https://insights.stackoverflow.com/survey/2020#correlated-technologies
(diagram going beyond Angular/JavaScript, to see how other languages/frameworks relate)

AndriiHubarievW:
Hi, @chrismclarke. 
 It is very useful to get this detailed feedback from you. I will consider all your advice for the future. Appreciate it so much!
 Thank you for your help and understanding. Special thanks for the links you provided. You do a great job.
  Also wish you the best for your future plans.

chrismclarke:
merge into #1161

PR #829, opened 2021-06-16
Fix/nested rendering

chrismclarke:
PR Checklist

- [ ] - Latest `master` branch merged
- [ ] - PR title descriptive (can be used in release notes)

## Update 2021-06-07
Hopefully now should be in much better shape, new updates include:

- Merged #830 to this branch for fixing child templates with same name as parent
- Merged #831 to remove duplicate re-renders following nav actions
- Updated change detection strategies to prevent excessive rerendering seen following #827 updates
- Improve additional change detection methods for nav events (fixes popup issue seen in `relax` templates)
- Improve change detection methods at component level (fixes immediate issue seen with `tools pages`)
- Fix slider bindings (fixes `debug_slider_with_template`)

## Description

- Add trackby methods to all display components to help rendering updates
- Change core row update methods to detect more changes
- Fix condition filter not being tracked for nested objects

## Git Issues

_Closes #_

## Screenshots/Videos

_If useful, provide screenshot or capture to highlight main changes_

PR #831, opened 2021-06-17
Remove duplicate nav re-renders

chrismclarke:
PR Checklist

## Description

Minor fix so that the automatic template reprocessing that happens after nav actions (e.g. popup close, back press) is only triggered from the top-most template (by default the action searches for the top-most template anyways and re-renders all templates, so this prevents duplicate rerendering)

This can be seen in the logs when navigating back to the weekly workshops page

## Git Issues

_Closes #_

## Screenshots/Videos


chrismclarke:
would be good to include in #829, so merging into master now

PR #838, opened 2021-06-17
WiP - Action list value updates

chrismclarke:
PR Checklist

## Description
As action lists are parsed and kept static, the values they refer to are always one step behind changes (if referring to self value in a row). Add workaround so that in cases where a row (e.g. `slider`) has an action (e.g. `set_field: some_field: @slider.value`), the self-reference is identified and populated when triggering the action (after the local reference updated)

## TODO
- Create example sheet for testing (e.g. workshop_setup qs)
- Check for any issues
- (optional) remove additional authoring added to duplicate set fields in nav buttons

## Git Issues

_Closes #_

## Screenshots/Videos


smborio (approved):
Will merge and test thoroughly

PR #846, opened 2021-06-19
new component help_icon

AndriiHubarievW:
PR Checklist

- [ ] - Latest `master` branch merged
- [ ] - PR title descriptive (can be used in release notes)

## Description
- added new component help_icon
- update documentation
- [feature_help_icon](https://docs.google.com/spreadsheets/d/1ARbNRGDer5vj9qSpRMZFrMkYifGkH3TLtDVp72YbaqU/edit#gid=1962267037)

_What this PR does_

## Git Issues

_Closes #_

## Screenshots/Videos

_If useful, provide screenshot or capture to highlight main changes_
Снимок экрана 2021-06-19 в 11 17 42

PR #847, opened 2021-06-19
Support full-stops at end of dynamic variables

chrismclarke:
PR Checklist

## Description

Add method to parser to check for dynamic expressions ending in full stops, and treat as a string full-stop instead of trying to access nested property

## TODO
- As the fix is to interpret the dynamic expressions slightly differently, any content authored as a workaround should ideally be updated (change spaces before full-stops). Although could add short term _hack_ to automatically replace ```(space).``` in sentences if useful. Otherwise should be good to merge

## Git Issues

_Closes #_

## Screenshots/Videos

image


chrismclarke:
(note build failing because of ```help_icon``` type issue in master, requiring merge of #846)

smborio (approved):
Approved by DS and CC in meeting

PR #875, opened 2021-06-23
Fix production build

chrismclarke:
PR Checklist

## Description
- Change how global functions are included in dynamic variable calculations as these are broken during optimisation step of production build
- Revert build script change to make ```production``` build by default
- Minor log change to exclude console logs from native builds (for improved performance)

## Git Issues

_Closes #_

## Screenshots/Videos
```v0.10.0``` - now appears to be working as intended
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "ng": "ng",
     "start": "ng serve --open",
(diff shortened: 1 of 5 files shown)


smborio (approved):
approved by david on a call

PR #884, opened 2021-06-30
Fix/tidy go to url

smborio:
PR Checklist

## Description

Added the function to open an external URL to the Nav services
Tidied the case for the go_to_url action
Added an if statement to identify and deal with urls that don't have http or https


smborio (commented):
functional test passed and code was reviewed with David, who approved it. I will merge with admin rights.

PR #886, opened 2021-07-06
change null to "" if the parameter for left_icon_src is left empty

smborio:
PR Checklist

## Description

Change the value for a left_icon_src parameter in the tile_component to "" from null. This is required as if it is null then the `*ngIf="left_icon_src"` in the file \parenting-app-ui\src\app\shared\components\template\components\tile-component\tile-component.component.html evaluates as true if the parameter is null. The consequence of having null is that for tiles that don't have the parameter left_icon_src the icon is not found as the location is "null" and it displays a default value if it exists in the authoring or a broken image if there is no default value. One example is the quickstart round icons in the home screen as shown below.

## Screenshots/Videos
Screenshot of quickstart buttons going back to default values if null:
</img>
Screenshot of quickstart buttons going back to default values if "":
</img>


smborio (commented):
Functional test passed. Code looks OK.

PR #897, opened 2021-07-13
Moved the reset app function to the settings services and created an …

smborio:
…action to launch the function

PR Checklist

Moved the reset app function to the settings services.
Added the action reset_app.
Defined the action to launch the function.


smborio:
David and I worked on this PR together and it's working well but we are not sure why it is. @chrismclarke could we review this together?

PR #917, opened 2021-07-21
Feat/campaigns updates

### LINKED ISSUE #912: Improvements to campaigns testing/debugger
**Details**
The current debugger for campaigns is very good but it would be much better to test if it could be improved.
Particularly, The following points would be useful:
1. to be able to change or set dates for any field that may be used for activation criteria 
2. to be able to redefine the first Launch day globally and not just while on the debugger 
3. to be able to set set any current date globally or while in the debugger.

For example, in [campaign_unlock_workshops](https://docs.google.com/spreadsheets/d/1droR9Op9sTzZMrLXkVmHlWjzwXMN3bq--2PKhocmlCY/edit#gid=1874917575) I need to be able to change the date when the field w_self_care_started became true or change the date in the app to 'pretend' that a number of days have elapsed in order to test whether workshops are being unlocked correctly through the unlock_workshops campaign.

Making global changes and not changes while in the debugger would be particularly useful to test that campaigns are actually working as expected and not just testing whether the activation or deactivation conditions are met. We currently don't have any way to test the actions are working as we would expect.

chrismclarke:
PR Checklist

## Description

Fix various issues with current campaigns system implementation and evaluation methods, and provide a better debugger with the ability to set persisted app fields and db data

## Git Issues

Closes #912

## Screenshots/Videos

Debugger screen now splits by activation status
image

Debugger row now uses checkboxes to show satisfied conditions
image

Variables editor now links to page where variables required by a campaign row can be directly modified. Additionally app first launch date and day can now be set to database (previously only mocked so non-persistent), and field value first dates
image

### DIFF
--- a/packages/plh-data/data/data_list/data_list.campaign_rows.ts
+++ b/packages/plh-data/data/data_list/data_list.campaign_rows.ts
@@ -46,7 +46,8 @@ const data_list: FlowTypes.Data_list[] = [
             condition_type: "field_evaluation",
             condition_args: {
               field_evaluation: {
-                evaluate: "w_self_care_completed",
+                field: "w_self_care_completed",
+                value: true,
               },
             },
             _raw: "get_field | w_self_care_completed : true",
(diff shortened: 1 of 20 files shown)


smborio (approved):
Functional test was good. Will keep testing after merging.

PR #935, opened 2021-07-30
Add new translations system

### LINKED ISSUE #918: Files and information re multilingual system
**Details**

1. JSON file containing the input file sent to translators in English: [example_original.json](https://drive.google.com/file/d/1tl8FBppsrYlb_pN97GUJ1jA5tWOehG8P/view?usp=sharing)
2. JSON file containing a rough translation (done very quickly and poorly) into Spanish in the structure we expect to get it back from translators: [example_translated](https://drive.google.com/file/d/1C0AIwx7xzoKxFjcjovDm_Lph4GIQz6JR/view?usp=sharing)
3. JSONs taken from the app sync process used to generate file 1. These are a subset of the corresponding data files in the app with only the relevant sections included. They refer to content in the google spreadsheet [example_languages](https://docs.google.com/spreadsheets/d/1vv1DbyXEkOEHVEAMwjtQW0vS23DDUBK1oOw24moe64M/edit?usp=sharing), which includes a global, a data list, a tour and four templates (which are inter-related)
a. [input_template.json](https://drive.google.com/file/d/1Jd7NBnwrTro5Ch6oYPHPoYkHVVakzPj7/view?usp=sharing)
b. [input_global.json](https://drive.google.com/file/d/1qxY40hGykNH0MKHFRegMbrJOeHCZiuDM/view?usp=sharing)
c. [input_data_list.json](https://drive.google.com/file/d/1FURJXbjD9d0C3p3IG0BsPYIaC8ES0NXD/view?usp=sharing)
d. [input_tour.json](https://drive.google.com/file/d/1Tkwfl6iK1EuIWTPXdY9sl3101kRWFSel/view?usp=sharing)

**Note**
These sheets have been switched to draft so that they don't lead to data sent for translation incorrectly. If testing, please release and then draft again. For the sake of this demonstration we can't just put the usual exclude_from_translation as true for everything, as we do in all other example sheets.

chrismclarke:
PR Checklist

## Description
Quite a large pr, relevant notes as follows:

### Breaking Changes
- New dependencies added will require `yarn install`

### Translation Folders
- PLH translations folder structures changed, feel free to delete everything except `from_translators` folder in `plh-data/translations` folder (image below). All other folders will be created during compilation

- Any new files received from translators should be put in the `from_translators` folder, named as `{filename}.{langCode}.json`. These will be compiled EVERY TIME the sync script is run (realised would cause too many git changes if only run on demand, as default scripts would overwrite compiled translation changes if ran standalone).

image

### Template Changes
- Once run, the standard data sync script populates template files with a `_translations` field (where specified) to provide a placeholder that this field should be translated, alongside language codes with current translation status (`TRUE` indicates translation currently available, `FALSE` means translation expected but doesn't currently exist - default fallback will be English value)

- Strings used in the app for lookup are now output to the `plh-data/data/translation_strings` folder (as if they were a flow_type), instead of integrated directly into templates to keep better separation and fewer future changes. 

### Python Scripts
A big sticking point of the current implementation is how strings for translations are processed outside of the app. Information such as fields to exclude from translation are hardcoded into these scripts, but required when trying to import translations back in (resulting in messy workarounds for local compilations scripts). 

Similarly the python scripts try to identify dynamic texts for data lists and global variables using a fixed/hardcoded list of known variables names - this is likely to be highly inconsistent as data changes and cumbersome to maintain (all lists and more importantly generic methods to identify dynamic variables already exist in the app scripts)

Ideally the scripts should be rewritten in JavaScript and implemented in this workspace (really wouldn't be a huge issue, the scripts are pretty basic and a much cleaner implementation is possible). Alongside the tidier implementation, one improvement I would suggest would be replacing inline dynamic texts with static placeholders to both avoid the need for translator notes and also enable translations to not easily be affected by variable name changes. I.e. 
Before
```
sourceText: "hello @local.some_var"
note: "The following should not be translated: @local.some_var
```
After
```
sourceText: "hello [REF1]"
refs: ["@local.some_var"]
```
(with translators told they never need to worry about refs)

## Known Issues
- some strings just have partial text extracted (e.g. combo_list ```value:some_value; text:some_text``` with no means to put back into the correct format. So individual methods have to be added to do runtime checks for possible translations within things like parameter lists. I've added these for combo_boxes but no other component types (will need future identification and work)

- Some fields don't actually show translated content change when translation button clicked (due to some individual components relying on methods called only once at start). I found this issue with the title component and fixed it, but would not be surprised if there are others, so worth watching out for. This will likely be a non-issue if the change language button is put somewhere away from complex templates (e.g. only available from link on the home screen or settings page)

## Future TODOs
- The translations system no longer needs `_translatedFields` references and so should be removed, but as they are currently used by the python conversion scripts will retain

## Git Issues

Closes #918

## Screenshots/Videos

https://user-images.githubusercontent.com/10515065/127581327-d382485e-c204-4a0b-85e9-5dbf172098a4.mp4


volloholic (approved):
I have looked through the code base and suggest this gets merged in as is. 
@chrismclarke we should follow up with a discussion as there are a couple of things which I would like us to revisit in the longer term.

PR #946, opened 2021-08-16
Feat/template startup

chrismclarke:
PR Checklist

## Note 1
Type definition updates to support feature already merged with #945 to support testing

## Note 2
The implementation is somewhat hacky, as currently the logic that processes and renders templates is coupled tightly with the template display component itself (because traditionally we would process templates in the order that they're rendered on the page, allowing deep nesting and more complex use cases).

So for now, to allow programmatic rendering we essentially create a dummy display component which is never rendered, and use that to process the logic. In the future we ideally would want to refactor to separate all the processing logic out from the display component, and call from either a display component or programatically.

## Description
Add support to programmatically parse template rows, so that all logic is processed (including set_variable, set_field etc.) statements, but no rows rendered. Templates can either be processed via `process_template` action, or automatically on app start via `process_on_start` content_list column

- Create [example_startup sheets](https://docs.google.com/spreadsheets/d/1ww3DZBygZH9iT_tIfyrV_-DrML6zaA9UKHF1M8DQifc/edit#gid=1745157248)
- Add support for `process_on_start` content_list column
- Add support for `process_template` action 

## Git Issues

Closes #

## Screenshots/Videos

_Startup Processing_
image

_Triggered Processing_
image

Example field initialised from startup.
Fields set in startup_1 have been overwritten by higher-priority fields set in startup_2 (see individual sheets for code)

image

### DIFF
--- a/src/app/shared/components/template/services/template-process.service.ts
+++ b/src/app/shared/components/template/services/template-process.service.ts
@@ -1,0 +1,79 @@
+import { Injectable } from "@angular/core";
+import { ActivatedRoute, Router } from "@angular/router";
+import { FlowTypes } from "packages/plh-data/model";
+import { SettingsService } from "src/app/pages/settings/settings.service";
+import { TEMPLATE } from "src/app/shared/services/data/data.service";
+import { TourService } from "src/app/shared/services/tour/tour.service";
+import { TemplateContainerComponent } from "../template-container.component";
+import { TemplateNavService } from "./template-nav.service";
+import { TemplateTranslateService } from "./template-translate.service";
+import { TemplateVariablesService } from "./template-variables.service";
+import { TemplateService } from "./template.service";
+
+/**
+ * The template process service is a slightly hacky wrapper around the template container component so that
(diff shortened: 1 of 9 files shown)


src/app/shared/components/template/services/template-process.service.ts
```diff
@@ -0,0 +1,79 @@
...
+    this.container.templateRowService.templateRowMap = {};
+    // process the template as if it were rendered
+    // TODO - should filter out template rows to only include those used programatically (e.g. set_variable, set_field etc.)
+    await this.container.templateRowService.processContainerTemplateRows();
+  }
+
+  private async initialiseStartupTemplates() {
+    const startupTemplates = TEMPLATE.filter((t) => t.process_on_start).sort(
+      (a, b) => a.process_on_start - b.process_on_start
+    );
```
volloholic, line 70: This looks like ascending order I thought we agreed descending? 
The ordering doesn't actually matter for now but this would be good to review together later.

volloholic (approved):
We may need to look at the ordering together but this is not important for now so I am merging and we will revisit later.

PR #964, opened 2021-08-24
Non-display items's Space Correction

zulkifelff:
PR Checklist

## Description
Due to the unset value of the 'set_variable' type in Display Group, empty spaces are been created.  We have introduced a check that will restrict the rendering of 'set_variable'.

## Git Issues

https://github.com/IDEMSInternational/parenting-app-ui/issues/925

### DIFF
--- a/src/app/shared/components/template/components/layout/display_group.ts
+++ b/src/app/shared/components/template/components/layout/display_group.ts
@@ -12,11 +12,13 @@ import { getNumberParamFromTemplateRow, getStringParamFromTemplateRow } from "..
   >
     <div [style.marginBottom.px]="-offset" class="offset" [ngSwitch]="type">
       <ng-container *ngSwitchCase="'default'">
-        <plh-template-component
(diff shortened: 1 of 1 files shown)


chrismclarke (changes requested):
So I looked more carefully at this issue, and realise there's 2 issues with the proposed solution:

1. There are multiple component type that could accidentally render a non-display row, including set_field, set_global, set_theme etc., and potential for more components in the future. So we need a solution that will lookup the existing list of components and handle display based on whether a display component exists or not (instead of specifying `set_variable` specifically.

2. There are also multiple display component types that could have the same problem, not just display groups (e.g. animated_section, group, accordion_section etc.). So we also need a solution at a higher level that can take into account any component that renders further components.

The existing codebase makes this slightly tricky to do, but I have a suggestion that I will open a PR for

chrismclarke:
Alternate PR opened: #971

@zulkifelff - please review the PR to see that it makes sense to you and fixes the problem as identified. If yes then we can merge that one and close this instead.

PR #995, opened 2021-08-31
Cyress tests for example_actions

saqlainrasheed:
PR Checklist

## Description

- Added cypress tests for example_actions which include 7 templates (example_go_to_1, example_go_to_2, example_go_to_3,example_emit,example_text,example_update_actions, example_pop_ups).

## Git Issues

Closes #


chrismclarke (approved):
As a base this is a good start for sure. I realised one major issue we have is that most tests will only pass if we have completed the first user workflow, and so will need to find a way of either disabling or skipping for tests. I will have a think about that and create either a new issue or a PR. 

We also will want to start think about how to automate more of the test creation (e.g. taking a single list of all templates to visit and a way that content authors can specify what should be seen in the test), but again I can create a follow-up issue for that

PR #1003, opened 2021-09-07
markdown syntax issue 

### LINKED ISSUE #960: [BUG] Markdown syntax is only recognised in text type
**Describe the bug**
Markdown is recognised and interpreted correctly in rows with text type, but does not yet work in other types. Ideally, every place in the app that displays text would interpret markdown syntax. Minimally, I'd like this to work in title / subtitle / button / tile_component / simple_checkbox / toggle_bar. 

**App version**
v0.11.2 (but also didn't work in earlier versions)

**To Reproduce**
See debug template.

**Debug Template**
[debug_markdown](https://docs.google.com/spreadsheets/d/14bVDjfCbh4_2NSMpgMuDZQ_IKUzl3FMQxnYu0Lgu6Uc/edit#gid=224996377
)

**Screenshots and videos**
image
image

**Additional context**
Beside using markdown, text formatting can also be specified using the parameter **style** in the parameter_list. This affects the entire text, whereas markdown enables highlighting part of the text. The style parameter and markdown should interact as they currently do for text type.

saqlainrasheed:
PR Checklist

## Breaking Changes
(tbc)

## Description

- Markdown rendering Issues are now resolved on the ```debug_markdown``` template.

## Git Issues

closes #960

## Screenshots/Videos

markdown

### DIFF
--- a/src/app/shared/components/template/components/button/button.component.html
+++ b/src/app/shared/components/template/components/button/button.component.html
@@ -3,8 +3,6 @@
   [class]="'full standard medium' + ' ' + style + ' ' + buttonAlign"
   [disabled]="disabled"
   (click)="triggerActions('click')"
-  *ngIf="innerHTML"
-  [innerHTML]="innerHTML | markdown"
 >
-  <span *ngIf="innerHTML" [class]="textAlign" [innerHTML]="innerHTML | markdown"> </span>
+  <span *ngIf="innerHTML" [class]="textAlign" [innerHTML]="_row.value | markdown"> </span>
 </ion-button>
--- a/src/app/shared/components/template/components/button/button.component.html
+++ b/src/app/shared/components/template/components/button/button.component.html
@@ -3,5 +3,8 @@
   [class]="'full standard medium' + ' ' + style + ' ' + buttonAlign"
   [disabled]="disabled"
   (click)="triggerActions('click')"
-  ><span *ngIf="innerHTML" [class]="textAlign" [innerHTML]="innerHTML"></span
-></ion-button>
+  *ngIf="innerHTML"
+  [innerHTML]="innerHTML | markdown"
+>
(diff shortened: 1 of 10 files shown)


chrismclarke:
As discussed on the call, we identified the `feature_formatting` template as important for testing changes to markdown implementation (as it was added to use direct html in buttons, which has to be implemented differently when adding markdown support). 

Comparing the current PR with master we can see a few visual differences that possibly should be resolved ahead of merge (e.g. padding added to components, button text align)

image

chrismclarke (changes requested):
As per comment above, further work needed to avoid too many visual changes to existing templates

chrismclarke:
@saqlainrasheed - please can you correct the description to link to the git issue. As mentioned previously, there is a specific list of words that can be used to link a pull request to an issue (full list here: https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests)

You have used the word `issue` which is not one of the keyword, I suggest you change it for `closes` (as recommended in the PR template). You also have left a space after the hashtag that breaks the reference, so you need to change `issue# 960` to `closes #960` (or equivalent).

chrismclarke (approved):
Visually I think this looks much more consistent now and so I expect it is ready to merge. I do still have some hesitation for knock-on effects however, and so I will recommend we hold off merging until after next release (and potentially wait until we can test against screenshots in #1004). 

But from the code side I think we can approve for now and revisit later if required. I'll add @EsmeeIDEMS to the review so that she is aware that this code could be merged if a high priority, but also my recommendation to wait if not.

EsmeeIDEMS (approved):
Functional test passed. Initially I noticed some erratic bugs with pop-ups not showing, but I could not reproduce this later.

PR #1009, opened 2021-09-08
add support for global default translations

chrismclarke:
PR Checklist

## Description
As discussed on the call, an issue exists where fields initialised with default values (e.g. `user_name_default: awesome parent`) are not updated when language changed because fields are only ever initialised once.

This PR provides a solution to the issue, by allowing `field_default` rows to point to `global_default` rows (actually no change required as global rows were always processed directly as strings and not evaluated), and adding `global_default` rows to the translation process with update on language change. 

In the user_name case, the field `@fields.user_name_default` is initialised with the value `@global.user_name_default`, so it will always show the updated global value unless the user has replaced with their own name. Even if the name the user replaced with exists in translations, it will not be changed on language change.

- Update global defaults to process translations
- Add bindings to update global defaults on language change
- Tidy code that initialises field and global default values
- Update example templates and clarify comments

## Git Issues

Closes #

## Screenshots/Videos

**[workshop_options](https://docs.google.com/spreadsheets/d/1FXmKOQh-7iqi5JAYmJEQeqa-7C2ISpY4yLghFUV5md0/edit#gid=2025264137) sets field and global defaults**

image

**Example update in [example_lang_template_1](https://docs.google.com/spreadsheets/d/1vv1DbyXEkOEHVEAMwjtQW0vS23DDUBK1oOw24moe64M/edit#gid=1595229908)**

https://user-images.githubusercontent.com/10515065/132480889-0aa086d5-d112-4c6a-a32d-455b0ee6ccc5.mp4

### DIFF
--- a/packages/plh-data/translations/from_translators/translated.es.spa.json
+++ b/packages/plh-data/translations/from_translators/translated.es.spa.json
@@ -208,5 +208,10 @@
     "SourceText": "sad",
     "text": "triste",
     "type": "template"
+  },
+  {
+    "SourceText": "awesome parent",
+    "text": "Padre asombroso",
+    "type": "template"
   }
 ]
--- a/src/app/shared/components/template/services/template-translate.service.ts
+++ b/src/app/shared/components/template/services/template-translate.service.ts
@@ -1,5 +1,6 @@
 import { Injectable } from "@angular/core";
 import { TRANSLATION_STRINGS } from "plh-data";
+import { BehaviorSubject } from "rxjs";
 import { LocalStorageService } from "src/app/shared/services/local-storage/local-storage.service";
 import { FlowTypes } from "../models";
 
@@ -15,6 +16,9 @@ const DEFAULT_LANGUAGE = "eng";
  * are already pre-populated in the column (so just a case to return the current language field)
  */
 export class TemplateTranslateService {
+  /** Provide an observable so services can subscribe and respond to language changes*/
+  app_language$ = new BehaviorSubject<string>(null);
+
   app_language: string;
   translation_strings = {};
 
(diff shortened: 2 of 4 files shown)


EsmeeIDEMS (commented):
I've done a functional test, this seems to work well in general. Only remaining issue:

- [ ] We need a syntax to set a field back to @global.something (without evaluating the global!). 

The action set_field : user_name : @global.default_user_name evaluates the global. I've tried using "@global.user_name_default" or '@global.user_name_default' but neither worked. 

See the template [example_lang_user_name](https://docs.google.com/spreadsheets/d/1vv1DbyXEkOEHVEAMwjtQW0vS23DDUBK1oOw24moe64M/edit#gid=2088861141)

Should this be a separate issue or do you want to include this in this PR?

chrismclarke:
I think we should create a separate issue as this could get a little bit complicated/unclear. I understand the issue, my concern comes from finding a way to distinguish:

- `@field` `@global` `@local` will always evaluate in a context where we show the value
- `@field` `@local` will always evaluate in a context where we set a value
- `@global` does not evaluate in context where we set a value

It's that combination of sometimes behaving the same and sometimes differently that could make for some ugly code implementation unless carefully thought out. My instinct would actually be more to add a way for the author to bypass evaluation, e.g. 

`set_field: some_field: "@global.some_value"` 

to behave differently to

`set_field: some_field: @global.some_value` 

So that the onus is on the author to provide an exception rather than have such subtle differences in the code. I guess the argument for subtle differences is that is how it works for field defaults, although again the reason for that is more the legacy issue of processing globals outside the regular templating system. 

So let's create an issue and tag for discussion. If merging this PR you can just reference in the new issue (e.g. see discussion in #1009)

volloholic (approved):

PR #1016, opened 2021-09-09
Fix/server configs

chrismclarke:
PR Checklist

## Breaking Changes
This changes processes used create new server deployments, automating and correcting steps that were performed manually on the current server installation. As such the existing server can likely be left as-is (not syncing latest update on remote server), until there are future updates requiring server upgrade (e.g. adding new API endpoints or updating database schema). In that case standard upgrade practice should be applied, e.g.:

- Schedule server downtime
- Backup existing server
- Pull latest changes
- Rebuild local docker API container
- Spin server back up
- Test

There should be no impact on the frontend code or the existing app

## Description

Whilst planning work to integrate matomo analytics into the app a few minor server issues have been identified and corrected:

- Database initialisation scripts (used in clean installs) had the wrong line ending (`CLRF` instead of `LF`) preventing the bash scripts from running, so corrected locally and should now be tracked via a `gitattributes` file to ensure correct ending.
- Some inconsistencies in environment variable naming corrected
- Documentation updated and improved
- Automated one of the manual steps for app db creation (had been planned, but not working due to init script issue)
- Issue fixed where database migration scripts were not being processed automatically (requiring manual call from command line). This was because the migration scripts look for definitions in `.ts` files, which exist locally but are compiled into `.js` files on the server (so file filter needed update). Additionally, the server creates type definition files with extension `.d.ts` which being processed (incorrectly), so are now filtered out. Better error logging added to identify such issues in the future.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/api/.env.example
+++ b/packages/api/.env.example
@@ -1,18 +1,18 @@
 WEBSOCKET_PORT=3000
 API_PORT=3000
-# if running on reverse proxy, e.g. /api/
-API_BASE_PATH=""
+# If running within docker containers this should be set to reverse proxy, e.g. /api/
+# If serving locally this should be left blank
+API_BASE_PATH="/api/"
 DB_HOST=localhost
 DB_PORT=5432
 
-# Admin DB (used to create client DB)
-DB_ADMIN_NAME=postgres
-DB_ADMIN_USER=postgres
-DB_ADMIN_PASSWORD=demo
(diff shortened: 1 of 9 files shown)


volloholic (approved):

volloholic:
There needs to be work on the production pipeline soon as well. In particular removing the login credentials.

PR #1021, opened 2021-09-13
Remove old home page

### LINKED ISSUE #965: Remove old homepage (/home)
**Details**
The homepage (/home) is redundant, it is currently hardcoded to embed the template /templates/home_screen. We need the following changes:
1. Update the app routing so that / and /home redirect to /template/home_screen
2. Update the app-menu and app-back buttons so that only the menu shows on the /template/home_screen page (like the current /home ) page
3. Remove all code from /feature/home

**Screenshots and videos**
image
image

saqlainrasheed:
PR Checklist

## Description

- Changed the old home page routing to the Home_screen template route.

## Git Issues

closes #965

### DIFF
--- a/src/app/shared/components/plh-main-header.ts
+++ b/src/app/shared/components/plh-main-header.ts
@@ -2,6 +2,7 @@ import { Location } from "@angular/common";
 import { Component, Input, NgZone, OnDestroy, OnInit } from "@angular/core";
 import { NavigationEnd, NavigationStart, Router } from "@angular/router";
 import { Subscription } from "rxjs";
+import { filter } from "rxjs/operators";
 
 @Component({
   selector: "plh-main-header",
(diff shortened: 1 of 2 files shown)


zulkifelff (approved):

EsmeeIDEMS (changes requested):
While doing a functional test I noticed that the left-top corner of the home screen still sometimes shows a hamburger menu (e.g. when reloading http://localhost:4200/template/home_screen), and sometimes shows a left chevron (e.g. when clicking ParentApp in the header), even though the hyperlink is the same. This should always show the hamburger menu. 

https://user-images.githubusercontent.com/74557272/135998122-d229d005-aeb5-45b1-bc88-af52290ab49d.mp4

src/app/shared/components/plh-main-header.ts
```diff
@@ -59,15 +61,24 @@ export class PLHMainHeaderComponent implements OnInit, OnDestroy {
...
-    // if wanting to access route params directly (not currently required)
-    const HOME_ROUTE = "/home";
-    // track if home page, allowing case where hosted from subdirectory (e.g. our pr preview system)
-    this.isHomePage = location.pathname.endsWith(HOME_ROUTE);
-    this.isSettingsPage = location.pathname.endsWith("/settings");
+    // if wanting to access route params directly (not currently required)4
+    const HOME_ROUTE = "/home_screen";
+
+    if (this.backHistory.length > 0) {
+      this.isHomePage = false;
```
chrismclarke, line 68: I'm not sure if this will work as intended. If a user goes `Page1 -> Page2 -> Page3 -> Page1` we still have a backHistory but are on the home page, so the check below will say homePage `false`. As well as clicking back buttons we have times when the user might be redirected back to the home page automatically, e.g. after they finish a workshop.

I think we should avoid trying to handle the full backHistory as it becomes quite complicated, and just keep a simple check for whether we are on the home page to show the hamburger menu and keep back button everywhere else

chrismclarke (changes requested):
I haven't tested fully, but I suspect there will be cases the behaviour won't work quite as intended. Could you please check my comment and see if changes are required are not?

chrismclarke:
This was implemented alternatively in #1112, closing here

PR #1027, opened 2021-09-16
rewrite tour introjs implementation

chrismclarke:
PR Checklist

## Breaking Changes
- Dependency updates require `yarn install`

## WiP TODO Items
- Hasn't been tested very extensively so would be good to try against issues previously found with old system

## Description
The current implementation of intro.js for the tour system was having issues and not clear how to resolve. I've re-written most methods and tried to improve handling of more of our custom implementation scenarios, namely:

- Improve overall flow and calculation of tooltips when page transitions required
- Remove repeated recalculation calls and provide smarter optimisation
- Add `_waitForElement` function to check if a component exists in the dom to avoid use of as many arbitrary timeouts
- Add check for duplicate target elements (in case of duplicated navigation stack pages), and attempt to resolve for currently visible
- Add bindings to support custom calculation of tooltip positioning
- Separate out code for most previously inlined functions (e.g. step generation and change callback)
- Improve example template
- Add better comments to functions and type definitions

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "firebase": "^8.6.8",
     "globalthis": "^1.0.2",
     "howler": "^2.2.1",
-    "intro.js": "^3.2.1",
+    "intro.js": "^4.2.2",
     "jasmine-ts": "^0.4.0",
     "karma-typescript": "^5.5.1",
     "lottie-web": "^5.7.11",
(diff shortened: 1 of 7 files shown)


EsmeeIDEMS (commented):
I've done some testing and came across the following: The tour does not seem to work well on scrollable pages - the underlying page does not scroll down, so the item that is being highlighted might not be visible on the page. See screenshots below from [example_lang_tour](https://docs.google.com/spreadsheets/d/1vv1DbyXEkOEHVEAMwjtQW0vS23DDUBK1oOw24moe64M/edit#gid=985180059), third and fourth steps. 

image
image

EsmeeIDEMS:
Also I observed issue #1036 which you may want to resolve in this PR.

PR #1028, opened 2021-09-16
add support for raw values

### LINKED ISSUE #1018: [FEATURE] Syntax to save a field to an unevaluated global
**What?**
We need a syntax to set a field (as an action) back to @global.something (without evaluating the global!).

**Why?**
The user name (similar for group name) should be set back to its default when the text_box is made blank. For translation purposes, this default has to be the unevaluated global _@global.user_name_default_ rather than its value _awesome parent_.  However, the action set_field : user_name : @global.user_name_default evaluates the global. I've tried using "@global.user_name_default" or '@global.user_name_default' but neither worked.

**How?**
[example_lang_user_name](https://docs.google.com/spreadsheets/d/1vv1DbyXEkOEHVEAMwjtQW0vS23DDUBK1oOw24moe64M/edit#gid=2088861141)

**Other details**
See discussion in #1009.

chrismclarke:
PR Checklist

## Breaking Changes
Assume none. In theory, any strings that use backticks already with dynamic fields e.g. `` `@fields.some_value` ``   would have previously been parsed and rendered between backticks, but no other backticks found in search on plh data so assume non-issue. 
Regular text can still use backticks if required.

## Description

- Add syntax so that dynamic fields wrapped in backticks,e.g. `` `@fields.some_value` `` are not parsed
- Add examples to [example_set_fields_in_action](https://docs.google.com/spreadsheets/d/1YRCwIrsPkTw8S6NRZPyRax0jE8jzHXKy7EoftTK-YXE/edit#gid=41221518)
- Minor improvements to parser regex comments

## Git Issues

Closes #1018

## Screenshots/Videos

[example_set_fields_in_action](https://docs.google.com/spreadsheets/d/1YRCwIrsPkTw8S6NRZPyRax0jE8jzHXKy7EoftTK-YXE/edit#gid=41221518)
image

image

### DIFF
--- a/packages/data-models/constants.ts
+++ b/packages/data-models/constants.ts
@@ -14,6 +14,7 @@ export const DYNAMIC_PREFIXES = [
   "campaign",
   "calc",
   "item",
+  "raw",
 ] as const;
 
 /**
--- a/src/app/shared/components/template/services/template-variables.service.ts
+++ b/src/app/shared/components/template/services/template-variables.service.ts
@@ -143,7 +143,7 @@ export class TemplateVariablesService {
       context.calcContext = calcContext;
 
       // If a raw evaluator exists for any part of expression, return full expression unparsed
-      // e.g. "Example field name syntax is `@field.my_name`"
+      // e.g. "Example syntax is `@field.my_name`" -> "Example syntax is @field.my_name"
       if (type === "raw") {
         return evaluator.fullExpression.replace(/`/gi, "");
       }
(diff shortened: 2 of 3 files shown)


volloholic:
I thought we had decided to implement this as set_field_raw or something rather than with backticks?

volloholic (approved):
I was surprised this was implemented like this rather than as agreed but I now understand why. If @EsmeeIDEMS is happy with this authoring then it looks good to go.

EsmeeIDEMS (approved):
I'm happy with this authoring. Functional test passed.

PR #1029, opened 2021-09-17
Add support for automated campaign scheduling and translation

chrismclarke:
PR Checklist

## Update 3 (2021-09-25)
Reverting to draft following issue identified for required updates to APIs. Likely the PR will be merged with another more substantial update

## Update 2 (2021-09-24)
I realised a core limitation in the current system where actions defined in a notification `click_action_list` would not be triggered for 2 main reasons:

1. We did not have bindings that could be used to notify services that notifications had been triggered
2. The capacitor API does not provide any events to respond to when notifications have been triggered while running on web
3. The capacitor API does not persist notifications scheduled on web

So to work around these a custom system has been implemented that tries to keep track of all notifications that have been scheduled in localstorage, and provide an observable event stream that services can use to respond to triggered notifications. This event stream stores a list of any notifications that have never previously been marked as sent (i.e. triggered either since the current app start OR between the current session start and the previous app session end) 

Also includes list of recently sent
image

## Update 1 2021-09-17)
- Added support for randomisation, so that if multiple satisfied rows have same priority one will be picked at random (new screenshot/video added at bottom)
- Fixed issue where the notification schedule defined in new format wasn't being processed correctly (required adding support to convert nest fields in data lists (e.g. `{time.hour:4}`) into json format (`{time:{hour:4}}`)

## Breaking Changes
There is one notification scheduled under the old system via `m_inactive_campaign` that will not be removed when new notifications are scheduled, and so will still trigger monthly inactive reminder one month after last set. This is because the old system never actually tracked which campaign had been used to set a notification (only the id of the notification itself), and was only ever rescheduled each app load for one month later.

## TODO (authors)
- [x]  The new `notification_schedule` sheet should be used to contain all timing information for campaigns. Currently this is merged with the `notification_schedule` column in individual campaign rows, and so duplicate content should be able to be removed from individual rows. 

NOTE - for now we still want to retain the column as a means to provide title/text for the notification (but will likely refactor later to own columns, e.g. notification.title and notification.text or similar).

- [x] Created sheet [debug_notification_schedule](https://docs.google.com/spreadsheets/d/1Jf38D4pkoBO0QIUcw1IFAqlEokrUzhOESi3hNlW0vZM/edit#gid=998700467) which should be moved to a proper campaign location and updated as required

- [ ] A few previously hardcoded strings have been identified (default values for notification title/text) and moved to a separate list of `APP_STRINGS` in the  `data-models/constants.ts` file. At some point in future we should plan better how these are used/modified/translated (although for now less important as all notifications have title/text provided and so defaults not used)

## Description
Previously scheduling of push notifications was hardcoded, just for the `m_inactive` campaign. The PR adds support for automated scheduling of any campaigns that have a defined schedule. Additionally it fixes the way campaigns are rescheduled (previously old notifications would not be removed), and adds support for translated content. Specifically:

- Add support for data lists with `notification_schedule` subtype used to apply standard scheduling across campaigns
- Refactor and update startup scripts to handle lookup of all campaigns with notification schedules from the campaign feature service, and automatically process any push notifications
- Update scheduling methods to include campaign_id of the campaign that scheduled a push notification, so that when re-evaluated any old notification can be removed
- Add support to translate notification title and text

## Git Issues

Closes #

## Screenshots/Videos

**Notifications now scheduled from multiple campaigns**
image

**[debug_notification_schedule](https://docs.google.com/spreadsheets/d/1Jf38D4pkoBO0QIUcw1IFAqlEokrUzhOESi3hNlW0vZM/edit#gid=998700467) example to provide auto-scheduling across campaigns and replace hardcoded timing content in campaign rows**
image

**Translated notifications rescheduled after language change (and reload)**
image

**Randomised debug notification from [debug_campaign_rows](https://docs.google.com/spreadsheets/d/1Jf38D4pkoBO0QIUcw1IFAqlEokrUzhOESi3hNlW0vZM/edit#gid=1344817192)**

https://user-images.githubusercontent.com/10515065/133786282-7037e58d-387f-4b60-bec9-fde9f0096a4d.mp4

image

### DIFF
--- a/packages/data-models/constants.ts
+++ b/packages/data-models/constants.ts
@@ -30,3 +30,12 @@ export const APP_FIELDS = {
   SERVER_SYNC_LATEST: `${FIELD_PREFIX}._server_sync_latest`,
   APP_LANGUAGE: `${FIELD_PREFIX}._app_language`,
 };
+
+/**
+ * Some specific strings are currently hardcoded into the app
+ * TODO - not all strings included, should add to when required
+ */
+export const APP_STRINGS = {
+  NOTIFICATION_DEFAULT_TITLE: "Notification",
+  NOTIFICATION_DEFAULT_TEXT: "You have a new message from PLH",
+};
--- a/packages/scripts/src/plh-data-convert/parsers/data_list/data_list.parser.ts
+++ b/packages/scripts/src/plh-data-convert/parsers/data_list/data_list.parser.ts
@@ -1,3 +1,4 @@
+import { setNestedProperty } from "../../../../../../src/app/shared/utils";
 import { FlowTypes } from "../../../../types";
 import { extractConditionList, parsePLHCollectionString } from "../../utils";
 import { DefaultParser } from "../default/default.parser";
@@ -14,6 +15,11 @@ export class DataListParser extends DefaultParser {
       }
       if (field.endsWith("notification_schedule")) {
         row[field] = parsePLHCollectionString(row[field]);
+      }
+      // assign nested structures, e.g {time.hour : 4} => {time: { hour: 4 }}
+      if (field.includes(".")) {
+        row = setNestedProperty(field, row[field], row);
+        delete row[field];
       }
     });
     return row;
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -79,7 +79,6 @@ export class AppComponent {
       }
       this.analyticsService.init();
       this.renderAppTemplates = true;
-      this.scheduleCampaignNotifications();
       this.scheduleReinitialisation();
     });
   }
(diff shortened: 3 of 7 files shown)


packages/data-models/constants.ts
```diff
@@ -30,3 +30,12 @@ export const APP_FIELDS = {
...
 };
+
+/**
+ * Some specific strings are currently hardcoded into the app
+ * TODO - not all strings included, should add to when required
+ */
+export const APP_STRINGS = {
+  NOTIFICATION_DEFAULT_TITLE: "Notification",
+  NOTIFICATION_DEFAULT_TEXT: "You have a new message from PLH",
+};
```
volloholic, line 41: @EsmeeIDEMS Can you define these with Anna it is important that even this text gets approval at least form the content authors

volloholic (approved):
Looks good pleased to have this in. @EsmeeIDEMS can you review in particular the default values

packages/data-models/constants.ts
EsmeeIDEMS: I've contacted Anna about this, but it might take a while before I hear back from her. I'm happy for this to be merged with the current defaults, I can change them afterwards.

EsmeeIDEMS (commented):
While doing a functional test, I noticed that the notification in the [campaign inactive_day](https://docs.google.com/spreadsheets/d/1RztJVUiuQZhXLN5lI0BUWcDqL9sBAz8_MbQwbWvfNb4/edit#gid=2010298881) does not disappear after it has been sent (not even after reloading the page). @chrismclarke Could you confirm that this is the expected behaviour?

https://user-images.githubusercontent.com/74557272/134008809-67a66207-e5df-4dbc-abe6-f245ce57f5af.mp4

PR #1069, opened 2021-11-05
fix parent point flicker

chrismclarke:
PR Checklist

## Description

After various other discussions and PRs (see slack thread), the cause of the screen flicker issue seems to be linked to the implementation of the parent point tiles (namely the celebration animation)

Previously every parent point renders a celebration animation component, and toggles a start play function when called. For whatever reason this proved problematic in recent versions of chrome (possibly because of too many animations on the page, how lottie implements non-started animations, the fact the animation is duplicated for each tile or any other combination of factors etc.)

This pr changes the celebration implementation to conditionally render the animation, only when it is triggered (to remove the duplicate animations from the dom). It appears to fix the screen flicker image when testing locally.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/components/points-item/points-item.component.ts
+++ b/src/app/shared/components/template/components/points-item/points-item.component.ts
@@ -11,7 +11,6 @@ import { TemplateBaseComponent } from "../base";
 import { FlowTypes, ITemplateRowProps } from "../../models";
 import { getBooleanParamFromTemplateRow, getStringParamFromTemplateRow } from "../../../../utils";
 import { AnimationOptions } from "ngx-lottie";
-import player from "lottie-web";
 import { getImageAssetPath } from "../../utils/template-utils";
 
 @Component({
@@ -26,7 +25,6 @@ export class TmplParentPointBoxComponent
 {
   @Input() template: FlowTypes.Template;
   @ViewChild("star", { static: false }) star: ElementRef;
-  @ViewChild("celebretionAnim", { static: false }) celebretionAnim: ElementRef;
   @ViewChild("item", { static: false }) item: ElementRef;
   icon_src: string | null;
   lottie_src: string | null;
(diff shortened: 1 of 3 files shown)


EsmeeIDEMS (changes requested):
Looks great, functional test passed. Somehow the "celebration stars" animation has moved behind the parent point tiles - would it be possible to move this back to the foreground? 
image

src/app/shared/components/template/components/points-item/points-item.component.ts
```diff
@@ -65,7 +64,7 @@ export class TmplParentPointBoxComponent
     this.animCelebrationOptions = {
       path: getImageAssetPath("/plh_lottie/habits/cascading_stars.json"),
       name: "celebration",
-      autoplay: false,
+      autoplay: true,
       loop: false,
       rendererSettings: {
         // svg scaling options: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/preserveAspectRatio
```
volloholic, line 70: @chrismclarke Presumably to solve Esmee's issue of the Stars being behind we just need to bring them to the front. I don't know exactly where but guess that it's just a question of inserting z-index: 999, or something in the right place. is that here?
```suggestion
        // svg scaling options: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/preserveAspectRatio
        z-index: 999,
```

volloholic (approved):
Looks generally good except as Esmee identified the stars relative position has changed.

src/app/shared/components/template/components/points-item/points-item.component.ts
chrismclarke: You're right - previously the code had a toggle where it added an extra class when the play started which updated the z-index. Now that the element only exists this is not necessary, but I hadn't made 'play' state z-index the default

Fixed with fb30e7caffb09abbf8d805211415b79d7de19c06

EsmeeIDEMS (approved):
Functional test passed.

PR #1077, opened 2021-11-12
Minor bug fix for template  toggle button

### LINKED ISSUE #1017: [BUG] Toggle bar needs refresh before displaying correctly
**Describe the bug**
The first time a toggle bar is rendered the icons are not displaying in the right place. A refresh resolves this. 

**App version**
v0.11.2

**To Reproduce**
Open [https://plh-teens-app1.web.app/template/feature_toggle_bar](https://plh-teens-app1.web.app/template/feature_toggle_bar) in a new window. The icons will display in the wrong place. Refresh. The icons now display in the correct place. 

**Debug Template**
[feature_toggle_bar](https://docs.google.com/spreadsheets/d/1ARbNRGDer5vj9qSpRMZFrMkYifGkH3TLtDVp72YbaqU/edit#gid=550703021)

**Screenshots and videos**
image

https://user-images.githubusercontent.com/74557272/132857764-5f84e891-747a-4cca-a769-13e61aa44c1b.mp4

thisishamza:
PR Checklist

- [x] - PR title descriptive (can be used in release  @notes)

## Description

This PR fixes the toggle button issue. Now, the toggle buttons appear in the right place and page does not need to be refreshed.

## Git Issues

Closes #1017 

## Screenshots/Videos

-Tested for Web
web_toggle_ss

-Tested for Android
mobile_ss

-Tested for iPhone
iphone_toggle_ss


EsmeeIDEMS (approved):
Functional test passed. @chrismclarke Can you review code? 

When switching to this branch I need to run `yarn install` before the app works correctly. I also noticed this error in the build: 
image

Not sure if either of these two things is problematic.

src/app/shared/components/template/components/toggle-bar/toggle-bar.ts
```diff
@@ -39,7 +39,7 @@ export class TmplToggleBarComponent
 
   constructor(private platform: Platform) {
     super();
-    this.mode = this.platform.is("android") ? "md" : "ios" || "ios";
+    this.mode = this.platform.is("ios") ? "md" : "android" || "android";
```
chrismclarke, line 42: I don't understand what this code is doing, we don't have ios deploy targets so that part is unecessary and mixing ternary (`? ... :`) with logical or (`||`) is a bit of an anti-pattern (both are trying to do the same thing of handling if-else scenarios. If it can't be made to work with one then just use an `if` statement for clarity

chrismclarke: https://stackoverflow.com/questions/42026158/precedence-logical-or-vs-ternary-operator

chrismclarke:
I think the fix provided wasn't ideal as it didn't explain the issue or improve overall code quality. The reason the issue exists is because the icons are overlayed on top of the base ionic components with hardcoded positions, but the ionic component sizing varies with ios/android builds (different default padding values)

The simple solution which I think was provided was to just ensure that we always use android padding values (although very ambigous from the code). I've made this a much clearer option now without the need for any conditional logic.

The second issue is layering lots of hardcoded css on top of the base component (what was previously done by other developers). I don't have the time to fully refactor the implementation, although I have much simplified the style sheets so that we don't have so much duplicate code or hardcoded values. I also replaced the background tick/cross images with standard characters as they typically appear more clearly and don't require as much overhead to render.

@EsmeeIDEMS - Let me know if looks reasonable to you and feel free to merge if so.

chrismclarke:
Updated screenshot
image

EsmeeIDEMS:
@chris I preferred the custom icons but since there's a reason not to have them this also works. I've replaced the "x" symbol by 
"✗" so that it is a cross-symbol similar to the tick "✓". 

image

EsmeeIDEMS (approved):
Functional test passed

PR #1085, opened 2021-11-13
fix server sync timestamp

chrismclarke:
PR Checklist

## Description

Previously the contact field that records last sync was outdated on the server because it only tracked the last saved value (which only is saved after a successful sync).

This PR changes the behaviour so that the time sync initiated is recorded, and if successful written locally so that server and local fields will match

## Git Issues

Closes #

## Screenshots/Videos

After clean start, initial sync now has a populated contact field for `_server_sync_latest`.

image


volloholic (approved):

EsmeeIDEMS:
Functional test passed on the released app v0.12.5.

PR #1098, opened 2021-11-25
Feat/notification schedules

chrismclarke:
PR Checklist

## Description
General improvements and refactoring to notifications system to separate out scheduling of campaigns from the main campaign rows. Specifically:

- Create `campaign_schedule` datalist subtype for storing scheduling info
- Add support for schedule-level activation/deactivation, fixed start/end dates, notification day of week and time
- Add contact field and debug toggle for opting in to debug notifications
- Remove support for inline scheduling information (may be added back in future as a means to override defaults)
- Improve campaign debug page
- Update example templates

## For Authors
- All the examples currently sit in [debug_campaigns](https://docs.google.com/spreadsheets/d/1Jf38D4pkoBO0QIUcw1IFAqlEokrUzhOESi3hNlW0vZM/edit#gid=1682205958). The schedule list has been marked as `draft` to avoid breaking type changes, but should be changed to `released` when testing branch locally and after merge

- There are still 2 non-debug schedules in the debug sheet (`inactive_week` and `inactive_month`). These could possibly be moved elsewhere. I've also disabled `inactive_day` as I think this might be temperamental (not tested properly)
There are also multiple additional sheets currently marked as draft which I think may no longer be relevant, but we may wish to review in case there's any useful bits of code in them to add to examples

- Start_date and end_date can be used to constrain date ranges that schedules will run. These should be stored in a date format in the sheets. By default excel might display in US format (mm/dd/yyyy), this can be changed from the sheet under `file -> settings -> locale`

- Custom conversion is required for the `start_date` and `end_date` columns. These are applied in the default parser for any column ending `_date`, so ideally no future sheets should use this convention (fine for field names etc., just not for entire column)

- Schedule.day_of_week can be used to specify a fixed day of the week (1-per-week). If left blank notifications will be scheduled for the next specified time (either same day or next day) at the time of scheduling (e.g. app open or after 15 min app refresh). There is currently no way to schedule other than once-per-day or once-per-week within the same schedule.

- All date and times will be relative to a user's own timezone. There is currently no supported way to send a notification at a universally fixed time.

## For Devs
- `start_date` and `end_date` are stored in google sheets via date format. When this is synced to the app it's converted into a datevalue (e.g. `40128`) which represents the number of days since January 1900. A script has been added to automatically convert this and any other columns ending `_date`, however one minor caveat is that if planning to support conversion directly from excel (i.e. not google sheets -> excel), excel on mac formats these differently as number of days since January 1904 (so could cause future potential issue).

- An extra `hack` function has been added which removes all existing notifications before rescheduling current. This is to allow clearing out of any notifications that may have been set under a different system (e.g. previous version with different campaign names). A better solution could be found for handling this in the future

## TODO (future dev)
- Debugger page is a bit messy, ideally more information should be inlined into each schedule dropdown instead of loading from notifications/bottom of page

- Add support for scheduling in batch so that multiple notifications for a campaign can be scheduled without need for reinitialisation in the app.

- Add support for more advanced scheduling options such as selecting from list at random, allow/avoid reschedule of same notification if sent (global option instead of relying on `notification_id.sent` fields), max_notifications to send in a single campaign etc.

- Efficiency improvements for scheduling/rescheduling processes

## Git Issues

Closes #

## Screenshots/Videos

https://user-images.githubusercontent.com/10515065/143499035-e6e99a0d-02a3-4db7-8239-53fa3546b5a0.mp4

[debug_campaign_schedules](https://docs.google.com/spreadsheets/d/1Jf38D4pkoBO0QIUcw1IFAqlEokrUzhOESi3hNlW0vZM/edit#gid=998700467)

[debug_campaign_rows]
image

image

### DIFF
--- a/packages/data-models/flowTypes.ts
+++ b/packages/data-models/flowTypes.ts
@@ -192,32 +192,29 @@ export namespace FlowTypes {
     habit_list: string[];
   }
 
-  export interface Campaign_listRow {
-    _id: string;
-    activation_condition_list: DataEvaluationCondition[];
-    deactivation_condition_list: DataEvaluationCondition[];
+  export interface Campaign_listRow extends RowWithActivationConditions {
+    id: string;
     campaign_list: string[]; // ids of campaigns where to run
     priority?: number; // higher numbers will be given more priority
-    notification_schedule?: NotificationSchedule;
-    _activated?: boolean; // all activation criteria satisfied
-    _deactivated?: boolean; // any deactivation criteria satisfied
-    _active?: boolean; // activated and not deactivated
-
     // additional fields for current data_list but not required
     click_action_list?: TemplateRowAction[];
     icon?: string;
     text?: string;
-
     // placeholder for any extra fields to be added
     [field: string]: any;
   }
-  export interface NotificationSchedule {
-    text?: string;
-    title?: string;
-    time?: { minute?: string; hour?: string }; // specified time for notification, e.g. 19:30
-    delay?: { days?: string; hours?: string; minutes?: string }; // delay until first notification, e.g. 7 day
-
-    _schedule_at?: Date; // calculated from above info
+  export interface Campaign_Schedule extends RowWithActivationConditions {
+    id: string;
+    /** specified time for notification, e.g. 19:30. Day indicates day of week, starting with 1 = monday */
+    time?: { minute?: string; hour?: string; day?: number };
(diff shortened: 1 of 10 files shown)


EsmeeIDEMS (approved):
Functional test passed. @volloholic Can you review code?

volloholic (approved):

PR #1103, opened 2021-12-02
Feat/notification fixes

### LINKED ISSUE #1101: [BUG] Default time for notifications is midnight
**Describe the bug**
Default time for notifications (when no time.hour is given) is currently midnight. A more reasonable default would be 12noon. 

**Debug Template**
[debug_campaigns_schedules](https://docs.google.com/spreadsheets/d/1Jf38D4pkoBO0QIUcw1IFAqlEokrUzhOESi3hNlW0vZM/edit#gid=998700467)

**Screenshots and videos**
image

chrismclarke:
PR Checklist

## Description

- Add default time (see screenshot below - can be changed by authors if required)
- Add `debug_december_early` example (6am notifications) to test if schedule correctly for next day
- Fix scheduling so that if schedule time already passed for the day (or week if day_of_week provided) selects next day (or week)

## Git Issues

Closes #1101

## Screenshots/Videos

- No time notification now triggers at 12pm
- Early notification now schedules for next day (time already passed for today)

image

Updated notification defaults (can be modified from `packages\data-models\constants.ts`
image

### DIFF
--- a/src/app/feature/campaign/campaign.service.ts
+++ b/src/app/feature/campaign/campaign.service.ts
@@ -1,6 +1,16 @@
 import { Injectable } from "@angular/core";
 import { addDays } from "@fullcalendar/angular";
-import { addHours, addMinutes, isBefore, setISODay } from "date-fns";
+import {
+  addHours,
+  addMinutes,
+  addWeeks,
+  endOfDay,
+  isAfter,
+  isBefore,
+  isFuture,
+  isPast,
+  setISODay,
+} from "date-fns";
 import { NOTIFICATION_DEFAULTS } from "packages/data-models/constants";
 import { Subscription } from "rxjs";
 import { TemplateTranslateService } from "src/app/shared/components/template/services/template-translate.service";
@@ -174,6 +184,8 @@ export class CampaignService {
       return;
     }
     const notification_schedule = this.evaluateCampaignNotification(schedule);
+    // may return null if schedule would be outside permitted timeframe, in which case do not schedule
+    if (!notification_schedule) return;
     let { title, text } = row;
     let { _schedule_at } = notification_schedule;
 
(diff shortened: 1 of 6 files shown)


EsmeeIDEMS (approved):
Functional test passed

src/app/feature/campaign/campaign.service.ts
```diff
@@ -306,16 +318,34 @@ export class CampaignService {
...
+    }
+    // Set notification time, add 1 day if time already passed
+    d.setHours(Number(time.hour));
+    d.setMinutes(Number(time.minute));
+    if (isPast(d)) {
+      d = addDays(d, 1);
+    }
+    // Schedule notification day of week, add 1 week if time already passed
     if (schedule.day_of_week) {
       d = setISODay(d, schedule.day_of_week);
```
volloholic, line 339: @chrismclarke This is the one bit I don't know if I fully understand can this go backwards? If so do we need to check against the start date again? If not why are we checking against the past again? 
Worst case scenario is that we may have a rare bug where it sends a reminder before the start date, which is not a serious problem at the moment and would be easy to catch so I will merge anyway.

volloholic (approved):

src/app/feature/campaign/campaign.service.ts
chrismclarke: I think everything should be forward moving. 

The start date is today (now), and we move it forward iff the specified start_date is in the future (we still want to schedule for the current day if we are in between start/end dates):
```
let d = new Date(); // (today)
if (schedule.start_date && isFuture(new Date(schedule.start_date))) {
      d = new Date(schedule.start_date); // (future start date)
    }
```
Then when it comes to setting time we check against the past because if it is currently 1pm and I'm scheduling for 7am, I could be scheduling for earlier in the same day (in which case we add a day to make it tomorrow at 7am).
```
    d.setHours(Number(time.hour));
    d.setMinutes(Number(time.minute));
    if (isPast(d)) {
      d = addDays(d, 1); // (set date one day later if the time scheduled is in past, i.e. earlier in same day)
    }
``` 
We do similarly for day_of_week, e.g. if it is currently Wednesday and we are setting for Monday it will default to Monday of current week (which is in the past), so we add on a week
```
if (schedule.end_date) {
      if (isAfter(d, endOfDay(new Date(schedule.end_date)))) {
        return null;
      }
    }
```
At least that's how things should be working, may be a case I missed somewhere...

PR #1131, opened 2021-12-16
refactor/core-styles

chrismclarke:
PR Checklist

## Breaking Changes
- Updated ionic components requires `yarn install`
- Various component refactoring will likely cause conflicts with outstanding style update PRs (to resolve post merge)

## Description

Previous style implementations had been heavily focused on specific plh screens/components, with highly mixed levels of code quality and organisation.

This PR tries to assert a more organised system of separation between core component styles and plh-specific modifications via a theme overrides folder, and starts the process of refactoring some of the core components. 

It is only the start of the process and further work will be required, however it should cover many of the largest/most-critical breaking changes to allow for smoother full migration over time.

**Main Changes**

- Reduce amount of nesting used to display all components. Previously an unnecessary number of containers/wrappers meant most components were nested a minimum of 3 and often 5-6 levels deep. Not only does this have performance implications, but it also makes styling harder to resolve the correct elements.

- Migrate all code (where identified) that addressed particular relative nest levels, and refactor either for updated nesting levels or to cleaner methods (where applicable). E.g. some components that would search for styles from `parent.parent.parent.parent.parent`

- Remove most shared style files as the system was not maintainable, and also led to many style conflicts. Alternative methods for sharing common styles were already in place, so attempted to migrate most code to alternative system.

- Implement cleaner way of applying styles to a particularly template/component, instead of using general class selectors that could apply anywhere (e.g. `.parent_point`), use attribute selectors that are more specific (e.g. `plh-tmpl-display-group .display-group-wrapper[data-param-style="parent_point"]`)

- Upgrade ionic components from 5 to 6. This introduces some breaking changes, particularly for datepicker and popup/modal elements.

- Various fixes applied to make screens in new system replicate what had been entirely hardcoded, particularly for templates, home_screen, w_self_care_stepper, parent_centre, weekly_workshops 

- Remove a significant amount of unnecessary/untidy CSS code. Of the overall file changes roughly 20% represent removed CSS (~200 lines) and replaced with a large number of additional comments (~100 lines)

## Review Notes

- Use output from visual test system as basis for core component changes. Most components will have small changes to things like spacing/padding, however it would be good to identify any screens where more significant changes are detected

- Breaking changes were introduced to systems that handle popups, so would be good to double-check where possible whether still looks correct from various contexts (e.g. start-up pages, combo box selects etc.)

- Check whether hidden disabled properties applied properly (could not find easy template to test)

## Author Notes
- Previously various variables had been intended to handle things like padding, but incorrectly implemented so the style changes were not included and then further built upon with custom styling. Now that the variable system has been fixed, many components will have subtle changes to things like padding, margins, alignment etc. as these take priority over some of the other hardcoded changes. Of particular note are buttons, which were meant to have min/max heights set from variables but were actually just using default values (now that they do read from variables most appear larger)

- Many templates had style authoring that did not fully match proposed mockups, and so developers then provided various custom overrides in the codebase to ensure they did. As a result there is still some inconsistency across the authored templates and outputs produced. At some point in the future additional work should be done to make clear what dev overrides are still in place, and provide better documentation/assistance to replace with correct authoring.

- Previously spacing between rows was handled in very mixed/messy ways. Some elements always left gaps above and below (e.g. images, buttons), others relied on authors manually supplying `margin` styles, some exposed custom style types (e.g. `margin-t-large` and others were just hardcoded at the template level to look correct. 
The new system removes some of the ambiguity by imposing a system that automatically adds spacing between every component, which can be manually removed via authoring. It is likely some of the existing templates might need minor tweaks to adjust spacing under new system.

## Dev Notes
General principles to follow for future code tidying and new component additions:
- Use individual scss files for components instead of inline/js styling (scss not compiled properly from ts files and can lead to inconsistent handling of things like variables)
- If using variables, maintain consistency in naming starting with style property (e.g. --padding-large instead of --large-padding)
- Avoid ambiguous names like style: large, prefer text-large, image-large etc.
- Use data attributes for defining styles based on parameters or rownames, ideally with support for chaining (e.g. `data-param-style~="parent_point"`)
- Use theme folder for all plh-specific code and keep component folders as clean as possible
- Provide strong type definitions for all custom styles
- Expect to remove lines of CSS more often than adding (codebase still quite messy and could be further improved).
- Use mixins over shared classes
- Avoid use of dynamic font sizing/scale-factors. Currently we do not have a good enough system of breakpoints implemented, so leave to authoring.

## Future TODOs

- Remove/replace instances of `setProperty` functions that set styles from javascript. These are generally confusing and not necessary
- Extract and document all known custom styles used in components to create a list of what can be removed and/or what requires further documentation. Could use vscode search regex with terms such as `type: "image"[\s\S\r]+?style: "`, or script methods

- Review margin-t-large (previously applied via `margin-bottom: -0.75em`
- feature_tile (Sort image borders)
- Tidy variable naming structures
- Finish migrating all component custom code to override styles folder
- Add styling override to build pipelines
- Create grid display primitive (include params like min/max content, column count etc.) and remove workarounds like `two_column`
- Improve display_group flex system to document and expose min/max content (to imply flex wrap)
- Convert existing docs to standalone pages and add example of ideal docs (e.g. use tables to list params, include full example, possible screenshot/demo)
- Add basic docs for best practices (e.g. negative margins, typography and global changes)
- Methods to auto-doc styles etc. (or maybe manually doc)
- Consider templating system refactor to remove containers altogether (now more possible with angular 12 change to dynamic dependency injection)
- Determine syntax for future shared/common styles (and how to implement, e.g if container level not display block), e.g. sizging (small/medium/large/tiny or 1/2/3/4). Consider naming, e.g. margin-t-regular/margin-top-regular/mt-regular
- Consider  using a utility-class/styled-system already documented? E.g. tailwind, bootstrap, (rebass equiv)
- Remove legacy button styles (e.g. make_me_smile, get_me_going etc.) and refactor rest of button component
- Finish moving all tmpl-components-common code
- Use a single `params` class property that is already extracted and available to all components

## Git Issues

Closes #

## Screenshots/Videos

Folder structure handling style code now separated out from core components
image

Home screen, style mostly preserved (and made more consistent)
image

General improvements to two_column image layouts
image

Language select also improved for larger (and smaller) screen sizes
image

Button changes now min-height values being read from variables
image


EsmeeIDEMS (changes requested):
I did a functional test, looks good in general. Pop-ups and the hidden column seem to work fine.

I noticed that the debug_mode does not work: for any template I tested, when the debug toggle is set to true nothing is displayed.
image

Also I noticed a number of display issues which we'll need to iterate on - I'll post these separately

EsmeeIDEMS:
Bugs: 
- [ ] In some dimensions the text "Parent Library" on the home_screen overlaps the tile border
image
- [ ] The [banner](https://docs.google.com/spreadsheets/d/1ZkAXSGDgCQrDMUbyVVgJrdMUTnuZOicvecP6qytA9sY/edit#gid=57567680) template shows an unwanted background (strangely enough it does work as expected inside the [workshop_activity](https://docs.google.com/spreadsheets/d/1ZkAXSGDgCQrDMUbyVVgJrdMUTnuZOicvecP6qytA9sY/edit#gid=1771109624) which, as far as I can tell, is authored in the same way...)
image
- [ ] Navigation buttons spacing when back-button is displayed
- [ ] Tiles inside Parent Library no longer appear side by side 
- [ ] Tiles inside Essential tools have no horizontal padding

chrismclarke:
Thanks @EsmeeIDEMS , really useful list to go through

> In some dimensions the text "Parent Library" on the home_screen overlaps the tile border
**Fixed** - due to some hardcoded authoring that I've overridden for now and we can remove once merged

> The banner template shows an unwanted background (strangely enough it does work as expected inside the workshop_activity which, as far as I can tell, is authored in the same way...)
**Fixed** - there was some messy hardcoded function that searched for background colours much higher up the page and adjusted accordingly (explaining the discrepancies). I never really understood what it did and had left it in, but removing now things appear more consistent - if we find any new knock-ons as a result I can re-implement in a more sensible way, although checking the feature_display_group template that has good examples they seem fine.

> Navigation buttons spacing when back-button is displayed
**Wont fix** - I noticed this also, the existing implementation was so confusing I found it impossible to recreate in a reasonable way. But I think can be resolved much more easily at the authoring level so suggest we convert to a standalone issue after merge (or remind me and I can quickly resolve). 

> Tiles inside Parent Library no longer appear side by side
This should just be down to screen size. I set a minimum content width of 200px as this seemed reasonable for the text + image, so you only get multiple around 480px width (200px + 200px +32px spacing between + 48px padding outside)

> Tiles inside Essential tools have no horizontal padding
I hadn't looked at this template, agree doesn't look very good. One of the problems with 2-column layout is when screens get smaller it still forces alongside even if natural fit wouldn't be great (old implementation had some custom logic to handle this)

EsmeeIDEMS:
Bugs:
- [ ] The dashed box display group should always be full width. Currently it seems that the width adapts to the content.
image
- [ ] Sometimes there is no vertical padding between the before-last component on the page and the nav buttons.
image

Can test both by navigating to any relaxing activity (e.g. template [activity_relax_1](https://docs.google.com/spreadsheets/d/1kW0e_ynEg40j8uWs81dKB-73fSVCu0ajeM8fnAqJYf0/edit#gid=1246218094).

PR #1143, opened 2021-12-21
Fix/svg icons

### LINKED ISSUE #1127: [BUG] Radio group changes svg images
**Describe the bug**
The radio group component occasionally interprets svg images and changes colours. 

This behaviour seems slightly erratic (e.g. the same component is displayed differently in debug_radio_group and workshop_options)

**App version**
v0.12.5 + some content edits

**Template Name**
[workshop_options_page](https://docs.google.com/spreadsheets/d/1tmtdyeSYcqGH6joqZiA-wPtH0QH6td8hA3qGp62AcHs/edit#gid=1204310080)

**Debug Template**
[debug_radio_group](https://docs.google.com/spreadsheets/d/1pSFFOv9B9x0oUSuSJDp6RNePJ1XSP6yVSjXA8jzBclQ/edit#gid=1595229908)

**Screenshots and videos**
image
image

**Additional context**
This seems to be a leftover of previous behaviour of the radio buttons, when they always interpreted images.

chrismclarke:
PR Checklist

## Description

The system in place for rendering svgs has a bug where when multiple svgs are on a page some of the styles from one can clash with another (repo issue: `https://github.com/czeckd/angular-svg-icon/issues/73`). We previously highlighted this issue and suggested some ways that content creators could author to avoid, however that is only a temporary fix (and one that's hard to catch or follow-up on)

Instead the PR removes the package entirely to render svgs within native image components. The main benefit of the previous system was the ability to provide custom overrides to svgs with classes, for example to recolour all the lines. However, as we have not implemented such a system yet and similar could be achieved with CSS filters assume fine to replace for now.

**Note** - while testing it is clear that the radio buttons have many other style issues that will need sorting in the future, but I've held off making any other changes in this PR to make it easier to check visual regression from the change (plus it's late!)

## Git Issues

Closes #1127

## Screenshots/Videos
image


EsmeeIDEMS (changes requested):
Certain images can no longer be found, and the navigation buttons have lost their in-between padding. 

image

chrismclarke:
> Certain images can no longer be found, and the navigation buttons have lost their in-between padding.
> 
Fixed, usually all images need an extra piece of code `| plhAsset` that automates loading images from the plh assets folder instead of the local app assets folder. The radio group was missing this for some images, which previously wasn't caught as the custom svg icon component had a fallback to run a second check in case missing. 

I also did a manual check of all other components to see if any other images were missing the plhAsset code and all others looked fine, although there is a bit of inconsistency because some components do use hardcoded images from local assets (e.g. timer icons) and others hardcoded plh assets (e.g. star icons), so that might be something to tidy up from the dev side in the future (i.e. have some hardcoded default icons that can be overridden)

volloholic (approved):

chrismclarke:
Reviewed on call, changes implemented, merging

PR #1157, opened 2022-01-06
Improve template service dependency handling

chrismclarke:
PR Checklist

## Description
Refactor template services to provide slightly cleaner and more scalable sharing of code both within the templating system and external modules

## Dev Notes
The templating system is broken down into various components and services tasked with different aspects of the system, such as processing templates, rendering rows or triggering actions. One of the downsides of this modularity is that sharing code between services can be tricky given the nature of cyclic dependencies (e.g. the template-processor service can't import from template-actions if template-actions also has a call to the template-processor, e.g. trigger re-process)

Another layer of complexity is added when we take into account that some services exist globally (single instance), whilst others are created at a per-template container level (because this vastly simplifies doing things like adding local context to services; it also provides a pathway to calling template services from outside the core templating system, such as programmatically processing a template that has not been rendered). 

Longer term the general architecture could do with a bit of a rethink (likely leveraging more event messaging/callback systems of communication and/or adding full separation between template rendering and processing logic), however some small steps are required short term to start enabling more required functionality (namely triggering actions from notifications using the existing template-action system)

This PR takes a small step to differentiate between services which are created globally, and services which are instantiated multiple times. It also changes how instantiated services call global services, previously relying on having the service specifically passed as an argument when instantiating (leading to very long and messy lists of args), to using a custom injector that pulls from the global scope instead.

image
_Distinction between global and per-template (instanced) - differences always existed in codebase, but made explicit with folder structure_ 

See inline comments for more explanation of the code.

## Review Notes
The code changes are only to with processing, and so should not result in any visual diffs except in case where a component is no longer rendered (unlikely). 

The main area to functional test would be logic that triggers/cascades through multiple levels of templates, such as steppers and popups.

## Git Issues

Closes #

### DIFF
--- a/src/app/shared/components/template/components/layout/layout.ts
+++ b/src/app/shared/components/template/components/layout/layout.ts
@@ -49,7 +49,9 @@ export class TemplateLayoutComponent implements ITemplateRowProps, OnInit {
   public scrollToTop() {
     let parent = this.parent;
     while (parent) {
-      parent.elRef.nativeElement.scrollTop = 0;
+      if (parent.elRef) {
+        parent.elRef.nativeElement.scrollTop = 0;
+      }
       parent = parent.parent;
     }
     const ionContentContainers = document.querySelectorAll("ion-content") || [];
--- a/src/app/shared/components/template/services/instance/template-action.service.ts
+++ b/src/app/shared/components/template/services/instance/template-action.service.ts
@@ -157,7 +157,10 @@ export class TemplateActionService extends TemplateInstanceService {
         return;
       case "process_template":
         // HACK - create an embedded template processor service instance to process template programatically
-        const templateToProcess = await this.templateService.getTemplateByName(args[0]);
+        const templateToProcess = await this.templateService.getTemplateByName(
+          args[0],
+          this.container?.row?.is_override_target
+        );
         const processor = new TemplateProcessService(this.injector);
         return processor.processTemplateWithoutRender(templateToProcess);
       case "emit":
--- a/src/app/shared/components/template/services/instance/template-action.service.ts
+++ b/src/app/shared/components/template/services/instance/template-action.service.ts
@@ -1,11 +1,17 @@
 import { BehaviorSubject } from "rxjs";
 import { takeWhile } from "rxjs/operators";
 import { FlowTypes } from "src/app/shared/model";
-import { TemplateContainerComponent } from "../template-container.component";
+import { TemplateContainerComponent } from "../../template-container.component";
 import { SettingsService } from "src/app/shared/services/settings.service";
-import { TemplateProcessService } from "./template-process.service";
+import { TemplateProcessService } from "../template-process.service";
 import { ServerService } from "src/app/shared/services/server/server.service";
 import { AnalyticsService } from "src/app/shared/services/analytics/analytics.service";
+import { Injector } from "@angular/core";
+import { TemplateInstanceService } from "./template-instance.service";
+import { TemplateNavService } from "../template-nav.service";
+import { TemplateService } from "../template.service";
+import { TourService } from "src/app/shared/services/tour/tour.service";
+import { TemplateTranslateService } from "../template-translate.service";
 
 /** Logging Toggle - rewrite default functions to enable or disable inline logs */
 let SHOW_DEBUG_LOGS = false;
@@ -17,16 +23,27 @@ let log_groupEnd = SHOW_DEBUG_LOGS ? console.groupEnd : () => null;
  *
  *
  */
-export class TemplateActionService {
+export class TemplateActionService extends TemplateInstanceService {
   private actionsQueue: FlowTypes.TemplateRowAction[] = [];
   private actionsQueueProcessing$ = new BehaviorSubject<boolean>(false);
 
-  constructor(
-    public container: TemplateContainerComponent,
-    private settingsService: SettingsService,
-    private serverService: ServerService,
-    private analyticsService: AnalyticsService
-  ) {}
+  private settingsService: SettingsService;
+  private serverService: ServerService;
+  private analyticsService: AnalyticsService;
+  private templateNavService: TemplateNavService;
+  private templateService: TemplateService;
+  private tourService: TourService;
+  private templateTranslateService: TemplateTranslateService;
+
+  constructor(public container: TemplateContainerComponent, injector: Injector) {
+    super(injector);
+    this.settingsService = this.getGlobalService(SettingsService);
+    this.serverService = this.getGlobalService(ServerService);
+    this.analyticsService = this.getGlobalService(AnalyticsService);
+    this.templateNavService = this.getGlobalService(TemplateNavService);
+    this.templateService = this.getGlobalService(TemplateService);
+    this.tourService = this.getGlobalService(TourService);
+  }
 
   /** Public method to add actions to processing queue and process */
   public async handleActions(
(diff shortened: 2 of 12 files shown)


src/app/shared/components/template/components/layout/layout.ts
```diff
@@ -49,7 +49,9 @@ export class TemplateLayoutComponent implements ITemplateRowProps, OnInit {
   public scrollToTop() {
     let parent = this.parent;
     while (parent) {
-      parent.elRef.nativeElement.scrollTop = 0;
+      if (parent.elRef) {
+        parent.elRef.nativeElement.scrollTop = 0;
+      }
       parent = parent.parent;
```
chrismclarke, line 55: When evaluating a template that is not rendered it will not have an element reference. This PR doesn't change that, but just adds a catch for code that was missing

src/app/shared/components/template/services/instance/template-action.service.ts
```diff
@@ -18,17 +24,29 @@ let log_groupEnd = SHOW_DEBUG_LOGS ? console.groupEnd : () => null;
...
+  constructor(public container: TemplateContainerComponent, injector: Injector) {
+    super(injector);
+    this.settingsService = this.getGlobalService(SettingsService);
+    this.serverService = this.getGlobalService(ServerService);
+    this.analyticsService = this.getGlobalService(AnalyticsService);
+    this.templateNavService = this.getGlobalService(TemplateNavService);
+    this.templateService = this.getGlobalService(TemplateService);
+    this.tourService = this.getGlobalService(TourService);
+    this.templateFieldService = this.getGlobalService(TemplateFieldService);
+  }
```
chrismclarke, line 49: This is the change to how services are shared to instantiated services (the template-actions service being one)
Instead of passing all services via args (what is in the `constructor` script), it instead defines placeholders which are then populated via a `getGlobalService` method

src/app/shared/components/template/services/instance/template-action.service.ts
```diff
@@ -40,7 +58,7 @@ export class TemplateActionService {
     const res = await this.processActionQueue();
     await this.handleActionsCallback([...unhandledActions], res);
     if (!this.container.parent) {
-      await this.container.templateNavService.handleNavActionsFromChild(actions, this.container);
+      await this.templateNavService.handleNavActionsFromChild(actions, this.container);
     }
```
chrismclarke, line 62: this replaces slightly awkward script that would access some global by calling from the parent component (now instantiated correctly above)

src/app/shared/components/template/services/instance/template-action.service.ts
```diff
@@ -143,23 +157,8 @@ export class TemplateActionService {
...
-          this.container.tourService,
-          this.container.router,
-          this.container.route,
-          this.container.templateNavService,
-          this.container.settingsService,
-          this.container.serverService,
-          this.analyticsService
-        );
+        const templateToProcess = await this.templateService.getTemplateByName(args[0]);
+        const processor = new TemplateProcessService(this.injector);
```
chrismclarke, line 161: this is an example of now only having to pass one arg to create a standalone template processor instead of 11 imported services

src/app/shared/components/template/services/instance/template-instance.service.ts
```diff
@@ -0,0 +1,10 @@
+import { Injectable, Injector, ProviderToken } from "@angular/core";
+
+@Injectable({ providedIn: "root" })
+export class TemplateInstanceService {
+  constructor(public injector: Injector) {}
+
+  getGlobalService<T>(token: ProviderToken<T>) {
+    return this.injector.get(token);
+  }
+}
```
chrismclarke, line 10: This is the common method used by instance services. The `injector` provides a link to angular's system for handling dependency injection, and is used to get reference to existing global services

src/app/shared/components/template/services/instance/template-process.service.ts
```diff
@@ -21,38 +14,19 @@ import { TemplateFieldService } from "./template-field.service";
...
-    serverService: ServerService,
-    analyticsService: AnalyticsService
-  ) {
+  templateService: TemplateService;
+  templateNavService: TemplateNavService;
+  constructor(injector: Injector) {
+    super(injector);
+    this.templateService = this.getGlobalService(TemplateService);
+    this.templateNavService = this.getGlobalService(TemplateNavService);
     // Create mock template container component
```
chrismclarke, line 25: Further examples of the much-reduced amount of boilerplate required to create template processors and template containers

src/app/shared/components/template/template-container.component.ts
```diff
@@ -224,13 +206,15 @@ export class TemplateContainerComponent implements OnInit, OnDestroy, ITemplateC
...
   /** Query params are used to track state across navigation events */
   private subscribeToQueryParamChanges() {
-    this.route.queryParams
-      .pipe(takeUntil(this.componentDestroyed$))
-      .subscribe(async (params: any) => {
-        this.debugMode = params.debugMode ? true : false;
-        // allow templateNavService to process actions based on query param change
-        await this.templateNavService.handleQueryParamChange(params, this);
-      });
+    if (this.route) {
```
chrismclarke, line 209: another catch added for cases where we are processing a template in standalone mode - we do not have access nor want to subscribe to the changes to the currently active url ( `route`)

src/app/shared/components/template/services/instance/template-action.service.ts
volloholic: Why is the row (this.container.row) no longer needed on line 160?

volloholic (approved):

src/app/shared/components/template/services/instance/template-action.service.ts
chrismclarke: Really good catch, we had approved but not merged the PR for the update to the template override system which caused a conflict and I must have accepted the wrong version. The reason we pass the row is to identify if the current template is already an override (and so not to re-override and result in infinite loops). 

I've tidied the code now to more explicitly accept the `is_override_target` arg instead of the general row, and included here. I _think_ the code would still have been fine as we are processing a template that is not actually rendered on the page (the workaround action that we created to handle processing outside of the templating system), and so there is no container and hence no risk of infinite loops (we're not rendering rows, instead just things more like processing startup scripts and calc sheets) - but still better to be explicit.

EsmeeIDEMS (changes requested):
Bug with multiple actions: 
- Open http://localhost:4200/template/parent_centre_relaxes
- Click any button, e.g. the top one "Quick pause" to launch activity_relax_1
- Navigate through the activity by clicking next 3 times gets you back to the list.
- Click next, open and close the pop-up on the second screen by clicking the "read and relax" button, then click next 2 times. In this case the last next button does not take you back to the list.

https://user-images.githubusercontent.com/74557272/148773485-0730d217-463f-4ee5-bf29-5f5b4cb1369e.mp4

chrismclarke:
Thanks for catching this @EsmeeIDEMS. I was a bit confused at first as to why the code changes I made meant we lost that functionality, but then think I managed to figure out an issue with how the old system would technically allow multiple popups of the same content and know to close all, whereas the new system is only ever expecting to close one. 

The change to fix this is a bit of a sticking plaster still, and given how much else is already in this PR (and it is preventing review/merge of a couple others) I'm going to move ahead with merging here and opening a new PR to fix.

Issue #1129, opened 2021-12-16
[FEATURE] Push notification buttons and images


EsmeeIDEMS:
**What?**
- [ ] (essential) Push notifications should be able to take you to a particular template in the app, possibly by having a button. 
- [ ] (desired) Push notifications should be able to have an image. 

**Why?**
Rather than just having a couple of notifications for inactive users, the content team now wants to expand the use of notifications to all users.

PR #1158, opened 2022-01-06
Feat/notification actions

chrismclarke:
PR Checklist

## Description
Previously, notifications only supported actions in a limited way via a hardcoded copy of just the `set_field` action. This pr improves greatly on the previous system by enabling notifications to call methods from the templating system, and hence provide support for _most_ of the same actions as the templating system (previously only `set_field`). See further notes below.

## Author Notes
Usually we use the actions system alongside templates, allowing actions like `set_local` which updates a variable local to the template. As these actions are performed outside the context of any specific templates, only those actions which are not template-dependent will work (e.g. `force_reprocess` and `set_local` will not work, but things like `go_to`, `pop_up` and `set_field`) all should, as well as including dynamic variables for parsing in the args (e.g. @data.some_value).

As a result, the previous method for carrying out complex lookups by using intermediate local variables will not work, and so if required more work will be needed in the future to enable better support (e.g. @data.@data.some_value.another_value).

One implementation detail of note, is that technically it is not only the actions for the tapped notification that will be triggered, but also any action for notifications that had previously been dismissed. This could mean in the case of multiple notifications sent and still active, that there could be multiple navigation events in quick succession. This handling should probably be reviewed in the future, but for now authors should aim to not have too many parallel notifications with nav/popup events

## Review Notes
It might be good to also do some testing on device to see how things interact when doing things like tapping notifications when app closed.

## Dev Notes
This implementation provides a novel example of using the services designed for use within the templating context from the outside. Previously this could have caused cyclic dependency issue (notification module would not be able to call template actions if template actions contained any references to the notification module). This however can now be supported by decoupling various parts of the templating system in #1157 and providing a better framework for importing one module into another

Longer-term more work should be done to better separate out what is available both in and out of template contexts, and sharing more services outside of the templating system, but for now, this provides a handy workaround that could also be applied elsewhere if required.

## Git Issues

Link to #1129 (closes essential part)

## Screenshots/Videos
https://user-images.githubusercontent.com/10515065/148328465-1f8b21cb-32ea-4f86-8ca6-33530af757a1.mp4


src/app/feature/campaign/campaign.service.ts
```diff
@@ -97,21 +98,33 @@ export class CampaignService {
...
-  public triggerRowActions(row: FlowTypes.Campaign_listRow) {
+  public async triggerRowActions(row: FlowTypes.Campaign_listRow) {
     if (row.click_action_list) {
+      // make a dynamic call to TemplateVariablesService as it also has handling
+      // for @campaigns which would otherwise result in cyclic dependency in constructor
+      const templateVariablesService = this.injector.get(TemplateVariablesService);
+
+      // Process any dynamic variables that might be present in args
+      const parsedActions = [];
       for (const action of row.click_action_list) {
```
chrismclarke, line 111: previous implementation hardcoded only set_field action. New implementation evaluates text for any dynamic conditions before executing within the template-action service

src/app/shared/components/template/services/instance/template-action.service.ts
```diff
@@ -57,7 +57,7 @@ export class TemplateActionService extends TemplateInstanceService {
     unhandledActions.forEach((action) => this.actionsQueue.push({ ...action, _triggeredBy }));
     const res = await this.processActionQueue();
     await this.handleActionsCallback([...unhandledActions], res);
-    if (!this.container.parent) {
+    if (!this.container?.parent) {
       await this.templateNavService.handleNavActionsFromChild(actions, this.container);
```
chrismclarke, line 61: Various catches (conditional chain `?`) are added in case actions are triggered from outside a template, where things like a container may not exist

src/app/shared/components/template/services/instance/template-action.service.ts
```diff
@@ -37,7 +37,7 @@ export class TemplateActionService extends TemplateInstanceService {
   private templateTranslateService: TemplateTranslateService;
   private templateFieldService: TemplateFieldService;
 
-  constructor(public container: TemplateContainerComponent, injector: Injector) {
+  constructor(injector: Injector, public container?: TemplateContainerComponent) {
     super(injector);
```
chrismclarke, line 41: constructor orders swapped as we always need to pass an injector (which provides access to global services), but only sometimes a container. Various code snippets above and below to handle swap

EsmeeIDEMS (approved):
Are notification actions performed at the moment a notification is scheduled, or at the moment a notification is tapped? I would prefer the second but it seems to be the first. 

I will properly test on a phone after merge and release.

chrismclarke:
> Are notification actions performed at the moment a notification is scheduled, or at the moment a notification is tapped? I would prefer the second but it seems to be the first.

Neither, in order for an action to be triggered the app has to be open, so therefore do the following to ensure actions are still performed even if the notification is dimissed:

1. A single notifications is scheduled in advance for specific campaigns, and rescheduled as required (re-evaluated on a timer while open)
2. Assume while app is closed notification triggered and user receives a push notification.
case a. User dismisses notification. Continue to (3) next time user opens app
case b. User taps notification, is brought into app.
3. App checks for any notifications that have been sent since last app was opened/re-evaluated. There may be multiple in case where user has been dismissing notifications received.
4. Process the action list for each sent notification (may be multiple)

The upside to this is we get consistency with things like marking notifications as `sent`. Even if dismissed the action is triggered.
The downside to this is we get slightly unpredicatble behaviour if the user opens the app following multiple dismissed notifications. The action lists could have conflicting actions (e.g. multiple go_to actions), which will all be executed in turn. This could result in lots of navigation in quick-succession, although I'm yet to discover the impact this has on user (whether noticeable or not).

In the future I suggest we fine-tune the action handlers to distinguish between notification `tapped` and `dismissed` to trigger different actions accordingly, but for now we just have the single `click` action list which is applied in both cases. I expect we will only want to trigger certain actions (e.g. go_to) in response to tapped interactions, but keep the likes of set_field even if dismissed.

(if agreed would be good to create an issue referring to this thread)

volloholic (approved):

EsmeeIDEMS:
@chrismclarke It might make sense to merge this PR and relate any further improvements to issue #1178.

PR #1161, opened 2022-01-06
Feat/feedback module frontend

chrismclarke:
PR Checklist

## Description
The feedback module aims to provide a way for the app to send contextualised feedback or general information to the server. 
The frontend consists of a few main parts

**Context Menu Module**
Context menus are small popups that can be dynamically generated in response to specific user actions such as a right-click, long-tap or text highlight, replacing default behaviour. A generalised system has been put in place so that any module can listen to these events and provide their own custom options.

**Feedback Module**
This builds on top of the context menu to provide links to trigger feedback events, as well general bindings to be used as part of the templating system. 

**Templates as forms**
An early proposal was created to create a `form` component type for managing linked data. A component was created and exists already in the templating system, however it is incomplete and doesn't support any real data management. Instead of using this component I've enabled it so that any template can be used as a form, in the context that it will be shown as a popup window that passes back values when the `emit:completed` action is triggered.

## Author Notes
1) The form implementation is somewhat limited, in that right now only a single value can be passed, provided as an additional argument from the emit statement, e.g. `emit:completed:@local.some_value`
In order to emit more complex data it will first need to be merged into an intermediate value, e.g. `@local.form_data` which can be assigned to `@local.feedback_item_1; @local.feedback_item_2;` and passed as a string. 

It is intended to improve on this by reimplemented the `form` component to automatically track and pass all nested child rows as a single json object, however that is beyond the scope of this PR.

2) I've created some default templates to act as placeholders in the feedback module [gsheet link](https://docs.google.com/spreadsheets/d/1IsjWr7NcPspsUWstdFKzTRD6iKmNxCutipQ5a0vvLsU/edit#gid=1060209527). These can be changed/replace, and linked to the context system in the data-models `packages\data-models\constants.ts`. Further notes are in the documentation (included as screenshot below)

image

## Review Notes
1) Content requires (breaking) changes to data-models and so all kept as draft for now. I've synced the changes to another temporary PR branch to allow preview in #1165

2) Some minor changes were made to full screen popups to allow passing more customisation options. These should be reviewed where used elsewhere (believe only the startup templates which are triggered as fullscreen popups)

3) In order to allow emit statements to pass generic data, a few minor modifications were made to remove some hardcoded actions that assumed any additional args after `emit:[arg1];[arg2]` should be treated as part of the navigation system (originally so that `emit:completed:template_1` could be used to emit completed for one template from another). 

This functionality doesn't appear to be used and is also broken anyway as we now have use cases that use multiple emit arguments not related to the nav system. A quick search for all `emit` statements with multiple `:` (using a regex search for `emit:(.)*:`) gave only language operations (not completed/uncompleted)
image

But this could still be reviewed/checked for knock-ons, particularly given that all emit statement handlers have been updated to receive a JSON object including name + possible data instead of previously a string that was only the name.

## Dev Notes
The reason for separation between feedback module and context menu module is 2-fold. Firstly, because the the process of intercepting click actions has knock-ons to the entire app it makes sense the core code for this is extracted into the core modules. Secondly, it means that the feature module does not have to interact with the main app components, and can remain fully contained for inclusion/exclusion in other builds.

## Next TODOs
- local sync of feedback data
- server sync of feedback data
- server triggers to enable notification of received feedback data
- long-press implementation (earlier attempt does not work well enough so removed)

## Future TODOs
- Track and display list of submitted feedback
- Implement some means to link all data in a template into a single variable for passing (likely linked to planned work on ODK integration).

## Git Issues

Closes #
Follows from some early work done in #826

## Screenshots/Videos
**Debug page demo - page available at `/feedback` (no menu button currently)**

https://user-images.githubusercontent.com/10515065/149060754-95144991-96cf-4968-9570-95ea56da3f26.mp4

**Example feedback metadata**
image

**Feedback Docs**
feedback-docs

**Example text-select sheet**
https://docs.google.com/spreadsheets/d/1IsjWr7NcPspsUWstdFKzTRD6iKmNxCutipQ5a0vvLsU/edit#gid=700972401
image


src/app/app.component.ts
```diff
@@ -116,7 +117,9 @@ export class AppComponent {
     for (const initAction of APP_INITIALISATION_DEFAULTS.app_first_launch_actions) {
       switch (initAction.type) {
         case "template_popup":
-          await this.templateService.runStandaloneTemplate(initAction.value);
+          await this.templateService.runStandaloneTemplate(initAction.value, {
+            showCloseButton: false,
+          });
```
chrismclarke, line 122: minor tweaks made in future section to allow passing additional options to standalone templates (such as whether to include a close button or not)

src/app/shared/components/template/components/layout/popup.ts
```diff
@@ -83,25 +88,19 @@ import { TemplateContainerComponent } from "../../template-container.component";
...
  * the template directly as a regular template-container element
  */
-export class TemplatePopupComponent implements ITemplateContainerProps, OnInit {
-  @Input() name: string;
-  @Input() templatename: string;
-  @Input() parent?: TemplateContainerComponent;
-  @Input() row?: FlowTypes.TemplateRow;
-  /** Standalone mode shows in fullscreen and dismisses on completed/uncompleted emit events */
-  @Input() standalone?: boolean;
+export class TemplatePopupComponent {
```
chrismclarke, line 91: The fullscreen popup component has been both simplified and extended, so that instead of multiple configurable inputs it receives just one, but that is a json object with more options.

src/app/shared/components/template/components/layout/popup.ts
```diff
@@ -83,25 +88,19 @@ import { TemplateContainerComponent } from "../../template-container.component";
...
-  ngOnInit() {}
-
   /**
    * When templates emit completed/uncompleted value from standalone popup close the popup
    * NOTE - we do not want to respond to non-standalone templates as this is done through template nav-actions
    * */
-  handleEmittedValue(value: string) {
-    if (this.standalone) {
-      if (["completed", "uncompleted"].includes(value)) {
+  handleEmittedValue(value: { emit_value: string; emit_data: any }) {
```
chrismclarke, line 100: previously standalone popups only emitted their emit_value (e.g. completed/uncompleted) on dismiss. To support the feedback/form implementation support added to also include optional data on emit (such as form data)

src/app/shared/components/template/services/instance/template-action.service.ts
```diff
@@ -164,17 +166,9 @@ export class TemplateActionService extends TemplateInstanceService {
         const processor = new TemplateProcessService(this.injector);
         return processor.processTemplateWithoutRender(templateToProcess);
       case "emit":
-        const [emit_value, emit_from] = args;
```
chrismclarke, line 167: this was the previous incorrect implementation, that assumed all emit statements were of the format `emit:[emit_value]:[emit_from]`. Instead we will assume a more general `[emit_value]:[emit_data]` where data could represent any number of things depending on context

src/app/shared/modules/context-menu/context-menu.module.ts
```diff
@@ -0,0 +1,21 @@
...
+@NgModule({
+  imports: [CommonModule, FormsModule, IonicModule],
+  declarations: [ContextMenuComponent],
+})
+/**
+ * The contextMenuModule handles tracking events such as right-click and text selection that can
+ * be used to trigger a custom context menu, allowing for registered actions to be made available.
+ * This should be imported into main app.module to allow tracking across the app
+ */
+export class ContextMenuModule {
```
chrismclarke, line 18: Most of the logic for handling context menus is in the `contextMenuService`, however we are combining this into a core `module` as it also includes a default display component (used to render the menu buttons). This is the first time we have specified a core `module` as opposed to `service`, however something I think likely we will continue to do in the future where relevant (previously things have been separated to have shared components and shared services that refer to each other but not collated into a module)

src/app/shared/modules/context-menu/context-menu.service.ts
```diff
@@ -0,0 +1,130 @@
...
+import { Observable } from "rxjs";
+
+@Injectable({ providedIn: "root" })
+/**
+ * The context menu service tracks interaction with right-click and text-selection actions
+ * and presents a custom context menu where required
+ * TODO - add mobile long-press (and test on mobile). E.g. via getstureController, hammerjs or touch start/end events
+ * https://ionicframework.com/docs/utilities/gestures#overview
+ * https://levelup.gitconnected.com/making-hammerjs-work-with-angular-9-81d289159320
+ */
```
chrismclarke, line 22: This is the service that handles listening for events and triggering actions accordingly. The most interesting part is the definition of handlers (the rest is mostly just organisation of subscription services)

src/app/shared/modules/context-menu/context-menu.service.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  /**
+   * Listen to right click (contextmenu) and text select (selectionchange)
+   * events to trigger event handlers
+   **/
+  private addEventListeners() {
+    const eventListeners: {
+      [context in IContextMenuType]: Observable<{ ev: PointerEvent; data?: any }>;
+    } = {
+      // Long-press (TODO)
```
chrismclarke, line 96: long-press requires implementation

src/app/shared/modules/context-menu/context-menu.service.ts
```diff
@@ -0,0 +1,130 @@
...
+      longPress: fromEvent(document, "TODO").pipe(map((ev: PointerEvent) => ({ ev }))),
+      // Right-click
+      rightClick: fromEvent(document, "contextmenu").pipe(map((ev: PointerEvent) => ({ ev }))),
+      // Text select
+      // Events continously triggered so include additional mapping to detect start/end of selection
```
chrismclarke, line 101: text-select is an interesting/slightly more challenging situation as events are constantly triggered every time any character is changed (instead of waiting to the end of selection). This is handled using various utility methods that swap from listening to start of text selection to waiting for mouse-up event, and then retrieving selected text

src/app/shared/services/event/event.service.ts
```diff
@@ -18,12 +18,14 @@ export class EventService {
     DB: new Subject<IEvent>(),
     FLOW: new Subject<IEvent>(),
     LOCAL_NOTIFICATION: new Subject<IEvent>(),
```
chrismclarke, line 20: The feedback and template actions system have the usual issue where both want to reference each other (e.g. feedback module triggers actions and actions can trigger feedback module). Recently a nice workaround was found via dynamic imports, although this time round we instead are using the event emitting system to a) see if it works any more nicely and b) provide easier access to the db system which is already hooked up to events

volloholic (approved):

PR #1164, opened 2022-01-12
Fix/popup nav

chrismclarke:
PR Checklist

## Description
Improvements to templating system meant that a single template-nav service is created instead of multiple, which has the knock-on of not correctly dismissing all popups in the case where multiple popups of the same template are created.

This PR adds a tracking variable to the template-nav service to check if a popup already exists before recreating, and just makes visible if that is the case (by default open popups are hidden when navigating to new pages from the popup so that we can restore the popup to pre-nav state when returning to the original page, or reaching it anew via navigation from a->b->c->a.

## Git Issues

Closes #1157

## Screenshots/Videos

https://user-images.githubusercontent.com/10515065/149046609-7842c955-462e-4a8f-ad9a-5fb03288d26b.mp4

### DIFF
--- a/src/app/shared/components/template/services/template-nav.service.ts
+++ b/src/app/shared/components/template/services/template-nav.service.ts
@@ -27,6 +27,15 @@ export class TemplateNavService {
     private route: ActivatedRoute
   ) {}
 
+  /**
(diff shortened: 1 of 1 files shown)


EsmeeIDEMS (approved):
Functional test passed.

volloholic (approved):

PR #1167, opened 2022-01-13
Feat/feedback backend

chrismclarke:
PR Checklist

## Important
- Follows on from from #1161 and should only be merged after (will require re-merge to show correct diffs just for backend changes)
- Requires server updates and so cannot be tested locally unless running debug server (plan to demo on call)

## Description
Adds backend infrastructure and bindings to support recording feedback to local DB and then syncing to server

## Review Notes
- App version bumped to handle db updates (v0.14.1). If previewing locally will not be able to revert to older versions without cache clear (recommended preview online)

## Dev Notes
1) We currently have no standard way in the app to sync from local to server DBs. A proposal was previously made to use dexie-sync protocol however this only enables complete 2-way sync of data so would sync all user records to all devices. This PR tries to standardise methods for handling a case where bulk local table records sync to single shared server app table (using a combination of user id and local record id to ensure globally unique). This could be further built on to handle other cases (e.g. refactoring current contact field sync methods to be a generic single user record case of the above)

2) It also raises a question whether the current approach of sending data via specific apis (e.g. `/app_feedback`) is sensible enough, or whether we want to provide more generic API endpoint or even a direct link from app db to server db. Possibly could be discussed.

3) Previously we had a placeholder `_synced` field which was intended for tracking db sync and was marked true/false, however this had 2 issues. Firstly boolean fields can't be indexed with dexie and so are less efficient when querying for non-synced data (although not noticeably) and more importantly it doesn't distinguish between data that hasn't been synced and data that shouldn't be synced. As such I've added a new `_sync_status` field that currently holds values `pending` `ignored` or `sent` , and we will need to plan how best to deprecate the old field.

## Next TODOs
- Add app bindings to trigger automated sync on schedule/demand
- Add triggers system to allow notification to email/slack when new feedback received.

## Future TODOs
- server service and db-sync service should ideally be combined as they are trying to achieve mostly the same thing

## Git Issues

Closes #

## Screenshots/Videos

**Demo recording feedback and syncing to server**

https://user-images.githubusercontent.com/10515065/149266783-a0065d4f-6260-44bb-b089-b326c68053b6.mp4

(Note - nested `additional` field in demo now fixed to remove nesting and `ev` fields, and add template path meta)

**Generic schema - storing multiple user records in a shared table includes identifier columns with a single data record field**
image


chrismclarke:
Agreed on call to merge and release to server to allow any additional debugging/feedback

PR #1169, opened 2022-01-14
Fix and document display of item loops

### LINKED ISSUE #1168: [BUG] Items do not have padding
**Describe the bug**
- [ ] Components that are authored using "items" do not have padding between them. 
- [ ] The go_to action of buttons authored using "items" will go to the template associated to the last item. 

**App version**
v0.13.0 + some content edits

**Debug template**
[debug_items
](https://docs.google.com/spreadsheets/d/1nrTs-din5K6Zjf5ZfzOmeH4TskgRbbQnNazFlNk3iH0/edit#gid=669305399)
image

**Screenshots and videos**
image

chrismclarke:
PR Checklist

## Description
A datalist can be iterated over and used to create a dynamic/repeating list of components via the `items` component type. One caveat however is that the items each render within an _item group_ component, and so don't behave as expected when placed inside a display group or when left to render standalone.

This PR refactors the rendering to cut out the intermediate group, and instead render the child components directly. So if iterating a list of 3 items each with 2 components the output will be a list of 6 components (instead of previously 3 groups each with 2 components). This means they will space correctly when rendered using existing methods and tools.

I've also fixed a bug noticed where items would perform the wrong click action (annoying javascript memory optimisations could leave the wrong items in memory when iterating over the list, fixed by forcing cleaner variable allocation).

**Main changes**

- Refactor item rendering method so that they do not render inside a particular container (to allow use within existing display groups)
- Fix memory allocation bug where items would use the incorrect click action
- Add [feature_items](https://docs.google.com/spreadsheets/d/1LnKtclWcnbFs8vQKSUVCFd50IG81iZChb3VgRujpOrg/edit#gid=1745157248) demo sheets to show different ways to display and group items
- Add documentation to explain how to use item loops

## Review Notes
- This changes the way items are displayed, so could have knock-ons if used elsewhere if previous workarounds had been used to fix item display

## Git Issues

Closes #1168

## Screenshots/Videos
**New documentation added**
127 0 0 1_8000_authors_advanced_looping-data_

**Fixed click actions**

https://user-images.githubusercontent.com/10515065/149435408-96ae43bf-1b78-4380-bc80-5d9c7d7fed68.mp4

**Feature_items demo**
image


packages/data-models/flowTypes.ts
```diff
@@ -377,7 +377,6 @@ export namespace FlowTypes {
     | "parent_point_box"
```
chrismclarke, line 377: remove previous `group` template type that was used to display items (no longer required as items do not render in groups, and also the component is too easily be confused with display_groups)

src/app/shared/components/template/services/instance/template-row.service.ts
```diff
@@ -195,7 +195,12 @@ export class TemplateRowService extends TemplateInstanceService {
       const processed = await this.processSingleRow(preProcessedRow, isNestedTemplate);
       // only include rows that do not return undefined (e.g. set_variable)
       if (processed) {
-        processedRows.push(processed);
```
chrismclarke, line 198: minor change required to allow for processing that returns more than one row for each input. Could be useful in the future if planning to render nested templates in the same way (currently render in containers but might be nicer if all components sit at same level as more akin to how they were designed in excel)

src/app/shared/components/template/services/instance/template-row.service.ts
```diff
@@ -327,30 +333,32 @@ export class TemplateRowService extends TemplateInstanceService {
...
+      }
+    }
+    return loopItemRows;
   }
   /** Update the evaluation context of a row and recursively any nested rows */
   private setRecursiveRowEvalContext(
     row: FlowTypes.TemplateRow,
     evalContext: FlowTypes.TemplateRow["_evalContext"]
   ) {
-    const { rows, ...rest } = row;
```
chrismclarke, line 346: memory allocation fix, when it comes to operating on deeply nested objects issues can arise in edge cases where re-assigning and referring to deeply nested objects with the same may sometimes keep the wrong item in memory. Hacky fix is to convert to string and back again as this always creates a new memory allocation.

EsmeeIDEMS (approved):
Functional test passed

volloholic (approved):

Issue #1173, opened 2022-01-14
[FEATURE] - Define more sensible style conventions for authoring


chrismclarke:
**What?**
Most component 'style' variations are defined through hardcoded CSS, and can be hard to understand or link to the authoring system (e.g. some components accept a `large` style which changes the font size to an arbitrary size, whilst others don't accept the style and some use `large` to change height/width). 

It is suggested to try and move away from as many hardcoded styles in CSS and expose a system that makes it easier to author styles.

**How?**
Recommend adding existing style utility library tailwind CSS as starting point for styling.
https://tailwindcss.com/docs/installation

In order to keep authoring simpler, an additional data list type (perhaps `style_list`) could be used to define names of specific styles and where to scope. E.g. as a rough example to create a `large` button for use within all templates the following could be defined

style_name | component_scope | template_scope | tailwind_class
-- | -- | -- | --
large | button | global | m-1 min-w-[100px]

The tailwind class would apply automated margins around the button (size `1` corresponds to a fixed pixel size, and could be increased to `m-2` if more spacing around required), and assign a minimum width of 100px. Authors can then still just use `style: large` while authoring buttons but be able to access clearer understanding of how that translated to code.

**Other details**
Column names just quick suggestion, open to suggestion

chrismclarke:
@volloholic @EsmeeIDEMS 
Feel free if you want to add any feedback on the proposal

chrismclarke:
See additional comments in #1183

PR #1183, opened 2022-01-18
Fix/dg style overrides

chrismclarke:
PR Checklist

## Description

Rasied in [sheets comment](https://docs.google.com/spreadsheets/d/19qLXy1PF6BOH9GZ-qpcZJ3-bOfxXAgj7J8H67Nfp9MM/edit?disco=AAAAUI2xa2I), style overrides applied to display groups do not have the intended effect. This is because the overrides apply to the main container element, but for display groups most of the display logic is handled within a child div.

This PR provides a means to apply the overridden styles to a specific part of a component, and applied to display groups to allow override of flex behaviour.

This is only a temporary solution (see issue #1173), but does raise a larger challenge for when we do move to the class-based system, how best to define in the syntax a way to specify where exactly the override styles/classes should be applied. Possibly could adapt from ionic/shadow-dom concept of 'slots' and/or create a custom authoring syntax.

## Review Notes
Could impact on existing templates that use custom style overrides on display groups, have added visual test to check

## Git Issues

Closes #

## Screenshots/Videos

**Display group tiles now inherit style overrides correctly**
image

image


EsmeeIDEMS (changes requested):
**Ignore this review, I was using a different branch.** 

I'm getting some errors when trying to run the app here, can it be that master needs to be merged in here?

chrismclarke:
Just merged master into branch and all looking fine to me so hopefully should be working as intended.

I also realised the quick solution to get tiles to sit in the middle of each space is to used another CSS property `justify-content: space-around`. See screenshot below of updated sheet

image

EsmeeIDEMS (approved):
Functional test passed. @volloholic can you review code?

chrismclarke:
I think probably good to merge. The code changes are only superficial (changing which level styles are applied to) and larger discussion of improving styling captured elsewhere

PR #1191, opened 2022-01-22
Feat/app minimize

chrismclarke:
PR Checklist

- [x] - Latest `master` branch merged
- [x] - PR title descriptive (can be used in release notes)

## Description

Capacitor (finally) released functionality to allow the app to be minimised (briefly could only be fully closed). As a result when pressing back button on the home screen running on an android device nothing happens.

This PR changes behaviour so that if back button is pressed on the home screen the app minimises (as is common for most other apps).

## Review Notes
As this can only be tested on a physical device recommend using generated link to test via appetize.io

## Dev Notes
Various core plugins have been updated to include the new feature. These are all minor version updates and so not expected to have any breaking changes.

## Git Issues

Closes #

## Screenshots/Videos

https://user-images.githubusercontent.com/10515065/150627860-fe02ca1a-ad7b-4d86-b893-d8ccb80b537e.mp4

PR #1192, opened 2022-01-22
feat: add appetize gh action

chrismclarke:
PR Checklist

## Description

Adds a github action to generate an android build and deploy to appetize.io for testing on a web-based android emulator.

## Review Notes
Adding the label `Test - appetize` will trigger the process. This usually takes around 5-10 minutes and a preview URL will be recorded in the pr comments section.

The workflow is currently configured so that ONLY 1 build will be available at a time. That means if the label is added to multiple PRs only the one most recently updated will be the live build (the same URL will be shared for both). 

For testing, appetize.io provide a free tier that allows 100 minutes per month. I've used my own personal account/key for now, but in the future we should create a new account for the project and change the key.

The app does not appear to work on the emulator android 8.1 or 9 (works on local emulator so assume issue with version of chrome/webview installed on the emulators), and so by default the URL points to an emulator running android 10

## Git Issues

Closes #

## Screenshots/Videos

https://user-images.githubusercontent.com/10515065/150627211-043a9d26-6ce1-4271-8b65-06e6195cdd64.mp4

### DIFF
--- a/.github/workflows/test-appetize.yml
+++ b/.github/workflows/test-appetize.yml
@@ -8,7 +8,7 @@ on:
       - master
 jobs:
   build:
-    if: contains(github.event.pull_request.labels.*.name, 'Test - device appetize')
+    if: contains(github.event.pull_request.labels.*.name, 'Test - appetize')
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


chrismclarke:
Merging to test in #1191

PR #1193, opened 2022-01-23
Feat/clicked notification actions

### LINKED ISSUE #1178: [FEATURE] Distinguish clicked and dismissed notifications
**What?**
Action handlers should distinguish between whether a notification was clicked or dismissed. 

**Additional comments**
Discussed on call and in comment thread of #1158.

chrismclarke:
PR Checklist

## Description

- Add support for `sent` notification actions
- Alter notification `click` actions to only be triggered if notification clicked
- Add debug template

## Review Notes
A brief template was created to test click/sent actions, however this has been marked as draft (due to breaking type changes) and not included in the PR (as there are too many other content changes to sync). 
https://docs.google.com/spreadsheets/d/1Jf38D4pkoBO0QIUcw1IFAqlEokrUzhOESi3hNlW0vZM/edit#gid=1437448788

## Author Notes
Many of the existing `click` actions will no longer be triggered, and so should be changed to `sent` actions if required to always be triggered.

## Dev Notes
This only addresses the immediate issue of differentiating between sent and interacted notifications, will still need another refactor to support click across future list of possible actions (possibly action:action_id handler or similar - although might make typing more difficult).

Originally we had planned to separately track `dismissed` notifications, however this was tricky to do and led to confusing code as as dismissing a notification doesn't invoke an action in the same way that clicking does. Therefore the only way to track dismissed notifications is to track all sent and cross-check against those that have been clicked, however, the order in which notifications are marked as sent vs clicked varies depending on whether the app is open or not. Therefore notifications may be incorrectly marked as dismissed if registering the sent action before the clicked action (or vice-versa). Separately tracking sent and clicked notifications however is much more simple/consistent, and could be used to manually calculate dismissed if required.

## Git Issues

Closes #1178

## Screenshots/Videos

https://user-images.githubusercontent.com/10515065/150669857-e0d71745-0d5c-4d79-8f9e-1ea374e0564e.mp4

### DIFF
--- a/packages/data-models/flowTypes.ts
+++ b/packages/data-models/flowTypes.ts
@@ -432,7 +432,8 @@ export namespace FlowTypes {
     | "audio_pause"
     | "audio_end"
     | "audio_first_start"
-    | "nav_resume"; // return to template after navigation or popup close;
+    | "nav_resume" // return to template after navigation or popup close;
+    | "sent"; // notification sent
 
   export interface TemplateRowAction {
     /** actions have an associated trigger */
--- a/packages/scripts/src/commands/app-data/convert/utils/app-data-action.utils.ts
+++ b/packages/scripts/src/commands/app-data/convert/utils/app-data-action.utils.ts
@@ -29,6 +29,7 @@ export function parseAppDataActionString(actionString: string): FlowTypes.Templa
     completed: true,
     uncompleted: true,
     nav_resume: true,
+    sent: true,
   };
   if (!Object.keys(actionTriggers).find((t) => actionString.startsWith(t))) {
     actionString = `click | ${actionString}`;
--- a/src/app/feature/campaign/campaign.service.ts
+++ b/src/app/feature/campaign/campaign.service.ts
@@ -47,6 +47,7 @@ export class CampaignService {
 
   private _handledNotifications = {};
   private _notificationUpdates$: Subscription;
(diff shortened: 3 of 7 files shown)


EsmeeIDEMS (approved):
Functional test passed. Two authoring comments: 

- Should the _click_action_list_ column be renamed _action_list_ now that it has both click and sent actions? 
image
- Given that the other events are _changed, sent, completed_, I was wondering if _click_ should be renamed _clicked_. This is does appear all over the place so changing in the spreadsheets is a bit of a pain...

chrismclarke:
> * Should the _click_action_list_ column be renamed _action_list_ now that it has both click and sent actions?

I was thinking that myself when making the update. I assume yes (it makes it more consistent with other places), but again a slight annoyance for authoring. I might propose a small change to the parser to temporarily support both while we deprecate click_action_list in favour of action_list

> * Given that the other events are _changed, sent, completed_, I was wondering if _click_ should be renamed _clicked_. This is does appear all over the place so changing in the spreadsheets is a bit of a pain...

I have less issue with this one, we have quite a lot of different actions that have different tenses (e.g. `nav_resume` following a popup close, `audio_play`, `audio_paused` etc.), so I think it will be a bit harder to be consistent throughout - but I have no strong opinion either way.

volloholic (approved):

PR #1197, opened 2022-01-24
chore: rename click_action_list to action_list

chrismclarke:
PR Checklist

## Description

Following comments in #1193, added a `LEGACY_FIELD_RENAME` to the default parser that can be used to convert all fields named one thing to another as a means to temporarily support both while deprecating legacy field names. This is currently set to rename any `click_action_list` to `action_list`.

## Author Notes
We don't have a fixed date for removing support (expect won't be any time soon), but recommend renaming click_action_list to action_list on at least the main example/debug sheets at first so that anyone using or modifying them has the correct column names.

There are a limited number of cases where this will not run, searching the codebase for `click_action_list` after implementing the changes shows a handful of debug sheets where the field is referred to as dynamic text. These will need to be updated once merged.

image

## Review Notes
The easiest time to review will be just after content merge so that the only changes seen will be from the script updates and not just from new/edited forms. 

## Git Issues

Closes #

## Screenshots/Videos

Small warning message displayed in parser. Eventually we will want to better collate these and present an overall summary at the end (possibly as linked json file)
image

### DIFF
--- a/packages/scripts/src/commands/app-data/convert/index.ts
+++ b/packages/scripts/src/commands/app-data/convert/index.ts
@@ -38,7 +38,7 @@ export default program
 
 class AppDataConverter {
   /** Change version to invalidate any cached conversions */
-  public converterVersion = 1;
+  public converterVersion = 1.1;
 
   private activeDeployment = getActiveDeployment();
   private paths = {
--- a/packages/data-models/flowTypes.ts
+++ b/packages/data-models/flowTypes.ts
@@ -205,7 +205,7 @@ export namespace FlowTypes {
     campaign_list: string[]; // ids of campaigns where to run
     priority?: number; // higher numbers will be given more priority
     // additional fields for current data_list but not required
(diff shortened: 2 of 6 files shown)


packages/scripts/src/commands/app-data/convert/index.ts
```diff
@@ -38,7 +38,7 @@ export default program
 
 class AppDataConverter {
   /** Change version to invalidate any cached conversions */
-  public converterVersion = 1;
+  public converterVersion = 1.1;
```
chrismclarke, line 41: internal code used simply to invalidate previously converted files and re-convert (to process with new legacy mapping update)

EsmeeIDEMS (approved):
Looks good to me. Functional test passed.

volloholic (approved):

EsmeeIDEMS:
@chrismclarke Can you merge this? There are some conflicts.

PR #1200, opened 2022-01-24
fix(1199): injected translate service

### LINKED ISSUE #1199: [BUG] - App language select not working from main splash screen
**Describe the bug**
See issue described in #1191

chrismclarke:
PR Checklist

## Description

Following previous core template service refactor the translate service was no longer included with the action service, breaking the language select sheet. Fixes the issue and attempts to add an additional warning to help catch some similar issues in the future.

## Review Notes
Minor fix working locally for me. I'm also tempted to suggest this might be the time to create another issue to add some more basic end-to-end tests that automate a few core functions and cover a reasonable amount of our manual testing strategy to make sure they always work. E.g. start app, select language, navigate to specific template, open popup, click button etc.

We already have a large part of the core in place needed for this so would hope it would only be a day or so's work to add some specific testing and action integration. We could also consider creating a small syntax for authoring explicit testing templates that cover similar case (e.g. adding a `test_step` row type alongside existing action list to manage `go_to`, `click` etc.).

## Git Issues

Closes #1199

## Screenshots/Videos

https://user-images.githubusercontent.com/10515065/150842617-8466a5b0-56ef-403f-b48f-b4ec39c11fdc.mp4

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -23,6 +23,7 @@ import { AnalyticsService } from "./shared/services/analytics/analytics.service"
 import { LocalNotificationService } from "./shared/services/notification/local-notification.service";
 import { APP_INITIALISATION_DEFAULTS, APP_SIDEMENU_DEFAULTS } from "packages/data-models/constants";
 import { TemplateFieldService } from "./shared/components/template/services/template-field.service";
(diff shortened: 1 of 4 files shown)


EsmeeIDEMS (approved):
Functional test passed.

volloholic (approved):

PR #1218, opened 2022-02-03
fix: fullscreen popup scroll

### LINKED ISSUE #1217: The terms and conditions page is not scrollable
Described in #1214 , separated for easy close via pr

chrismclarke:
PR Checklist

## Description

Fix height style issue preventing scroll in full-screen popups

## Git Issues

Closes #1217

## Screenshots/Videos

https://user-images.githubusercontent.com/10515065/152430672-90311a65-c661-403e-b313-f2eb27054ce7.mp4

### DIFF
--- a/src/app/shared/components/template/components/combo-box/combo-box-modal/combo-box-modal.component.scss
+++ b/src/app/shared/components/template/components/combo-box/combo-box-modal/combo-box-modal.component.scss
@@ -25,6 +25,7 @@
   height: 100%;
   padding: 20px 20px 0 20px;
   overflow: auto;
+  max-height: 90vh; // match ion-modal.combo-box-modal
   .header {
     font-size: 20px;
     text-align: center;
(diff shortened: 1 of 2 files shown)


EsmeeIDEMS (approved):
In the "first app launch template sequence" the template **organisation_registration** no longer fills the screen - but I actually think it looks good this way so I'm not too worried

image

image

chrismclarke:
@EsmeeIDEMS - thanks for catching this. Although I agree with you that I prefer the style with things in the center, I think it still makes sense that the template layout be respected in the same way it is for the actual template. It also makes it easier to have things like buttons consistently in the same space at the bottom of the screen.

I've made a minor change so that the full-screen content now always fills the full available space. In the future we could try add a better mode for the display group (or alternatives) to make it more easy to include vertically centered content.

I also noticed an issue where the combobox content doesn't scroll when picking my group. Not related to this PR, but included as a fix here also

volloholic (approved):

EsmeeIDEMS (approved):
Functional test passed.

Issue #1013, opened 2021-09-09
[BUG] Manual rerender needed after changing language or entering developer mode


EsmeeIDEMS:
**Describe the bug**
- [ ] After changing language, some texts on the Weekly Workshop page appear in the previous language. This can be resolved by refreshing the page (if in browser). 
- [ ] After changing entering developer mode (i.e. changing user_mode to false), the left side-menu still shows as for user mode. This can be resolved by refreshing the page (if in browser).

**App version**
v0.11.2

**To Reproduce**

**Template Name**
[weekly_workshops](https://docs.google.com/spreadsheets/d/1tmtdyeSYcqGH6joqZiA-wPtH0QH6td8hA3qGp62AcHs/edit#gid=1277645488)
[language_select](https://docs.google.com/spreadsheets/d/1NUCgonMc3K4y9bFjYH7LjquRZyiIN2AappLP_0mlH5Q/edit#gid=823624970)
[app_menu_enter_dev_mode](https://docs.google.com/spreadsheets/d/1NUCgonMc3K4y9bFjYH7LjquRZyiIN2AappLP_0mlH5Q/edit#gid=656974851)

**Screenshots and videos**

PR #1220, opened 2022-02-03
feat: add force_restart emit action

chrismclarke:
PR Checklist

## Description

We have 2 (confusingly named) actions `force_reload` and `force_reprocess` that both trigger re-evaluation of templates (with reload re-evaluating fully and reprocess just checking for variable changes). Neither, however re-evaluate content that is sat outside the current template page, such as the app menu.

As an interim solution I've added a `force_restart` action that does a more heavy-handed full page reload and redirect to the home page (like when running app reset).

I've also updated the authoring to perform a restart when changing between dev and user mode from the settings screen.

## Review Notes
There's no easy way to review the changes, because when running locally the app is forced into dev mode regardless. So we will need to sync the latest content with the authoring changes and deploy to the web version to test. 

But by temporarily blocking the local app overrides all behaves as expected in video below

## Git Issues

Closes #

## Screenshots/Videos

https://user-images.githubusercontent.com/10515065/152440306-b304171b-8794-4b4c-b5aa-e1ac1b58ee22.mp4

### DIFF
--- a/src/app/shared/components/template/services/instance/template-action.service.ts
+++ b/src/app/shared/components/template/services/instance/template-action.service.ts
@@ -183,7 +183,14 @@ export class TemplateActionService extends TemplateInstanceService {
         // TODO - trigger emit in shared service to allow individual services/components to subscribe
         // themselves instead of relying on hardcoded calls here
 
-        // Handle a forced rerender
+        /* Emit actions
+            emit:force_restart    // reload app window to home page
+            emit:force_reload     // full re-render of all templates in current page
+            emit:force_reprocess   // recalculate existing rows (not set_variable/set_nested)
+        */
+        if (emit_value === "force_restart") {
+          location.href = "/";
+        }
         if (emit_value === "force_reload") {
           await this.container?.forceRerender(true);
         }


EsmeeIDEMS (changes requested):
This relates to issue #1013, do you want to close it or should we leave it open since this is only an interim solution? 

I wanted to use this also on the [language_select](https://docs.google.com/spreadsheets/d/1iSG1kzK4OTHcoiwWoZ25LZxrJqERg2_iP9369HTi_bQ/edit#gid=823624970) template (so that the side menu updates to the correct language when the language is changed). This could be either on the radio group directly
`changed | emit: force_restart`
or on the nav_buttons template
`completed | emit:force_restart`
However, neither seems to work (the console says `force_reload` in one case and `force_restart` in the other, but I never get redirected to the home screen)
image

chrismclarke:
I think best to leave open, the language select is a little bit tricker because the way things are currently authored changing the language during the startup screen would reload the app but skip the rest of the slides. Really I just need to add a better means to update the menu, so good to keep open

volloholic (approved):

EsmeeIDEMS (approved):

PR #1228, opened 2022-02-05
feat: use largeBody for notification text

chrismclarke:
PR Checklist

## Description

Converts all notifications to use `largeBody` for text by default. 

## Author Notes
This is a slight trade-off as notifications will now show less content when contracted (just one line instead of previous character limit), but more when expanded (previously not possible).

The notification API has support for both `summary` and `largeBody`, so that we could experiment more with displaying more/different content in the contracted view, however I hope for now having the option to support more text is sufficient.

## Review Notes
I've included appetizer preview, however as this does not have the changes required for scrollable terms display only works if you change to a larger device. Also I'm not sure if there are any good notifications from the last content sync to test (i.e. that have large content) - I tested on local android emulator with some content from `nf_weekly_workshops` as these notifications seemed longer (screenshots below), although again they're not available in the preview because the content has not been merged. 

## Git Issues

Closes #

## Screenshots/Videos

Example notification expanded
image

Example notification not expanded
image


volloholic (approved):

EsmeeIDEMS (approved):
Functional test passed

PR #1241, opened 2022-02-12
Feat/campaign batch notifications

### LINKED ISSUE #1227: [FEATURE] - Allow schedule of multiple notifications in advance
**Why?**
As notifications only schedule when the app is open, daily notifications will only ever send on the first day if the user chooses to dismiss the notification and not re-open the app.

**How?**
Previously drafted a `batch_size` column used for this purpose. Should implement and document

chrismclarke:
PR Checklist

## Description

Add campaign schedule support for `schedule.batch_size` as a means to schedule multiple notifications for a campaign. This allows daily notifications to be scheduled for a set number of days to ensure they will trigger without need to re-open the app and re-evaluate conditions.

## Author notes
By default, only notifications that satisfy activation/deactivation conditions will be scheduled. So if `notification_2` requires `notification_1.sent` it will not be scheduled even if `batch_size` set to 2. Instead, the recommended approach is to omit activation conditions and instead use a combination of batch_size, priority, deactivation criteria to schedule both and deactivate as required.

## Review notes
I've updated the debug-daily notifications to include `schedule.batch_size` which should demo the functionality. I also updated the campaign debug page to list multiple pending notifications. I think things look as expected, although I haven't tested receiving notifications across multiple days (might be one for beta-testers to try on device next release)

## Git Issues

Closes #1227

## Screenshots/Videos

Debug_daily notifications schedule with `batch_size: 5`
image

### DIFF
--- a/documentation/docs/authors/notifications.md
+++ b/documentation/docs/authors/notifications.md
@@ -22,6 +22,7 @@ Here is an example of two rows added to a `==content_list==` for creating notifi
 | schedule.start_date           | Date of earliest notification. If in past will be ignored |
 | schedule.end_date             | Date of latest notification. If in past will be ignored |
 | schedule.day_of_week          | Send notifications weekly on a given day of the week (numbered from 1 - Monday, to 7 - Sunday) |
+| schedule.batch_size           | Maximum number of notifications to schedule, e.g. if scheduling weekly specify number of weeks to preschedule |
 | time.hour                     | Hour of day to send notificaitons on, in 24h format       |
 | time.minute                   | Minute of day to send notification on |
 | delay.days                    | Offset to schedule notfication a given number of days from today |
--- a/src/app/feature/campaign/campaign.service.ts
+++ b/src/app/feature/campaign/campaign.service.ts
@@ -1,16 +1,6 @@
 import { Injectable, Injector } from "@angular/core";
 import { addDays, addSeconds } from "date-fns";
-import {
-  addHours,
-  addMinutes,
-  addWeeks,
-  endOfDay,
-  isAfter,
-  isBefore,
-  isFuture,
-  isPast,
-  setISODay,
-} from "date-fns";
+import { addHours, addMinutes, addWeeks, endOfDay, isAfter, isBefore, setISODay } from "date-fns";
 import { extractDynamicFields } from "packages/data-models";
 import { NOTIFICATION_DEFAULTS } from "packages/data-models/constants";
 import { Subscription } from "rxjs";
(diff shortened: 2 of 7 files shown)


EsmeeIDEMS:
> By default, only notifications that satisfy activation/deactivation conditions will be scheduled. So if notification_2 requires notification_1.sent it will not be scheduled even if batch_size set to 2. Instead, the recommended approach is to omit activation conditions and instead use a combination of batch_size, priority, deactivation criteria to schedule both and deactivate as required.

I am confused by this author note. Which of the two below is correct? 

> By default, only notifications that satisfy activation/~~deactivation~~ conditions will be scheduled. So if notification_2 requires notification_1.sent it will not be scheduled even if batch_size set to 2. Instead, the recommended approach is to omit activation conditions and instead use a combination of batch_size, priority, deactivation criteria to schedule both and deactivate as required.

or 

> By default, only notifications that satisfy activation/deactivation conditions will be scheduled. So if notification_2 requires notification_1.sent it will not be scheduled even if batch_size set to 2. Instead, the recommended approach is to omit activation/**deactivation** conditions and instead use a combination of batch_size, priority, ~~deactivation criteria~~ to schedule both and deactivate as required.

EsmeeIDEMS (changes requested):
Not sure if I'm doing something silly but to me it seems that all notifications in the batch are scheduled for the same day/time 
image

chrismclarke:
> Not sure if I'm doing something silly but to me it seems that all notifications in the batch are scheduled for the same day/time 

I had the same issue but fixed shortly after with commit 6c1aa7a, I'm not sure if maybe you tested before that update came in? Could you retry to confirm if the issue persists?


> I am confused by this author note. Which of the two below is correct?
> 
>> By default, only notifications that satisfy activation/deactivation conditions will be scheduled. So if notification_2 requires notification_1.sent it will not be scheduled even if batch_size set to 2. Instead, the recommended approach is to omit activation conditions and instead use a combination of batch_size, priority, deactivation criteria to schedule both and deactivate as required.
> 
> or
> 
>> By default, only notifications that satisfy activation/deactivation conditions will be scheduled. So if notification_2 requires notification_1.sent it will not be scheduled even if batch_size set to 2. Instead, the recommended approach is to omit activation/deactivation conditions and instead use a combination of batch_size, priority, deactivation criteria to schedule both and deactivate as required.

To clarify - by activation/deactivation I mean satisfy all activation criteria and not satisfy any deactivation criteria. So what I am proposing is to reduce any activation criteria that depend on another being marked as `previous_notification.sent`. You can still use the `.sent` property to deactivate a notification (to prevent resending same notification), and rely on priority to prevent something like `notification_2` being sent before `notification_1`

EsmeeIDEMS:
> I had the same issue but fixed shortly after with commit [6c1aa7a](https://github.com/IDEMSInternational/parenting-app-ui/commit/6c1aa7a335030093108e77073fd8932b30c4b745), I'm not sure if maybe you tested before that update came in? Could you retry to confirm if the issue persists?

I tested (and now tested again) after that commit, but the issue persists. 
image

chrismclarke:
Very strange, probably best to just try quickly debug on tomorrows call

EsmeeIDEMS (approved):
Functional test passed. Timezone issue resolved.

PR #1244, opened 2022-02-14
fix: add db sync to server_sync action

chrismclarke:
PR Checklist

## Description

There are two different mechanisms in the app for syncing user data. A hardcoded/standalone system used to sync app user data (created to sync device info and contact fields), and a more generic system for syncing the database more generally (used for app_feedback).

This PR adds the second db sync to the action that manages the other user data sync (`emit:server_sync`). 

## Review Notes
Will require adding feedback and checking if sync successful from setting sync button. If working could also consider additing the `emit: server_sync` action to be triggered after submitting feedback for instant upload.

## Dev Notes
At some point in the future it would make sense to combine, but considered lower priority for now

## Git Issues

Closes #

## Screenshots/Videos

https://user-images.githubusercontent.com/10515065/153935321-f163b976-d456-43ee-9de8-4bdb8cdf0f55.mp4

### DIFF
--- a/src/app/shared/components/template/services/instance/template-action.service.ts
+++ b/src/app/shared/components/template/services/instance/template-action.service.ts
@@ -14,6 +14,7 @@ import { TourService } from "src/app/shared/services/tour/tour.service";
 import { TemplateTranslateService } from "../template-translate.service";
 import { TemplateFieldService } from "../template-field.service";
 import { EventService } from "src/app/shared/services/event/event.service";
+import { DBSyncService } from "src/app/shared/services/db/db-sync.service";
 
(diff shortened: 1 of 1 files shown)


EsmeeIDEMS:
For some reason I am not able to reproduce what you show in the video using the preview URL. The `feedback`, `_created` and `_sync_status` tabs all remain empty even though I've authored 4 test feedback messages.  

image

chrismclarke:
I think you might need to manually press the refresh button for the indexeddb to see changes, you should be able to see it when on a table (small 🔃 style button ). Otherwise I think this should be good to go, so will merge to allow on-device testing next week

image

PR #1250, opened 2022-02-18
feat: add deployment name field mapping

chrismclarke:
PR Checklist

## Description

- Add bindings to automatically populate a `_deployment_name` field with the current deployment name
- Add example template to highlight known hardcoded fields

## Review Notes
Debug sheet should show as working correctly, can also try changing app deployment name (to something more desirable) and confirm working. Although if changing a corresponding change will need to be made to github secrets so let me know if that's the case (see dev note below)

## Author Notes
I can't remember if any of these hardcoded variables are documented elsewhere, if not might be good to make a note somewhere. 

## Dev Notes
Currently the scripts that build and deploy the app from the cloud work on the assumption that whatever is in the repo is what we want built. However, in order to access the deployment name we need to access the specific deployment json file which is not committed (generated dynamically with `yarn scripts deployment set`). So now the cloud builder specifically needs to know which deployment to build and is now passed as an environment variable from github itself (currently `PLH ZA`).

This was always going to be a requirement to support multiple app builds in the future so not a fundamental issue/change, more just one to be aware of. Short term it means that  if we want to change the `PLH ZA` name I need to update the environment variable stored on github, longer term we will want to think up a better way of handling (part of longer term cloud build system)

## Git Issues

Closes #

## Screenshots/Videos

[example_hardcoded_fields](https://docs.google.com/spreadsheets/d/1X10C5YWsqJ7NRJp-pSiaYjdAoV4hl5-scvisiBThmq8/edit#gid=460679842)

image

Running locally shows my debug deployment
image

Running through cloud build preview shows plh za deployment
image

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -45,6 +45,10 @@ jobs:
         run: echo $GOOGLE_SERVICES_JSON > android/app/google-services.json
       - name: Install Dependencies
         run: yarn install
+      - name: Set Deployment
+        run: yarn scripts deployment set $APP_DEPLOYMENT_NAME
+        env:
+          APP_DEPLOYMENT_NAME: ${{ secrets.APP_DEPLOYMENT_NAME }}
       - name: Build Angular Project
         run: yarn build
       - name: Sync Android Files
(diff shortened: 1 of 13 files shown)


EsmeeIDEMS (approved):
Functional test passed.

PR #1258, opened 2022-02-25
refactor: add app default language to constants

chrismclarke:
PR Checklist

## Description

Refactor app to allow default language to be specified from constants

## Author Notes
The default language can now be changed from the `packages\data-models\constants.ts` file. This will only affect users that have not already specified a language. The default value has currently just been set as `en` which does not exist as a translation, and so defaults to the source strings (which are in generic English)

## Git Issues

Closes #

## Screenshots/Videos
Example loading app for first time with `tz_sw` specified as the default language
```
export const APP_LANGUAGES = {
  /** Language used during first load. If translations do not exist will default to source strings (en) */
  default: "tz_sw",
};
```

https://user-images.githubusercontent.com/10515065/155628336-21b3b73f-8727-4b5f-adae-7439cba63af4.mp4

### DIFF
--- a/packages/data-models/constants.ts
+++ b/packages/data-models/constants.ts
@@ -35,6 +35,11 @@ export const APP_FIELDS = {
   DEPLOYMENT_NAME: `${FIELD_PREFIX}._deployment_name`,
 };
 
+export const APP_LANGUAGES = {
+  /** Language used during first load. If translations do not exist will default to source strings (en) */
+  default: "en",
+};
+
 /**
  * Some specific strings are currently hardcoded into the app
  * TODO - not all strings included, should add to when required
--- a/packages/data-models/constants.ts
+++ b/packages/data-models/constants.ts
@@ -37,7 +37,7 @@ export const APP_FIELDS = {
 
 export const APP_LANGUAGES = {
   /** Language used during first load. If translations do not exist will default to source strings (gb_en) */
-  default: "en",
+  default: "gb_en",
 };
 
 /**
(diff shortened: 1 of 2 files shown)


packages/data-models/constants.ts
```diff
@@ -35,6 +35,11 @@ export const APP_FIELDS = {
   DEPLOYMENT_NAME: `${FIELD_PREFIX}._deployment_name`,
 };
 
+export const APP_LANGUAGES = {
+  /** Language used during first load. If translations do not exist will default to source strings (en) */
```
EsmeeIDEMS, line 39:
```suggestion
  /** Language used during first load. If translations do not exist will default to source strings (gb_en) */
```

packages/data-models/constants.ts
```diff
@@ -35,6 +35,11 @@ export const APP_FIELDS = {
...
+  default: "en",
```
EsmeeIDEMS, line 40:
```suggestion
  default: "gb_en",
```

packages/data-models/constants.ts
EsmeeIDEMS: I'm not sure what "generic" English would be. Since the app content always uses British spelling I think it's clearer to call this `gb_en`

EsmeeIDEMS (changes requested):
Functional test passed.

packages/data-models/constants.ts
chrismclarke: I think I was trying to distinguish between languages that we actually have specific overrides for vs just using the _source_ strings. So equally I could have just said the first language was _default_ or _source_, but I have no issue with calling it gb_en

EsmeeIDEMS (approved):

PR #1278, opened 2022-03-15
Feat/e2e test authoring

### LINKED ISSUE #1211: [FEATURE] - E2E Test Authoring
**What?**
In the app we currently have a system for visual regression testing, that visits each authored templates, takes screenshots and compares to previous state to test for changes. Whilst good at catching a lot of potential issues, it has a couple limitations:
1) Number of false positives on pages where things like animations will display differently depending on exact timing
2) Lack of test for interactive functionality, such as button clicks, popups etc.

We want to add an additional system to be used alongside that allows authors to create explicit sets of end-to-end tests to carry out. E.g. visit a template, click certain buttons, check what is expected to appear actually appears.

**Why?**
Reduce need to repeatedly manually test common features/functionality and add better catches for breaking changes. 

**How?**
Main tasks:
- Identify current state of testing system (what works, what doesn't)
- Coordinate with content/qa team to identify roughly what should be included in e2e tests (e.g. specific workflows to visit, how to interact (e.g. click button, wait for element, check storage state etc.). Basically what happens when the app is sent for manual 'functional testing'?
- Propose a syntax for authoring tests. E.g. google sheet with columns for step-by-step actions and expectations that map reasonably to cypress + whatever assertion library is in use
- Add required bindings to parse the sheet and run as a test
- Create example e2e test sheet that covers available functions
- Add integrations to run e2e tests from github actions
- Create documentation that provides rough outline/explanation of e2e authored tests
- On-board authors to write a few tests

**Other details**
We started adding rough support for end-to-end testing via cypress in the `test-e2e` workspace, however this was abandoned in a fairly rough state. We will need to revisit to assess what functionality is available, and build on it to support an authored testing system. The visual test system uses puppeteer instead of cypress, undecided which will be the best approach for the proposed testing system (I assume cypress but open to suggestion)

chrismclarke:
PR Checklist

## Description
Merge WiP updates to e2e system from @thewhodidthis mirror repo.

## PR Feedback
Thanks again for you work on this @thewhodidthis. Overall I think the update is in pretty good shape so merging into main as a better base for future iterations.

See specific comments in-line. I've also added one minor commit to change the cli to process input/output parameters instead of args and remove updates for synced content to avoid conflicts.

## Git Issues

Closes #1211 (new issue to be created for future work)

## Screenshots/Videos


.github/workflows/test-e2e.yml
```diff
@@ -0,0 +1,57 @@
+# Compile and run E2E tests using Cypress.
+name: Test E2E
+on:
+  # Run manually for now.
+  workflow_dispatch:
+  # NB: Dicuss whether this workflow should be configured to run once
```
chrismclarke, line 6: Nice to have the option to triggering as a workflow - I expect likely we will later integrate with the labelling system to define when/if to run but for now this is more than sufficient

packages/scripts/src/commands/e2e-data/index.ts
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env node
+import { Command } from "commander";
+
+import parseCmd from "./parse";
+import { logProgramHelp } from "../../utils";
+
+const program = new Command("e2e-data").description("Parse E2E data");
```
chrismclarke, line 7: I think it makes sense to link up with the existing commands system so that we don't have to install all the added dependencies into the test-e2e repo (and will help linking in the future with existing parsing methods) - also really great to see you were able to see how this side of the codebase functions to make some sensible addition

packages/scripts/src/commands/e2e-data/mappings.ts
```diff
@@ -0,0 +1,40 @@
+export function click(input = "") {
```
chrismclarke, line 1: This mapping is very useful to have, although it may feel a little verbose for now (e.g. pathname just calls pathname) there's a good chance we will need to add more abstraction to the authoring layer so this gives us a place to manage

packages/scripts/src/commands/e2e-data/parse.ts
```diff
@@ -0,0 +1,92 @@
...
+  input: string;
+  output: string;
+}
+/** [WIP] Helps turn XLSX data into cypress compatible test runs. */
+const program = new Command("parse");
+export default program
+  .description("Parse worksheet into JS!")
+  .requiredOption("-i --input <string>", "path to input xlsx file (relative to project root)")
+  .requiredOption("-o --output <string>", "path to output spec file (relative to project root")
+  .action((options: IProgramOptions) => {
```
chrismclarke, line 20: I made these changes (previously input was an arg and output piped to file via script). The main reasons for doing so are:
1) A bit easier to scale if the list grows (can add type interface and order not important)
2) Piping output makes it harder to add debug logs (would be printed to the file) and can have incompatibilities depending on os and client (`>` operator may or may not exist)

But I did think it was a really need solution to essentially use the logs to write line-by-line. I had to find another workaround instead by storing an array of responses and concatenating at the end

packages/scripts/src/commands/e2e-data/snippets.ts
```diff
@@ -0,0 +1,18 @@
+// NB: Might a Map be better in this case?
```
chrismclarke, line 1: I think this is mostly just a case of personal preference. I know some people have a strong opinion on named vs default exports (and how to go about it) - probably the best write-up I know in favour of named is https://basarat.gitbook.io/typescript/main-1/defaultisbad. But what you have here should still fully support type inference so I think fine either way

packages/test-e2e/README.md
```diff
@@ -0,0 +1,9 @@
+# Test E2E
```
chrismclarke, line 1: I've added this as a placeholder readme with minor script change for input/output paramaters

packages/scripts/src/commands/e2e-data/parse.ts
```diff
@@ -0,0 +1,92 @@
...
+    // Setup Input and Outputs
+    const { input, output } = options;
+    const inputPath = path.resolve(ROOT_DIR, input);
+    const outputPath = path.resolve(ROOT_DIR, output);
+    fs.ensureDirSync(path.dirname(outputPath));
+    const workbook = XLSX.readFile(inputPath);
+    const outputs = ["/** This file has been automatically generated. Do not modify directly */"];
+
+    // Convert XLSX to JSON
+    const worksheets = ["Run", "Specify", "Describe"].reduce((cargo, v) => {
```
chrismclarke, line 30: At first I was confused why we weren't using the existing xlsx conversion but actually it makes a lot of sense for this to be handled independently right now given that we want to link data across the different sheets and create individual (not merged/collated) outputs. It also means development on the e2e feature won't be blocked by limitations with the existing parsers, so definitely good to keep for the short term at least

Issue #1283, opened 2022-03-22
[BUG] App takes long time to open on TECNO ITEL AC13 phones


EsmeeIDEMS:
**Describe the bug**
Testers observed that the TECNO ITEL AC13 smartphones take a significant time to open the app (the screen recording below shows 22 seconds from clicking the icon to having a fully rendered home screen, but the testers reported a case of it taking 3 minutes). 

The splash screen appears immediately, but then it disappears and for 8 seconds the screen is white. After that the screen turns light blue, and the home screen renders. Once the app is open, templates render at a reasonable speed. 

Would it be possible to display something during this wait time to indicate to users that something is happening? 

**App version**
v0.14.2 (Play Store release `international.idems.plh_teens`) 

**Screenshots and videos**
Here's a screen video taken on a TECNO ITEL AC13 phone. Opening the app takes 22 seconds (the app icon is pressed at 0:04 and the home screen is fully rendered at 0:26). The screen is white from 0:09 to 0:17. 

https://user-images.githubusercontent.com/74557272/159519303-beecb052-6e19-4faf-ad76-d7d5c0f44bc4.mp4

**Additional context**
For reference here's a screen video taken on a 5yo high end smartphone, in which opening the app takes 8 seconds. 

https://user-images.githubusercontent.com/74557272/159518787-dc80db67-2823-426f-8d87-a357697d7000.mp4

EsmeeIDEMS:
Just noticed the Play Console pre-launch report picked up on slow rendering as well.

<img width="956" alt="image" src="https://user-images.githubusercontent.com/74557272/160607389-09e97c37-9b09-4e04-91b5-9f0944c9eab2.png">

chrismclarke:
Thanks for the recordings, really handy. I expect the act of running the recording software probably also slows the device down somewhat, but would still be good to improve on this. 

The white-screen delay is when the app is running all its start-up logic. I think there might be some short term gains to be had just by letting the splash screen stay up a little longer (it's meant to autohide after 3s, although it seems to stay a bit longer already anyway) and trying to reduce the number of scripts run on startup

The blue-screen delay is more down to the rendering of templates themselves, and so this is a harder one to improve on without larger changes to how template data is retrieved and displayed. 

As for the performance reports - nice to have although I expect they will be slightly misleading as any startup goes into the various splash screen and logo images page and not the main app screen so it's hard to know exactly what it is measuring.

PR #1312, opened 2022-04-01
Perf/home page

chrismclarke:
PR Checklist

## Description

- Increase the max time splash screen could be shown from 3s to 7s
- Try to defer some of startup logic until after main template displayed
- Remove unused survey package

## Review Notes
It's hard to measure what impact this will have on the specific devices (we don't have any automated timing tests), but would be good to get what feedback we can from the testers of #1283 after merge and release

As this does change the startup it would be good to do a quick functional test from a clean device for the first load (can use the appetizer link), and then closing the app and trying a subsequent load (just to ensure things like the initial splash screens display once and only once, and the templates load correctly)

## Git Issues
Progress towards #1283

## Screenshots/Videos

The package removed was a particularly bulky one (weighing in at roughly 1.4MB / 10MB for the total javascript first parsed when app loaded). This doesn't have a huge impact as the android version has all js in-built (so doesn't have to request from a server), but may help a small amount with some of the initial rendering time.

Likely larger gains would be found by removing the sheets data from being preloaded to read on demand (discussed previously with plans to migrate to individual json files instead of grouped ts files) 

image


volloholic (approved):
@chrismclarke I have approved but we should discuss the survey component at some point. I assume at some point we felt we would use this but can't remember what the decision points were.

chrismclarke:
> I have approved but we should discuss the survey component at some point. I assume at some point we felt we would use this but can't remember what the decision points were.

@volloholic - The survey component and service were originally designed to wrap around the survey-js system for designing forms (originally user on-boarding questions), however serves little use now as we were able to extend the templating system to achieve most of what we needed short term, and long term we would be planning to create a mapping from odk -> template instead of authoring -> surveyJS -> template

The only reason why I haven't removed totally is because at the time for convenience we used the survey system as a way to show the initial fading intro slides (because the survey system was already set up to display things in full-screen at the start), but really no reason that shouldn't also be handled by the templating system (todo item that will likely need prioritising when we next want to change the slides for another app or update)

EsmeeIDEMS (changes requested):
Seems to be working fine in general. I noticed that the navigation button on the organisation select template in the start up screens is no longer positioned at the bottom of the screen.

image

EsmeeIDEMS:
I noticed also this page looks quite bad on Appetize: 
image
I'm expecting this is unrelated to this PR?

chrismclarke:
@EsmeeIDEMS - I ran visual tests locally and all fine from the code side so I'm guessing it's an issue with the emulators themselves. At some point in the future it might be worth testing on appetizer to see what device/android version combinations gives us the best results, but for now I'll merge here and we can check once deployed live on a physical device.

PR #1313, opened 2022-04-02
feat: add crashlytics, custom error handler, perf monitor

chrismclarke:
PR Checklist

## Description

- Add support for automated crash reporting via firebase crashlytics
- Add support for custom error reporting to firebase crashlytics
- Add support for automated performance monitoring to firebase performance monitor
- Cloud build fixes (updating node & java versions)

## Review Notes
There shouldn't be any noticeable change for frontend users, but as this makes a number of core android testers it is strongly recommended that after release to internal testing we encourage users on a range of devices to ensure the app loads as expected and do a very brief functional test of a few main screens. After launch it would also be good to check the firebase dashboard reports

## Author Notes
I can't remember if we already mention in privacy policy/terms any reference to firebase. The new tools (crashlytics and performance monitoring) should be covered by the same documentation, but probably should have some explicit reference just to say that we are using them: https://firebase.google.com/support/privacy

## Developer Notes
The automated performance monitoring is mostly just app startup times and possibly screen freeze. There are integrations that allow for specific tracking during actions, e.g. we could add a manual detection from when the splash screen is dismissed until we mark the home page template ready for display, or from start of a template render until complete, however this would need a bit of extra consideration to avoid being too obtrusive. Another option might be to include performance timings as part of automated testing, to track speed across specific user journeys.

## Git Issues

Closes #

## Screenshots/Videos

Crashlytics dashboard available at https://console.firebase.google.com/project/plh-teens-app1/crashlytics. Includes android crash reports as well as runtime javascript errors (existing data during dev testing)
image

Crash/error reports also contain core device info
image

Performance monitoring dashboard with some basic initial metrics
image


EsmeeIDEMS (commented):
I don't think there's much here to test, in any case I did a quick functional test and the app seems to be running fine. 

- The crashlytics page only shows for the ZA app, not for TZ. Is that what you'd expect at this point? 
image
- We're already mentioning Firebase in the privacy policy, however we're referring to this page https://firebase.google.com/policies/analytics . Should we include the other link explicitly as well?

volloholic (approved):
@EsmeeIDEMS can you edit the T&C to include the link.

PR #1327, opened 2022-04-13
Feat/latex component and pipe

### LINKED ISSUE #1281: [FEATURE] - Add component that renders latex
**What?**
A specific component that accepts text formatted as latex and renders

**Why?**
Useful to have, particularly for maths app

**How?**
Likely easiest way would be to add a package like [katex](https://katex.org/), and use a standalone component that renders. I'd probably also recommend using a pipe like we do with markdown (see `src\app\shared\components\template\pipes\markdown.pipe.ts`) to handle any value updates and give the option if we later want to include latex support for existing components instead of just standalone

**Other details**
Documentation page for adding components: 
https://idemsinternational.github.io/parenting-app-ui/developers/adding-components/

jfmcquade:
PR Checklist

## Description

Adds a new `latex` component type that can be used to render LaTeX.
Also adds a new `latex` pipe for use in other components.

**Currently in a draft state pending input.**

## To discuss:
- How to avoid including the data files in the commit – I merged my updated master branch before pushing my feature branch
- The component uses an existing angular module `ng-katex`, which is neat but only accepts one kind of LaTeX "math mode" delimiters, namely `$...$` for in-line maths, and `$$...$$` for displayed equations.
- The pipe is written independently of the `ng-katex` module, and can additionally accept LaTeX input which uses alternative delimiters, namely `\(...\)` and `\[...\]`.
- Is it important for the component to accept both LaTeX syntaxes?
- What is the value of using an existing package vs adapting it to our own code?
- If we want the component to accept both forms of LaTeX input, would it be bad practice to have the component work via an application of the pipe?
- Possible security issues with using `domSanitizer.bypassSecurityTrustHtml` in this case

## Git Issues

Closes #1281

## Screenshots/Videos
Component demo
Screenshot 2022-04-13 at 18 07 55

[Example Sheet](https://docs.google.com/spreadsheets/d/1qPNAFQbrEQX8ZahmTr3LGhx3uoaooQJqVOAF2HHwmcg/edit#gid=3937162)
Screenshot 2022-04-13 at 18 09 34

Documentation
To be added


chrismclarke (commented):
Really great you could get this working!

I think you've raised some good points in the description worth talking through on a call. A couple quick notes:

> How to avoid including the data files in the commit – I merged my updated master branch before pushing my feature branch

Yeah managing the data files independent of code is a pain. We have mid-term plans to update the way things are handled so that content is in standalone repos, but until that happens the only tip I would have would be using `flow_subtype` as a means to provide a bit more separation (e.g. `component_demo_latex` would put in a standalone file) and then being a bit more selective on what is committed/discarded. Resolving won't be a big issue in any case, I can always just rebase onto a clean branch if we don't merge the content changes into master in the meantime

> The component uses an existing angular module ng-katex, which is neat but only accepts one kind of LaTeX "math mode" delimiters, namely $...$ for in-line maths, and $$...$$ for displayed equations.

Funnily enough it looks like the `ng-katex` module includes `extract-math` as a dependency, so there should be a way to use any delimters that katex supports. That being said, it looks like the [ng-katex repo](https://github.com/garciparedes/ng-katex/tree/master/projects/ng-katex/src/lib) is actually just a very thin wrapper around vanilla katex anyway (the service just provides to methods to call katex render methods), and given that it hasn't been updated in a couple years (whilst katex itself seems to be under active development) I'd be tempted to say we just use katex directly and provide our own wrappers as necessary (you've already done most of the work with your pipe system anyways which is nice)

> Is it important for the component to accept both LaTeX syntaxes?

Always a good question. My feeling is we should start with whatever is easiest and add complexity as it arises. Given that the user will explicitly using a `latex` type column I think we can be opinionated with what that means for typesetting instead of trying to match everything (expose whatever default options katex offers). If we have more complicated use cases likely the workflow scripts would be a better place to convert formats to reduce overhead on the app itself.

> What is the value of using an existing package vs adapting it to our own code?

Depends on a few factors, as mentioned in 3 I think this is a good case for just writing our own methods. The (minor) red flags for me are just that ng-katex hasn't been updated in a while, and I can see open issues for angular 13 support (and prs for various security updates). 

> If we want the component to accept both forms of LaTeX input, would it be bad practice to have the component work via an application of the pipe?

Pipes are quite useful because they handle change detection. E.g. if you are relying on ngOnInit hook for processing content then it won't reprocess if the component receives new data. There are ways to deal with this, but I find pipes are perfectly places for content transformations

> Possible security issues with using domSanitizer.bypassSecurityTrustHtml in this case

Converting latex will always mean injecting dynamic html, so will require bypassing angular dom sanitizer. As the only content is coming from our own excel sheets the security risk is much reduced (compared to a db that anyone could write to) so a valid consideration but one that I probably wouldn't be too worried about (a pretty common use pattern). Specifically for katex, I'm guessing if a user did try to include malicious `<script>` tags the katex parser will just throw error instead of rendering raw (could be tested). As this also applies to the `html` component a more secure solution would probably to again leverage build workflows and parse any content through a 3rd party sanitizer (e.g. dompurify) to check for malicious code - but again, this would be coming from our trusted authors so more a nice-to-have rather than required essential.


Overall though definitely really good progress. I think the next steps will probably be:
- Tidy up master or rebase this PR for easier future dev
- Consider replacing ng-katex with pipe methods
- Figure out how best to manage typesetting options (depending on whether using ng-katex or not)

src/app/shared/components/template/components/latex/latex.component.ts
```diff
@@ -0,0 +1,23 @@
...
+@Component({
+  selector: "template-latex-component",
+  template: ` <span [innerHTML]="latexInput | latex"> </span> `,
+})
+/** Render raw LaTeX **/
+export class TmplLatexComponent extends TemplateBaseComponent implements OnInit {
+  latexInput: string;
+
+  constructor() {
+    super();
```
chrismclarke, line 13: You only need to make calls to the `super` class if using the constructor (e.g. injecting another service). So you can simply omit here

src/app/shared/components/template/components/latex/latex.component.ts
```diff
@@ -0,0 +1,23 @@
+import { Component, OnInit } from "@angular/core";
+import { TemplateBaseComponent } from "src/app/shared/components/template/components/base";
+
+@Component({
+  selector: "template-latex-component",
+  template: ` <span [innerHTML]="latexInput | latex"> </span> `,
```
chrismclarke, line 6: It would probably be easier to just bind directly to `_row.value` instead of re-casting to the `latexInput` variable. This also removes any need for `ngOnInit` and `getParams` so actually everything will just be handled by the pipe

src/app/shared/components/template/pipes/latex.pipe.ts
```diff
@@ -0,0 +1,59 @@
...
+
+@Pipe({
+  name: "latex",
+})
+export class LatexPipe implements PipeTransform {
+  segments: Segment[] = [];
+
+  constructor(private domSanitizer: DomSanitizer) {}
+
+  transform(latexInput: string): SafeHtml {
```
chrismclarke, line 15: Overall I think this is a nice and tidy implementation, and definitely better than having to import the whole `ng-katex` library. It would be good to have a comment either above the `transform` function or the class name just to give a quick overview of what this does (i.e. checks for latex within text string with various delimeter formats, and renders as html)

src/index.html
```diff
@@ -1,29 +1,34 @@
...
+    <!-- add to homescreen for ios -->
+    <meta name="apple-mobile-web-app-capable" content="yes" />
+    <meta name="apple-mobile-web-app-status-bar-style" content="black" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500&display=swap"
+      rel="stylesheet"
+    />
+    <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet" />
+    <link href="https://cdn.jsdelivr.net/npm/katex@0.15.3/dist/katex.min.css" rel="stylesheet" />
```
chrismclarke, line 28: We don't want to use cdn imports as that will require all devices to access the internet when running the app. 
The two alternatives would be:
1. Download the css file manually, put in the assets folder and link from there
2. Use the `angular.json` file which has configuration option to auto-copy/inject styles from `node_modules`

My preference would be option (2) as that means the styles will automatically be updated as we update the package.json. You can see the example of how we imported the styles when using `introjs.css`

This also flags a few extra imports which I've pretty sure are no longer used/required (I expect legacy of default angular repo), I'll open another PR to remove

chrismclarke (changes requested):
Nice work overall, I think the changes to bypass ng-katex and add the dompurify santizer all look good. I've tidied up the PR to just reflect the main changes and added a few comments inline to address. Once changes made let me know either on slack (or re-request review) and I think should be ready for merge

chrismclarke (approved):
Changes look great to me, thanks for taking all the feedback items into consideration.
I think we are good to merge here and add any documentation as a follow-up

PR #1342, opened 2022-04-21
Fix/populate assets

chrismclarke:
PR Checklist

## Description

Since refactoring to use json data from assets instead of in-memory ts files angular has been unable to detect changes made when new templates synced as the app is running. This is due to the way its compiler copies and caches data populated from our `app-assets` folder.

This PR removes the reliance on angulars tooling to populate the `app-data` at runtime, and instead manually copies on demand which should fix the issue of angular caching.

A few additional minor fixes include:
- Hooks to automatically copy the app-data as part of install process (in case user does not sync after checking out repo or ci-build process)
- Improvements to the way we calculate what data needs to be copied into the app (previously all data would be cleared and repopulated, now better folder replication methods copy only files with content changes)
- Sorting output of gdrive asset download to keep contents files in better order

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade (approved):
Can confirm that the templates do update as expected when the data is synced when the app is running. I don't feel confident to comment on the code, but it works as expected

PR #1343, opened 2022-04-21
feat: add translator_mode action

chrismclarke:
PR Checklist

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/data-models/constants.ts
+++ b/packages/data-models/constants.ts
@@ -36,6 +36,7 @@ const APP_FIELDS = {
   SERVER_SYNC_LATEST: `${FIELD_PREFIX}._server_sync_latest`,
   APP_LANGUAGE: `${FIELD_PREFIX}._app_language`,
   DEPLOYMENT_NAME: `${FIELD_PREFIX}._deployment_name`,
+  TRANSLATOR_MODE_ENABLED: `${FIELD_PREFIX}._translator_mode_enabled`,
 };
 
 const APP_LANGUAGES = {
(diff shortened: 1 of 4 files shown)


EsmeeIDEMS (commented):
You mentioned in the notes that this does not work perfectly with dynamic references - I guess both of the issues below are instances of this.
- example_lang_translator_mode: I don't understand why the text in `enabled_text` does not show `(none) ·`
image

- [topic_intro](https://docs.google.com/spreadsheets/d/1WXhNzkP-_740aw-3i-VtNF3mfKiuJMgAOrE0-qyDh_A/edit#gid=1013546368): the string `Hear @global.topic_intro_name's message` appears translated in Afrikaans as `Hoor Sbo se boodskap` (I'm actually surprised that works, the translation only exists when the global is already evaluated to Sbo). When displaying in translator mode things get messy, it shows `Hear (none) · Sbo's message` - so none must refer to the global, and the complete string does not get translated. 
image

EsmeeIDEMS (changes requested):
I found the following bug: 
https://user-images.githubusercontent.com/74557272/164434850-c5d4ad12-47c8-469b-b051-75af761193fb.mp4
It seems that going to / popping up a template that is set through a local variable no longer works, since the local variable string is adapted.

chrismclarke:
Thanks for testing @EsmeeIDEMS 
I'm going to revert this back to draft as I just can't find a way to get things working reasonably.

There's too many combinations of factors to make this feature possible, namely:
- Text can be defined statically within a sheet or dynamically loaded from globals or data lists. Particularly for the case of data lists we don't know where in the template the data will be used, so can't be selective about making sure to translate text that will be displayed vs not translating variables that will be included in an action list
- Translation methods are called from multiple locations, including when rendering templates as well as pre-processing things like global lists
- Often dynamic text goes through multiple stages of parsing
- A lot of text displayed sits within additional fields such as paramter_lists (not just value)
- Translation happens before dynamic evaluation (leaving things like @global as @global), and so we can't impose much change in the translation layer

I think if we want something that will work better we would really need to redesign from the ground up with those aims in mind.

PR #1219, opened 2022-02-03
fix: redirect to home page after reset

chrismclarke:
PR Checklist

- [x] - Latest `master` branch merged
- [x] - PR title descriptive (can be used in release notes)

## Description
Previously reset app would reload on the current page which could lead to some unexpected behaviours. Now reset redirects to home screen

## Git Issues

Closes #

## Screenshots/Videos

https://user-images.githubusercontent.com/10515065/152435566-11be77e6-ccda-47fb-8223-d84cdcb89875.mp4

PR #1348, opened 2022-04-27
fix: strip bold markdown from notifications

### LINKED ISSUE #1314: Strip markdown out of notifications
**Details**
We can't interpret markdown in notifications for now, but then at least strip it out. Related to #1249

chrismclarke:
PR Checklist

## Description

Strips any markdown `**` from notifications as this does not show correctly
Small workaround to #1219

## Author Notes
As we need to manually specify what markdown to strip I've just focused on bold text (the only example I could see). Are there any others we need to remove for now? I'm aware of not being too heavy-handed in case another author wants to use markdown characters as text (e.g. hashtags and underscores)

## Git Issues

Closes #1314

## Screenshots/Videos

Scheduling bold-markdown text shows the text that gets modified before sending

https://user-images.githubusercontent.com/10515065/165499396-6eaea873-e3c1-4155-832e-e7cc0354f466.mp4

Here is the message that arrived on my local device (ignore the bold heading, that isn't due to markdown just because the notification title and text are the same)

image

### DIFF
--- a/src/app/feature/campaign/campaign.service.ts
+++ b/src/app/feature/campaign/campaign.service.ts
@@ -235,6 +235,10 @@ export class CampaignService {
     if (!notification_schedule) return;
     let { _schedule_at } = notification_schedule;
 
+    // HACK - remove markdown form title and text as not currently supported in capacitor notifications
+    row.text = this.hackStripNotificationMarkdown(row.text);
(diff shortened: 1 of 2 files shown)


EsmeeIDEMS (approved):
Functional test passed.

PR #1351, opened 2022-04-27
feat: convert splash screens to templates

### LINKED ISSUE #1331: [FEATURE] - Rework legacy intro splash survey for templating system
**What?**
The current plh intro splash is hardcoded into the app and run via the legacy _survey_ service. It would be good if this could be refactored to be linked up with the templating system and run either as a new `animated_slides` component, or integrated with the existing `animated_section` system and extended to provide the full functionality

**Why?**
We want to avoid deployment-specific content hardcoded into the app, and would also be nice to be able to provide a smoother slides/faded transition component type

**How?**
Likely can base from or extend the logic seen in both the existing implementation for the survey as well as the animated_section template component

**Other details**
A similar PR was drafted a long time back, but no idea if still relevant or not (https://github.com/IDEMSInternational/parenting-app-ui/pull/675)

Code for the existing survey can be found in: `src\app\feature\survey\surveys\plh-intro-splash.survey.ts`
image

This creates the intro slides as seen in
image

Once refactored the entire survey module/service should be able to be removed, as well as removing the `run_survey` launch action defined in ` packages\data-models\constants.ts` and implemented in `src\app\app.component.ts`

chrismclarke:
PR Checklist

## Description
Rebase of #1347

Converts legacy splash survey into a component for use in template system.

Template here for intro splash slides, adapted from existing template including legacy comments:
https://docs.google.com/spreadsheets/d/1MJzzdTYDZg0VkS5zxrXY4lKV5Oi954pw1sIVJUOOzWM/edit?pli=1#gid=578428718

Component demo for `animated_slides` here: https://docs.google.com/spreadsheets/d/15eQ1B7BFQKQVdAM1mD1qPa_vGrudfylYVVDOtPenkWI/edit#gid=569531329

Also adds 3 new styles for `display_group` to help with layout. Can discuss whether these are sensible: 
1. `centre_image` - Centres image within its container and sets `img` max-width to 200px
2. `tight` - An optional modifier for the style `two_columns_images`, which tightens up the display grid to fit images in smaller space
3. `two_columns_images_overflow` - To be used on the last image that would otherwise be in a `two_columns_images` display_group when there are an odd number of images. Centres this image and styles it like the others in the grid to maintain symmetry

TODO:
- [x] Remove legacy splash survey and whole survey service/module

## Git Issues

Closes #1331 

## Screenshots/Videos

New intro slides generated from template:

https://user-images.githubusercontent.com/64838927/166953611-418e6f59-1000-4451-9245-d4a7868d6a5f.mov


chrismclarke:
I've just removed the now-deprecated survey code and taken a look through the rest. All looks really good to me, a tidy implementation that looks to work well.

I might have one (non-blocking) question around the `tight` style param for the two-column displays (a bit unclear what that actually means, if we are talking about the max width I'd probably go for something like `width-narrow` or if it is the gutter size between grid items maybe something like `grid-spacing-sm` or similar.... But I think that is likely a moot point because really the two-image style isn't great and should be replaced by a more generic `grid` style with clear customisation options that relate to the css grid spec

As I've taken ownership of the PR I'll mark @jfmcquade to add own review, and then hand over to @EsmeeIDEMS for functional test and merge

jfmcquade (approved):

EsmeeIDEMS (changes requested):
Looks great on sensible screen dimensions. When changing the screen to be very narrow the components (text wrap, image size) do not behave the same as they do in a regular template: 
- text does not wrap, instead it runs off the screen on the right hand side
- image does not shrink, instead it runs off the screen on the right hand side

See terms and conditions page that launches after the intro for comparison. 

https://user-images.githubusercontent.com/74557272/169017896-064a32cb-f12e-4461-aac6-f7f5f15129f0.mp4

EsmeeIDEMS:
Not sure if this is important, but FYI when I move to this branch from master I have the following issue: 
- Two files appear as "changed". 
image
- I cannot discard these changes nor move to another branch in the usual way. The only way I found to move back to master is by running `git checkout -f master`

chrismclarke:
Looking at the sizing issue, yes I think there's some hardcoded minimum values in there, however I think we should still go ahead and merge this first and create as a minor issue for follow-up

The minimum width we consider for devices is 320px which does have some cut-off. This still represents very few devices so I'd suggest it can be a low-priority issue to resolve

image

EsmeeIDEMS (approved):

PR #1355, opened 2022-04-28
style: add 1024px max template width

chrismclarke:
PR Checklist

## Description

Add a max-width of 1024px so that content fits a bit nicer on large screens (desktop).
Whilst this is still probably a bit on the large side, recommend keeping as a reasonable default for now and can refine individual templates in the future for fitting different screen sizes

## Review Notes
I don't expect this to have knock-ons as most of our work is focused on devices with smaller screens, however as I did change a bit of the styles that handle templates filling space it would be good to test via visual test that there are no accidental regressions (e.g. things suddenly becoming much wider or thinner in some template contexts)

Note - this only applies to custom templates, so some of the existing dev screens will still fit full page (this is to allow things like the header bar to stay full width and keep scroll bars on the outer edge of the page)

## Git Issues

Closes #

## Screenshots/Videos
E.g. on my desktop (1920px screen)

Before
image

After
image

### DIFF
--- a/src/app/shared/components/template/template-component.scss
+++ b/src/app/shared/components/template/template-component.scss
@@ -14,6 +14,12 @@ plh-template-component[data-hidden="true"] {
   display: none !important;
   flex: 0 !important;
 }
+plh-template-container {
+  width: 100%;
+  max-width: 1024px;
+  margin: auto;
+}
+
 /// Ensure all directly nested template rows have spacing between elements,
 /// except in case where only one row (e.g. deep-nested)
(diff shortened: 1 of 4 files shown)


EsmeeIDEMS (approved):
Functional test passed, this improves the look of the app on a desktop screen indeed. I think it's worth merging even when leaving as is, but I have these comments: 

- The narrower screen size does not seem to affect tours
image

-The components inside a pop-up do become more narrow, but not the white background rectangle
image

- Also for some reason the "Thank you for your feedback" pop-up bar (that shows up after submitting feedback) has a different width
image

PR #1357, opened 2022-04-29
feat: add support for full screen popups

chrismclarke:
PR Checklist

## Description

- Add support for passing a 3rd arg to action lists which can contain multiple key-value parameters
- Add support for fullscreen popups
- Add example fullscreen popup to [example_pop_ups](https://docs.google.com/spreadsheets/d/1K9JZ9Glnj73CFJmpfGIRa5dRA2TBxSsJmaguV6GL8Pc/edit#gid=402992224) sheet

## Review Notes
I've purposefully included all templates which had knock-on changes due to the parser updates (individual json files makes this so much easier now!). As you can see the only ones are debug templates which included some additional authoring of additional args that aren't currently being used in practice (now included as `params` in the sheet rows).

I don't expect there to be any further knock-ons, but might be good to check a couple pop_ups to be safe.

## Dev Notes
It would have been possible to add a new `pop_up_fullscreen` action but decided better to use the existing action and provide authoring option for passing additional parameters. This had not previously been possible, so conversion scripts modified slightly to allow processing a third argument as parameters, e.g.
 ```
 click | pop_up:example_template | param_1:value_1, param_2:value_2
 ```
Whilst it might be a little messy to have both args and params for action lists (where params are technically just the 3rd arg passed), for now it should keep things interoperable and provide a bit more flexibility moving forward.

## Git Issues

Closes #

## Screenshots/Videos

Demo

https://user-images.githubusercontent.com/10515065/165985403-96452ee0-80e3-4a3f-bacd-e8d78d8a5328.mp4

Updated template
image


packages/app-data/sheets/template/debug/debug_nesting_conditional.json
```diff
@@ -165,7 +165,10 @@
           "action_id": "set_local",
           "args": [],
           "_raw": "click | set_local | var_toggle: !@local.var_toggle",
-          "_cleaned": "click | set_local | var_toggle: !@local.var_toggle"
```
EsmeeIDEMS, line 168: Will syntax 
`changed | set_field: @local.slider_tracker: @local.slider; changed | set_local : slider_answered : true` still work or should this be refactored everywhere to be `changed | set_field | @local.slider_tracker: @local.slider; changed | set_local | slider_answered : true`?

volloholic (commented):
I am keen to be involved in the discussion of Esmee's question as I think there is an interesting point there about the distinction or not between our arguments and parameters.

packages/app-data/sheets/template/debug/debug_nesting_conditional.json
chrismclarke: I've just copied the above and added line breaks to try and see the difference. 
```
changed | set_field: @local.slider_tracker: @local.slider; 
changed | set_local : slider_answered : true
```
```
changed | set_field | @local.slider_tracker: @local.slider; 
changed | set_local | slider_answered : true
```
I wasn't aware we are using two different syntaxes for authoring set_local, but seeing as it has only been picked up in the one example template I'm assuming this is actually a typo (?). In either case this shouldn't change the handling of existing set_field and set_local operations, but does open the door if we want to add support for providing a list of values to change e.g.
```
changed | set_field | field_a:value_a, field_b:value_b, field_c: value_c;
```
This would still require some additional code changes to the set_field and set_local actions (and anywhere else we wanted to provide), but could be useful to have. In this case we would aim to keep both options available at least in the short term

chrismclarke:
> I am keen to be involved in the discussion of Esmee's question as I think there is an interesting point there about the distinction or not between our arguments and parameters.

Agreed. It's a bit of a tricky tradeoff as args always remove one step of authoring (the position implies the prefix, e.g. `changed |` instead of `trigger: changed`), however we then are typically limited with how much additional info we can pass (given that args will always need to be in order and the code gets very messy very quickly when saying some actions can read args 1->n and others 2->n depending). 

I think focusing solely on the case of action lists having two patterns is not too unmanageable, namely
```
trigger | action : arg
```
e.g. `click | goto : template_1`
and
```
trigger | action : arg | params
```
e.g. `completed | pop_up : template_1 | fullscreen:true, showCloseButton: false`

But agreed it still isn't very clear what we mean by an argument or parameter in this case, and whether there should be. Perhaps replacing params with `options` might be semantically a bit clearer, or some other means to distinguish what information is required vs what is optional

packages/app-data/sheets/template/debug/debug_nesting_conditional.json
EsmeeIDEMS: (I just checked the file history and indeed it looks like this was just an authoring typo)

PR #1367, opened 2022-05-11
feat: add error message if tech support request doesn't sync immediately

### LINKED ISSUE #1360: [FEATURE] Notification when sync didn’t go through
**What**
Users should be notified when a sync didn’t go through.

**Why?**
We've had a case of a user submitting a request for technical support, which did not get synced to the database until 3 weeks later. 

**App version**
v0.15.2

jfmcquade:
PR Checklist

## Description

Attempts to sync data at the point of the user submitting a tech support request. If this fails, they receive a custom error message

## Git Issues

closes #1360 

## Review Notes
Testing from `/feedback` page (manually input into url bar as no button in menu)
Feedback will fail to submit when running locally (no server used locally) but should pass from preview link

## Screenshots/Videos
https://user-images.githubusercontent.com/10515065/168126635-85344073-369f-4cb5-9c31-c3d30ff1cd09.mp4


chrismclarke (approved):
As discussed on the call I've refactored where these messages are stored and updated the frontend bindings. I've also added #1372 for the issue of integrating these hardcoded strings into translation systems

Otherwise the code looks good to me, thanks for the addition.

EsmeeIDEMS (approved):
Looks good to me. Only issue I noticed is that I don't seem to be able to sync to the server on  the preview link, but since this is also true on the master preview link I think it's not related to this PR. I recorded this issue separately in #1381

PR #1368, opened 2022-05-11
chore: runs prettier on all existing files

jfmcquade:
PR Checklist

## Description

Applies prettier's style conventions to across all .ts .json .html and .scss files in the repo. The original aim was to enforce `crlf` line endings on all files, but this also applies prettier's other specifications, like quote mark type, across the board.

## Git Issues

Closes #


chrismclarke (approved):
Thanks for this, all looks good to me - will definitely be useful to avoid large diffs in the future and keeping the codebase more consistent

PR #1378, opened 2022-05-17
Feat: Google auth sign-in (updated)

chrismclarke:
PR Checklist

## Description

Builds on #1374, with following updates:
- Add deployment config options to enforce sign-in
- Tidy auth methods and add option to wait for user sign in
- Add methods to allow programatic dismiss of popup template (e.g. dismiss auth sign in template after sign in recorded)
- Integrate native auth methods
- Refactor firebase integrations to support v9 changes
- Unify web/native implementation methods

## Dev Notes
Testing locally will require native build and manually registering local debug signing SHA-1 on firebase project

## Review Notes
There is currently no easy way to review. This is because firebase auth needs additional permissions specified depending on environment, for web we need to authorize the deployment web domain (which changes every deploy preview), and on native we need to authorize the app build signing key (which is different for preview builds compared to main deployment).

So we will likely need to create an internal release, test and adapt as required.

## Author Notes
There is a new option to enforce login available to deployment config
```
config.app_constants.APP_AUTHENTICATION_DEFAULTS.encorceLogin = true;
```
It should be noted however that simply enabling this will not work on other deployments until manual steps are taken to record the play store build in the list of approved native auth provides (via play store sha1 key) and web domain also. 

Additionally there is currently no good way to make the auth popup appear at any other place than the start, due to the issue previously identified (that first-launch-actions are bypassed on reload) and because if we wanted to show the screen after it would also be after the tour which would be a bit weird (showing home screen tour and then loading additional popup). All of this can be tidied in a future pr once we confirm the core is working as expected.

## Git Issues

Closes #1374

## Screenshots/Videos

Device demo - auth enforced even after restart. After login success no longer shown on restart.

https://user-images.githubusercontent.com/10515065/168823663-c1941edf-03ee-4be7-8a8e-f764b9ce2164.mp4

Example google play store page where production sha-1 key needs to be retrieved to allow a specific app access to firebase native auth
image

Example firebase console page where sha-1 keys registered for specific app
image

Additional firebase auth page where permitted domains registered to allow access to firebase auth on web (localhost always allowed)
image


jfmcquade:
I've found quite a specific bug where you can't get past the sign in screen if it's the first time you've set up your Google account on the phone _and_ that Google account requires that you set up a lock screen for the device. It's solved if you close and reopen the app. See video.

I've tried logging in on a fresh emulator with a new Google account that doesn't enforce a lock screen, and I got the expected behaviour where, after logging in, I was returned to the app and the lock screen was bypassed.

It seems that in this instance the `authService.waitForSignInComplete()` promise is not resolving properly? Could it even be timing out?

Not sure if this is a blocking bug as (as far as I can tell) the conditions are very specific.

https://user-images.githubusercontent.com/64838927/168836073-89d63a24-b472-4ab9-9d5c-6d9af680c7c4.mov

chrismclarke:
That is indeed weird - I'd say for now we call it a feature :D 
(anyone downloading the app will need play store so will also need to have registered a google account on their device)

jfmcquade:
> That is indeed weird - I'd say for now we call it a feature :D (anyone downloading the app will need play store so will also need to have registered a google account on their device)

Yeh that makes sense. The native sign in experience is really nice if you've already got a registered Google account.

One more thing – do we need to associate the Google User email address/other info with our user meta data and store it in the database?

EsmeeIDEMS (commented):
I'm assuming `encorceLogin` is a typo and should be replaced by `enforceLogin`?

I'm able to run the app on this branch without problems, but will test further on a release once this is merged.

chrismclarke:
Typo fixed and conflicts resolved

EsmeeIDEMS (approved):

jfmcquade (approved):
LGTM

PR #1379, opened 2022-05-17
Update EFM web-preview 17/05/2022

EsmeeIDEMS:
PR Checklist

## Description
PR updates the web preview content and code as of 17/05/2022.


EsmeeIDEMS:
@chrismclarke I added the tags `test-visual` and `test-appetize` here but they don't seem to have worked. Is this because I'm not merging into master?

chrismclarke:
That is correct, the test-visual needs to compare against previously generated screenshots which we only track on master branch.

I think we probably could fix the appetize link to work on any branch, however I'm still somewhat reluctant to make too much use of it as the appetize link is specific to a personal account so I wouldn't expect to be sharing with any wider testing team (60min monthly quota would disappear quite quickly). More just to test when we really need to check a specific native feature which can also be done manually by people like myself of Johnny

PR #1394, opened 2022-05-24
Fix/template overrides

### LINKED ISSUE #1391: [BUG] Deployment specific template overrides do not work
**Describe the bug**
Deployment specific template overrides no longer work. Global overrides do. 

**App version**
v0.16.0

**To Reproduce**
E.g. the language list (at first app launch and found under hamburger menu > settings): the deployment `plh_global` should show the template `language_select`, `plh_tz` should show `language_select_tz` (= only English and Kiswahili) and `plh_za` should show `language_select_za` (= all languages but Kiswahili). Currently all deployments show `language_select`.

**Additional information**
It seems that template overrides do work in the example `example_temp_override_default`

chrismclarke:
PR Checklist

## Description

Fix issue where overrides only assigned if in same excel file - now all files are collated together before assigning overrides. 
I've also synced the latest content which should have the correct overrides now populated.

## Review Notes
If running as deployment preview can manually change the `_rp-contact-field._deployment_name` to `plh_tz` or `plh_za` via the developer console and then open the `language_select template` (or any other with overrides) and should show correct. Alternatively can run locally and just change the deployment via `yarn scripts deployment set`

## Author Notes
I noticed a few additional flow types that exist which I assume we want to remove from the deployment version (e.g. `campaign_rows_debug`, `component_demo`, `example_hardcoded`. I've updated the config files to exclude, although would be good to confirm that the `campaign_rows_debug` are indeed not required (I remember at one time we had a few templates there which we did use, but think no longer need)

image

Any other types can be included in the za/tz config files via the `sheets_filter_function`
image

## Git Issues

Closes #1391

## Screenshots/Videos

### DIFF
--- a/.idems_app/deployments/plh/tz.config.ts
+++ b/.idems_app/deployments/plh/tz.config.ts
@@ -4,7 +4,10 @@ import DEFAULT_CONFIG from "./global.config";
 
 const config = DEFAULT_CONFIG;
 config.name = "plh_tz";
-config.app_data.sheets_filter_function = (flow) => !["debug"].includes(flow.flow_subtype);
+config.app_data.sheets_filter_function = (flow) =>
+  !["debug", "component_demo", "example_hardcoded", "campaign_rows_debug"].includes(
+    flow.flow_subtype
+  );
 config.translations.filter_language_codes = ["tz_en", "tz_sw"];
 
 // Override constants
--- a/.idems_app/deployments/plh/za.config.ts
+++ b/.idems_app/deployments/plh/za.config.ts
@@ -4,7 +4,10 @@ import DEFAULT_CONFIG from "./global.config";
 
 const config = DEFAULT_CONFIG;
 config.name = "plh_za";
-config.app_data.sheets_filter_function = (flow) => !["debug"].includes(flow.flow_subtype);
+config.app_data.sheets_filter_function = (flow) =>
+  !["debug", "component_demo", "example_hardcoded", "campaign_rows_debug"].includes(
+    flow.flow_subtype
+  );
 config.translations.filter_language_codes = ["za_en", "za_zu", "za_xh", "za_tn", "za_st", "za_af"];
 
 // Override constants


packages/app-data/sheets/index.ts
```diff
@@ -2734,6 +2734,10 @@ export const SHEETS_CONTENT_LIST: ISheetContents = {
       flow_type: "template",
       flow_name: "language_select",
       _xlsxPath: "global/core_templates/core_templates_settings.xlsx",
+      _overrides: {
```
chrismclarke, line 2737: This was previously missing but now should show correctly

volloholic (approved):
The changes in packages/app-data/sheets/template/component_demo/comp_slider.json do not seem relevant to this fix or am I missing something?
Given the functional test passed this should probably be merged in so that we can release. But I would like to discuss this in our next call as there are some code segments which are not what I was expecting so it would be good to understand the reasoning.

EsmeeIDEMS:
> The changes in packages/app-data/sheets/template/component_demo/comp_slider.json do not seem relevant to this fix or am I missing something?

No, that is a file that Cabrine has been working on. It's here because Chris did a content sync.

chrismclarke:
Sounds good. As a quick overview of the process and override mechanics, we are focused on the scripts which do roughly the following things:

1. Download content as excel sheets (multiple files containing mixed flow types)
2. Process each sheet one-by-one. This forms the basis of our individual output json files as a 1:1 mapping, and is also used for caching
3. Make a second pass of all sheets for any cases where we need to provide reference from one sheet to another. This cannot be done during step (2) as there is no guarantee the referenced sheet will have been processed yet (depends on file name/order, hence the bug we had). 
 
The only use case we currently have for the last step is to populate a reference on the original sheet in the case of template overrides. This is so that when loading a template such as `language_select`, the template can inform us of any conditions where an override exists instead of having to manually search through all sheets just in case.

PR #1395, opened 2022-05-24
Add documentation of two questions

EsmeeIDEMS:
PR Checklist

## Description
PR adds Q & A section to the documentation website and adds two questions to this section.

@jfmcquade @chrismclarke I don't think there is _need_ for you to review this, but if you have any suggestions for improvements before I move forward and add more questions let me know.

## Screenshots
image
image

### DIFF
--- a/documentation/docs/questions/images/display_group.png
+++ b/documentation/docs/questions/images/display_group.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:9ddd3a1d58608377d5936d7a4a619d622dfc331ce3ac4b96cc044c763707ad21
+size 90507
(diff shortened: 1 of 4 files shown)


chrismclarke:
Looks good to me, one suggestion I would include is hidden vs condition column.

I think right now the distinction I have is to always use a condition as this does not populate any content to the page. There may be some occasions where dynamic content is not updated correctly when updating condition that do display correctly with hidden (I can't remember the exact case anymore), however these should be identified as bugs and fixed in the future. 

We can also include a related Q&A about updating dynamic content. Typically content that depends on dynamic variables should update automatically if it is in the same sheet (and I _think_ also any nested child templates of the template that triggered the action), however if a child template makes changes that affect the parent content (e.g. setting a field) then an additional `emit: force_reprocess` or `emit: force_reload` action may be required. For the case of app-wide changes (e.g. language change, enabling developer mode etc. a full `emit: force_restart` action is required that will restart the app.

chrismclarke (approved):

PR #1404, opened 2022-05-30
bug: T&Cs bypass

### LINKED ISSUE #1382: [BUG] T&C screen can be bypassed
**Describe the bug**
At the moment, it seems that the first-app-launch screens can be bypassed by closing the app and reopening. This is problematic in particular because this allows people that install the app to use it without having agreed to the T&Cs. 

**App version**
v0.16.0

**Screenshots and videos**

https://user-images.githubusercontent.com/74557272/169290995-02d4616d-ce09-475e-a50e-fcd789df2a65.mp4

jfmcquade:
PR Checklist

## Description

The T&Cs screen could be bypassed by reloading the app, because the `accept_terms` template was shown as part of the app-specific `app_first_launch_actions`, which only ran the first time the app was launched. This PR addresses this bug by storing a `@field` that tracks whether or not the user has accepted the terms (for a given version of the T&Cs), and doesn't allow progress past the `accept_terms` screen until they have done so.

This PR also reworks the way launch actions are handled so that they can be authored via the templating system. See the template for setting launch actions here: https://docs.google.com/spreadsheets/d/1MJzzdTYDZg0VkS5zxrXY4lKV5Oi954pw1sIVJUOOzWM/edit#gid=1682205958
Other "lifecycle actions" could also be authored in the future, triggered by other "lifecycle events" such as `app_close`, `app_minimize`, `app_first_start` etc.

Each launch action has an associated `@field` that can be referenced in the conditions that determine whether or not the action will run on launch, meaning that the sequence of launch actions effectively "picks up where the user left off" – previously they would all be bypassed on reload/relaunch.

## Git Issues

Closes #1382

### The launch_actions template
Screenshot 2022-06-09 at 13 32 49
From a discussion it was suggested that the `set_field` actions be triggered by the child templates, with the thought that variables could be used within the templates further down the line to help with authoring. I have put the `set_field` for accepting the terms and conditions in the relevant child template, as this is a case of a strict rule for how the page can be bypassed, but for now I have left the other `set_field` actions within this `launch_actions` template, as they are all handled correctly, with the exception of the `start_tour` action, which runs before the tour is completed or dismissed. This was agreed to be acceptable as it is no worse than the previous behaviour of the app. All of the `set_field` actions could be moved to their child templates if there is good reason to do so.

### Launch actions behaviour, including T&C screen bypass bug fix
Does the initial launch now take longer?

https://user-images.githubusercontent.com/64838927/172849858-64f4ec3c-efb9-480d-b54f-07d02af5cd46.mov

### DIFF
--- a/.idems_app/deployments/plh/global.config.ts
+++ b/.idems_app/deployments/plh/global.config.ts
@@ -2,39 +2,6 @@ import { IDeploymentConfig } from "data-models";
 import { DEFAULT_CONSTANTS } from "data-models";
 
 const app_constants = DEFAULT_CONSTANTS;
-/** Define default startup actions for plh app */
-app_constants.APP_INITIALISATION_DEFAULTS.app_launch_actions = [
-  {
-    type: "template_popup",
-    value: "splash_screens",
-    first_launch_only: true,
-    condition: null,
-  },
-  {
-    type: "template_popup",
-    value: "accept_terms",
-    first_launch_only: false,
-    condition: "terms_not_accepted"
-  },
-  {
-    type: "template_popup",
-    value: "language_select",
-    first_launch_only: true,
-    condition: null,
-  },
-  {
-    type: "template_popup",
-    value: "organisation_registration",
-    first_launch_only: true,
-    condition: null,
-  },
-  {
-    type: "tour_start",
-    value: "home_screen_tour_intro",
-    first_launch_only: true,
-    condition: null
-  },
-];
 
 /**
  * The default config should ideally be a superset of any extended configs
--- a/.idems_app/deployments/plh/global.config.ts
+++ b/.idems_app/deployments/plh/global.config.ts
@@ -1,7 +1,7 @@
 import { IDeploymentConfig } from "data-models";
-import { DEFAULT_CONSTANTS } from "data-models";
+import { getDefaultAppConstants } from "data-models/constants";
 
-const app_constants = DEFAULT_CONSTANTS;
+const app_constants = getDefaultAppConstants();
 
 /**
  * The default config should ideally be a superset of any extended configs
@@ -9,7 +9,7 @@ const app_constants = DEFAULT_CONSTANTS;
  */
 const config: IDeploymentConfig = {
   name: "plh_global",
-  app_constants: DEFAULT_CONSTANTS,
+  app_constants,
   google_drive: {
     sheets_folder_id: "1n221Zv9LYuwxmjhiboq8vhQg67_K9L5f",
     assets_folder_id: "1dp9QAQ84m8pm0IBQTSPXe1ramyynKPNn",
--- a/.idems_app/deployments/plh/global.config.ts
+++ b/.idems_app/deployments/plh/global.config.ts
@@ -3,26 +3,36 @@ import { DEFAULT_CONSTANTS } from "data-models";
 
 const app_constants = DEFAULT_CONSTANTS;
 /** Define default startup actions for plh app */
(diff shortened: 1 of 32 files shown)


src/app/shared/services/userMeta/userMeta.service.ts
```diff
@@ -28,6 +28,10 @@ export class UserMetaService {
     return this.userMeta[key];
   }
 
+  getUserMetaUuid() {
```
jfmcquade, line 31: This is needed to avoid a typeError as `getUserMeta()` can now potentially return a boolean, so the result can't be assigned to a `uuid`. There might be a better way around this

chrismclarke:
I think making them more general app_launch and passing conditions is definitely the way to go. I think if we are taking the time to tidy this it might also be worth going one step further and just integrate with the templating system, so that we can re-use condition and dynamic field handling. 

I would probably suggest we make as a fairly generic data list for now, possibly with a `flow_sub_type` of `launch_actions` or `lifecycle_actions` or similar. We can use something a `lifecycle` column to define `app_start` actions (maybe later also `app_open` `app_close` `app_minimize` `app_first_start` etc.), and then standard `condition` and `action_list` columns to define what to trigger. 

Would be keen to discuss/have input from @EsmeeIDEMS and @volloholic

.idems_app/deployments/plh/global.config.ts
```diff
@@ -3,26 +3,36 @@ import { DEFAULT_CONSTANTS } from "data-models";
...
     type: "template_popup",
     value: "splash_screens",
+    first_launch_only: true,
+    condition: null,
   },
   {
     type: "template_popup",
     value: "accept_terms",
+    first_launch_only: false,
+    condition: "terms_not_accepted"
```
chrismclarke, line 17: Ideally we should be tracking via a localstorage variable, e.g. `@fields.terms_v1_accepted`, populating that field on template condition. Ideally this can be handled all within authoring if we move this code to templating syntax instead of hardcoded

packages/data-models/flowTypes.ts
```diff
@@ -319,6 +319,7 @@ export namespace FlowTypes {
     | "audio_pause"
     | "audio_end"
     | "audio_first_start"
+    | "accept_terms"
```
chrismclarke, line 322: I think this is a bit too content-specific to be an action type, probably better to just use `set_field` action and manage via fields

src/app/app.component.ts
```diff
@@ -156,10 +156,19 @@ export class AppComponent {
...
    **/
-  private async handleFirstLaunchDataActions() {
-    for (const initAction of APP_INITIALISATION_DEFAULTS.app_first_launch_actions) {
+  private async handleLaunchDataActions(user) {
+    for (const initAction of APP_INITIALISATION_DEFAULTS.app_launch_actions) {
+      if (initAction.first_launch_only && user.first_app_open) {
+        continue;
+      }
+      if (initAction.condition) {
+        const conditionMet = this.checkLaunchActionCondition(initAction.condition, user);
```
chrismclarke, line 167: There is `evaluateConditionString` method in the template service that can be used to parse any condition (containing dynamic text or not). This would probably be better than define another condition handler, and enable using `@fields.some_field="some_value"`.

EsmeeIDEMS (approved):
Functional test passed.

chrismclarke:
A bit slow to the re-review, just wanted to say thanks @jfmcquade for taking on all the feedback, the updated implementation looks a lot cleaner, good to see this going in

PR #1413, opened 2022-06-16
Feat/android assets workflow

### LINKED ISSUE #1280: [FEATURE] - Populate android assets as part of workflow
**What?**
Add scripts that allow a deployment to populate custom android assets including app icon and splash screen

**Why?**
Android assets are hardcoded into the repo, and so overwriting across deployments will cause conflicts. Additionally the process of changing app icons is quite technical (requires use of android studio to generate), so would be good to have something that could be automated.

**How?**
This is non-trivial as icons typically have a large number of variations, based on the separate foreground and background images and different screen sizes. Likely to avoid having to use android studio either we will need a script that will call underlying android code that generates icons (if available) and notes in the readme of additional software required for install (later can be run through docker container or similar), or using 3rd-party tooling such as [cordova-res/capacitor-assets](https://github.com/ionic-team/capacitor-assets) (probably my recommended approach)

**Other details**
First requires merge of #1210 which has core infrastructure for deployment workflows

Details of how to generate in android studio: https://developer.android.com/studio/write/image-asset-studio

Example generated icon found in `android\app\src\main\res\mipmap-hdpi\ic_launcher.png`
image

Stub in place for creating new android-specific tasks to include in workflows
image

jfmcquade:
PR Checklist

## Description

Adds 3 new workflows:
1. `set_splash_image`, which takes a `splash.png` image file, generates Android's required splash screen files and populates the appropriate folders
2. `set_launcher_icon`, which takes an `icon.png` image file, and optionally `icon-background.png` and `icon-foreground.png` files for creating adaptive icons, generates Android's required launcher icon files and populates the appropriate folders
3. `populate_android_assets`, which runs both of the above operations

By default, these workflow actions expect a deployment's source android assets to be found inside the directory `assets/android`. For the PLH app, directories that sit directly inside `assets` must be named after a language code, and so I've set the path for the android assets to `assets/global/android` in the global config. This is mirrored in the folder structure on Google Drive.

This PR also tweaks the settings for the splash screen in the capacitor config to stop the splash image from being stretched, which should look better.

TODO
- Add support for inputting a colour instead of a background image for generating adaptive icons (can currently be achieved just by supplying an `icon-background.png` of a solid colour)

## Testing

You can test the workflows by syncing the assets to get the source files from google cloud, and then running `yarn workflow sync`. You should see the new files generated in Source Control and populated to the appropriate folders.
To test more thoroughly, try replacing the source files on Google Drive first to confirm that the newly generated files look correct, and then run the app in emulation to confirm that the populated Android assets are being used correctly.

## Git Issues

Closes #1280 

## Screenshots/Videos

Screenshot 2022-06-16 at 10 19 29


.idems_app/deployments/plh/global.config.ts
```diff
@@ -14,7 +14,9 @@ const config: IDeploymentConfig = {
     sheets_folder_id: "1n221Zv9LYuwxmjhiboq8vhQg67_K9L5f",
     assets_folder_id: "1dp9QAQ84m8pm0IBQTSPXe1ramyynKPNn",
   },
-  app_data: {},
+  app_data: {
+    android_assets_subpath: "global/android"
```
chrismclarke, line 18: I would actually think of this the other way round and have a top level `android` config item that can be used to specify `icon_asset_path` and `splash_asset_path`

That shoule help keep code related to android-specific workflows a little more contained. In the future we could also expand to include other android-specific settings, like app id and name, and create additional workflows for configuring android deployment fully

packages/data-models/workflows/misc.workflows.ts
```diff
@@ -27,6 +27,51 @@ const workflows: IDeploymentWorkflows = {
       },
     ],
   },
+  // Generate Android assets from source images (splash.png, icon.png and, optionally, icon-foreground.png and icon-background.png)
+  // Icon images must be at least 1024×1024px, splash image must be at least 2732×2732px
+  // Further specifications provided here: https://www.npmjs.com/package/cordova-res
+  populate_android_assets: {
```
chrismclarke, line 33: The workflows all look good to me, although I think probably they are substantial enough that we can move them to their own `android.workflows.ts` file

packages/scripts/src/tasks/providers/android.ts
```diff
@@ -1,11 +1,92 @@
+import * as path from "path";
+import { Options, run } from "cordova-res";
+import fs from "fs";
+import { ROOT_DIR } from "../../paths";
+import { logError } from "../../utils";
+
+const set_splash_image = async (assetsFolderPath: string, subPath: string) => {
+  const androidAssetsPath = path.join(path.resolve(assetsFolderPath), subPath);
```
chrismclarke, line 8: This may need a tweak if you go ahead and allow users to specify icon and splash asset paths explicitly in config. The config generator automatically hardcodes relative paths, however I expect it will miss the app-data/assets part of the path so that might need fixing

chrismclarke (changes requested):
Overall looks really good to me, just the one main change requested related to how we specify where to pick the icon/splash images from.

Would it also be possible to add a very brief documentation page somewhere, just saying how to use the config to specify a path to icon/splash assets, and possibly size/format restrictions. Bonus points if you want to include example of using foreground/background and single images?

jfmcquade:
Thanks, I've made some changes based on your comments, including adding some docs

documentation/docs/developers/android-assets.md
```diff
@@ -0,0 +1,56 @@
+# Android Assets
```
chrismclarke, line 1: Thanks for adding this, reads very well and handy to have all the extra links and info sources too

.idems_app/deployments/plh/global.config.ts
```diff
@@ -14,9 +14,13 @@ const config: IDeploymentConfig = {
     sheets_folder_id: "1n221Zv9LYuwxmjhiboq8vhQg67_K9L5f",
     assets_folder_id: "1dp9QAQ84m8pm0IBQTSPXe1ramyynKPNn",
   },
-  app_data: {
-    android_assets_subpath: "global/android"
+  android: {
```
chrismclarke, line 17: This looks perfect, thanks for making the changes

chrismclarke (approved):

PR #1417, opened 2022-06-17
Feat: Add subtitles to video component

### LINKED ISSUE #1375: [FEATURE] Video component to show subtitles
**What?**
- The video component should be able to display subtitles. The subtitles are input in an `.srt` file with the same name as the video it corresponds to, which sits in the same folder as the video. 
- Subtitles should be able to be translated. This should work the same as the rest of the app: when translations to the selected language exist they show, otherwise the global default (English) shows. I _think_ a line-by-line translation of the subtitles is most convenient technically, rather than having the full subtitles of the video as one big string - but let me know if you disagree.

**Why?**
To make videos more accessible, also to non-English speakers. 

**Other details**
The app currently only contains one video. This video, together with it's subtitles, can be found [here](https://drive.google.com/drive/folders/12Fcs1NDx_WPhjUyKx_sv5lGwskxj3O3Z).

jfmcquade:
PR Checklist

## Description

Adds subtitles to a video component, where the subtitles file is provided (in .vtt format) via the `subtitles_src` parameter.

When switching language, if the translated subtitles file exists in the relevant folder, this will be used instead. For example, if the the source is set to `subtitles_src: plh_video/lets_slow_down.vtt`, and a `lets_slow_down.vtt` file exists at both `plh_assets/global/plh_video` and `plh_assets/tz_sw/plh_video`, then the relevant file will be used for the respective choice of app language. Otherwise the default `global` file will be used.

Also adds a tool to manually generate translated .vtt files, assuming the string translations already exist. This can be run with `yarn workflow translate_vtt_files`. Currently there is only one subtitles file and one translation of it (into tz_sw), but this tool can be used to generate the translated .vtt files as the original file is translated to other languages.
This tool could doubtless be improved, but it is useful as a stopgap before rethinking and reworking the existing translation system to handle assets more generally.

## Git Issues

Closes #1375 

## Screenshots/Videos

Template:
Screenshot 2022-06-17 at 16 53 58

Subtitles displayed (app language set to en):
Screenshot 2022-06-17 at 16 55 27

Subtitles displayed (app language set to sw):
Screenshot 2022-06-17 at 17 14 47


EsmeeIDEMS (approved):
Functional test passed. For some reason the subtitles display slightly differently on my screen than they do on yours (the grey banner behind the subtitles does not cover the full width of the video). Not sure if it depends on the browser, I'm using Chrome. In any case, I think either display is fine.

image

jfmcquade:
> 

Interesting, I'm using Chrome too (on Mac). It is a native HTML element so indeed there is default styling applied specific to the browser,. The only thing I tweaked was increasing the font size slightly, but the default styling should be fine on any device and be considerate of accessibility etc.

package.json
```diff
@@ -82,6 +82,7 @@
     "nouislider": "^14.6.3",
     "rxjs": "^6.6.7",
     "stacktrace-js": "^2.0.2",
+    "subtitles-parser-vtt": "^0.1.0",
```
chrismclarke, line 85: As the subtitles-parser is only used by the scripts workspace we can move it to the package.json there instead (or add via `yarn workspace scripts add subtitles-parser-vtt`. 

This won't show any error because yarn also checks the top node-modules for packages (usual behaviour if for yarn to hoist node_modules from workspace folders to the top folder where possible to avoid package duplication, although I've disabled this as often leads to issues if one workspace relies on something like webpack 4 and the other webpack 5)

I'll quickly push the change directly.

packages/scripts/src/tasks/providers/subtitles.ts
```diff
@@ -0,0 +1,120 @@
+import fs from "fs";
+import * as path from "path";
+import parser from "subtitles-parser-vtt";
+import { ROOT_DIR } from "../../paths";
+import { recursiveFindByExtension, logError, logOutput, promptOptions } from "../../utils";
+import { TRANSLATIONS_CONTENT_LIST } from "../../../../app-data/translations";
```
chrismclarke, line 6: This import may prove problematic as we are moving app-data folders to be deployment specific. So instead I'll probably recommend importing from the root `src/assets/app_data/translations` folder instead for now, as that path will remain fixed (included in gitignore)

I'll push a quick fix for this also

chrismclarke (approved):
Such a cool feature! I never knew how seamless it was to integrate with web video tags
I've added a few minor bits of housekeeping in fa3ab304c8f8239d318f98dd06800ad47b23e6f8, let me know if all seems reasonable to you @jfmcquade and otherwise I'd say good to go.

EsmeeIDEMS (changes requested):
Although this was working properly before, I'm now getting double subtitles in tz_sw: 

image

EsmeeIDEMS (approved):
Ignore that, the duplication is resolved by a content sync so I probably just resolved the conflicts in the vtt file incorrectly. Thanks!

PR #1438, opened 2022-07-04
chore: remove auth data

chrismclarke:
PR Checklist

## Description

Remove auth data previously copied from firebase to idems db. Now only unique firebase `uid` retained which can be used for lookup on firebase server

## Postdeployment Notes
This will not remove existing data from the database or any additional databases derived from it, and so will require somebody with access and knowledge of such things to clean existing data 

## Git Issues

Closes #

## Screenshots/Videos


EsmeeIDEMS (approved):
Looks good, @jfmcquade can we do this database clean up together?

PR #1439, opened 2022-07-07
refactor: begin system to define colour variables from theme colours

jfmcquade:
PR Checklist

## Description

WIP
Started making system to generate colour variables from given colours. Includes a tool to automatically set the text colour (black or white) based on the best contrast with the background colour. 

## Git Issues

Closes #

## Screenshots/Videos


src/theme/variables.scss
```diff
@@ -19,26 +24,120 @@ $tertiary: hsl(47, 100%, 92%);
 $quaternary: hsl(204, 88%, 94%);
```
chrismclarke, line 24: I hope `quaternary` can be removed... but should probably check first

src/theme/variables.scss
```diff
@@ -19,26 +24,120 @@ $tertiary: hsl(47, 100%, 92%);
...
+  // primary-gradient
+  --ion-gradient-primary-start: #{map-get($colorPalette, "gradient-primary-start")};
+  --ion-gradient-primary-end: #{map-get($colorPalette, "gradient-primary-end")};
+  // secondary-gradient
+  --ion-gradient-secondary-start: #{map-get($colorPalette, "gradient-secondary-start")};
+  --ion-gradient-secondary-end: #{map-get($colorPalette, "gradient-secondary-end")};
+}
+
+.theme-default {
+  @include ionColorPalette;
```
chrismclarke, line 103: incredible how quickly a class gets reduced down to a single line of code. Although I'm wondering within this class we should be defining the $primary and $secondary variables, generating the pallets and passing to the mixin. That way everything for a theme can be encapsulated within a single class.

We can also then move that class to a separate folder (e.g. `themes/default.scss`) and use as a template for scaffolding out further themes

src/theme/variables.scss
```diff
@@ -19,26 +24,120 @@ $tertiary: hsl(47, 100%, 92%);
...
+  //  Borders
+  --ion-border-standard: 2px solid #{map-get($colorPalette, "color-primary")};
+  --ion-border-thin-standard: 1px solid #{map-get($colorPalette, "color-primary")};
+  --ion-border-color-secondary: 2px solid #{map-get($colorPalette, "color-secondary")};
+  --ion-border-light: 1px solid #{$light};
+  --ion-border-light-thicker: 2px solid #{$light};
+  --border-dashed: 2px dashed #{map-get($colorPalette, "color-primary")};
 
-  // Margins //
+  //   // Margins //
```
chrismclarke, line 132: I don't understand the double quoting, in some editors this will show as a crossed-out comment.

chrismclarke (changes requested):
Thanks for this, definitely taking a big step in the right direction, nice to see how we can start to use generator functions to define all we need from a small set of variables.

I've given a few comments regarding folder and file structures which are non-blocking, the main change required will be not to deprecate too early variables that are in active use. Ideally with a feature overhaul this large we will want to be able to migrate in stages, so keeping fidelity with the existing apps would be a strong requirement.

As messaged in slack, we can go a long way to discover what variables are still in use via regex searches. I know we use the ion-color primary, primary-contrast and secondary a lot, so quick regex for any other `--ion-color` produces a handy shortlist

[regexr.com/6pa5b](http://regexr.com/6pa5b)

image

image

namely: medium, quaternary, help-tooltip, slider, primary-lighter, danger

These _legacy_ variables can be collected together with a note that we should migrate in the future.

Similarly we should do regex searches for other variables, e.g. all `var(--ion-` that are not `color` variables:
image

This shows a few more to keep for now, and then all `var(--` that are not `ion` variables
image

Each of the above searches can also be further cascaded to help isolate variables to keep

src/theme/variables.scss
jfmcquade: Ah yeh, just an artefact of commenting out a whole block (including comments) and then uncommenting some of it

jfmcquade:
Thanks, very helpful comments. I've sorted the file structure in a way that I think makes sense. And I've reverted all the variables to active, and then deprecated some that are unused.

I think these changes should be now able to be merged without affecting anything, pending any thoughts you have about the code. 

There are a few things to consider next as I see it:

1. How we treat global colour variables that are currently defined like the following:
```--tile-button-background: #{color.adjust($primary, $saturation: 14%, $lightness: 60%)}```
It would be simple enough to keep this custom variable with a custom adjustment to our defined primary colour, but it would be better if those colour variations are defined via adjustments in the colour palette itself, and then assigned to global variables outside of the colour palette definitions. In which case it might be time to decide on a systematic way to define colour variations.

2. We have a lot of cases similar to this, where colours are defined as variations of our input colours, and are then fed in as values for a gradient which defines another global variable:
```--tile-primary-gradient-start: #{color.adjust($primary, $saturation: 17%, $lightness: 39%)}; // #3aaff8
  --tile-primary-gradient-end: #{color.adjust($primary, $saturation: 2%, $lightness: 13%)}; // #13669a
  --tile-primary-gradient: linear-gradient(
    170.54deg,
    var(--tile-primary-gradient-start) 0.43%,
    var(--tile-primary-gradient-end) 74.07%
  );
```
Again, it would be simple to define the colours as part of our automatically generated palette. But we need to think a bit about where the gradient should be defined – for example, what if one theme doesn't want this variable to be a gradient at all? This is related to thinking about where and how we handle the non-colour-related variables.

It would be good to discuss the next steps on a call soon, I'll see if we can arrange a time this week

chrismclarke (approved):
Thanks Jonny, it's looking really good! I like the folder structure and definitely feels like a big step in the right direction. I particularly like how you've gone about using the spec to generate strong levels of contrast and just how minimal the default.scss file ends up being after all of this.

I expect the next things todo will be:

- deprecate unused variables (as per regex suggestions)
- clarify any additional calcs required (e.g. light/lighter/lightest vs 50 - 900 weight system), following the point you flagged for the tile buttons
- figure out gradients and component-specific overrides (again as you mentioned)
- refactor legacy calcs to use defined pallet instead
- test with another deployment to see how the theming looks
- start adding support for fonts, spacing etc.
- fix up or deprecate theme service depending on what we want to do with it

chrismclarke:
@EsmeeIDEMS - I'm happy with the code improvements but I can see there are still a few potential knock-ons with the visual tests. The purpose of this PR is to lay some groundwork without breaking anything so it would be good if you could work with @jfmcquade to pass functional testing

I've just picked up and fixed one for the background of display groups, but would be great if you guys can check for any others

jfmcquade:
This is ready for review now @EsmeeIDEMS, see Chris's short comment above.

There are still 2 differences showing up in the [latest screenshots from the visual tests](https://github.com/IDEMSInternational/parenting-app-ui/actions/runs/2746224780) (besides the false positives produced by the animations etc which are presumably always present). But I think these are both acceptable.

1. Difference in the text and border colour for `dashed_box` with `style: alert`. 
As far as I can tell, we don't actually use this element in the app outside of comp_demo/debug sheets. Previously the colour was being pulled from a custom hard-coded colour that shared the name "secondary", but is now the actual secondary colour defined globally. Screenshots here, the more yellow colour is before, the more orange colour is after. This change seems fine to me, in fact the more orange text is easier to read on a white background.
Screenshot 2022-07-27 at 12 54 05
Screenshot 2022-07-27 at 12 53 54

2. Slight difference in the audio component play button background.
This is due to the fact that our previous way of adjusting the hue of a colour (using a percentage) has been deprecated in favour of another (using a value in degrees). So while the resulting colour should be identical, there's apparently a slight difference. This is pretty imperceptible, but will affect this component across the board.
example_widget_nesting_audio after
example_widget_nesting_audio before

EsmeeIDEMS (changes requested):
I noticed a couple of changes in the start up templates and tour.

Start-up templates: 
- previously orange nav buttons are now white
- previously blue buttons are now white
- radio buttons and combo box lost their border
image

Tour:
- bullet outlines have disappeared
- previously orange "next" button is now white
- "previous" button lost its border
image

jfmcquade:
@EsmeeIDEMS Ah, thanks. I was putting too much faith in the visual tests. I'll revert the colour variables I deprecated and then do so again more carefully

chrismclarke:
> I was putting too much faith in the visual tests
The visual tests only actually test against debug templates, not every template in the app. The reason for this is partly it would just take an awful long time, and partly because we want the core codebase to function independently of specific content.

Ideally we would have good example templates that roughly cover a kitchen-sink approach (all components and variations), I can see we've started building these but the code hasn't been updated to include so I'll make a quick PR for that now. I expect there will always be some things they do not catch however, such as popup containers, tour containers or any other hardcoded interactive sections. 

We do also include a few specific templates which seemed to cover a reasonable range of content (namely home_screen and weekly_workshops). You can see the full config at [packages\app-data\test\index.ts](https://github.com/IDEMSInternational/parenting-app-ui/blob/master/packages/app-data/test/index.ts), it might be worth thinking following this experience if there's anything else worth adding

EsmeeIDEMS:
> it might be worth thinking following this experience if there's anything else worth adding (@EsmeeIDEMS )

The issues I found were with the tours and the first-launch templates (which are regular templates but look different when launched in the first-launch sequences). I don't know if there is any way to include these in the visual test?

jfmcquade:
> The issues I found were with the tours and the first-launch templates (which are regular templates but look different when launched in the first-launch sequence). I don't know if there is any way to include these in the visual test?

This was the issue, those launch action templates aren't children of our `ion-app` element so weren't having the styling applied via the new theme system.

@chrismclarke I've fixed that with [this commit](https://github.com/IDEMSInternational/parenting-app-ui/pull/1439/commits/04091b4eb3e3d98b0215773dbdc6de0f40fe5b92), let me know what you think of that solution. We will be revisiting the whole theme service soon so potentially this is a stop-gap, but using the theme service seemed like the appropriate way to apply a theme-based class to the document body.

Also the new visual tests seemed to work well, took 11 minutes to run which seems fine. They do show some additional false positives I think, namely these incorrect sizings of icons in 2 places (both in `answer_lists`s). But when I look manually check, there is no difference and the icons appear as they should. I imagine this could be caused by the visual tester taking its screenshots before the elements have finished rendering and having their styles applied?
Screenshot 2022-07-28 at 12 07 45
Screenshot 2022-07-28 at 12 06 46

src/app/feature/theme/theme-service/theme.service.ts
```diff
@@ -28,6 +28,17 @@ export class ThemeService {
 
     this.currentTheme = this.getCurrentTheme();
     this.applyCSSVariablesForTheme(this.currentTheme);
+    this.applyTheme(this.currentTheme.name);
+  }
+
+  public applyTheme(themeName: string) {
+    document.body.classList.add("theme-" + themeName);
```
chrismclarke, line 35: nit(non-blocking) - it might be tidier to just use a data attribute (e.g. data-theme='default') to save adding and removing classes each time

chrismclarke (approved):
Thanks for making the updates, and thanks both for catching the additional issues.
I've left one non-blocking comment which could also just be future tidying, but otherwise I'm happy from the code side.

Thanks also for checking the visual tests. The false positives are a little annoying (likely related to render not fully complete as suggested), but I'd say probably not worth diving too much deeper into for now. I would probably also suggest as a small future update that we also order the files in the output by diff (e.g. `1. example_template`) and possibly limit to only the top 10 or so just to make it easier to manually check if required

EsmeeIDEMS (approved):
Functional test passed, looking good!

PR #1444, opened 2022-07-21
bug: fixed text sizing issue on overflow funder on splash_funders template

### LINKED ISSUE #1443: [FEATURE] Small text on Funders page
**What?**
The text about the anonymous partner on the funders page needs to appear small - in it's current size it draws too much attention. 
 (I think) I don't currently have the means to influence the text size from the spreadsheet level.

**Why?**
_Describe why this feature would be useful and its intended purpose._
_Does this feature solve an existing problem? If so, please add a description or link to a bug report._

**How?**
Possibly using the style list or hard-coding. Only needs to be resolved in this one instance quickly, the ongoing work on the theming system will hopefully provide a neater, more sustainable solution for the longer term. 
 
**Template** 
[splash_funders](https://docs.google.com/spreadsheets/d/1MJzzdTYDZg0VkS5zxrXY4lKV5Oi954pw1sIVJUOOzWM/edit#gid=289625878
)

**Screenshot**
image

jfmcquade:
PR Checklist

## Description

The final funder in the list is represented by text rather than a log. This text needed to be styled to fit with the other funders' logos, but there was no way to do this from the template level.
This PR styles the text correctly by making use of the `display_group` with style `two_columns_images_overflow`.

## Notes

Because the `text` component has the class `large` applied by default, it doesn't seem possible to resize text from the template level, either by using a parameter, e.g. `style: small`, or an explicit style in `style_list`, e.g. `font-size: 50%`. Therefore I used the `subtitle` component, for which the text size can be affected by `style: small`. This seems like a workaround and something that can be addressed with the theming system – I'm not sure why the `large` class is forced on all text, but it means text size can't be affected by parent elements of `text` components, unless I'm missing a way to do it. @chrismclarke we could discuss at some point, probably in the context of theming, and how to apply theme variables relating to text.

## Git Issues

Closes #1443 

## Screenshots/Videos

(The previous styling can be seen in the screenshot in the issue)
Screenshot 2022-07-21 at 12 28 59

### DIFF
--- a/packages/app-data/sheets/template/splash_funders.json
+++ b/packages/app-data/sheets/template/splash_funders.json
@@ -147,84 +147,94 @@
       "_nested_name": "global_funders"
     },
     {
-      "type": "items",
-      "name": "loop_funders_description",
-      "value": "@local.funders_data",
+      "type": "display_group",
+      "name": "global_funders_overflow",
+      "parameter_list": {
+        "style": "two_columns_images_overflow grid-spacing-sm"
+      },
       "rows": [
         {
-          "type": "text",
-          "name": "@item.id",
-          "value": "@item.description",
-          "_translations": {
-            "value": {}
-          },
-          "condition": "@item.description",
-          "parameter_list": {
-            "style": "center"
-          },
-          "_nested_name": "loop_funders_description.@item.id",
-          "_dynamicFields": {
-            "name": [
-              {
-                "fullExpression": "@item.id",
-                "matchedExpression": "@item.id",
-                "type": "item",
-                "fieldName": "id"
+          "type": "items",
+          "name": "loop_funders_description",
+          "value": "@local.funders_data",
+          "rows": [
+            {
+              "type": "subtitle",
+              "name": "@item.id",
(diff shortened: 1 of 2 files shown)


EsmeeIDEMS (changes requested):
Would it be possible to set the small font to be smaller than this? The feedback I got was that it needs to be "much smaller" than the regular text size. Maybe a size similar to the "Lego Foundation" text?

jfmcquade:
> Would it be possible to set the small font to be smaller than this? The feedback I got was that it needs to be "much smaller" than the regular text size. Maybe a size similar to the "Lego Foundation" text?

Yes that's pretty straightforward, I've changed "small" in the spreadsheet to "tiny", so it now displays like this:
Screenshot 2022-07-21 at 13 52 12

EsmeeIDEMS (approved):
Thanks! I wasn't aware there was an option for "smaller than small". I think I now found the full list of options that are available to the spreadsheets: 
`src\theme\deployment\_overrides.scss` which refers to `src\theme\variables.scss`

Issue #1284, opened 2022-03-23
[BUG] Answer list with global does not translate correctly


EsmeeIDEMS:
**Describe the bug**
An item in an answer list that contains a global and also more text (e.g. `Text and @global.language_global_1`) does not translate correctly. The global does get translated, but the text outside the global doesn't. This happens both when used in a radio group and in a combo box.

**App version**
v0.14.2

**To Reproduce**
Launch the template [debug_lang_global_1](https://docs.google.com/spreadsheets/d/1Vp9V0JCmd49ZSXQpC9YftRgDVrkA0xVfJlQ-H1n5pts/edit#gid=1069066475), set language to Spanish. Observe that the first item in the 3rd radio group and in the 3rd combo box is not translated (it says `Text and Esta es un variable global determinada en una planilla global` whereas it should say `Texto y Esta es un variable global determinada en una planilla global`).

**Debug Template**
[debug_lang_global_1](https://docs.google.com/spreadsheets/d/1Vp9V0JCmd49ZSXQpC9YftRgDVrkA0xVfJlQ-H1n5pts/edit#gid=1069066475)
<img width="325" alt="image" src="https://user-images.githubusercontent.com/74557272/159690499-1e99d2de-b265-4f8e-870a-e28f75943e2c.png">

**Template Name**
For instance, [w_instruct_question_1](https://docs.google.com/spreadsheets/d/1mbkICdQZc4RdgpVcLpKuEzAmFIQ10o3rMJaklb46DYM/edit#gid=708552731) with language set to Afrikaans. In the combo box on the second screen the 2nd and 3rd items are not translated. 
<img width="299" alt="image" src="https://user-images.githubusercontent.com/74557272/159690926-360d0dc3-4280-4626-b9fd-ff2a414f0820.png">

EsmeeIDEMS:
Closing this issue as this can be resolved with authoring the answer list names through local variables (for details see #1458)

PR #1458, opened 2022-07-27
Update debug radio globals

chrismclarke:
PR Checklist

## Description
Update debug radio globals to show bug more clearly and include authoring workaround

## Additional Info (from slack)
Just as a quick update, I've looked more into the issue and at the core is the fact that the parameter list passed to the radio-group component is stored as strings retaining | characters, i.e.
0: "name:happy | text: Text and @global.language_global_1"
1: "name:ok | text:ok"
2: "name:sad | text:sad"
So everything that can be translated in a generic way is done so before, but the final form that arrives is more like "name:happy | text: Text and @global.language_global_1" , so the translation doesn't match as intended.

The obvious fix is to move all separation of parameters to a higher level where they can be processed in a more general way, e.g. converted to json {"name": "happy", "text": "Text and @global.language_global_1"}  - this in turn can be translated correctly part by part. However this would be a mid-size refactor so I'm not sure if it's worth undertaking unless we need it more urgently than anything else

Otherwise it is a simple authoring fix by defining the text as a local variable instead (which should translate correctly)

I've updated the example template to include the authoring solution (PR 1458). If we're happy with it for now I'd suggest lowering priority and I'll link this message in the PR for future ref, otherwise we can organise priorities on Friday.

## Git Issues

Relates to #1284

## Screenshots/Videos

image

### DIFF
--- a/packages/app-data/sheets/template/debug/debug_lang_global_1.json
+++ b/packages/app-data/sheets/template/debug/debug_lang_global_1.json
@@ -40,7 +40,9 @@
       "name": "text_2",
       "value": "Text and @global.language_global_1",
       "_translations": {
-        "value": {}
+        "value": {
+          "es_sp": true
+        }
       },
       "_nested_name": "text_2",
       "_dynamicFields": {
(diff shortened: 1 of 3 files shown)


EsmeeIDEMS (approved):
Thanks for providing this authoring solution, tested, works.

PR #1460, opened 2022-07-30
feat: add support for sheet defaults in parser

chrismclarke:
PR Checklist

## Description

**Main Changes**
- Add support for using `@default` top-row in any sheet to populate as default values throughout
- Add support for `@include_default` and `@omit_default` to append or omit default values for a given row
- Create example datalist [example_sheet_defaults_list](https://docs.google.com/spreadsheets/d/1witVIPxVBJsok1c2CtmEOSIfBgm73RNlxdp2RyoDE9w/edit#gid=1694318910)
- Create example template [example_sheet_defaults](https://docs.google.com/spreadsheets/d/1witVIPxVBJsok1c2CtmEOSIfBgm73RNlxdp2RyoDE9w/edit#gid=409059068)

**Additional Changes**
As part of testing I included setting an entire action_list from a datalist as part of an item loop (e.g. `@item.data_list`) and discovered that this was not actually supported (only able to set string values, such as `action_list`  = `click | set_field : @item.field : @item.value`

To add support a moderate refactor of the default parser was required, to tidy up and allow more control over the order of processes (e.g. to allow passing a string as an action list and ignoring subsequent processing)

As this only impacts templates parsed during the sync process it was possible to make changes and cross-check all templates that they still converted in the same way as before (so should have no functional knock-ons)

## Review Notes
All changes are to parser only, so a simple way to check for potentially breaking changes (which I've manually done) is to sync and process content on main branch (ideally committing latest content changes), and then trying to resync and process on this pr branch to confirm the sheets do not change

## Additional Notes
It would be good if somebody could document this feature once approved and merged

## Git Issues

Closes #

## Screenshots/Videos
**Example data list showing how to apply default values, override, append, remove (omit)**

image

**Example template with logged output following button clicks**

image

image

### DIFF
--- a/packages/app-data/sheets/data_list/example_sheet_defaults/example_sheet_defaults_list.json
+++ b/packages/app-data/sheets/data_list/example_sheet_defaults/example_sheet_defaults_list.json
@@ -1,0 +1,129 @@
+{
+  "flow_type": "data_list",
+  "flow_name": "example_sheet_defaults_list",
+  "status": "released",
+  "flow_subtype": "example_sheet_defaults",
+  "rows": [
+    {
+      "id": "item_1",
+      "text": "ID: item_1",
+      "action_list": [
+        {
+          "trigger": "click",
+          "action_id": "set_field",
+          "args": [
+            "example_sheet_default.item_1.completed",
+            true
+          ],
+          "_raw": "click | set_field : example_sheet_default.item_1.completed : true",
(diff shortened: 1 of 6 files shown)


volloholic (approved):

EsmeeIDEMS (approved):
Makes sense. Functional test passed.

PR #1463, opened 2022-08-04
Feat: add string templating utility

chrismclarke:
PR Checklist

## Updates 2022-08-11
- Syntax approved on call
- Additional methods and tests added to complete parsing process
- Refactored to shared module for use both within scripts and runtime templating

**Example Test Output**
image

**Example Test Context**
```ts
 row: {
      first_name: "Bob",
      last_name: "Smith",
      firstname_lookup_field: "first_name",
      Bob_Smith: "Bob_Smith Lookup Row",
    },
```

## Description
This introduces a new utility script that can be used to extract variables that are contained within delimeters. 
The expected use case is to compile nested or concatenated expressions down into simpler ones, e.g. `{@row.some_field}_completed` -> `@row.some_value_completed`

I've also tidied up and integrated unit-testing for the scripts workspace so that we can add minimal tests for utilities

## Review Notes
This does not impact any parsing or runtime applications (yet), just providing a method for later use, so no functional testing required.

For code, a [test spec](https://github.com/IDEMSInternational/parenting-app-ui/pull/1463/files#diff-0f3d151d1fe2212ae6721f1e4db846571d7b3415bfc6c9deac32914104bb8d20) has been created which also outlines some expected inputs and corresponding outputs. All tests currently pass, so the only check required here is to see if the test cover a reasonable number of use-cases. You can see the first example below, with the input (text) and generated output (expected)

```ts
{
      text: "1. {@row.first_name}-{@row.last_name}",
      expected: {
        value: "1. [$1]-[$2]",
        variables: {
          "[$1]": { value: "@row.first_name" },
          "[$2]": { value: "@row.last_name" },
        },
      },
    },
```

## TODO

- [x] Decide on best delimiters for authoring and generated syntax (current authoring `{`...`}`), generated `[$1]`)
- [ ] Integrate with `@row` syntax for use during initial parsing
- [ ] Integrate with dynamic variables for use at runtime

Note - for the authoring delimiter single characters work best (e.g. `{`,`(`,`[`), or character combinations could also work (e.g. `<%`). In all cases the text will only be interpreted as dynamic if it also include `@` after the start delimter, e.g. `{@row...}`, so should be fairly resistant to conflict with regular authored text

As for the generated delimiter, this will always be a simple string (no nesting) and will be contain a representation of the nested variable (e.g. `[$1]`, `[$1.1]`). The only thing to watch out for would be conflicts with regular text authored alongside dynamic variables. This is again quite unlikely with current design which uses both square brackets and a dollar symbol

## Git Issues

Closes #

## Screenshots/Videos

Output from testing utility
image


chrismclarke:
Approved on call, will merge and create new PRs to implement the logic as part of scripts and templating workflows

Issue #1468, opened 2022-08-04
[CHECKLIST] Theme system ongoing changes


chrismclarke:
Just to track issues emerging from ongoing PRs and discussions

Follwing #1439 
- [ ] deprecate unused variables (as per regex suggestions )
- [ ] clarify any additional calcs required (e.g. light/lighter/lightest vs 50 - 900 weight system), following the point flagged  for the tile buttons
- [ ] figure out gradients
- [ ] refactor legacy calcs to use defined pallet instead
- [ ] test with another deployment to see how the theming looks
- [ ] start adding support for fonts, spacing etc.
- [ ] fix up or deprecate theme service depending on what we want to do with it

Follwing #1466
- [ ] deprecate tertiary colour (replace with fixed gradients or fixed colour variables, e.g. --color-yellow or --gradient-yellow)
- [ ] consider system for exposing and overriding specific variables
- [ ] integrate override into authoring

PR #1471, opened 2022-08-08
refactor: reworked non-gradient component-specific global variables to use new theme variables, with optional overrides

jfmcquade:
PR Checklist

## Description

Reworks all non-gradient component-specific global variables to use a default value of a global variable provided by the generated theme. These values can be overridden at the theme definition level, and the migrated variables are currently listed in `default.scss` with override values that match their previous, legacy colour values (commented out).

There are a few cases where the colour value has changed slightly, but within a range that is imperceptible or acceptably small.

Also adds some new values to the base theme generator: `--ion-color-yellow`, `--ion-color-red`, `--ion-default-box-shadow`, `--ion-color-primary-transparent` and `--ion-color-secondary-transparent`.

## Visual tests
The results of the visual tests are as expected, except for a false positive in the `comp_round_button`(also appears on `feature_home`), which shows the wrong colours – screenshots below. Manually checking the template shows that the colours are displaying correctly, so I'm not sure what causes the visual tests to show the wrong colours, but presumably it's a case of the screenshot being taken before the page has finished loading fully.

## Git Issues

Closes #

#### Visual tests
`comp_round_button` before and after according to the visual tests:
visual tests- before and after

Actual before and after:
live before and after


src/theme/variables.scss
```diff
@@ -475,15 +451,4 @@ $horizontal: 90deg;
     var(--combo-box-gradient-passive-start) 34.6%,
     var(--combo-box-gradient-passive-end) 105.68%
   );
-
-  --ion-color-help-tooltip: #{color.adjust($primary, $hue: -8%, $saturation: 12%, $lightness: 9%)};
-  --ion-color-slider: #{color.adjust($primary, $hue: -8%, $saturation: 12%, $lightness: 9%)};
-  --ion-btn-dark-box-shadow: 0px 4px 4px #00000040;
```
jfmcquade, line 481: this is actually identical to `--ion-default-box-shadow: 0 4px 4px rgba(0, 0, 0, 0.25)` – one is using a hexadecimal value to define the alpha property and the other is using a proportion of 1. So have deprecated `--ion-btn-dark-box-shadow`

src/app/shared/components/template/components/points-item/points-item.component.scss
```diff
@@ -1,5 +1,7 @@
 @use "/src/theme/mixins";
 
+$points-item-bg: var(--points-item-bg, var(--ion-color-primary-200));
```
chrismclarke, line 3: I wonder if this might be a good time to start adding consistency to how we refer to things like background. E.g. should it be `points-item-bg` or `points-item-background` (thinking about what is used elsewhere).

The benefit to this is we can a later stage offer a much simpler override system, where an author can specify at the component level something like `background: 'secondary-light'` and we can interpret as setting the corresponding variable `points-item-background: var(--color-secondary-light)` or similar

So the two things to keep in mind for this trying to keep component-specific prefixes and standard suffixes reasonably consistent where possible.

Doesn't have to all be done now, just something worth flagging

src/app/shared/components/template/components/round-icon-button/round-icon-button.component.scss
```diff
@@ -1,12 +1,16 @@
 @use "/src/theme/mixins";
 
+$round-icon-button-light: var(--round-icon-button-light, var(--ion-color-yellow));
```
chrismclarke, line 3: Here I'd probably recommend the default always being one of primary or secondary colours (probably best to stick mostly with shades of primary) and make the override at the theme level to the yellow colour (which I think is sensible with the way you have added and renamed).

src/theme/deployment/components/_display-group.scss
```diff
@@ -1,3 +1,10 @@
+$tool-one-bg: var(--tool-one-bg, var(--ion-color-secondary-400));
```
chrismclarke, line 1: I know this is something inherited, but it really seems off to me that there is so much plh-specific code within this very general display group component. Do you think it would be possible to refactor in some way to make it a bit cleaner?

src/theme/themes/default.scss
```diff
@@ -1,11 +1,35 @@
...
 
 /** Authoring variables **/
-$color-primary: hsl(203, 76%, 21%);
-$color-secondary: hsl(31, 95%, 57%);
+$color-primary: hsl(203, 76%, 21%); // #0d3f5e
+$color-secondary: hsl(31, 95%, 57%); // #e5943e
+
+/** Authoring component overrides **/
+$component-variables: (
+  // E.g.:
```
chrismclarke, line 10: Thanks for tracking these, very handy to have the list

src/theme/variables.scss
chrismclarke: sounds good, yeah given the choice I much prefer use of rgba or hsla colours where possible, nice to be able to do colour operations on and have a reasonable expectation of what will happpen

chrismclarke (approved):
Code changes look good to me, I've added a few comments that should all be taken as nit(non-blocking), more just identifying more pain points in the existing system and a few thoughts to slowly improve on. But looks like another good step forward

EsmeeIDEMS (commented):
Noticed a bit of a change to some shades of blue, I'm undecided whether this is "too much". Is there a sensible way to have these be more similar? If not I would say leave as is.

image

EsmeeIDEMS (approved):

jfmcquade:
Thanks @chrismclarke, I've added the points I haven't addressed yet to the checklist: #1468 

I have corrected the behaviour of the round-button to use the `-yellow` value as an override rather than by default. And in the process noticed that the `--ion-color-secondary` colour was still being populated by the legacy theme service. This was also prompted by Esmee noticing that the orange colour palette doesn't seem to draw from the more yellow colour of `--ion-color-secondary` as it appears on the `/theme` page. So I've switched off the part of the theme service that sets the custom variables. See screenshots showing before and after deprecating the theme service:
theme service - before and after

This caused the visual test to identify hundreds of changes, revealing that the page background colour was still being overridden by the theme service too. To avoid all of these visual changes at once, I've set the background colour to match the value it had previously, using the in-built optional argument of the new them generator. We can remove this override in a future PR when we're less interested in the other results of the visual tests.

There was another value that was only being populated by the theme service, the `--ion-item-background` variable (see the `colorIdToCSSVarName()` function in `theme.model.ts`). I've moved this over to the new system too, including an override to preserve visual continuity. This limits the visual changes to what we'd expect.

chrismclarke:
Thanks @jfmcquade - I've gone one step further and just removed the code you deprecated/bypassed in the theme service with a comment in the refactor notes (we have a history on git so no real benefit keeping unused code lingering too long)

Otherwise all looks good to me

PR #1473, opened 2022-08-12
Feat/template literals parser

chrismclarke:
PR Checklist

# Breaking changes
- Dependency updates require `yarn install`

## Description
- Integrate templated string parser into sheet parser (to replace `@row` references
- Add [example_templating_list](https://docs.google.com/spreadsheets/d/1kVLixzeWCgwYn4KRaJeDt9q9UDm5drxrfDQTxCPDwv0/edit#gid=1082016302) data list

**Additional changes**
- Refactor to improve performance when calculating repeated replacements
- Add support to parse not only strings but data in any format (not required for sheet parser as all data comes in as string, but will be useful for runtime parsing where data can be in different forms)
- Improve non-templated data parse methods (i.e. `@row` statements without curly braces)

## Review Notes
This only impacts downloaded sheet parsing (no runtime changes), so only needs to check for changes following sync command.

I've updated the parser cache version to force all sheets to be re-processed, I see a handful of changes but most look like they are picking up things previously marked as `undefined` for whatever reason, e.g.

image

In any case recommend taking a closer look, might be easier after latest content merged and then trying resync from this branch

## Git Issues

Closes #
Builds on #1471

## Screenshots/Videos

**[Example sheet](https://docs.google.com/spreadsheets/d/1kVLixzeWCgwYn4KRaJeDt9q9UDm5drxrfDQTxCPDwv0/edit#gid=1082016302)**
image

**Parsed values**
```json
"rows": [
    {
      "id": "example_1",
      "text": "Hello example_1"
    },
    {
      "id": "example_2",
      "text_2": "example_2_completed"
    },
    {
      "id": "example_3",
      "text": "Hello example_3",
      "text_2": "@row.text - with additional message"
    },
    {
      "id": "example_4",
      "action_list": [
        {
          "trigger": "click",
          "action_id": "set_field",
          "args": [
            "example_templating_example_4_completed",
            true
          ],
          "_raw": "click | set_field : example_templating_example_4_completed: true",
          "_cleaned": "click | set_field : example_templating_example_4_completed: true"
        }
      ]
    }
```

**Additional test (bottom) now also supports parsing nested data**
image

**Note** the last 2 tests are added support for legacy data. Within the new system `@row.first_name.sent` should really be rewritten as `{@row.first_name}.sent` as `.sent` is a suffix (and now `.` is a reserved character whereas it previously wasn't), but a legacy method in case a match isn't found to check for the closest possible (i.e. @row.first_name.sent doesn't match but @row.first_name does) and append the rest as a string


packages/scripts/src/commands/app-data/convert/parsers/default/default.parser.ts
```diff
@@ -231,21 +232,10 @@ class RowProcessor {
...
-        const rowReplacements = [...this.row[field].matchAll(/@row.([0-9a-z_:]+)/gim)];
-        for (const replacement of rowReplacements) {
-          const [expression, replaceField] = replacement;
-          const replaceValue = this.row[replaceField];
-          this.row[field] = this.row[field].replace(expression, replaceValue);
-        }
-      }
-    });
+    const context = { row: this.row };
+    return new TemplatedData(context).parse(this.row);
```
chrismclarke, line 238: Legacy methods that searched generically for `@row...` statements have now been refactored and merged into a more general TemplateData class parse method

packages/shared/src/utils/templating.ts
```diff
@@ -5,15 +5,85 @@ export interface ITemplatedStringVariable {
...
+
 /**
+ * Utility class for parsing data containing templated values
  *
- * @param value
- * @param context
+ * @param context - json object specifying specific values for replacement
+ * E.g. {row:{id:'example_1'}} will replace `@row.id` with 'example_1`
+ */
+export class TemplatedData {
```
chrismclarke, line 16: New generic class for taking any data which may or may not contain templated data in some form (e.g. @row statements), and provide methods to parse

EsmeeIDEMS (approved):
As you requested I updated content into master and then pulled master into this branch. 

When now running a content sync all changes to the parsed `.json` files look fine (actually the changes to the "undefined" strings helped me track down a typo in a number of spreadsheets, thanks!)

PR #1474, opened 2022-08-12
refactor: gradients generated by new theme system; handled gradient variables

jfmcquade:
PR Checklist

## Description

Previously, there were a lot of gradients that were very similar to each other, but which fell into a limited number of categories. So it made sense to rationalise this system and draw from a limited suite of gradients auto-generated by the theme (i.e. based solely on the primary and secondary colours of the given theme).

As well as enabling the new theming system, this has some advantages for the look of the app:
- Unifies the look and feel of different components across the app
- It will make it much easier to choose gradients when making new components
- It's potentially an advantage to avoid using colours which don't strictly fit with the defined palette (many of the previous gradients defined custom colours)

But it does mean that some of the gradients will change subtly, so I think we need to consider whether these changes are acceptable on a case-by-case basis.

See screenshots for details.

## Git Issues

Closes #

### New gradients
These are the new gradients generated by the theme, which can be applied across the app. I've set the parameters of these in a way that tries to maximise visual similarity with the gradients we currently use, but they can be tweaked. They are generated systematically to cover a `-light`, `-mid` and `-dark` version for each source colour, as well as one that is based around the absolute lightness of the source colour itself, with horizontal and vertical versions of each. Some of these we don't currently use for this deployment, e.g. `gradient-secondary-dark`, which comes out as a muddy brown when generated from our secondary colour.
gradients - primary
gradients - secondary

### Comparisons between the old and new gradients
The values ending in `-old` are the gradients previously used in the app, and the values above them, without a suffix, are their proposed replacements. To my eyes, these replacements look acceptable. But we might need to assess some of the values in situ, and add custom overrides for any that we think need to stay closer to how they are now.
gradients comparison 1
gradients comparison 2
gradients comparison 3

### Homepage override
Currently the only custom override I've applied is for the homepage, as without it the characters on the second button were more difficult to distinguish.

#### Old home screen
home screen - before

#### New home screen without override
home screen 2 - without override

#### New home screen with override
home screen - after


src/app/shared/components/template/components/audio/audio.component.scss
```diff
@@ -42,7 +45,7 @@
   }
   .progress-block {
     .audio-range {
-      --bar-background-active: var(--ion-btn-control-bg);
+      --bar-background-active: #{$ion-btn-control-bg};
```
chrismclarke, line 48: nit(blocking) - just to double-check - do we need to interpolate these variables using `#{$ ... }` syntax? 
I don't remember doing this in the other components but might just be my mind blanking (I know you can just use variables directly like `$ion-btn-control-bg` , but do the requirements change if referencing var statements??

src/app/shared/components/template/components/audio/audio.component.scss
```diff
@@ -1,4 +1,7 @@
 @use "/src/theme/mixins";
+
+$ion-btn-control-bg: var(--ion-btn-control-bg, var(--ion-color-primary-500));
```
chrismclarke, line 3: nit(non-blocking) - Something I've started noticing is a lot of the previous naming conventions seem to confuse what are variables coming from or overriding those provided by ionic (e.g. `--ion-color-primary`) vs ones we have created and are using and have just given a `--ion` prefix to for whatever reason. 

I really think it would be good to decide a way to start disambiguating (maybe add as a future action item on the existing checklist?). Really we should only use `--ion` for things designed to interact with and override ionic components. Anything else should either just be simple names (ionic components themselves mostly just refer to `$background` or similar before being overwritten by global variable), or possibly prefixed in a standard way (e.g. `--theme-button-control-background`)

src/app/shared/components/template/components/layout/form.ts
```diff
@@ -19,23 +19,7 @@ import { Device } from "@capacitor/device";
...
-        min-height: 71px;
-        padding: var(--regular-margin) var(--regular-padding) 0;
-        --border-radius: 15px;
-        --box-shadow: var(--ion-default-box-shadow);
-        min-width: 100%;
-        --background: var(--ion-btn-primary);
-      }
-    `,
-  ],
+  styleUrls: ["./form.scss"],
```
chrismclarke, line 22: Thanks for tidying this, agree that inline styles get quite messy quite quickly (particularly when we start needing variables)

src/theme/themes/utils/generate-theme.scss
```diff
@@ -43,27 +39,74 @@
...
+        color.change($base, $lightness: math.percentage(1 - math.div($weight, 1000)))
+    );
     $variants-map: map.merge($variants-map, $map);
   }
   $map: (color-#{$baseName}-transparent: rgba($base, 0.5));
   $variants-map: map.merge($variants-map, $map);
   @return $variants-map;
 }
 
+@function getGradientPalette($colorPalette) {
```
chrismclarke, line 53: great to have these so succinctly generated, much appreciated!

chrismclarke (changes requested):
One minor clarification requested from me but otherwise looks good

src/app/shared/components/template/components/audio/audio.component.scss
jfmcquade: We do in these cases where we're declaring a `--` css custom property, explanation here: https://sass-lang.com/documentation/style-rules/declarations#custom-properties

chrismclarke (approved):

EsmeeIDEMS (approved):
Functional test passed, looks great. Only non-blocking issue I could find described below.

Hands of blue blob character are a bit less visible in the new gradient. However already in the old gradient they were not too visible so I think it's mainly due to the design choice blue-on-blue and not something we need to fix pre-empting the discussion (e.g. we can fix for this particular image but if next time the character's hands are in a slightly different position the same issue might reoccur).
Before - After:
imageimage

PR #1475, opened 2022-08-15
Feat/yarn auto install

chrismclarke:
PR Checklist

## Description
A very common breaking change is when we update npm dependencies requiring a `yarn install` command. Communicating this is a bit tedious and is often overlooked, however expecting newer devs to know to do this as required is a bit of a challenge.

This adds a small script and set of hooks to automatically check for changes to dependencies 

- Add _husky_ package to make it easier to work with git hooks (previously used simple-git-hooks)
- Add hooks to detect post-checkout, merge and rewrite
- Add method to track yarn.lock install state and auto-install as required when updating git

## Review notes
This is a bit tricky to test as the hooks only take effect after the first time and install has been manually completed from the branch, and which point it no longer needs to auto-install. But just for reference you can try:

- change to pr branch
- run `yarn install`
- change back to master branch
- change back to pr branch

You should see a `yarn.auto-install` file populated in your top-level directory tracking a timestamp of the latest install (referencing the lock file most recent modified date). In the future when checking out PRs that have updated dependencies this auto-install should also happen

Note - you won't see any messages as the file creates if clicking to change branch in vscode, you will see more output if doing things via command line but not essential as the file should populate in either case

## Git Issues

Closes #

## Screenshots/Videos
File created for tracking last auto-install time
image


jfmcquade (approved):
Very nice. Seems to be working well.
I needed to modify the permissions of the `.husky/` scripts to get it to run locally – I ran `chmod +x .husky/*` and committed the script files. I've pushed my changes so have a look at whether that seems right

PR #1483, opened 2022-08-18
refactor: Rename theme variables; deprecate set_theme template action

jfmcquade:
PR Checklist

## Description

- Renames "intermediate" theme variables (those which can be overridden at the theme definition level) according to some conventions, making explicit to which component(s) they apply and which aspect of the component they affect, and removing the `ion-` prefix where it is redundant. E.g. `ion-banner-secondary` becomes `display-group-background-banner-secondary`, `ion-btn-control-bg` becomes `audio-control-background`. This isn't intended as the final step in rationalising these variables, but a move towards a better system.
- Deprecates the `set_theme` template action which was intended to switch components between "active" and "passive" styles, but which was never fully integrated. This functionality could be built from the new theming system in the future if required. 

## Testing

Functional testing is required for the `combo-box` and `radio-group` components in particular, as they previously integrated the "active" and "passive" themes into the logic that applies their internal styles.

## Git Issues

Closes #


EsmeeIDEMS (approved):
Functional test passed.

jfmcquade:
> 

Yes, I think that's captured the intended naming convention well, I'll add to the style-guide doc.

And I agree that the variable names are still far from ideal. It made sense to me to hold off from renaming ones like `-tool-1` until we can also rename the class/data-param-style that they connect to – which might not be until we're ready to implement a new style system. Although I appreciate it could still be a good idea to rename these variables before that point.

Issue #1467, opened 2022-08-04
[FEATURE] Writing list in single cell from data list


EsmeeIDEMS:
**What?**
I would like to be able to write a cell entry like 

```
name: item_1 | text: first; name: item_2 | text: second; name: item_3 | text: third
```
or 
```
 first; second; third
```
from a data list of the form

| id     | text                |
| --------- | ------                      |
|item_1| first                |
|item_2| second	        |
|item_3| third       |

**Why?**
Useful for creating an answer list (for a radio group or combo box) based on what's in a data list. 

**How?**
Possibly using an items environment. This does not work currently: 
1. See [example_items_radio](https://docs.google.com/spreadsheets/d/1n82prWaYx-HWKW0ibxv-sO_Vuw9WxvuNLRJ_jI_WXtk/edit#gid=1427752706).
<img width="273" alt="image" src="https://user-images.githubusercontent.com/74557272/182905997-f305f721-dfaa-42ce-8b70-0aa8c58d6a93.png"><img width="395" alt="image" src="https://user-images.githubusercontent.com/74557272/182906055-fdc8f281-25cf-49f7-9757-5711a11cb00d.png">
2. See [example_items_nav](https://docs.google.com/spreadsheets/d/1n82prWaYx-HWKW0ibxv-sO_Vuw9WxvuNLRJ_jI_WXtk/edit#gid=1360411654) 
<img width="307" alt="image" src="https://user-images.githubusercontent.com/74557272/182907904-6c4c2c96-aea6-4bfc-b3ca-26a8764b8ed5.png"><img width="466" alt="image" src="https://user-images.githubusercontent.com/74557272/182907954-acaa552e-4568-4438-88d7-c4683386a818.png">

EsmeeIDEMS:
On further thought I decided I could also go around this issue by authoring an extra cell in the data list which uses an excel formula to concatenate the strings (separated by `;`s). 

However, that does not work - see example added to [example_items_nav](https://docs.google.com/spreadsheets/d/1n82prWaYx-HWKW0ibxv-sO_Vuw9WxvuNLRJ_jI_WXtk/edit#gid=1360411654)
<img width="212" alt="image" src="https://user-images.githubusercontent.com/74557272/184341720-d1b3b66b-cde1-42be-b38b-7bb31a6a25f8.png">

chrismclarke:
@EsmeeIDEMS 
[...]

PR #1487, opened 2022-08-19
feat: nav_group can receive a data_list as value and extract template names

jfmcquade:
PR Checklist

## Description

Provides a workaround for Issue #1467, in the case of a `nav_group` component using template names from a data list.

A `nav_group` can now be passed a whole data list as its `value` using `@data`, and if that data list contains a `template` column then the values of that column are used to populate the template array of the `nav_list`.

## Screenshots/Videos

From [example_items_nav](https://docs.google.com/spreadsheets/d/1n82prWaYx-HWKW0ibxv-sO_Vuw9WxvuNLRJ_jI_WXtk/edit#gid=1360411654)
NB: example shows 4 "stepper" steps because the source data list contains a row for the CONCATENATE workaround tested in #1467, but this row could be removed
Screenshot 2022-08-19 at 12 48 08
Screenshot 2022-08-19 at 12 48 26
Screenshot 2022-08-19 at 12 06 34

### DIFF
--- a/src/app/shared/components/template/components/layout/nav_group.ts
+++ b/src/app/shared/components/template/components/layout/nav_group.ts
@@ -87,25 +87,25 @@ export class NavGroupComponent extends TemplateLayoutComponent {
   }
 
   modifyRowSetter(row: FlowTypes.TemplateRow) {
-    if (Array.isArray(row?.value)) {
-      // Check if value is an object referring to a whole data_list with a "template" column,
-      // and if so, assign an array of the values of this column to templateNames
-      const templateArray = [];
-      if (row?.value.length === 1) {
-        const dataObject = row?.value[0];
-        if (isObject(dataObject)) {
-          for (const property in dataObject) {
-            if (dataObject[property].hasOwnProperty("template")) {
-              templateArray.push(dataObject[property].template);
-            }
-          }
+    // Check if value is an object referring to a whole data_list with a "template" column,
+    // and if so, assign an array of the values of this column to templateNames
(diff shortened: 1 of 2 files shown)


EsmeeIDEMS (approved):
Functional test passed. From a spreadsheet authoring point of view I'm happy with the proposed solution. I'll leave it to @chrismclarke  to comment on code. 

I quickly checked whether inputting a data list through a local variable works in the case of an answer_list (radio group) and it doesn't seem to. It would make sense to me if this would behave the same - i.e. both answer_lists and template_lists can be authored as a data list in a local variable or neither. 

image
image

src/app/shared/components/template/components/layout/nav_group.ts
```diff
@@ -86,8 +87,30 @@ export class NavGroupComponent extends TemplateLayoutComponent {
...
-      this.templateNames = row.value;
+    // Check if value is an object referring to a whole data_list with a "template" column,
+    // and if so, assign an array of the values of this column to templateNames.
+    // Handle case where object is passed in directly, or where inside a "_list" local variable
+    const templateArray = [];
+    const dataListObject = isObject(row?.value)
+      ? row.value
+      : Array.isArray(row?.value) && isObject(row?.value[0])
+      ? row.value[0]
+      : null;
```
jfmcquade, line 98: I would prefer these last 2 lines to be indented for clarity, but the indent gets removed by the linter on commit

chrismclarke (approved):
much clearer to me, thanks

Issue #1492, opened 2022-08-25
[Checklist] App Data Parsing


chrismclarke:
Following #1482

- [ ] Parser integration
- [ ] replace instances of `evaluateJSExpression` with new jsEvaluator class
- [ ] consider replacing templateVariablesService methods with newer classes where possible
- [ ] consider moving all shared methods to new package (e.g. app-templating)
- [ ] consider tidying up naming of models, e.g. making models all `contextModels` that extend a common base with setContext and evaluate/parse methods
- [ ] decide how best to handle runtime evaluation (e.g. what to do if fields context not provided)
- [ ] add documentation
- [x] add normalize function to ensure all columns added to datalist in json with incomplete values
- [ ] integrate normlize function to all existing methods

Additional

- [ ] Add better support/handling for 'id' column - use as [index on dataframes](https://danfo.jsdata.org/api-reference/dataframe/dataframe.set_index) (where exists) for operations like concat and merge. Allow override as required.

PR #1494, opened 2022-08-28
Feat/data pipe parser integration

chrismclarke:
PR Checklist

## Description
Integrate data pipe parsing with compile-time sync methods. This makes it possible to write `data_pipe` flowtypes and generate corresponding data

**Additional Changes**
- Addition of queue system for sheet processing and ability to defer (reschedule) conversion operations for case of data_pipes that depend on other data list items to be converted
- Another layer of abstraction was added to make the flow converter behave in ways similar to the xlsx converter. Whilst this greatly simplifies the code (allowing for common caching and logging methods), it makes it a bit harder to wrap your head around the full process (multiple levels of inheritence). Likely the best short-term fix for this will be to improve documentation, rather than attempting much further refactor (complex processes will likely always be a little complex)
- Various updates to tests, error-tracking/handling and logging methods. For the most part the error tracking moves away from throwing breaking errors to logging errors, as typically it is difficult to effectively propogate the error if thrown from a deeply nested process. So the error log files become more significant (indicated after processing)

## Review Notes
The data_pipe feature contains a significant number of moving parts, it would be good to add some additional data_pipe debug/example sheets for functional testing, and carefully inspect the output to see if everything behaving as expected. Ideally these should grow in complexity, from single input sources and operations to multiple sources, use of intermediates, output targets etc.

Example sheets have been created at:
https://docs.google.com/spreadsheets/d/18CUEEFHYs6FqJmaAiXzTRspWxV1alqSkcr1wuEOHfNw/edit#gid=345767231

When functional testing it would also be good to ensure changes update as expected when making changes not only to the data_pipe sheet, but also any data_lists used as input sources (in case of caching issues)

## Future TODOs
Tracked in #1492

## Git Issues

Closes #

## Screenshots/Videos


volloholic (approved):

EsmeeIDEMS (changes requested):
I'm trying to author a data-pipe sheet that uses `map` followed by `concat`, however, this doesn't work (nothing appears in the generated subfolder). My current hypotheses (might be wrong) are that 
1. A `concat` row expects a blank `input_source`, but when it's not the first row on a `data_pipe` sheet blank means "`input_source` equals `output_target` from previous row" so things might get confused
2. It looks like `concat` doesn't work with generated lists

See the sheet [workshop_tasks_pipe ](https://docs.google.com/spreadsheets/d/1kdhunjNQ6xM9cnK7M0J4UneQ9zvasagWGAT9LEH2Xr0/edit#gid=640656823)

EsmeeIDEMS (approved):
I realised the issue I came across previously was actually explained in the documentation...

packages/scripts/src/commands/app-data/convert/processors/xlsxWorkbook.spec.ts
```diff
@@ -37,18 +36,36 @@ describe("XLSX Workbook Processor", () => {
...
+/** Utility class to allow direct access to parsed flows for use in tests */
+export class TestXLSXWorkbookProcessor extends XLSXWorkbookProcessor {
+  constructor() {
+    super(paths);
+  }
+  async getProcessed(sheetPath: string, flowname: string) {
+    const parsedWorkbooks = await this.process([{ relativePath: sheetPath } as any]);
+    const matchingFlow = parsedWorkbooks[0].find((f) => f.flow_name === flowname);
+    return matchingFlow;
+  }
```
jfmcquade, line 70: Add underscore to flow_name

PR #1498, opened 2022-09-01
Fix: Notifications now properly scheduled on initial app launch

### LINKED ISSUE #1496: [BUG] Not receiving notifications unless force sending one
**Describe the bug**
It seems that no notifications are scheduled/pending/sent/received unless a first one has been "forced" by sending it through the campaigns list. This works when this first one is part of the campaign `nf_inactive_day` or `nf_inactive_week` but not when it's part of `debug_html` - maybe being debug or not is the deciding factor? 

**App version**
v0.16.4

**Additional context**
I noticed that lately I was receiving notifications from the ZA app but not from the TZ app on my phone. When looking into this more closely I came across this issue which might explain it (I have for sure sent myself a test notification on the ZA app before, but maybe not on the TZ app). 

If this is truly the case though I don't understand why no-one has reported this issue (most people will have never force sent themselves a notification)... Would be good if one of you could check and see if they can replicate the issue.

**To Reproduce**
- Go to the list of notifications (hyperlink `/notifications` or on the Play Store app in the hamburger menu after entering developer mode using the code `app_dev`). This didn't show any pending or sent notifications in my case. 
- Go to the list of campaigns  (hyperlink `/campaigns` or on the Play Store app after entering developer mode).
- Send a notification
- Reaccess the list of notifications, there are now pending notifications.

**Screenshots and videos**

https://user-images.githubusercontent.com/74557272/187503066-97ad8d88-2ec0-42f0-8294-aaee979309a0.mp4

jfmcquade:
PR Checklist

## Description

Fixes a bug where notifications were not properly scheduled on initial app launch, see Issue #1496. 

This was due to the fact that the "local notifications" service is required by the "campaign" service to schedule notifications, but the latter was being initialised before the former. Switching the order of these initialisations doesn't seem to have any negative effects from a quick functional test, but perhaps @chrismclarke could assess whether we would expect any knock-on effects relating to changing the order of initialisation of the various core services.

(The fact that the services need to be initialised in this order is actually mentioned in the function description for `initialiseCoreServices`)

## Explanation of the behaviour in the issue

As outlined in Issue #1496, no notifications were initially scheduled when launching the app, but manually "forcing" a notification to be sent via the campaigns list did prompt the "pending notifications" to be populated. I think this can be explained by the fact that the method used to schedule a notification includes a trigger to reload all notifications (including "rescheduling any notifications that have been listed in the DB but do not appear as pending"). So any campaign notification being scheduled, whether through a manual trigger or not, should have properly populated the list of pending notifications.

## Git Issues

Closes #1496

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -140,13 +140,13 @@ export class AppComponent {
     await this.appDataService.init();
     await this.templateService.init();
     await this.templateProcessService.init();
-    await this.campaignService.init();
     await this.tourService.init();
 
     // Initialise additional services in a non-blocking way
     setTimeout(async () => {
       await this.localNotificationInteractionService.init();
       await this.localNotificationService.init();
+      await this.campaignService.init();
       await this.dbSyncService.init();
       await this.analyticsService.init();
       /** CC 2022-04-01 - Disable service as not currently in use */


chrismclarke (approved):
Nice catch! makes sense to me. I can't imagine any knock-ons, there shouldn't be any other services that depend on the campaign service (not a core service like notifications). 
In the future would be nicer to provide a better level of separation for what can/can't inherit other services (or find some other means of ensuring init), but for now seems like a good fix to me

EsmeeIDEMS (approved):
Tested on Appetize, indeed there are pending notifications.

PR #1500, opened 2022-09-05
feat: added a second 'flat' theme for use with second skin; refactored theme service

jfmcquade:
PR Checklist

## Description

- Adds a second 'flat' theme for use with second skin (based on [these figma designs](https://www.figma.com/file/OeuqW9ahqwRR5aXbblMHK0/new-skin-human?node-id=0%3A1)). Theme will likely need to be fleshed out along with the second skin.
- Adds buttons to `/theme` page to switch between available themes
- Refactors theme service to allow switching between themes using new theme system
- Deprecates legacy `theme-editor` component and `built-in-themes`

#### Example of theme switching (using a second theme with more extreme variation for demonstration purposes, not present in PR)
https://user-images.githubusercontent.com/64838927/188485992-5b425379-5462-4a12-b603-ca7fe8585990.mov

#### Example of a template (weekly_workshops) showing new "flat" theme:
Screenshot 2022-09-05 at 16 45 00


src/app/feature/theme/services/theme.service.ts
```diff
@@ -1,18 +1,44 @@
...
+      this.currentTheme$.next(themeName);
+      // Use local storage so that the current theme persists across app launches
+      this.localStorageService.setString("currentTheme", themeName);
+    } else {
+      console.error(`No theme found with name "${themeName}"`);
+    }
+  }
+
+  public getCurrentTheme() {
+    return this.localStorageService.getString("currentTheme");
```
chrismclarke, line 36: nit(blocking) - we should prefix this with an `_` to mark as a reserved variable

src/theme/variables.scss
```diff
@@ -1,15 +1,16 @@
 @use "sass:color";
 @use "sass:map";
 @use "./breakpoints";
-@use "./themes/default.scss";
+@use "./themes/index" as *;
```
chrismclarke, line 4: nit(non-blocking) - probably safer to keep namespacing, so include statements more like `themes.theme-default`, although I can see we have `theme-` prefixing so not essential

chrismclarke (changes requested):
Happy with the deprecations and new theme, one minor blocking comment on the theme service but otherwise looks good to me!

EsmeeIDEMS (changes requested):
Some things that I picked up when doing a functional test: 

**Text only radio group component** 
in the flat theme the background of the selected option is white and therefore indistinguishable from the non-selected ones. This should be blue (as is the behaviour for a radio group with images)

https://user-images.githubusercontent.com/74557272/188701751-a3383c7e-90a0-4347-875d-11743a974915.mp4

**Home screen tiles**
In the flat theme the home screen tiles appear to still have a gradient. 
image

chrismclarke (approved):
Thanks for the changes, code looks good to me, I'll leave it to @EsmeeIDEMS to resolve any functional issues with you

jfmcquade:
Thanks @EsmeeIDEMS, I'm glad you spotted the bug with the radio groups. That should be fixed now, along with "flattening" the home screen tiles. Ready to merge if all looks good.

EsmeeIDEMS (approved):
Functional test passed

PR #1503, opened 2022-09-06
fix: empty campaigns

chrismclarke:
PR Checklist

## Description
Add fallback for case where campaign returns no rows

## Git Issues

Closes #

## Screenshots/Videos

**Console now just shows error on empty campaigns**
image

**Previous error message**
image


EsmeeIDEMS (approved):
Looks good, indeed for me as well on master currently the home page doesn't load whereas on this branch it does. 

I don't understand though why these campaigns would not have any rows. I double checked [the corresponding spreadsheet](https://docs.google.com/spreadsheets/d/1pW420PnkUFFtwKqm5JBGIUBAUePOYVb9VV8WQ3UkGNo/edit#gid=142288605) which looks fine, and also when going to the campaigns list in the app there seems to be content found for the `weekly_workshops` campaign image

PR #1505, opened 2022-09-07
fix: app services init order

chrismclarke:
PR Checklist

## Description

Fix regression bug created in #1498. A lot of confusing dependency management going on, we had deferred campaign service evaluation to after local notifications, but realise that the campaigns can be called during the templateProcessService which is used to process some standalone (non-rendered) templates at startup with rows `@campaign.some_campaign`.

Longer term it would be better to ensure some sort of ready$ observable on services that can track if a service is initialisings or has been initialised on demand, so that the exact order the the app.component.ts file is not so important.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -139,14 +139,16 @@ export class AppComponent {
     await this.templateTranslateService.init();
     await this.appDataService.init();
     await this.templateService.init();
+    // ensure local notifications service available for campaigns service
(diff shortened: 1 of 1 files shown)


jfmcquade (approved):
looks good, makes sense

EsmeeIDEMS (approved):
Functional test passed.

PR #1490, opened 2022-08-22
Refactor/App data parser

chrismclarke:
PR Checklist

- [x] - Latest `master` branch merged
- [x] - PR title descriptive (can be used in release notes)

## TODO
**Following PR merge**
- Add better logging outputs
- Integrate data_pipe processing (with queue/requeue system)

**Future (add to checklist)**
- xlsxWorkbook processor could split out html methods to separate processor
- Improve data used as part of tests
- Further improve logging (add child file, filter by context, clear, generalise workspace etc.)

## Description
Following the work in #1482, new methods are required to integrate data_pipe parsing as part of xlsx conversion workflows. This presents a challenge - up till now all processing was handled in a linear fashion where no specific sheet processing depended on any other, and so order was not important. Data pipes however can both depend on parsed data_lists, and generate new data_lists (which other pipes could depend on), and so a significant overhaul of app data parsing methods is required to support this

**Main Changes**
- Completely rewrite main app data parse method to process as a pipeline, allowing improved readability and flexibility

- Develop a new strategy for caching results of processors (previously a single cache was used based on the timestamp of the xlsx file, however now individual caches are used on all pipeline levels. This should speed up cases where only a single sheet is changed in a workbook which would normally require the entire workbook to be reprocessed). This also includes more fine-tuned methods to invalidate caches for steps

- Make all pipeline methods more modular as `Processors` with supporting unit test infrastructure

- Integrate a better logging system for tracking errors and warnings across multiple processes

NOTE - despite the branch name a follow-up PR will be required to actually integrate the data-pipe parsing methods to the app data conversion workflow

## Review Notes
This is a significant refactor of data parsers, so all potential breaking changes should show as app-data diffs after sync workflow. Testing locally it appears to maintain full compatibility, but would be good to double check
- Clear all existing caches `yarn workflow clear_workflow_cache`
- Sync content `yarn workflow sync`
- Check diffs
- Update content
- Sync and check again (in case of caching issues)

As for code review - the new code lives `packages\scripts\src\commands\app-data\convert`. It will likely make sense to look at this folder (starting with the `index.ts` file) as a whole instead of git diffs, using spec test files to highlight key functionality. Specs can be run via `yarn workspace scripts test:watch`

## Git Issues

Closes #

## Screenshots/Videos

**Updated error tracking and output now links to generated log file which contains more details**
image
image

**Lots of new tests added for various behaviours**

image

image

PR #1514, opened 2022-09-23
Bug: Fix bug where template post-processing was not being properly applied

### LINKED ISSUE #1508: URGENT breaking parser bug
We believe there's an issue with the parser whereby variables/dynamic fields are not being handled correctly.

**Debugging steps we undertook**
Found it doesn't seem to be possible to run a `begin_items` `end_items` over a generated list. An example is given in example_items_generated. This results in the console error
image

Problem seemed to be that the value as it is received by the items row is the string (e.g. "@data.sheet_name") rather than the actual data list that is being referred to. 

We then investigated further and found that actually templates with items environments do work on the parenting app repo, but don't when moved over to a different Google Drive folder associated to our AE app fork. Hypothesis is that since these sheets haven't changed since the last content sync the parenting repo does not parse them again, and so no changes show, whereas they are new content to the AE repo and so they are parsed for the first time. See [differences](https://www.diffchecker.com/kXy6RflW) between the two parsed template jsons. 

We then worried that the issue might be more general affecting variables. Checking a simple template (just some texts set through globals) on the AE app fork we found that indeed global variables are not being found/evaluated. 

global sheet: [example_initialise_global](https://docs.google.com/spreadsheets/d/1YRCwIrsPkTw8S6NRZPyRax0jE8jzHXKy7EoftTK-YXE/edit#gid=1595229908)
image
template: [example_call_global_constants](https://docs.google.com/spreadsheets/d/1YRCwIrsPkTw8S6NRZPyRax0jE8jzHXKy7EoftTK-YXE/edit#gid=776720797) 
image
image

See [differences](https://www.diffchecker.com/ovhql81p) between the two template json files

jfmcquade:
PR Checklist

## Description

Fixes bug where template post-processing was not being applied, meaning that variables were not being evaluated and items loops were broken.

## Testing

Checkout this feature branch, then run `yarn workflow clear_workflow_cache` followed by `yarn workflow sync_sheets` to process all sheets with the updated parser. In Source Control, the only file differences should be genuine changes made since the last content sync. I suggest testing extensively on this branch, perhaps authoring sheets to test the new data-pipe functionalities.

## Further steps

Further investigation is required to look into the source of the issue: as far as I understand it, templates parsed since PR #1490 should have been parsed incorrectly, not receiving post-processing. This didn't seem to be the case. Potentially an issue related to caching? It's possible that nobody cleared their workflow cache since that PR was merged, although in that case newly authored or edited templates should still have failed to be parsed correctly, according to my current diagnosis.

## Git Issues

Closes #1508 

## Screenshots/Videos

The template `example_items_generated` now displays as expected:
Screenshot 2022-09-23 at 10 30 38

### DIFF
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.ts
@@ -10,7 +10,7 @@ import {
 } from "../../../utils";
 
 export class TemplateParser extends DefaultParser {
-  postProcess(row: FlowTypes.TemplateRow, nestedPath?: string) {
+  postProcessRow(row: FlowTypes.TemplateRow, nestedPath?: string) {
     // remove empty rows
     if (Object.keys(row).length === 0) {
       return;
@@ -59,7 +59,7 @@ export class TemplateParser extends DefaultParser {
 
     // handle nested rows in same way
     if (row.rows) {
-      row.rows = row.rows.map((r) => this.postProcess(r, row._nested_name));
+      row.rows = row.rows.map((r) => this.postProcessRow(r, row._nested_name));
     }
 
     if (row.exclude_from_translation) {
--- a/src/app/shared/components/template/services/instance/template-row.service.ts
+++ b/src/app/shared/components/template/services/instance/template-row.service.ts
@@ -261,10 +261,7 @@ export class TemplateRowService extends TemplateInstanceService {
     // Instead of returning themselves items looped child rows
     if (type === "items") {
       // extract raw parameter list
-      let itemDataList: { [id: string]: any } = row.value;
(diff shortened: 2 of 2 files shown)


EsmeeIDEMS (approved):
Functional test passed, thanks a lot for this! I'll keep on authoring content and testing on this branch until @chrismclarke has had time to review.

volloholic (approved):
Well done for finding this.

chrismclarke:
@jfmcquade @EsmeeIDEMS 
Looks sensible to me, but the real test should be if the content synced changes before/after the different working versions. 

The only other minor comment I have for @jfmcquade is not sure if you can remember some time back you asked about the `AppDataConverter` version code which I said can be used to invalidate cached data. So whilst fine to manually clear cache it would also be worth bumping this so that any other authors will be forced to reprocess all sheets with the updated parser

PR #1518, opened 2022-09-27
bug: fix template overrides

### LINKED ISSUE #1516: [BUG] Templates overrides not working
**Describe the bug**
When running a content sync I noticed that the json files corresponding to the templates `language_select` and `organisation_registration` had changed: The `_overrides` section has disappeared. 

**App version**
v0.16.4, `bug/parser` branch

**Screenshots and videos**
image

jfmcquade:
PR Checklist

## Description

A further bug fix following on from PR #1514.

The method used to assign template overrides was not being called by the sheets processor, since PR #1490. 
This PR makes the `postProcessFlows()` method for each parser explicitly called within `flowParser.postProcess()` to ensure that `templateParser.postProcessFlows()`, which handles assigning the template overrides, is called. There may be a more idiomatic way to achieve this, in the context of our parsers, but I believe this approach meets our current needs.

This PR also updates the version code for the `AppDataConverter` in order to invalidate any author's previous caches and force all the sheets to be reprocessed using the updated parser.

## Testing

Because the `AppDataConverter` version number is updated, you shouldn't need to manually clear the workflow cache: just checkout out this branch, sync the sheets and manually check that the changes look correct. In particular, ensure that the overrides are present and correct.

## Git Issues

Closes #1516

### DIFF
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/flowParser.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/flowParser.ts
@@ -68,7 +68,9 @@ export class FlowParserProcessor extends BaseProcessor<FlowTypes.FlowTypeWithDat
   public async postProcess(flows: FlowTypes.FlowTypeWithData[]): Promise<IParsedWorkbookData> {
     const flowsByType: IParsedWorkbookData = groupJsonByKey(flows, "flow_type");
     for (const [flowType, parser] of Object.entries(this.parsers)) {
-      flowsByType[flowType] = parser.postProcessFlows(flowsByType[flowType]);
+      if (flowsByType[flowType]) {
+        flowsByType[flowType] = parser.postProcessFlows(flowsByType[flowType]);
+      }
     }
     const flowTypesWithGenerated: IParsedWorkbookData =
       this.populateDataPipeGeneratedFlows(flowsByType);
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/flowParser.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/flowParser.ts
@@ -67,6 +67,9 @@ export class FlowParserProcessor extends BaseProcessor<FlowTypes.FlowTypeWithDat
    */
   public async postProcess(flows: FlowTypes.FlowTypeWithData[]): Promise<IParsedWorkbookData> {
     const flowsByType: IParsedWorkbookData = groupJsonByKey(flows, "flow_type");
+    for (const [flowType, parser] of Object.entries(this.parsers)) {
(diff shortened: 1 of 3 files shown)


packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/default.parser.ts
```diff
@@ -79,6 +79,11 @@ export class DefaultParser<
     return flow;
   }
 
+  /** Postprocess an array of flows */
```
chrismclarke, line 82: It would be better to comment why we have a function that essentially just returns its own input (as this comment is pretty much covered by the function name). I'm still a bit out of the loop on these changes, but assuming something along the lines of _Overridable method called by parser to apply any additional processing on the array of all processed flows. By default the original list is simply returned_

packages/scripts/src/commands/app-data/convert/processors/flowParser/flowParser.ts
```diff
@@ -67,6 +67,11 @@ export class FlowParserProcessor extends BaseProcessor<FlowTypes.FlowTypeWithDat
    */
   public async postProcess(flows: FlowTypes.FlowTypeWithData[]): Promise<IParsedWorkbookData> {
     const flowsByType: IParsedWorkbookData = groupJsonByKey(flows, "flow_type");
+    for (const [flowType, parser] of Object.entries(this.parsers)) {
```
chrismclarke, line 70: This feels sensible to me, can see the function comment asserts that postprocessing methods should be called by flowtype but appears to be omitted. 

What I think would be most useful is trying to capture some documentation/flowchart on the processing order. I think it should actually be fairly straight-forward as the processing order is just something along the lines of

- Apply xlsx converter to extract flows from xlsx files.
- Apply default parser by flow, applying any `@default` or `@row` statements, convert _list and _collection columns, removes meta fields, handles translated fields, processes nested groups etc.
- Apply methods from extended parsers (e.g. data_list, data_pipe). These assign overrides per row, per flow, and across all processed flows

There is also a side flow that explains how caching and queues are used, but again reasonably straightforward (generally tracked by file md5 checksum, so that if a file contents are unchanged cached results used; exceptions possible, currently used for data_pipes which depend on child inputs so cache ignored)

If not creating now, might be good to ensure an issue is created (possibly linking to this comment for some reference)

chrismclarke:
I don't have time to functionally test unfortunately, but see my comments above. If the git diff looks sensible then I'd say should be good to go. Thanks for diving into this @jfmcquade

jfmcquade:
@EsmeeIDEMS I've acted on @chrismclarke 's comments. So again, this one should be good to merge pending testing.

EsmeeIDEMS (approved):
Functional test passed.

PR #1519, opened 2022-09-28
Refactor: Changed oauth method used for authenticating author's google account to access gdrive 

### LINKED ISSUE #1517: [REFACTOR] Migrate Google Oauth flow away from deprecated method
**What?**
Implement a different Oauth flow for authenticating the developer's Google account in order to access Google Drive.

**Why?**
The "OOB" Oauth flow we currently use will be deprecated on October 3rd 2022. 

More details here: https://developers.google.com/identity/protocols/oauth2/resources/oob-migration

jfmcquade:
PR Checklist

## Description

In order for an author/dev to sync data from Google Drive, they are required to authenticate using their Google account. The method we previously used for this is being deprecated by Google on October 3rd (details [here](https://developers.google.com/identity/protocols/oauth2/resources/oob-migration)). This PR implements a new oauth flow (namely the [Loopback IP Address flow](https://developers.google.com/identity/protocols/oauth2/native-app#redirect-uri_loopback)) in order to generate and receive the necessary auth token.

Also updates readme to replace references to deprecated scripts.

## Testing

In order to test the auth flow:
1. Checkout this branch, `refactor/google-oauth-flow`
2. Run `yarn scripts config decrypt` to repopulate `packages/scripts/config/credentials.json` with updated credentials
3. Delete the auth token saved to `packages/scripts/config/token.json`
4. Run `yarn scripts app-data download --authorize` and complete the new authorisation flow
5. A new token should be saved to `packages/scripts/config/token.json`
6. Run `yarn workflow sync` and pay attention to the output to ensure that the authorisation has worked correctly and you are able to download from Google Drive (e.g. make a trivial change to a spreadsheet to ensure it gets synced correctly).

## Git Issues

Closes #1517 

## Screenshots/Videos

A video showing completion of the new authorisation flow:

https://user-images.githubusercontent.com/64838927/192834787-1fd90ae6-20f5-4aa7-b091-064526aa9c14.mov


packages/@idemsInternational/gdrive-tools/src/commands/authorize.ts
```diff
@@ -1,6 +1,10 @@
 import fs from "fs";
-import readline from "readline";
+import chalk from "chalk";
 import { driveactivity_v2, drive_v3, google } from "googleapis";
+import http from "http";
+import opn from "open";
+import url from "url";
+import destroyer from "server-destroy";
```
chrismclarke, line 7: This package doesn't seem to be maintained and is only 16 lines of code anyway. Better to just include manual methods to destroy server
https://github.com/isaacs/server-destroy/blob/master/index.js

packages/@idemsInternational/gdrive-tools/src/commands/authorize.ts
```diff
@@ -148,27 +152,52 @@ function getAccessToken(
...
+
+  // Open an http server to accept the oauth callback.
+  const server = http
+    .createServer(async (req, res) => {
+      try {
+        // The request endpoint of "/oauth2callback" is configured
+        // within Google Cloud's API Console Credentials page
+        if (req.url.indexOf("/oauth2callback") > -1) {
+          const qs = new url.URL(req.url, "http://localhost:8080").searchParams;
+          res.end(
```
chrismclarke, line 175: does the callback have to be served on port 8080 or can it be any port? If it can be any port I'd probably suggest picking one less likely to be in use by other programs. I can see the docs using port 3000 (would clash with react apps but should be fine for angular), or can just pick a more random one

jfmcquade:
@EsmeeIDEMS I have updated the code in response to @chrismclarke 's comments, so I think it should be ready to merge pending functional testing

EsmeeIDEMS (approved):
Functional test passed. I made a change in the template `example_dg` which is coming through correctly when syncing content on the `plh_global`  deployment. Strangely it does appear that this file, together with almost all other the "debug" templates, has been deleted (red D in VS Code). I'm guessing this is just VS Code misbehaving - maybe because these files were deleted when I synced content on a different deployment earlier. 

image

PR #1530, opened 2022-10-03
feat: add support for skins

jfmcquade:
PR Checklist

## Description

Adds support for skins. The current skin is tracked in the value of an app field: `_app_skin`. This value can be referred to in templates in order to apply template overrides that are conditional on the current app skin, as exemplified in [example overrides](https://docs.google.com/spreadsheets/d/1MpoH3BxhECZRmYM10HZ0pTOoe69FJ-fEW9FwzK-Q6yw/edit#gid=1745157248) (last 3 sheets).

Also exposes `_app_skin` (and `_app_theme`) to be updated via the `set_skin` (and `set_theme`) emit. As these are reserved app fields, they should be treated as readonly and no template action is offered to alter them

## Git Issues

Closes #

## Screenshots/Videos

Example of skin switching and overrides being applied. Template: [example_skin_override_default](https://docs.google.com/spreadsheets/d/1MpoH3BxhECZRmYM10HZ0pTOoe69FJ-fEW9FwzK-Q6yw/edit#gid=1226561114)

https://user-images.githubusercontent.com/64838927/193626580-cd2e89ea-5215-494d-901b-60294beef0ef.mov

### DIFF
--- a/src/app/shared/services/skin/skin.service.ts
+++ b/src/app/shared/services/skin/skin.service.ts
@@ -1,0 +1,48 @@
+import { Injectable } from "@angular/core";
+import { BehaviorSubject } from "rxjs";
+import { LocalStorageService } from "src/app/shared/services/local-storage/local-storage.service";
+import { APP_CONSTANTS } from "src/app/data";
+
+const { APP_FIELDS } = APP_CONSTANTS;
+
+// TODO - Where should these live and be added to? Should ultimately come from sheets?
+// Could live at deployment level, e.g. app_constants.compatible_skins
+const SKINS = ["default", "modular"];
+
+@Injectable({
+  providedIn: "root",
+})
+export class SkinService {
+  currentSkin$ = new BehaviorSubject<string>(null);
+  availableSkins = SKINS;
+
+  constructor(private localStorageService: LocalStorageService) {}
+
+  init() {
+    const currentSkinName = this.getCurrentSkin();
+    if (currentSkinName) {
+      this.setSkin(currentSkinName);
+    } else {
+      this.setSkin("default");
+    }
+  }
+
+  public setSkin(skinName: string) {
+    if (this.availableSkins.includes(skinName)) {
+      console.log("[SET SKIN]", skinName);
+      this.currentSkin$.next(skinName);
+      // Use local storage so that the current skin persists across app launches
+      this.localStorageService.setString(APP_FIELDS.APP_SKIN, skinName);
+    } else {
+      console.error(`No skin found with name "${skinName}"`);
+    }
+  }
+
+  public getCurrentSkin() {
+    return this.localStorageService.getString(APP_FIELDS.APP_SKIN);
+  }
+
+  public getAllSkins() {
+    return this.availableSkins;
+  }
+}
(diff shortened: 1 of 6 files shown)


EsmeeIDEMS (approved):
Functional test passed. I also added "Spanish translations" of the strings "Overridden value, es_sp" and "Overridden value, modular skin"
image

and these show correctly in combination with the skin overrides.

[Untitled_ Oct 5, 2022 8_48 AM.webm](https://user-images.githubusercontent.com/74557272/194008648-f3a944a8-d349-4f17-9653-bc11d9529aab.webm)

src/app/shared/services/skin/skin.service.ts
```diff
@@ -0,0 +1,48 @@
...
+  providedIn: "root",
+})
+export class SkinService {
+  currentSkin$ = new BehaviorSubject<string>(null);
+  availableSkins = SKINS;
+
+  constructor(private localStorageService: LocalStorageService) {}
+
+  init() {
+    const currentSkinName = this.getCurrentSkin();
```
chrismclarke, line 22: I'd probably say turn this into a 2-liner,

```ts
// Retrieve the last active skin with default fallback
const activeSkin = this.getCurrentSkin() ?? 'default'
// Set active skin
this.setSkin(activeSkin)
```

src/app/shared/services/skin/skin.service.ts
```diff
@@ -0,0 +1,48 @@
+import { Injectable } from "@angular/core";
+import { BehaviorSubject } from "rxjs";
+import { LocalStorageService } from "src/app/shared/services/local-storage/local-storage.service";
+import { APP_CONSTANTS } from "src/app/data";
+
+const { APP_FIELDS } = APP_CONSTANTS;
+
+// TODO - Where should these live and be added to? Should ultimately come from sheets?
+// Could live at deployment level, e.g. app_constants.compatible_skins
+const SKINS = ["default", "modular"];
```
chrismclarke, line 10: I think we need to clarify what level skins live at. Do themes dictate the skins available? Or deployments?

src/app/shared/services/skin/skin.service.ts
```diff
@@ -0,0 +1,48 @@
...
+    } else {
+      console.error(`No skin found with name "${skinName}"`);
+    }
+  }
+
+  public getCurrentSkin() {
+    return this.localStorageService.getString(APP_FIELDS.APP_SKIN);
+  }
+
+  public getAllSkins() {
```
chrismclarke, line 45: You've defined `availableSkins` as a public property anyway so this method is redundant

src/app/shared/components/template/services/template.service.ts
```diff
@@ -35,6 +37,10 @@ export class TemplateService {
     this.translateService.app_language$.subscribe(async (lang) => {
       await this.initialiseDefaultFieldAndGlobals();
     });
+    // Update default values when skin changed to allow for skin-specific global overrides
```
chrismclarke, line 40: Thanks for adding the clarification

chrismclarke (commented):
Belated review, see a few minor (non-blocking) comments inline.

Overall the code looks clean, although there's part of me wondering if we need to create a whole new service and a lot of boiler plate code for essentially populating and retrieving a single localstorage value. I guess time will tell if we need to further build-out with more skin-specific logic, but if not we might want to consider refactoring skin + theme to some sort of common config service

PR #1533, opened 2022-10-04
feat: add info button to 'parent point box' component

### LINKED ISSUE #1529: [FEATURE] Info button on parent_point_box component
**What?**
A button (or a clickable icon) to appear on top of the `parent_point_box` component that upon click launches a pop-up template with more information. 

**Why?**
Need to include a descriptive text with each ParentPoint and the component doesn't currently have the ability to do this. 

**How?**
The star counter should move to the top-left, so that the info button can appear in the top-right corner of the `parent_point_box`. (David said it feels wrong if the info button would appear in the top-left corner of the component). 
image

Since the action list column of the parent_point_box is already interpreted as the action of the full button, the action list of the info button should probably be authored in the parameter list column. This might need to be done through a local variable to allow for list nesting (the parameter list is already a list with separator `;`, and the action list of the info  button would be a list nested therein).

jfmcquade:
PR Checklist

## Description

Adds an optional "info" button to the top right of the `parent_point_box` component. The button is present iff an `info_icon_src` is provided to the component. The actions triggered on click of this icon are set at the template level, via the new `info_click` trigger in the action list of the `parent_point_box` component.
See [example_parent_point_box](https://docs.google.com/spreadsheets/d/1OzJtUnbn8b4km7x7eYmCajOugn9syQhdzWEUzuILIkc/edit#gid=63099847).

The star containing the counter is shifted to the left of the box in order to put the info button in its most natural place (even when there is no info button).

Adding a new type of action trigger, in this case `info_click`, is generally not desirable, but I think it is potentially justified here – it offers a very efficient way to achieve our current use case, and could be useful if we want to extend this functionality to other components. For example, if we want to move away from using ionic, then we may want to implement custom "info" tooltips and move away from ionic's in-built functionality (as currently in use, for example, on the slider component – see `comp_slider`).

## Git Issues

Closes #1529 

## Screenshots/Videos

[example_parent_point_box](https://docs.google.com/spreadsheets/d/1OzJtUnbn8b4km7x7eYmCajOugn9syQhdzWEUzuILIkc/edit#gid=63099847)

https://user-images.githubusercontent.com/64838927/193871589-b9b616a7-decf-4453-9f4e-5f3fee44ac5b.mov

Screenshot 2022-10-04 at 17 13 14

### DIFF
--- a/packages/data-models/flowTypes.ts
+++ b/packages/data-models/flowTypes.ts
@@ -338,7 +338,8 @@ export namespace FlowTypes {
     | "audio_end"
     | "audio_first_start"
     | "nav_resume" // return to template after navigation or popup close;
-    | "sent"; // notification sent
+    | "sent" // notification sent
+    | "info_click";
 
   export interface TemplateRowAction {
     /** actions have an associated trigger */
--- a/packages/scripts/src/commands/app-data/convert/utils/app-data-action.utils.ts
+++ b/packages/scripts/src/commands/app-data/convert/utils/app-data-action.utils.ts
@@ -30,6 +30,7 @@ export function parseAppDataActionString(actionString: string): FlowTypes.Templa
     uncompleted: true,
     nav_resume: true,
     sent: true,
+    info_click: true,
   };
   if (!Object.keys(actionTriggers).find((t) => actionString.startsWith(t))) {
     actionString = `click | ${actionString}`;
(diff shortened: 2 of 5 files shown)


EsmeeIDEMS (approved):
Looking good, functional test passed. I'm happy with the proposed syntax. Indeed I would expect we will eventually have this info icon appear on other components as well.

PR #1541, opened 2022-10-06
feat: Navigation bar in footer for modular skin

### LINKED ISSUE #1485: Header and footer (desirable: editable from spreadsheet)
Icons for the footer can be found [here](https://drive.google.com/drive/folders/1NCkT37b3q46K6x_pEV2wcGyVs_vSXYaU) (house_white, parent_child_white, book_white)

image

jfmcquade:
PR Checklist

## Description

This PR has 2 main elements:

1. Expands 'skin' functionality
  - Building upon #1530 
  - Skins can contain additional config, including footer settings
  - Configures footer to display a template, [navigation_bar](https://docs.google.com/spreadsheets/d/1I4KOfILcyeT9IX1rBpnTfQ4sZimCakYvelOP7ePLtf4/edit#gid=298399117), when the `modular` skin is active
  - Removes deprecated `IAppSkin` model
3. Adds a new component, `navigation_bar`
  - Currently receives a parameter, `button_list`, specifying `text`, `image` and `target_template` for each button to be rendered in the nav bar e.g.:
    Screenshot 2022-10-07 at 16 51 55
    - It was discussed in a meeting that a better authoring convention would be for the `navigation-bar` component to take a template as an argument rather than a custom parameter list. It was agreed that this approach is fine for now, but that it would better to use a different authoring convention – this should be discussed along with our current use of `answer_list`s
  - Naming
    - "Navigation bar" is the name used to refer to this type of component in the latest Material Design docs
      - Whilst we don't stick closely to material design principles, this seems a like a good, modern naming convention. But this can be discussed.
    - "navigation bar"
      - [Material Design 3](https://m3.material.io/components/navigation-bar/overview)
    - "bottom navigation"
      - [Material Design 2](https://material.io/components/bottom-navigation)

## Git Issues

Closes #1485 – Whilst that issue also mentions the header, including adding a bell notification icon from the modular skin designs, it was agreed that more discussion is needed to clarify what our current use for the bell icon would be, so this can be a new issue.

## Screenshots/Videos
Feature demonstration, using the templates [feature_skin_select](https://docs.google.com/spreadsheets/d/1I4KOfILcyeT9IX1rBpnTfQ4sZimCakYvelOP7ePLtf4/edit#gid=1921555013) and [navigation_bar](https://docs.google.com/spreadsheets/d/1I4KOfILcyeT9IX1rBpnTfQ4sZimCakYvelOP7ePLtf4/edit#gid=298399117)

https://user-images.githubusercontent.com/64838927/195083731-d95349a0-11f1-464a-802c-21fcd7fd86ae.mov


EsmeeIDEMS (approved):
Functional test passed. Looks good! I've replaced the button texts by global variables and that's all working fine. 

I'm not sure if the alignment "feels" right when some of the text subscripts wrap over multiple lines, but not all (which happens for instance in Kiswahili). I'm just checking with the designer if she has any strong feelings on this, but if so I think we can tweak this later - Good to go in like this for me.
image

EsmeeIDEMS:
I observed that the side menu opens in front of the footer. I'm not sure what's the standard behaviour for this, just thought I'd flag it. I don't think this looks bad. 

image

packages/data-models/constants.ts
```diff
@@ -48,6 +48,11 @@ const APP_LANGUAGES = {
   default: "gb_en",
 };
 
+const APP_SKINS = {
+  /** Default skin used if none other is explicitly set */
+  default_skin_name: "weekly_workshop",
```
chrismclarke, line 53: nit(blocking)
This feels like the wrong place to be defining plh-specific content. Also by making this a default it will have an immediate knock-on for other apps.

If we really need to force some overrides just for plh then lets do it from the plh deployment config files instead

src/app/app.component.html
```diff
@@ -25,4 +25,12 @@
       <ion-router-outlet id="main-content"></ion-router-outlet>
     </div>
   </div>
+  <ion-footer *ngIf="currentSkin.footerTemplateName">
```
chrismclarke, line 28: As mentioned in other files, really this should be linked in a similar way to sidemenu settings, such as a `footerMenuDefaults` variable which reads from main config (which in turn is modified from skins)

src/app/app.component.ts
```diff
@@ -42,6 +43,7 @@ export class AppComponent {
   APP_VERSION = environment.version;
   DEPLOYMENT_NAME = environment.deploymentName;
   sideMenuDefaults = APP_SIDEMENU_DEFAULTS;
+  currentSkin: IAppSkin;
```
chrismclarke, line 46: We shouldn't need to access the current skin here, instead what we want is a better way of modifying and updating the app_constants in response to skin updates

chrismclarke (changes requested):
Nice to see progress on this, although it feels like we're taking a slight step backwards in coupling design specifically for the PLH app vs exposing options that can be customised by any app.

Really we need to think of skinning as a child/subset of app deployments - essentially they are a way to modify a whole bunch of core settings at runtime. So we need to create the options to be exposed generally, such as template to display for an app footer, and then update the deployment-specific config and skin files to handle whatever configuration changes required.

I've given a few specific examples inline to try clarify.

I think there's also a second core issue with how we go about the footer. Is there a reason we can't just display a template like we do for the sidebar? I think probably we could decouple the 2 prs, one to implement footers like sidebars and then another for updating core config via skinning.

packages/data-models/constants.ts
jfmcquade: I think you're right to flag here that we don't want to be defining plh-specific content at the level of the default app constants. But I think this might be a feature of my current naming convention for skins: I've renamed the default skin from "default" to "weekly_workshop", which is a descriptive name to contrast it with the second, "modular" skin. (Likewise for themes, we currently have "blob" and "professional"). 

Assuming that all apps/deployments need a default skin, it makes sense to specify a skin which will be the default skin if no default skin is explicitly specified by the app deployment config. And this seems like the place to define that. If so, then currently our only option for the default skin is one of the two existing skins. 

I can see a few ways to fix this:
1. Require a deployment config to specify a default skin, meaning we don't have to provide one at the default app constants level
2. Change the name of the "weekly_workshop" skin to "default" (or "base" or equivalent)
3. Duplicate the "weekly_workshop" skin and name this third skin "default", to be used as the default skin when none other is specified. These two skins would initially be identical, but could diverge in the future. This would allow the plh app to switch between two descriptively named skins, and apply template overrides based on conditions relating to either skin. But it would also mean that other apps would have the same default skin that they do currently, but without using a plh-specific name

It's worth saying that "weekly_workshop" and "modular" aren't necessary plh-specific names. Our default skin, in combination with our default templates, is configured to display content in the form of weekly workshops. But it's true that the skin itself doesn't contain that information, it's the default templates that do: the skin only specifies not to apply any skin-specific template overrides.

jfmcquade:
Thanks for the helpful comments, I can see why it's better to think of skins as a way to modify core app settings at runtime, and I think it's a good idea to split the changes into multiple PRs.

I've been thinking first about enabling skins to update the core config.

We specifically care about the `APP_CONSTANTS`. As I understand it, these constants are currently pulled from static files that are built at the point of setting a deployment, before compile time, so the task is to allow runtime changes/overrides of these constants.

My initial thought is to make another(!) service, specifically for `app-constants`, that can provide the core app constants to anywhere they're needed, applying any overrides dynamically. For our current use case, applying the overrides would consist of getting the active skin (via the skin service) and applying the overrides specified by that skin. Does that sound like a sensible approach, or can you think of a better way?

I've also replied to [one of your comments above](https://github.com/IDEMSInternational/parenting-app-ui/pull/1541#discussion_r993513146) specifically.

packages/data-models/constants.ts
chrismclarke: I would argue that neither `weekly_workshop` or `modular` are descriptive outside the context of the plh app. If we are thinking of what a _default_ skin might be you can look at how ionic [templates ](https://ionicframework.com/docs/v3/cli/starters.html)it's apps - you basically get a choice of things like _blank_ , _sidebar_ or _tabs_ which are all ways to decide what sidebar/header/footer options are available (docs link a bit outdated but main idea is there). 

So yes, you are right that we are lacking a real _default_ option - for that I'd probably probably suggest we stick with a sidebar-enabled configuration, which will impose the need for matching header and sidebar templates (fine for now, in the future we can include a stock/placeholder template to at least include a header with the sidebar toggle button). If we want a more descriptive plh option name then we extend this default template and rename (even if changing no other options) - but my hunch is there will be some minor changes between what we eventually settle on for plh defaults and what we decide makes most sense for general defaults. 

So I think this is most in-line with options (2) and (3) above - we change the name back to default, then extend/duplicate to create a new weekly_workshops skin, which as you say will start identical but diverge over time. Importantly as it diverges it won't break other apps. The weekly workhsops skin can be set as the default skin for all plh deployment configs

chrismclarke:
> My initial thought is to make another(!) service, specifically for app-constants, that can provide the core app constants to anywhere they're needed, applying any overrides dynamically. For our current use case, applying the overrides would consist of getting the active skin (via the skin service) and applying the overrides specified by that skin. Does that sound like a sensible approach, or can you think of a better way?

Yeah I think that is mostly the direction I'm thinking of too. A couple minor caveats:

1) As these are no longer really _constants_ we should probably rename to `APP_CONFIG` and create as an `appConfigService` or similar

2) There's quite a few existing methods that read from the appConstants, but I doubt many of these will be responsive to changes. So likely we will either need to expose a few extra observables, or maybe more simply force changing the app constants to reload the app and ensure the constants are merged with overrides from skins on init. 

3) Changing the app config variables is just one of the tasks that skins can perform. It would be good to clarify/define the others at some point (I'm guessing they can also set available themes, and _maybe_ override templates (?)) - would be good to understand exactly where skin overrides end and theme/language overrides begin.

chrismclarke:
I'm going to revert this PR to draft pending the split into separate prs

PR #1542, opened 2022-10-07
fix: dev credentials for google drive access now remain valid indefinitely

jfmcquade:
PR Checklist

## Description

After PR #1519, for some devs/authors (me, but not Esmee), authenticating with `yarn scripts app-data download --authorize` provided an access token that was only valid for 1 hour, after which time they were required to authenticate again. This is because a "refresh_token" was only being received from Google the first time a dev/author authenticated the app. This PR ensures that a refresh token is received and saved locally with every authentication, so that re-authenticating is not be required.

## Testing

In order to test the auth flow:

1. Checkout this branch, `refactor/google-oauth-flow`
2. Run `yarn scripts app-data download --authorize` and complete the new authorisation flow
3. A new token should be saved to `packages/scripts/config/token.json`. Manually check that it contains a value for `"refresh_token"`.
4. Run `yarn workflow sync` and pay attention to the output to ensure that the authorisation has worked correctly and you are able to connect to Google Drive (e.g. ensure that the `Retrieving list of files` step finds folders and files).
  Screenshot 2022-10-07 at 12 39 10


EsmeeIDEMS (approved):
Functional test passed (refresh token populated and content syncing correctly).

PR #1552, opened 2022-10-11
chore: add authorize workflow

chrismclarke:
PR Checklist

## Description

Just running a belated re-review of #1519, all looks good to me although I've previously been trying to deprecate some of the `yarn scripts` commands in favour of workflows so added a quick PR here to migrate the code a little.

Namely
- add a `sync_authorize` workflow and update getting started readme accordingly (note, I don't think this actually needs to be explicit as first run will prompt if auth does not exist, although probably good to keep for reference if we ever want people to manually re-authenticate)

- update the authorize command to also remove existing auth (I found I wasn't being prompted to create new credentials but the old kept failing)

- code tidying on the `gdrive` tasks to add authorize command (to call from task instead of app-data), and refactor a bit of duplicate code for generating common args passed to the gdrive tools executable

## Review Notes
run `yarn workflow sync_authorize`, check that this prompts creation of a new auth token that works with sync

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/documentation/docs/authors/quickstart.md
+++ b/documentation/docs/authors/quickstart.md
@@ -33,7 +33,7 @@ All users need to connect to their personal Google Drive account to access files
 To connect to your Google account you must run:
 
 ```
-yarn scripts app-data download --authorize
+yarn workflow sync_authorize
 ```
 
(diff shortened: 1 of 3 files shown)


jfmcquade (approved):
Looks good, nice to have it as a workflow and to have a more concise command to run. 
I've tested re-authenticating and it's working for me

PR #1558, opened 2022-10-12
feat: added QR Code component

### LINKED ISSUE #1549: [FEATURE] QR code display component
**What?**
A component that can display a string as a QR code. 

**Why?**
This will need to display the user id, which needs to be read into an ODK survey on a tablet. Current process is using the tablet to take a photo of the smartphone screen, and then manually processing these photos to extract the user id strings.

jfmcquade:
PR Checklist

## Description

- Adds a component, `qr_code`, to generate and display a QR code for a given value
- Adds an Angular pipe, `qrCode`, that can be used across the app to transform a string to a Data URL for a QR code png 

## Testing

The QR code generated does seem to correctly match the input string from my testing. 
In order to test the [user_code](https://docs.google.com/spreadsheets/d/1iSG1kzK4OTHcoiwWoZ25LZxrJqERg2_iP9369HTi_bQ/edit#gid=1630264062) page, you'll need to have a value for the `_server_sync_latest` field. This can be achieved on localhost either by running the backend locally and executing a server sync (the "Sync now" button), or, more conveniently, by manually editing the `_server_sync_latest` field  – there may also be other ways to achieve this that I'm not aware of.

## Git Issues

Closes #1549 

## Screenshots/Videos

The [user_code](https://docs.google.com/spreadsheets/d/1iSG1kzK4OTHcoiwWoZ25LZxrJqERg2_iP9369HTi_bQ/edit#gid=1630264062) template:
Screenshot 2022-10-12 at 16 08 59

The [comp_qr_code](https://docs.google.com/spreadsheets/d/11kaUWIC_oYU5JBUUtEEfmChZASKVTHXVirbilW12xWQ/edit#gid=569531329) template:
Screenshot 2022-10-12 at 13 19 28

### DIFF
--- a/src/app/shared/components/template/components/qr-code/qr-code.component.ts
+++ b/src/app/shared/components/template/components/qr-code/qr-code.component.ts
@@ -1,0 +1,26 @@
+import { Component, OnInit } from "@angular/core";
+import { TemplateBaseComponent } from "../base";
+import QRCode from "qrcode";
+
+@Component({
+  selector: "plh-qr-code",
+  templateUrl: "./qr-code.component.html",
+  styleUrls: ["./qr-code.component.scss"],
+})
+export class TmplQRCodeComponent extends TemplateBaseComponent implements OnInit {
+  value: string;
+  QRCodeDataURL: string;
+
+  ngOnInit() {
+    this.value = this._row.value;
+    this.generateQR(this.value);
+  }
+
+  async generateQR(inputString: string) {
+    try {
+      this.QRCodeDataURL = await QRCode.toDataURL(inputString);
+    } catch (err) {
+      console.error(err);
+    }
+  }
+}
--- a/src/app/shared/components/template/components/qr-code/qr-code.component.ts
+++ b/src/app/shared/components/template/components/qr-code/qr-code.component.ts
@@ -1,26 +1,9 @@
-import { Component, OnInit } from "@angular/core";
+import { Component } from "@angular/core";
 import { TemplateBaseComponent } from "../base";
-import QRCode from "qrcode";
 
 @Component({
   selector: "plh-qr-code",
   templateUrl: "./qr-code.component.html",
   styleUrls: ["./qr-code.component.scss"],
 })
-export class TmplQRCodeComponent extends TemplateBaseComponent implements OnInit {
-  value: string;
-  QRCodeDataURL: string;
-
-  ngOnInit() {
-    this.value = this._row.value;
-    this.generateQR(this.value);
-  }
-
-  async generateQR(inputString: string) {
-    try {
-      this.QRCodeDataURL = await QRCode.toDataURL(inputString);
-    } catch (err) {
-      console.error(err);
-    }
-  }
-}
+export class TmplQRCodeComponent extends TemplateBaseComponent {}
--- a/src/app/shared/components/template/components/select-text/select-text.component.html
+++ b/src/app/shared/components/template/components/select-text/select-text.component.html
@@ -1,7 +1,4 @@
 <div class="box-wrapper" [attr.data-toggled]="toggle" (click)="triggerCopyContent($event, _row)">
-  <div *ngIf="QRCodeValue">
-    <img src="{{ QRCodeValue | qrCode | async }}" alt="{{ QRCodeValue }}" />
-  </div>
   <div class="text">{{ _row?.value }}</div>
   <div class="copy-button">
     <ng-container *ngIf="_row?.parameter_list?.copy_text && _row?.parameter_list.copied_text">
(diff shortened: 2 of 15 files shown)


chrismclarke:
Checking build stats, looks like it bundles an extra 22kb to the main build. Not significant for the app but a little more noticeable on web. At some point probably would make sense to make some specific components lazy-loadable, but for now I can think of better places to focus any optimisation efforts so I'd say no real issue 
image

chrismclarke:
Quick note on testing, I don't think you can edit a localstorage field that has never been populated, so workaround is to type directly into the developer console. I.e
image

`localStorage.setItem('rp-contact-field._server_sync_latest','2022-10-12')`

src/app/shared/components/template/components/qr-code/qr-code.component.ts
```diff
@@ -0,0 +1,26 @@
...
+export class TmplQRCodeComponent extends TemplateBaseComponent implements OnInit {
+  value: string;
+  QRCodeDataURL: string;
+
+  ngOnInit() {
+    this.value = this._row.value;
+    this.generateQR(this.value);
+  }
+
+  async generateQR(inputString: string) {
```
chrismclarke, line 19: nit(blocking)
You already have these methods in the pipe so no need to repeat here. That way all you need is the `<img` tag with the src piped from the qr pipe and `_row.value`, and can omit all the logic from this component file

The benefit of doing that way also means that the qrcode will update if the input value updates

src/app/shared/components/template/components/select-text/select-text.component.html
```diff
@@ -1,4 +1,7 @@
 <div class="box-wrapper" [attr.data-toggled]="toggle" (click)="triggerCopyContent($event, _row)">
+  <div *ngIf="QRCodeValue">
```
chrismclarke, line 2: I don't think I quite understand what is going on with the select-text component and why it receives QR codes as an input... Is there a reason why we're not just using the qrcode template component either before/after the select-text component?

src/app/shared/components/template/pipes/qr-code.pipe.ts
```diff
@@ -0,0 +1,14 @@
+import { Pipe, PipeTransform } from "@angular/core";
+import QRCode from "qrcode";
+
+@Pipe({
+  name: "qrCode",
+})
+export class QRCodePipe implements PipeTransform {
+  async transform(value: any, args?: any[]): Promise<any> {
```
chrismclarke, line 8: nice idea to have as a pipe, will make it easier to be responsive. The only downside is harder to defer loading, but that's an issue for another day

src/app/shared/components/template/pipes/qr-code.pipe.ts
```diff
@@ -0,0 +1,14 @@
...
+    if (value && typeof value === "string" && value.length >= 0) {
```
chrismclarke, line 9: nit(non-blocking) - if the value exists and is a string then you shouldn't need to also check its length. A zero-length string `""` is falsy so will fail the first clause, and so what you are left with will have length >0 regardless

chrismclarke:
A couple minor nits inline and one question more for my own knowledge but otherwise looks good to me! Nice to have the feature in so quickly

src/app/shared/components/template/components/select-text/select-text.component.html
chrismclarke: I think ask @EsmeeIDEMS if happy with the QR code sitting outside.

If yes then let's remove the QR-specific code from the select to keep it cleaner and merge

If no I'd suggest creating a display group with the QR and select-text components inside, removing the styles from select-text and instead applying at the display group level (so the display group adds a border around everything), so that again we do not need any qr-specific code in the select-text component

chrismclarke (changes requested):
@jfmcquade I can see you've re-requested review and added some commits, but could you also resolve the open conversations first?

For any blocking issues that you have resolved by adding a commit just simply mention as you resolve. For any general comments I added that don't require additional input feel free to simply resolve (just to say that you've read them and nothign required).For any not yet resolved just leave open. 

I think probably there are still some things to finish resolving so possibly not ready for re-review yet

src/app/shared/components/template/components/select-text/select-text.component.html
jfmcquade: OK, I'm happy with the QR code outside the box too, so I've updated the code and PR description, I think it's ready to merge.

I think that what you're describing, having "copy to clipboard" available as a template action, would be achievable. But I think it would be non-trivial, for the reasons you suggest as well as handling the custom styling that gets applied to the whole dashed box currently. Perhaps it's worth creating a low priority issue?

jfmcquade:
Sounds like we're all happy with the QR code appearing outside the dashed box, so I've updated the code and believe this is ready to merge

chrismclarke (approved):
Thanks for the updates, I think that simplifies everything much more nicely. All good from the code side, I'll leave it to @EsmeeIDEMS for any last functional test and merge

EsmeeIDEMS (approved):
Functional test passed.

PR #1567, opened 2022-10-13
fix: data list generated popluation

chrismclarke:
PR Checklist

## Description
- Clear caches via version update
- Remove `_processed` data from final data_pipe outputs
- Code tidying

Should fix issue raised in comments #1556

## Git Issues

Closes #

## Screenshots/Videos
There's too many changes pending elsewhere for proper compare, but it does look like the expected outputs are being shown
image


EsmeeIDEMS (approved):
Functional test passed - but I was no longer seeing the problem on master. 

I updated content in master and then merged master into this branch so that the changes coming out of a content sync would be a bit cleaner. Can see the "_processed" entries have disappeared.
image

PR #1571, opened 2022-10-14
Feat/skins can update app config

jfmcquade:
PR Checklist

## Description

Continues from the discussion in #1541 

- Renames `APP_CONSTANTS` -> `APP_CONFIG`
- Expands skins to include skin-specific overrides to values within `APP_CONFIG`
- Adds an `appConfigService` to handle applying skin-specific overrides to `APP_CONFIG`
- Migrates methods that consume 'APP_CONFIG` to get its value from the new service

## Git Issues

Closes #

## Screenshots/Videos

Example showing app config settings being updated at the skin level, in this case the header title:

https://user-images.githubusercontent.com/64838927/196936630-e6878805-091e-41e7-bc92-14030accc2f0.mov


.idems_app/deployments/plh/global.config.ts
```diff
@@ -28,10 +28,12 @@ const config: IDeploymentConfig = {
...
-config.app_constants.APP_SIDEMENU_DEFAULTS.title = "ParentApp";
-config.app_constants.APP_HEADER_DEFAULTS.title = "ParentApp";
-config.app_constants.NOTIFICATION_DEFAULTS.title = "New message from PLH";
-config.app_constants.NOTIFICATION_DEFAULTS.text = "You have a new message from ParentApp";
+config.app_config.APP_LANGUAGES.default = "gb_en";
+config.app_config.APP_SIDEMENU_DEFAULTS.title = "ParentApp";
+config.app_config.APP_HEADER_DEFAULTS.title = "ParentApp";
+config.app_config.NOTIFICATION_DEFAULTS.title = "New message from PLH";
+config.app_config.NOTIFICATION_DEFAULTS.text = "You have a new message from ParentApp";
+config.app_config.APP_SKINS_CONFIG.default = "weekly_workshop"
```
chrismclarke, line 36: Nicer to have this configured here for sure

packages/data-models/constants.ts
```diff
@@ -101,6 +106,7 @@ const APP_HEADER_DEFAULTS = {
 };
 
 const APP_SIDEMENU_DEFAULTS = {
+  enable_sidemenu: true,
```
chrismclarke, line 109: thanks for adding this, although as it's already the side_menu defaults probably just calling it `enabled` would be sufficient

packages/data-models/constants.ts
```diff
@@ -48,6 +48,11 @@ const APP_LANGUAGES = {
   default: "gb_en",
 };
 
+const APP_SKINS_CONFIG = {
```
chrismclarke, line 51: I'm not sure exactly what the `default` and `enabled` options refer to - can you add a couple comments to clarify?

packages/data-models/constants.ts
```diff
@@ -165,5 +172,13 @@ const APP_CONSTANTS = {
...
 // Export as a clone to avoid risk one import could alter another
-export const getDefaultAppConstants = () => clone(APP_CONSTANTS);
-export type IAppConstants = typeof APP_CONSTANTS;
+export const getDefaultAppConfig = () => clone(APP_CONFIG);
+export type IAppConfig = typeof APP_CONFIG;
+
+/** A recursive version of Partial, making all properties, included nested ones, optional.
+ * Copied from https://stackoverflow.com/a/47914631
+ */
+type RecursivePartial<T> = {
```
chrismclarke, line 181: Very cool, never thought of defining a partial recursively but makes a lot of sense

packages/data-models/deployment.model.ts
```diff
@@ -26,7 +26,7 @@ export interface IDeploymentConfig {
     icon_asset_background_path?: string;
   };
   /** Optional override of any provided constants from data-models/constants */
-  app_constants?: Partial<IAppConstants>;
+  app_config?: Partial<IAppConfig>;
```
chrismclarke, line 29: So should this now be renamed to `IAppConfigOverride`?

src/app/app.component.ts
```diff
@@ -25,13 +25,17 @@ import { TemplateTranslateService } from "./shared/components/template/services/
...
 import { AppDataService } from "./shared/services/data/app-data.service";
 import { AuthService } from "./shared/services/auth/auth.service";
 import { LifecycleActionsService } from "./shared/services/lifecycle-actions/lifecycle-actions.service";
+import { AppConfigService } from "./shared/services/app-config/app-config.service";
+import { IAppConfig } from "./shared/model";
 
-const { APP_FIELDS, APP_SIDEMENU_DEFAULTS, APP_AUTHENTICATION_DEFAULTS } = APP_CONSTANTS;
+// TODO - The APP_FIELDS should come from appConfigService.APP_CONFIG,
+// but currently they are being used before the core services are intialised
+const { APP_FIELDS } = APP_CONFIG;
```
chrismclarke, line 38: I think probably no harm moving app config init methods into the constructor (to ensure called on import), although then we'd still need to make sure it's imported into the app module so not sure if it would really fix this issue or not. It may be worth creating an `AppCore` module or similar to import into the app.module that initialises core services that will be required throughout

src/app/shared/services/app-config/app-config.service.ts
```diff
@@ -0,0 +1,47 @@
...
+  appConfig$ = new BehaviorSubject<IAppConfig | {}>({});
+  /** Static value to be read by methods that are not responsive to changes
+   * TODO: migrate all consumers of this value to be responsive to appConfig$ */
+  APP_CONFIG: IAppConfig;
+
+  constructor(private skinService: SkinService) {}
+
+  init() {
+    this.updateAppConfig();
+    this.skinService.activeSkin$.subscribe((activeSkin) => {
```
chrismclarke, line 25: nit(blocking)
In terms of hierarchy I think the appConfig service should sit above/independent of the skin service, so instead of listening for changes the skin service should call the appConfig service to update config instead

src/app/shared/services/skin/skin.service.ts
```diff
@@ -1,48 +1,57 @@
...
+import { APP_SKINS } from "src/skins";
+import { IAppSkin } from "src/skins/skin.model";
 
-const { APP_FIELDS } = APP_CONSTANTS;
-
-// TODO - Where should these live and be added to? Should ultimately come from sheets?
-// Could live at deployment level, e.g. app_constants.compatible_skins
-const SKINS = ["default", "modular"];
+// TODO - The skin service needs these values, but getting them from the
+// appDataService would cause a circular dependency. Possible solution: limit
```
chrismclarke, line 9: Ideally the appConfigService shouldn't import anything from the skin service (see previous comment), which should hopefully resolve the circular dependency issue

src/skins/index.ts
```diff
@@ -0,0 +1,33 @@
+import { IAppSkin } from "./skin.model";
+import { arrayToHashmap } from "src/app/shared/utils";
+
+// TODO: Eventually these skins should be authored from templates
```
chrismclarke, line 4: I think probably the skins should also be kept in the deployment folder, so that the various plh apps define their own set of skins available and enabled

chrismclarke (changes requested):
I'd say definitely moving in the right direction, with 2 main recommended changes:

1. Move the skin definition files to within the `.idems_app\deployments` folders, so that each deployment can specify the skins available and their configuration. The only knowledge the main appConfig has is the optional property of available skins and default.

2. Flip the logic so that the skinService calls update methods in the appConfig service directly on change, instead of having the appConfig service listen for changes to the skin service. This should hopefully fix some of the cyclic deps issues and also provide better separation.

jfmcquade:
> I'd say definitely moving in the right direction, with 2 main recommended changes:
> 
> 1. Move the skin definition files to within the `.idems_app\deployments` folders, so that each deployment can specify the skins available and their configuration. The only knowledge the main appConfig has is the optional property of available skins and default.
> 2. Flip the logic so that the skinService calls update methods in the appConfig service directly on change, instead of having the appConfig service listen for changes to the skin service. This should hopefully fix some of the cyclic deps issues and also provide better separation.

Thanks, I think I've addressed your 2 points. I think the relationship between deployments and skins is much clearer now.

Some additional points:

- The configuration in `.idems_app/.gitignore` never seemed to work for me: I was unable to add additional exceptions for paths to files I created within `.idems_app/deployments`. I've updated the `.gitignore` file, based on [this answer](https://stackoverflow.com/a/44568826), and it seems to be working correctly for me now

- There are still more files/methods to update to use the new `appConfigService`, but I wanted to check that my current approach was right before finishing the chore of refactoring the rest. My current approach is seen in the changes to the `campaign.service` and `feedback.service`, namely: subscribing to `appConfigService.appConfig$` in the constructor of each service and setting the dependent properties (the `appConfigService` initialises its values in its own constructor).
Something I'm not sure of is how to accomplish the equivalent within a module – see `app-routing.module`. Is this what you're suggesting solving by creating another module in your [comment above](https://github.com/IDEMSInternational/parenting-app-ui/pull/1571#discussion_r996608699)? Perhaps this can wait for a later date? Unless it's the best way of migrating all of the files that previously used the static value of `APP_CONSTANTS` to use the new values from `appConfigService`

src/app/app.component.ts
jfmcquade: Having the `appConfigService` call its init methods in its constructor does seem to work as a solution to this particular case.
I'm leaving this thread open because I think it may be relevant to using the dynamic `APP_CONFIG` values in a module. See my comment below for more details

.idems_app/deployments/plh/global.config.ts
```diff
@@ -1,15 +1,16 @@
 import { IDeploymentConfig } from "data-models";
-import { getDefaultAppConstants } from "data-models/constants";
+import { getDefaultAppConfig } from "data-models/constants";
+import * as SKINS from "./skins";
```
chrismclarke, line 3: This looks much nicer to me thanks, at some point in the future we could have some global skins in the main folders but for now I think makes a lot more sense to just link with the deployments and copy/paste code as required

.idems_app/deployments/plh/skins/index.ts
```diff
@@ -0,0 +1,27 @@
+import { IAppSkin } from "data-models";
+
+// PLH-specific skins
+
+/** Skins can be applied to an app to override the core APP_CONFIG.
```
chrismclarke, line 5: This comment can be moved to the data-models definition. if you put `@example` in the comment it should give nice output when hovering the definition

packages/data-models/constants.ts
```diff
@@ -48,6 +50,14 @@ const APP_LANGUAGES = {
   default: "gb_en",
 };
 
+/** Specifies the skins available to the current app build.
+ * "default": the skin used by default for a given build
+ * "available": an array of additional skins that are available to the current build */
+const APP_SKINS: { default: IAppSkin; available: IAppSkin[] } = {
+  default: { name: "default" },
+  available: [] as IAppSkin[],
```
chrismclarke, line 58: Thanks for adding the comment. Minor nit, as you have defined the `available` type above I expect should be fine to just leave `available: []` and ts should infer correctly

packages/data-models/constants.ts
```diff
@@ -48,6 +50,14 @@ const APP_LANGUAGES = {
   default: "gb_en",
 };
 
+/** Specifies the skins available to the current app build.
+ * "default": the skin used by default for a given build
+ * "available": an array of additional skins that are available to the current build */
+const APP_SKINS: { default: IAppSkin; available: IAppSkin[] } = {
+  default: { name: "default" },
```
chrismclarke, line 57: nit(blocking)
The default should probably just be a string containing the name instead of a partial skin. The skin will then be looked up from the list of available, applying the changes if found (possibly just logging error if not). Otherwise it could be tempting to pass a full skin into the default property

packages/data-models/skin.model.ts
```diff
@@ -0,0 +1,8 @@
+import { IAppConfigOverride } from ".";
```
chrismclarke, line 1: nit(blocking)
I know vscode autocompletes this way, but sometimes I have compiler issues when importing from the index (assuming this file will also be exported from it). Probably safer to put `./constants` (and maybe consider renaming the file to `appConfig` in line with the other changes)

src/app/app.component.ts
```diff
@@ -121,6 +125,16 @@ export class AppComponent {
     }
   }
 
+  /** Initialise appConfig and set dependent properties */
+  initialiseAppConfig() {
```
chrismclarke, line 129: nit(non-blocking)
Might be more declarative to rename to `subscribeToAppConfigChanges` or similar

src/app/feature/campaign/campaign.service.ts
```diff
@@ -48,8 +47,13 @@ export class CampaignService {
     private localNotificationService: LocalNotificationService,
     private templateTranslateService: TemplateTranslateService,
     private appDataService: AppDataService,
+    private appConfigService: AppConfigService,
     private injector: Injector
-  ) {}
+  ) {
+    this.appConfigService.appConfig$.subscribe((appConfig: IAppConfig) => {
```
chrismclarke, line 53: Thanks for adding this to track the updates. Might be better to move into a `subscribeToAppConfigChanges` function or similar

src/app/feature/feedback/feedback.service.ts
```diff
@@ -44,8 +44,14 @@ export class FeedbackService {
     private userMetaService: UserMetaService,
     private toastController: ToastController,
     private dbService: DbService,
-    private dbSyncService: DBSyncService
+    private dbSyncService: DBSyncService,
+    private appConfigService: AppConfigService
   ) {
+    this.appConfigService.appConfig$.subscribe((appConfig: IAppConfig) => {
```
chrismclarke, line 50: question
Just to check in with your understanding. Some services which use constructor methods will be initialised on import, and potentially before other services have had their init methods called. Could this cause problems if the appConfigService sets the initial config during init? Will we see a flash of old config/layout before new one, and potentially not applying the correct defaults to a service like this?

src/app/shared/components/plh-main-header.ts
```diff
@@ -38,8 +38,16 @@ export class PLHMainHeaderComponent implements OnInit, OnDestroy {
...
   hasBackHistory = false;
-  constructor(private router: Router, private location: Location) {}
+  constructor(
+    private router: Router,
+    private location: Location,
+    private appConfigService: AppConfigService
+  ) {}
   async ngOnInit() {
+    this.appConfigService.appConfig$.subscribe((appConfig: IAppConfig) => {
+      this.appConfig = appConfig;
```
chrismclarke, line 48: nit(blocking)
I don't quite understand why we need to keep the full appConfig instead of just the header_defaults, so probably a `headerConfig` variable would make more sense that takes the value of the appConfig header_defaults property.
Also as we now are dealing with a display component it would be good not to leave subscriptions open as they can lead to memory leaks, So either you will want to add an unsubscribe method/pipe into ngOnDestroy, or ignore the subscription altogether and just bind the display component to a public `appConfigService.appConfig.APP_HEADER_DEFAULTS.title | async` (as the async pipe automatically handles unsubscribe).

src/app/shared/components/template/services/template.service.ts
```diff
@@ -38,7 +38,7 @@ export class TemplateService {
       await this.initialiseDefaultFieldAndGlobals();
     });
     // Update default values when skin changed to allow for skin-specific global overrides
-    this.skinService.currentSkin$.subscribe(async (skin) => {
+    this.skinService.activeSkin$.subscribe(async (skin) => {
```
chrismclarke, line 41: nit(non-blocking)
Another question of awareness, would it make more sense for the skin service to re-initialise the default field and globals instead of the template service listening for the config change? I'm on the fence myself, but part of me thinks better to have more logic in the skin service as it can actually check whether we need to reinitialise before calling the method on the template service

src/app/shared/services/app-config/app-config.service.ts
```diff
@@ -0,0 +1,43 @@
...
+export class AppConfigService {
+  deploymentOverrides: IAppConfigOverride = (environment.deploymentConfig as any).app_config || {};
+  /** List of constants provided by data-models combined with deployment-specific overrides and skin-specific overrides */
+  appConfig$ = new BehaviorSubject<IAppConfig | {}>({});
+  /** Static value to be read by methods that are not responsive to changes
+   * TODO: migrate all consumers of this value to be responsive to appConfig$ */
+  APP_CONFIG: IAppConfig;
+  deploymentAppConfig: IAppConfig;
+
+  constructor() {
```
chrismclarke, line 20: I see you've now moved the initialisation to the constructor which I think makes sense and should hopefully avoid some of the issues flagged previously

src/app/shared/services/skin/skin.service.ts
```diff
@@ -1,48 +1,58 @@
...
+import { AppConfigService } from "../app-config/app-config.service";
 
 @Injectable({
   providedIn: "root",
 })
 export class SkinService {
-  currentSkin$ = new BehaviorSubject<string>(null);
-  availableSkins = SKINS;
+  // A hashmap of all skins available to the current deployment
+  availableSkins: { [skinName: string]: IAppSkin };
```
chrismclarke, line 14: nit(non-blocking)
Something I'm still not in the habit of as it never used to be part of the spec, but if preferred you can define a hashmap as type `Record<string,IAppSkin>
https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type`

src/app/shared/services/skin/skin.service.ts
```diff
@@ -1,48 +1,58 @@
...
-      this.setSkin(currentSkinName);
-    } else {
-      this.setSkin("default");
-    }
+    const skinsConfig = this.appConfigService.APP_CONFIG.APP_SKINS;
+    this.availableSkins = arrayToHashmap([...skinsConfig.available, skinsConfig.default], "name");
+    // Retrieve the last active skin with default fallback
+    const activeSkinName = this.getActiveSkinName() ?? skinsConfig.default.name;
+    // Set active skin
+    this.setSkin(activeSkinName);
```
chrismclarke, line 28: I think this floats back to my on-again off-again question of order of operations. Should hte skin service be initialised before the appConfig service so that the expected skin is loaded immediately? I'm not sure myself, but can imagine there being some potential confusing behaviour if we don't get this right (e.g. scheduling notifications using default title/text from the wrong skin, which then aren't rescheduled, or just duplicating all calls to other init functions)

chrismclarke (changes requested):
Thanks @jfmcquade , all the changes look good and I think we are 95% of the way there. The only major outstanding issue from me is the question on order of initialisation between skinService and appConfig service - to me it feels like the skin service should load everything as eagerly as possible (i.e. in constructor methods) and make sure the appConfig overrides are applied before any other services need to access them - but possibly easier said than done.

Otherwise I think just a few small bits of housekeeping and appConfig bindings to update, but nearly there!

src/app/shared/components/plh-main-header.ts
chrismclarke: The main thing to know about subscriptions is that they use memory when subscribed, so if you have a component that gets destroyed and recreated frequently and subscribes each time on init you quickly lead to memory leaks. This isn't really important for things like global services which hold single subscriptions (or arguably our header component which should mostly persist throughout), but good practice to consider when using display components.

The angular `async` pipe includes custom logic to automatically unsubscribe when a component is destroyed so usually the safest way to go about subscriptions, but definitely not the only way - most common is hooking into the ngOnDestroy method to manually unsubscribe. In order to do this all subscriptions need to be stored as a named variable for reference, and individually called with _unsubscribe()_ on destroy which can become a bit tedious if you have multiple, so a common simplification is to have a single _componentDestroyed$_ emitter (or similar) which can be piped into subscriptions as a means to unsubscribe via `takeUntil` - you can see examples of this in our codebase if you search `destroyed$`

As for variable naming - yeah this gets ugly pretty quickly. The easiest workaround is when using `ngIf` statements you can [rename the condition variable](https://angular.io/api/common/NgIf#storing-a-conditional-result-in-a-variable) (e.g. `ngIf="appConfigService.appConfig | async as config"` and refer to `config["APP_HEADER_DEFAULTS].title` or similar renaming more nested to headerConfig. Otherwise you can use ng-templates to pass variables between components with different names (a bit messy), or as a final quick solution just write a funtion in the `.ts` file to reference the variable, e.g.
```ts
get headerConfig(){
  return this.appConfigService.appConfig$.value.APP_HEADER_DEFAULTS`
}
```
and simply reference as `<div>{{headerConfig.title}}</div>`

chrismclarke: I can see you've gone ahead with adding the unsubscribe method which makes sense so feel free to mark as resolved

chrismclarke (approved):
Thanks for all the updates @jfmcquade 

I've added some minor code tidying commits but otherwise looks good to me. For the issue you raised, as app-routing config is read directly from source it won't be possible to make the config files respond to the skin change, but as was the case with appConfig and template service, the onus should  be the other way round with the skinService triggering changes (not the routing file). 

This is reasonably possible as the app-routing config can be directly accessed and modified from the `router.config` property available if you inject the `router` service.

I've started to scaffold out a PR for this although it adds a little more complexity (and would require some better testing data) so I'd say lets get this merged in first and then build on top

jfmcquade:
Functional test passed.

Notifications are disabled on the web preview, but I temporarily added some debug logs to check that the notification service was receiving the correct values on initialisation.

PR #1575, opened 2022-10-20
Feat/skins route updates

chrismclarke:
PR Checklist

## TODO

- [x] Builds on #1571 - should be viewed after merge for cleaner diffs
- [x] Testing
- [ ] Config types could be tidied (a lot of variables marked optional but populated by defaults)

## Description
- Add skin service methods to reconfigure app-routing defaults
- Adds debug skin for easier testing of skin overrides
- Add debug config to allow selection of debug skin
- Tidy skins imports to separate files

## Git Issues

Closes #

## Screenshots/Videos


chrismclarke:
@jfmcquade 
Just to quickly revisit this. I think a while back I suggested handing this over to you via slack, but don't think I made it clear on github. Would you be able to take a quick look to see if you think what is here still holds relevance and is should be opened for review/merge?

.idems_app/deployments/plh/skins/debug.ts
```diff
@@ -0,0 +1,18 @@
+import { IAppSkin } from "data-models";
+
+const debug: IAppSkin = {
+  name: "debug",
+  appConfig: {
```
jfmcquade, line 5: I've refactored this skin to be in line with how I saw skins as working: namely, rather than including the full default app config (from `getDefaultAppConfig()`) with some changes applied, the skin includes a partial `appConfig` containing only the values that differ from the default `appConfig`.
Applying the skin's config values is is handled by the `appConfigService.updateAppConfig()` method, which deep merges the `appConfig` from the deployment with the the skin's `appConfig`. Through the power of the deep merge, I was considering a skin's `appConfig` to be a minimal set of specific _overrides_, not as a full appConfig with some alterations like that of a deployment. 
See other comment for an example of why it makes sense to me to keep skins' `appConfig` as a minimal set of overrides.
Although there may well be downsides or issues with this approach that I haven't considered

src/app/shared/services/skin/skin.service.ts
```diff
@@ -57,13 +66,57 @@ export class SkinService {
...
         availableSkins: this.availableSkins,
       });
     }
   }
 
+  /** Override changes to config-dependent routing config inherited in app-routing.module */
+  private updateRoutingDefaults(newSkin?: IAppSkin, oldSkin?: IAppSkin) {
+    const newRouteDefaults = newSkin?.appConfig?.APP_ROUTE_DEFAULTS;
+    let routes = this.router.config;
+    if (newRouteDefaults) {
```
jfmcquade, line 81: Checks like this require that the skin's `appConfig` consists of a minimal set of overrides. If the skin's `appConfig` contains the whole default `appConfig`, then this check will always pass (assuming that the default appConfig includes `APP_ROUTE_DEFAULTS`). 
See other comment on the debug skin file

.idems_app/deployments/plh/debug.config.ts
```diff
@@ -0,0 +1,15 @@
...
+import { SKINS } from "./skins";
+
+/** Debug config extends the default config **/
+
+const config = cloneConfig(DEFAULT_CONFIG);
+config.name = "plh_debug";
+
+// Override constants
+config.app_config!.APP_SKINS!.defaultSkinName = SKINS.debug.name;
+// Limit available skins to only include debug skin, to force this skin to be applied on init
```
jfmcquade, line 12: This currently seems like the best way to ensure that the debug skin is applied when setting the active deployment to `plh_debug`. If, for example, `weekly_workshop` was an available skin for this deployment, then when switching to this deployment the `weekly_workshop` skin would remain active. However if we were to force a deployment's default skin to be applied on init, then the ParentApp would reset it's skin to `weekly_workshop` on every initialisation, overriding some users' choice of the ` modular` skin.

jfmcquade:
@chrismclarke 
I think this is still relevant/required as a way for the router defaults to be updated by skins. I've refactored a few things and left a few comments, I'd recommend looking at the changes since you last looked at it and see if it seems sensible.

Activating the `plh_debug` deployment, and thereby the `debug` skin, does successfully change the home route to "/template/test"
Screenshot 2022-11-28 at 13 26 24

.idems_app/deployments/plh/debug.config.ts
chrismclarke: I'm assuming this is a unique problem for local development, given that all other deployments sit either on their own domain or app, and so localstorage won't be cross-populated. Happy with this as a workaround

.idems_app/deployments/plh/skins/debug.ts
chrismclarke: Yeah that makes sense, easier to author that way and I remember we already have methods to support deep-merging of objects so good to use. As part of tests at some point it would be good to also include function overrides just to make sure they are also handled correctly (skin config will be converted to string, so can't remember if the deep merge will also re-parse)

chrismclarke:
Thanks for checking this @jfmcquade , all the changes you made look sensible to me (comments inlined). I'll mark as ready for review, would it be possible to add some notes to the description to specify how to functional test for Esmee?

jfmcquade (approved):
Functional test passed, I will merge after today's app release

Issue #1420, opened 2022-06-18
[BUG] No padding at bottom of pages


EsmeeIDEMS:
**Describe the bug**
Some pages that fill up the entire screen or are even longer do not have padding below the last component. There should be some padding. 

**App version**
v0.16.1 (EFM)

**Screenshots and videos**
<img width="288" alt="image" src="https://user-images.githubusercontent.com/74557272/174433529-f2d3d85a-58c9-4c8e-bc69-97f0f128232e.png">
<img width="284" alt="image" src="https://user-images.githubusercontent.com/74557272/174433544-40b9b7f1-965c-4e3a-b096-5576b5e34efc.png">

esmeetewinkel:
Laurie also noticed this bug on the FacilitatorApp deployment. Might be worth using this as an excuse to prioritise it now - @jfmcquade how much work would this be?

<img width="492" alt="image (18)" src="https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/c73d4330-4eaf-4caa-a002-7593dadf9a50">

esmeetewinkel:
Also sometimes controls get hidden behind the device's native buttons (screenshot is a full-screen pop-up with style:navigation display group opened on Android Chrome web browser)
![image](https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/da116a8f-8c3e-4a35-8470-2452e428267b)

PR #1576, opened 2022-10-20
Feat/footer configurable from skin; navigation-bar component

### LINKED ISSUE #1485: Header and footer (desirable: editable from spreadsheet)

jfmcquade:
PR Checklist

## Description

The second step of the two bits of work that were originally attempted together in PR #1541.

1. Adds an optional footer to be displayed on all pages. It is enabled at the skin level by specifying a template to display in the footer.

2. Adds a new `navigation-bar` component. Currently in use in the [footer](https://docs.google.com/spreadsheets/d/17ltRR9OyZRuBk2sXYGzYeXDZ-FvOkY6YPFJb4BH8pQM/edit#gid=156888762) template.
This could be expanded in the future to display any child rows in the required way, drawing on a template such as [this example](https://docs.google.com/spreadsheets/d/1V55cAOZ109P2FbEuGCE7I4HetZ1yJcCePnwdCUuStKM/edit#gid=337269833) rather than solely taking in a list of custom parameters.

## Functional Testing

The app should be tested to ensure that the footer behaves as a user would expect, and that having the footer display on every page does not interfere with any existing content. 
When launching the app preview on a new device, the user will be prompted to select a "skin", alongside other launch actions. The "Module" skin should be selected to enable the footer. (If this step is missed, the video below shows how to change the skin to "Module" via the Settings menu).
It would be helpful to do a spot check of many pages and ensure that everything seems to work and look as expected, including scrolling to the bottom of pages.

## Git Issues

Closes #1485

## Screenshots/Videos

Demonstration of enabling the footer by changing skin, and the footer navigation bar's functionality

https://user-images.githubusercontent.com/64838927/197207245-ac0b2021-85a5-4e3a-8af9-e29c5ca2eb19.mov


chrismclarke (changes requested):
Afraid I can't get it working locally. Ideally in the description or preview video can you include what template you initially navigate to - when searching for _skin_ I see a few slightly mixed up results (probably could do with a tidy)
image

Asssuming I use the `skin_select` template which looks most identical, when I enable I don't see any footer changes. I can't see any skins where the footer template is defined so guessing you are missing a pushed commit somewhere or possibly got overridden somewhere else?

src/app/app.component.html
```diff
@@ -31,4 +31,12 @@
       <ion-router-outlet id="main-content"></ion-router-outlet>
     </div>
   </div>
+  <ion-footer *ngIf="footerDefaults.templateName">
```
chrismclarke, line 34: question(blocking)
If the footer is appearing below the content div will it be covering the bottom of the content on pages with full scroll - none of the demos show scrolling to the bottom to see if the last text is available. Please confirm

src/app/shared/components/template/components/navigation-bar/navigation-bar.component.html
```diff
@@ -0,0 +1,14 @@
+<div class="navigation-bar-wrapper">
+  <div
```
chrismclarke, line 2: Generally it's best to use divs only when no other native components fit - as this is a link it probably is better to use an `a` tag (there's quite a few benefits in terms of accessibility and code quality using more fundamental elements.

Additionally angular provides a lot of tools for managing navigation so really we don't want to be using manual click methods when we can use [routerLink](https://angular.io/api/router/RouterLink), and we don't need to manually track if a route is already active because angular provides [routerLinkActive](https://angular.io/api/router/RouterLinkActive)

src/app/shared/components/template/components/navigation-bar/navigation-bar.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  target_template: string | null;
+  name: string | null;
+}
+
+@Component({
+  selector: "plh-navigation-bar",
+  templateUrl: "./navigation-bar.component.html",
+  styleUrls: ["./navigation-bar.component.scss"],
+})
+export class TmplNavigationBarComponent extends TemplateBaseComponent implements OnInit {
```
chrismclarke, line 19: I think fair enough for now to have a standalone component for this, as you are right - there is enough specific behaviour that might otherwise be hard to capture in display group components (namely applying class based on active link). 

For the most part this looks fine to me, although I expect most of the code will disappear when using routerLink and routerLinkActive in the template component instead as suggested

src/app/shared/utils/utils.ts
```diff
@@ -324,9 +324,7 @@ export async function asyncEvery(arr: any[], predicate: Function) {
...
 }
 
-/** A recursive version of Partial, making all properties, included nested ones, optional.
- * Copied from https://stackoverflow.com/a/47914631
- */
-export type RecursivePartial<T> = {
-  [P in keyof T]?: RecursivePartial<T[P]>;
-};
+/** Extract template name from a full router url */
+export function getTemplateNameFromUrl(urlString: string) {
```
chrismclarke, line 328: Again, this won't be required if using routerLink and routerLinkActive. Also if we ever did want to manually extract what template is used we can directly call the router to extract the `templateName` route parameter on routes such as `/template/my_template`
```ts
{
    path: ":templateName",
    component: TemplatePage,
  },
```
https://angular.io/guide/router#accessing-query-parameters-and-fragments

I'm also wondering why we no longer need the RecursivePartial type - seemed like a nice utility to have. Has it been moved or just no longer required?

src/app/shared/utils/utils.ts
```diff
@@ -324,9 +324,7 @@ export async function asyncEvery(arr: any[], predicate: Function) {
...
+  return urlString.split(/[?#]/)[0].replace(/^(\/template\/)/, "");
+}
```
chrismclarke, line 330: nit(non-blocking)
I'm also wondering why we no longer need the RecursivePartial type - seemed like a nice utility to have. Has it been moved or just no longer required?

src/app/shared/utils/utils.ts
```diff
@@ -324,9 +324,7 @@ export async function asyncEvery(arr: any[], predicate: Function) {
   return true;
 }
 
-/** A recursive version of Partial, making all properties, included nested ones, optional.
```
chrismclarke, line 327: nit(non-blocking)
I'm also wondering why we no longer need the RecursivePartial type - seemed like a nice utility to have. Has it been moved or just no longer required?

src/app/shared/utils/utils.ts
```diff
@@ -324,9 +324,7 @@ export async function asyncEvery(arr: any[], predicate: Function) {
...
- * Copied from https://stackoverflow.com/a/47914631
- */
-export type RecursivePartial<T> = {
```
chrismclarke, line 330: nit(non-blocking)
I'm also wondering why we no longer need the RecursivePartial type - seemed like a nice utility to have. Has it been moved or just no longer required?

src/app/shared/utils/utils.ts
chrismclarke, line 330: nit(non-blocking)
I'm also wondering why we no longer need the RecursivePartial type - seemed like a nice utility to have. Has it been moved or just no longer required?

jfmcquade:
Thanks for the comments.

Woops, you're right, I'd overwritten the skin config that enabled the footer. It should be working on the preview link now. I've also removed some unnecessary debug sheets. See the updated video – you can access the `skin_select` component via the settings menu (this was an existing change that came in with a content sync).

Using `routerLink` substantially cleans up the code, I knew there must be a native Angular way to achieve that but hadn't found it.

Your question about scrolling pages and the impact of the footer is important. From my testing it seems to work well – the scrollable area just shrinks to accommodate the footer, and the content inside it displays as expected. However it would be good to do some more in-depth functional testing, I'll tag Santiago, who has agreed to do some testing while Esmee is away if he has time.

Some scrolling pages do go right to the bottom of the page without any whitespace, as recorded in outstanding issue #1420. This is possibly more of a problem when the footer is enabled: in the video in this PR description, you can see this demonstrated on the "Home" and "Parent Library" pages. I have previously tried to fix this issue and was unable to find a way to consistently add extra whitespace at the bottom of every scrollable page. Let me know if you have any ideas of how this could be achieved, e.g. on Slack.

chrismclarke (approved):
Thanks for adding all the updates @jfmcquade , definitely makes the code a lot tidier for sure. Tested and working for me too, good that the bottom padding issue is captured elsewhere

PR #1577, opened 2022-10-20
feat: set theme when setting skin

jfmcquade:
PR Checklist

## Description

By default, changing skin now changes theme also.

Skins now specify a default theme and a list of available themes. The `set_skin` template action sets the active theme to be the target skin's default theme. 
If the target skin has no default theme, then the active theme is unchanged – unless the current theme is not available in the target skin, in which case the active theme is set to the first _available_ theme of the target skin.

Currently the base theme is still named "default", unlike in the case of skins following discussions on PR #1571. Leaving the theme name as "default" avoids many complications, including handling theme-specific assets that reference a theme name. Duplicating the theme and renaming the duplicate "blob" (or equivalent) is an option, but I believe it would slightly increase the package size as the .css files would be duplicated unnecessarily.

## Git Issues

Closes #

## Screenshots/Videos

Demonstration of setting skin also setting theme, using the [skin_select](https://docs.google.com/spreadsheets/d/1iSG1kzK4OTHcoiwWoZ25LZxrJqERg2_iP9369HTi_bQ/edit#gid=1717159007) template 

https://user-images.githubusercontent.com/64838927/197001593-e56b9a39-8383-46a7-b429-d98175c7026f.mov

### DIFF
--- a/packages/data-models/skin.model.ts
+++ b/packages/data-models/skin.model.ts
@@ -21,5 +21,4 @@ import { IAppConfigOverride } from "./appConfig";
 export interface IAppSkin {
   name: string;
   appConfig?: IAppConfigOverride;
-  availableThemes?: string[];
 }
--- a/packages/data-models/skin.model.ts
+++ b/packages/data-models/skin.model.ts
@@ -21,4 +21,5 @@ import { IAppConfigOverride } from "./appConfig";
 export interface IAppSkin {
   name: string;
   appConfig?: IAppConfigOverride;
+  availableThemes?: string[];
 }
--- a/src/app/feature/theme/services/theme.service.ts
+++ b/src/app/feature/theme/services/theme.service.ts
@@ -1,7 +1,6 @@
 import { Injectable } from "@angular/core";
 import { BehaviorSubject } from "rxjs";
 import { LocalStorageService } from "src/app/shared/services/local-storage/local-storage.service";
-import { THEMES } from "src/theme/themes";
 import { AppConfigService } from "src/app/shared/services/app-config/app-config.service";
 import { IAppConfig } from "packages/data-models";
 
@@ -10,8 +9,9 @@ import { IAppConfig } from "packages/data-models";
 })
 export class ThemeService {
   currentTheme$ = new BehaviorSubject<string>(null);
-  availableThemes = THEMES;
+  availableThemes: IAppConfig["APP_THEMES"]["available"];
   appFields: IAppConfig["APP_FIELDS"];
+  defaultThemeName: string;
 
   constructor(
     private localStorageService: LocalStorageService,
@@ -21,12 +21,9 @@ export class ThemeService {
   }
 
   init() {
-    const currentThemeName = this.getCurrentTheme();
-    if (currentThemeName) {
-      this.setTheme(currentThemeName);
-    } else {
-      this.setTheme("default");
-    }
+    // Retrieve the last active theme with default fallback
+    const currentThemeName = this.getCurrentTheme() ?? this.defaultThemeName;
+    this.setTheme(currentThemeName);
   }
 
   public setTheme(themeName: string) {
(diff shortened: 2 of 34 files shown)


packages/data-models/skin.model.ts
```diff
@@ -21,4 +21,5 @@ import { IAppConfigOverride } from "./appConfig";
 export interface IAppSkin {
   name: string;
   appConfig?: IAppConfigOverride;
+  availableThemes?: string[];
```
chrismclarke, line 24: nit(blocking)
the appConfig already has theme properties so is there any real reason to elevate a level higher here? feels a bit confusing to me

src/app/feature/theme/services/theme.service.ts
```diff
@@ -10,8 +9,9 @@ import { IAppConfig } from "packages/data-models";
 })
 export class ThemeService {
   currentTheme$ = new BehaviorSubject<string>(null);
-  availableThemes = THEMES;
+  availableThemes: IAppConfig["APP_THEMES"]["available"];
```
chrismclarke, line 12: nice to have this linked up with the config and no longer hardcoded themes

src/app/shared/components/template/services/instance/template-action.service.ts
```diff
@@ -222,6 +222,9 @@ export class TemplateActionService extends TemplateInstanceService {
         if (emit_value === "set_skin") {
           this.skinService.setSkin(args[1]);
         }
+        if (emit_value === "set_skin_only") {
```
chrismclarke, line 225: nit(blocking)
I don't really like this syntax, if it really is an override probably using additional args on set_skin is a better alternative (e.g. `emit: set_skin:my_skin_name | preserve_theme` or something to that effect). 

But even still that doesn't make much sense to me, if a skin defines a theme it should probably be applied (particularly if you are trying to preserve a theme which isn't officially supported by the skin) - if you really want a skin that doesn't change theme then you can just create a different skin without a default theme. So if you agree I think this can go

src/app/shared/services/skin/skin.service.ts
```diff
@@ -44,6 +58,11 @@ export class SkinService {
       if (!isInit) {
         this.templateService.initialiseDefaultFieldAndGlobals();
       }
+      // If setTheme OR if the current theme is not available in the target skin,
+      // then set the theme to the targetSkin's default theme
+      if (setTheme || !this.isCurrentThemeAvailableInTargetSkin(targetSkin)) {
```
chrismclarke, line 63: nit(non-blocking)
ideally this should be moved into its own method, e.g. `applySkinThemeChanges` (or similar) which can handle all the conditional and application logic. That way eventually the setSkin method can just call a list of knock-on effectors without having to worry too much about following conditional logic, and individual methods can manage the rest

src/app/shared/components/template/services/instance/template-action.service.ts
jfmcquade: OK, I agree. 

I was imagining the scenario of two skins that have different default themes but share an available (and currently active) theme. But I hadn't considered the idea that a skin doesn't need to specify a default skin – it makes sense that if it does then that theme should probably be applied along with the skin.

And you're right that there are ways around it that are probably more appropriate. Another way to override a skin's default theme would be to explicitly call the `set_theme` action from the template. Getting rid of this option cleans up the code in the skin service too.

jfmcquade:
Thanks, I've addressed your comments and removed the unnecessary `set_skin_only` template action, as suggested [here](https://github.com/IDEMSInternational/parenting-app-ui/pull/1577#discussion_r1000958032).

I think this should be to merge

chrismclarke (approved):
Thanks for making the changes, looks good to me!

EsmeeIDEMS:
@jfmcquade Just looking at this now. I don't like that one of the themes is called "default" (since it's only the default theme of one of the two skins, and we don't actually have a default/preferred skin so it doesn't make sense). However, I'm fine with this for now if - as you say - this avoids many complications.

PR #1579, opened 2022-10-20
chore: improve error logging during data pipe ops

### LINKED ISSUE #1578: [FEATURE] Improve feedback for data pipe errors
**What?**
During content authoring following error thrown (linked to concat operation of data pipe)
image

This is due to trying to create a dataframe indexed by id, but duplicate ids appearing in the data. Ideally we should pass back the names of the datalists where the error occured and the specific duplicate ids

chrismclarke:
PR Checklist

## Description

- Add log of explicit duplicate ids that are thrown in error when apply index to dataframe (e.g. during concat operation in data pipes)

## Review Notes
This is mostly just a quick fix to include a bit more error information. To provide full context (e.g. template names, path to excel etc.) it would need tighter integration with the current logging system which would need a larger PR. So suggest for now we move to merge these incremental changes and flag a new issue in the future if we find difficulty troubleshooting further errors. 

The only way to functionally test would be to author data pipes with known errors. This would probably be good to have more generally, but might be a bit time-consuming to do right now.

## Git Issues

Closes #1578

## Screenshots/Videos
When creating a dataframe that expects unique index values (e.g. _id_), it now explicitly states what values were found as duplicate
image

Additional logging also added to provide more information about the step being processed
image

Note - the above is generated from testing data with a data_list called _names_ (hence the repeated reference)

### DIFF
--- a/packages/shared/src/models/dataPipe/operators/concat.spec.ts
+++ b/packages/shared/src/models/dataPipe/operators/concat.spec.ts
@@ -31,7 +31,7 @@ describe("Concat Operator", () => {
   it("Throws if concatenating lists with duplicate ids", () => {
     const duplicateDf = new DataFrame([{ id: "id_3", first_name: "Duplicated" }]);
     expect(() => new concat(duplicateDf, ["names"], testPipe).apply()).toThrowError(
-      "names\nMultiple entries exist for index: id_3"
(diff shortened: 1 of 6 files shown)


jfmcquade (approved):
Looks good, and seems like a sensible improvement

EsmeeIDEMS (approved):

PR #1580, opened 2022-10-20
fix: add tests and fix invalid var name bug in pipes

### LINKED ISSUE #1573: [BUG] Filter data pipe doesn't work when data list has translatable column
**Describe the bug**
If a data list contains a column of the form `column_name::eng` and a `filter` data pipe operation is applied to that sheet, then whenever syncing content the console gives the following error: 

image

**App version**
v0.16.6

### LINKED ISSUE #1590: [BUG] Append_columns skips one row of a data list
**Describe the bug**
When appending two columns to a data list with 12 rows, for some reason the columns did only get appended to 11 rows. The row for which the appended values are missing differs from the other rows in having more blank cells in the original data list. 

Also, separately:  For a data list with data_list_name different from flow_name, when writing the data_list_name rather than the flow_name in the `input_source` column of a data_pipe `append_columns` row and syncing content, the data pipe .json does not appear nor does the generated sheet. I'm not sure if this is the expected behaviour. 

**App version**
v0.16.6

**Template Name**
original data list sheet: [workshop_data](https://docs.google.com/spreadsheets/d/1PNPiy9SiVMcwWTXGn2of5cvBYfCAPyyjtooczyTBKk0/edit#gid=559635134) 
data pipe sheet: [workshop_data_pipe](https://docs.google.com/spreadsheets/d/1Dtr4Ki6V2I2DFA4puAH5cz8rjuCel_DBOmtjlDUyVeQ/edit#gid=1083865127)
generated data list: workshop_tasks `packages\app-data\sheets\data_list\generated\workshop_tasks.json`
row id for which appended columns are missing: `w_celebrate`

**Screenshots and videos**
image
image
image

chrismclarke:
PR Checklist

## Description

When evaluating dynamic expression as part of data_pipe filter operations, the parser creates a context in which any row variables can be referred to within the filter statement (e.g. filter arg: `name === 'name_1'` ). This expects the `name` column to exist in the input data_list, and will filter all rows accordingly

As all columns are converted into variables, an error would be thrown if a column contained a name that is not valid javascript, e.g. `name::eng` as a variable cannot contain `:` characters. This PR fixes by:

- Adding a check when evaluating JS conditions that proposed variable names are valid (ignoring those that aren't)
- Add a general test to js evaluator spec for checking valid names
- Add a specific _bugs_ test list to test this issue more explicitly in the context of language columns, and as a placeholder for future issues along similar lines

## Review Notes
I've added local tests which previously failed and now pass so I hope the issue is resolved, however it would be good to also test with the data that originally threw the error

## Author Notes
Any columns containing invalid javascript syntax are simply ignored in evaluation. This will have no impact unless that specific column is being referred to in the statement, so be aware of the limitation that you could not have a statement such as `name::eng === 'some_name'`.

In addition there are a number of reserved keywords in  javascript which also will throw errors if used for column names during data_pipe filter operations. I haven't included code to test explicitly check and ignore this as the list is quite long (so will still throw an error), so for now should be considered an authoring limitation. For reference here is the list of reserved names:
https://www.w3schools.com/js/js_reserved.asp

Ideally both of these limitations should be noted somewhere for authors.

For both of these cases the simple workaround would be to either rename the column, or to use additional append/map column operators to rename as required before and after filter operations

## Git Issues

Closes #1573
Closes #1590

## Screenshots/Videos


EsmeeIDEMS (changes requested):
This does not resolve issue #1573. I realised that issue was maybe not written very clearly so I added a [comment](https://github.com/IDEMSInternational/parenting-app-ui/issues/1573#issuecomment-1292462102) with debug sheets. 

If I'm understanding the notes correctly, you're saying that I should never filter for a translatable column - which is fine with me. However, the issue is, filtering does not work on a data list that contains **any** translatable column. 

image

chrismclarke:
Thanks @EsmeeIDEMS - this issue was clear, just turns out there was a second issue, being that as well as needing to clean the input variable names I also needed to clean the input variable values which get set as global variables. 

When we have a filter statement like `id === 'some_id'` before evaluating I create variables for all the row entries which can be accessed during the evaluation context. E.g. if a row has the `id` column set to the value `id_1`, then referring directly to `id` in the filter expression will return that value. 

Normally this is fine, however a quirk of this type of evaluation is that it doesn't have support for nested objects (e.g. `{id: {nested: 'id_1'}}`. 
When authoring data_lists we most only author string/number/boolean values so it shouldn't be an issue, however when working with translated data an additional `_translated` column with nested data gets populated automatically which was throwing issues.

So I've just pushed another fix for this case, which should hopefully resolve the linked issue. If we do in fact need to reference nested data from the filter operations it would still be possible, but would require a slightly different authoring style (due to a subtle difference we have trying to evaluate `@row.id` vs `id` (within the context of the row)). 

The only case I can think of where we might need this, would be filtering with columns which are automatically converted to arrays, i.e. ending `_list` - NOTE - this is probably an additional bug faced, as I think this would also have failed data pipe parsing. 

In such a case, e.g. filter: `args_list.length > 2` there's 3 different ways we could go about this:

1. Create a new keyword to refer to the row, and use to reference, e.g. `@pipe_row.args_list[0] === 'some_value'`
2. Use the built-in `this` keyword which is javascripts own self-reference tool `this.args_list[0] === 'some_value'`
3. Use javascript to modify the (existing) string representation of the array: `JSON.parse(args_list)[0]==='some_value'`

But as I say, I don't think this is an issue with any of the current authoring, and now should stop throwing errors

chrismclarke:
Also added minor update to fix #1590

EsmeeIDEMS (approved):
Functional test passed.

volloholic (approved):
I would like to have a further discussion in our next meeting about this but it resolves the immediate issues so should be merged

PR #1584, opened 2022-10-25
Feat/drawer component

jfmcquade:
PR Checklist

## Description

Adds a `drawer` component for use on the home page for the modular skin.

Component template: [comp_drawer](https://docs.google.com/spreadsheets/d/1NY-FwPH7AZNlBoaS3Oi0E5j20ppdZwt0zbnqh72D1Tk/edit#gid=569531329)

Suggested to merge #1582 first and then add any custom styling to the drawer component to accommodate the new style of buttons

Possible TODOs:
- Add gesture support for opening/closing with a swipe
  - Currently supports click only
- Add a shade that appears over background content when the drawer is expanded

## Git Issues

Closes #

## Screenshots/Videos

Demonstration using template [home_screen_modular](https://docs.google.com/spreadsheets/d/1Whi-zledVspIAd_hmAiItv2OxWnmtOVs8PEuCl0Eu8U/edit#gid=1771109624):

https://user-images.githubusercontent.com/64838927/198039258-59e1bcf9-beed-4265-abe6-54c14eab246a.mov

Screenshot 2022-10-26 at 14 32 18


src/app/shared/components/template/components/drawer/drawer.component.ts
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+      drawer.style.transition = "transform .3s ease-out";
+      drawer.style.transform = `translateY(${-this.openHeight}px)`;
+      this.isOpen = true;
+    }
+  }
+
+  // WIP: open and close drawer with a swipe
+  // Requires `gestureCtrl: GestureController` in constructor
+  // async enableGestures() {
```
chrismclarke, line 48: nit(non-blocking)
Have you tested this? What part of the code is still WIP?

The challenge with using a swipe gesture is there isn't really any visual change while swiping, you kind of just do an entire gesture and then the component disappears after. It might be better using either something like an [ion-item-sliding](https://ionicframework.com/docs/api/item-sliding) to make the card title slideable and auto-dismiss (like most mobile email clients), or use swiperjs  to create some sort of stacking effect
https://swiperjs.com/demos#effect-cards

chrismclarke (approved):
Look fine to me, see inline comment about how I think it might be better to handle swipe gesture (assuming that becomes a dev follow-up)

src/app/shared/components/template/components/drawer/drawer.component.ts
jfmcquade: I think "swipe" is probably the wrong word – I was trying to enable the user to click and drag the top handle to open and close the drawer. This would also include a "flick" action that would send the drawer to its fully open/closed position without having to drag it all the way. Essentially the same behaviour as the bottom drawer in the Google Maps app, along the lines of [this React component](https://gorhom.github.io/react-native-bottom-sheet/).

It's proving quite complicated though, I've taken out the WIP code and can come back to it at a later date

chrismclarke: Ionic already has a component for that:
https://ionicframework.com/docs/api/modal#sheet-modal

If that doesn't work we could also consider:
https://panejs.com/

EsmeeIDEMS (approved):
(Non-blocking) three minor comments about the look: 
- I **sometimes** seem to have no padding between the drawer and the footer. I can't figure out exactly when this does / does not display correctly.
image

- Currently the scroll bar on the RHS of the screen appears in front of the drawer, but I think it would make more sense if it was behind the drawer. 
image

- When there's no footer the drawer should drop down to the bottom of the screen, taking up the footer's space. Not a use case we have at the moment, but it would make sense to me. 
image

PR #1585, opened 2022-10-26
feat: add support for sidebar split-pane routing

chrismclarke:
PR Checklist

## Description

To lay some groundwork for further feedback mode improvements, creates a new split-pane view which can be used to display independent content as a sort of split-screen.

This introduces a second router that can navigate independently of the main router. When the router is active its content will appear in the sidebar content display. The benefit of handling sidebar content in this way is that the feedback module doesn't need explicit access to the main app component to render in the space, and additionally provides the option for rendering additional sidebar content in the future

## Review Notes
- Feedback will only be displayed when content wider than 992px

- The sidebar only shows when a sidebar route is active. This can be achieved by adding `(sidebar:my-route-name)` to the end of the url. The example screenshot shows `http://localhost:4200/template/home_screen(sidebar:feedback)` , which renders the home_screen template in the main content alongside the feedback page in the sidebar

- Any sidebar routes have to be explicitly defined, so currently only `feedback` and `template` have been added

- I noticed a few knock-ons with the app menu button disappearing because technically the _home page_ isn't the same when the sidebar is on show (the url has extra `(sidebar:feedback)` text), but I fixed this so should display properly now. It would still be good to quickly check nothing else changed with regards to menus and main content (still fills/works as expected)

## Git Issues

Closes #

## Screenshots/Videos
**Home page with feedback sidebar**
image

**Example using template select page alongside home template**
image


src/app/shared/components/plh-main-header.ts
```diff
@@ -12,7 +12,7 @@ import { AppConfigService } from "../services/app-config/app-config.service";
   template: `<ion-header>
     <ion-toolbar color="primary">
       <ion-buttons slot="start" style="position:absolute">
-        <ion-menu-button *ngIf="showMenuButton"></ion-menu-button>
+        <ion-menu-button *ngIf="showMenuButton" autoHide="false" menu="start"></ion-menu-button>
```
chrismclarke, line 15: The left menu-button auto-hides when split-screen mode enabled as it assumes the menu is now in full display, however as we are actually keeping the left side-menu and only using split-screen for the right-sidebar content prevent this behaviour by setting `autoHide` to false and ensuring it knows to open the left (start) menu

src/app/app.component.scss
```diff
@@ -20,3 +20,11 @@
     height: 100%;
   }
 }
+#sidebar {
+  width: 500px;
+  min-width: 500px;
+}
```
volloholic, line 26: I can't see where the code has been added to determine when the split screen would be shown. I was expecting sidebarRouter.isActivated to be set somewhere I assume that does not already exist?
My understanding is that when active the sidebar is the left pane containing and would contain 'app'. If I have understood correctly then this 500px should be a variable that can be set dynamically from the right pane. And there might be value in centering it and not having 100% width below but that would be for a later discussion.

volloholic (changes requested):
This looks good I really like this router approach I can see multiple applications.

src/app/app.component.html
```diff
@@ -25,18 +26,27 @@
       ></plh-template-container>
     </ion-content>
   </ion-menu>
-  <div style="display: flex; flex-direction: column; height: 100%">
-    <plh-main-header></plh-main-header>
-    <div class="route-container">
-      <ion-router-outlet id="main-content"></ion-router-outlet>
+  <!-- Main content: shows in split-pane when sidebar route active -->
+  <ion-split-pane when="lg" contentId="main" [disabled]="!sidebarRouter.isActivated">
```
chrismclarke, line 30: @volloholic - here is the code that handles when/if to use split panes. Confusingly enough _disabling_ the pane means the content will still render (just not in split pane)

It uses `!sidebarRouter.isActivated` which is not actually defined in code, but instead read directly from the router outlet further below
```
<router-outlet name="sidebar" #sidebarRouter="outlet"></router-outlet>
```
Angular has a system where defining a `#id` on any element provides local reference, so this code exposes the underlying `outlet` via the variable `sidebarRouter` for access within the local html, and this is what we are reading in the code above.

By assigning the disabled property to activation state, this means currently the sidebar will always be enabled whenever the routing asserts a sidebar should exist (e.g. `/template/home(sidebar:somePage`) and not whenever there is no corresponding route active. 

There is the additional caveat of `when="lg"` which uses screen size breakpoints to only show on larger screens. All of these things can indeed be customised how we see fit including sizes of breakpoints and sizes of sidebar.
Using this default additionally means that we can still access the sidebar content on smaller devices as a sidemenu instead of sidepane (just like how the left menu can expand out)

src/app/app.component.scss
chrismclarke: I've added a comment to clarify how/when split screen is shown.

Yes you are correct that the left pane displays the app, not to be confused with the left menu which sits further outside. I.e. currently the building blocks are:

[left menu][content]

Where the [content] further subdivides to
[content-main router][content-side router]

The content-main-router will automatically expand to fit whatever space is available, and the content-side-router requires a static width (default 300px, I've increased to 500px) - but yes this can also be set programatically/dynamically, and both content can have max-widths imposed alongside things like centering if we wanted to, for example, toggle a display mode that rendered the content at a specific width

src/app/app.component.html
volloholic: Thanks I'd missed the when="lg".

src/app/app.component.scss
volloholic: I think it would be worth adding being able to change both the side-router width and the actual content width within that dynamically to be able to see what the content looks like on various devices without changing screen layout. Currently the RCT device is going to have 480 width so it would be good to have that as a default.

chrismclarke: Sounds good - I'll link the specific-width changes into the next PR more specifically for the feedback system (drafted in #1603)

EsmeeIDEMS (commented):
I thought it can be a bit confusing that pop-ups from either side of the screen open over the full screen - I would have expected them to only use their own side. 
image

EsmeeIDEMS (approved):

PR #1592, opened 2022-10-26
feat: Tasks - highlighted task is automatically evaluated; carousel focuses on the highlighted task group; various bug-fixes

jfmcquade:
PR Checklist

## Description

Requires #1594 

Implements the following changes to the task system and carousel, as discussed on a call with @EsmeeIDEMS and @volloholic 

- The carousel component has a new optional parameter, `task_groups_data`, to receive the name of a task group data list. If this parameter is provided, then upon rendering, the carousel will automatically centre on the card corresponding to the `_task_highlighted_group_id `, if the task appears in the list provided. Currently, this relies on the list used to generate the cards matching the `task_groups_data` parameter, so that the index of the card corresponding to the `_task_highlighted_group_id ` is the same in both lists. 
  - This is implemented in [example_task_carousel](https://docs.google.com/spreadsheets/d/1f_vq3KFGlExf1-g1PwPiDad9gOZkIOdRE5X99jCIRsM/edit#gid=1225992006):
  Screenshot 2022-10-28 at 15 58 10

    - Note that the value passed to `items` and the `task_groups_data` passed to the carousel both refer to the `workshop_tasks` data list. 
  - If desirable, the carousel could be configured to receive _just_ a value for a `task_groups_data`, and generate its constituent cards for each task group in that list. This would remove some customisability from the templating level, but may be desirable for our current use case.
  - This new carousel functionality would presumable usurp the design for the home page that includes 2 carousels (one for "upcoming modules" and one for "completed modules"), since the completed modules would remain in the carousel but be panned off to the left of the screen by default. If we want to keep 2 separate carousels, we would need a way to pass a filtered data list (the `workshop_tasks` data list filtered to contain either only completed or only uncompleted workshops) into both the items loop and the carousel component. This functionality exists for the items loop, but a workaround would be needed for the carousel component.
- Adds functionality to automatically evaluate the current highlighted task group and set the `_task_highlighted_group_id` field, defined as the id of the highest priority task group from the given list that is not completed and not skipped. The list of highlight-able task groups is currently hardcoded to be the generated data list, `workshop_tasks`.
- The `task_group_set_highlighted` template action now sets the respective `_skipped` field to "true" for each task group that is not completed and has higher priority than the target task group, then re-evaluates the `_task_highlighted_group_id`.

Notes on the code:
- Renames the field `highlighted_task` to  `_task_highlighted_group_id`, which is a more informative description and includes an underscore to indicate that it is a protected field.
  - I didn't want to name this field in universal `app_config`, as I'm not sure if it will have meaning outside of PLH. This field name could be skin-specific, but that might cause issues when switching between skins, so at the moment its name is set at the _deployment_ level.
- There is further discussion to be had about the extent to which "tasks" are a core structure to the app. Perhaps the task service should be further generalised in future, but for now it seems an expedient way to meet our current requirements. If we are accepting some potentially plh-specific terminology, then it might be clearer to refer to "workshop" as opposed to "task group", which feels clunky, e.g. `taskService.setHighlightedWorkshop()`, but this might be too much of a compromise.

## Git Issues

Closes #

## Screenshots/Videos

Demonstration of the carousel centring the highlighted task, using the templates [example_task_carousel](https://docs.google.com/spreadsheets/d/1f_vq3KFGlExf1-g1PwPiDad9gOZkIOdRE5X99jCIRsM/edit#gid=1225992006) and [feature_highlighted_task](https://docs.google.com/spreadsheets/d/1miWFEgue4B6kNQ0ZCXqQ72TkoBRUF2_f5bvDrPwXZ4I/edit#gid=1430609248).

https://user-images.githubusercontent.com/64838927/198281469-9a319fb4-c8c8-438e-b18e-db1667970d73.mov


packages/data-models/appConfig.ts
```diff
@@ -43,6 +43,7 @@ const APP_FIELDS = {
   APP_AUTH_USER: `${FIELD_PREFIX}._app_auth_user`,
   APP_SKIN: `${FIELD_PREFIX}._app_skin`,
   APP_THEME: `${FIELD_PREFIX}._app_theme`,
+  HIGHLIGHTED_DATA_ID: `${FIELD_PREFIX}._highlighted_data_id`,
```
chrismclarke, line 46: nit(blocking)
As this (for now) is pretty specific just to the tasks system which is subject to change let's I'd suggest just hardcoding this field into the task.service.ts field and remove from app config. If we really need it to change with skin/theme then you could always add as a suffix to the name. 

In addition we should be more semantic with what the _data_ is referring to, so probably a better name would be along the lines of `_task_highlighted_group_id` or similar

packages/data-models/flowTypes.ts
```diff
@@ -373,7 +373,7 @@ export namespace FlowTypes {
       | "track_event"
       | "process_template"
       | "google_auth"
-      | "set_highlighted_task";
+      | "set_highlighted_task_group";
```
chrismclarke, line 376: nit(non-blocking)
would be better to be prefix with `task` so easier to couple with functionality, e.g. `task_group_set_highlighted`

src/app/shared/services/task/task.service.ts
```diff
@@ -1,26 +1,117 @@
...
+    const taskGroupsNotCompletedAndNotSkipped = clone(this.taskGroups).filter((taskGroup) => {
+      return (
+        !this.templateFieldService.getField(taskGroup.completed_field) &&
+        !this.templateFieldService.getField(taskGroup.skipped_field)
+      );
+    });
+    // If all task groups are completed or skipped (e.g. when user completes final task group),
+    // then set highlighted task group to undefined
+    if (taskGroupsNotCompletedAndNotSkipped.length === 0) {
+      this.templateFieldService.setField(this.appFields.HIGHLIGHTED_DATA_ID, "undefined");
```
chrismclarke, line 52: nit(non-blocking)
Setting string undefined feels weird, can it just be set as a blank string instead?

src/app/shared/services/task/task.service.ts
```diff
@@ -1,26 +1,117 @@
...
+    return this.templateFieldService.getField(this.appFields.HIGHLIGHTED_DATA_ID);
+  }
 
-  public getHighlightedTask() {
-    return this.templateFieldService.getField("highlighted_task");
+  /** For a given task groups list, @return the index of the highlighted task within that list */
+  public async getHighlightedTaskGroupIndex(taskGroupsList: string) {
+    const taskGroupsDataList = await this.appDataService.getSheet("data_list", taskGroupsList);
+    const arrayOfIds = taskGroupsDataList.rows.map((taskGroup) => taskGroup.id);
+    console.log("arrayOfIds:", arrayOfIds);
```
chrismclarke, line 74: nit(blocking)
please remove development logs

src/app/shared/services/task/task.service.ts
```diff
@@ -1,26 +1,117 @@
...
+    console.log("HighlightedTaskGroup:", this.getHighlightedTaskGroup());
+    return arrayOfIds.indexOf(this.getHighlightedTaskGroup());
```
chrismclarke, line 76: nit(non-blocking)
Youd should make the call to getHighlightedTaskGroup at the start. That way if it is "undefined" (or assuming you update as suggested, "") you can return a null value immediately and in the carousel.component use the null value instead of checking for -1

EsmeeIDEMS (approved):
Functional test passed. Merging this even though still awaiting further review from Chris, since this fixed a load of bugs that are currently present on master. Hopefully we can improve from here.

PR #1596, opened 2022-10-27
fix: toggle bar change detection

chrismclarke:
PR Checklist

## Description
Fix issue where toggle bar would fire toggle twice

## Git Issues

Closes #

## Screenshots/Videos


chrismclarke:
> This works, including in my use case that was broken. However it does remove ionic's native animation from the toggle element. Obviously more important to have a working toggle bar for now

Yes, but I think it's the case that the toggle sits within the button no? i.e. `<ion-buton> .... <ion-toggle> .... </ion-button>`? 

image

This will always cause the ripple animation as you are in fact clicking the button to click the toggle that is inside it - the only thing we can interfere with is the following order of script execution, which starts from the inside outwards. If you want to avoid the ripple then you will need to move the toggle out of the button and reposition using absolute positioning, but that would be a bit of a refactoring pain.

PR #1599, opened 2022-10-27
fix: delimited contexts

### LINKED ISSUE #1572: [BUG] Review @row inconsistencies
**Describe the bug**
Flagged in #1569 - it appears that some nested properties replace @row in different ways, parsing string @row but not templated {@row} (but does remove the template braces). 

Ideally `@row` shouldn't appear in any final templates, except data_pipes which will then be processed during pipe parsing

It would be good to add tests to make sure all @row statements are converted as intended, including
- Use of both simple and templated syntax
- Use within action and parameter lists

**Screenshots and videos**
Before minor fix could see some @row statements still in final content which it should not be
image

After fix only instance found was in old debug sheet, but still suggests @row statements might not work in action lists
image

chrismclarke:
PR Checklist

## Description

When parsing some dynamic expressions the delimiters are lost, e.g. 
`@data.workshop.{@fields.current_workshop}.relax.title` ->`@data.workshop.@fields.current_workshop.relax.title`. 

This is due to the fact that the parser attempts to evaluate the inner `{@fields.current_workshop}`, although fields should only be processed at runtime. When the parser recognises this it returns the original value `@fields.current_workshop`, however it has lost the delimeters.

This PR fixes and adds additional tests so that the sheet parser does not alter additional `@` contexts that it is unaware of (i.e. only be `@row`). Everything else is handled at runtime

## Git Issues

Closes #1572

## Review Notes
- Changes affect dynamic parsing during sheet processing, so ideally should check before/after on a clean(ish) content sync

- Changes won't be seen properly in the frontend, as we have not integrated delimited syntax evaluation e.g.
`@data.{@fields.some_field}.id`
Currently this is only integrated for `{@row}`), so this should be created as a follow-up issue. 
I've created [debug_delimeters](https://docs.google.com/spreadsheets/d/1jr04BjxLRzKD2QTBoCq0iEF2-JcZ3EINbWsLgs3ZoDM/edit#gid=1276574703) while testing this behaviour

## Screenshots/Videos
Example parsed `common_workshop_task_gs` which now preserves `{@fields}` within the more complex expression
```
{
      "id": "relax",
      "individual": true,
      "together": true,
      "priority": 3,
      "has_single_task": true,
      "title": "@data.workshop.{@fields.current_workshop}.relax.title",
      "_translations": {
        "title": {}
      },
      "_translatedFields": {
        "title": {
          "eng": "@data.workshop.{@fields.current_workshop}.relax.title"
        }
      }
    },
```


EsmeeIDEMS (approved):
Works for me as Chris described. Indeed to resolve the issue I was having we still need the front-end addition.

EsmeeIDEMS:
In fact the syntax `@data.debug_delimiters_list[@local.target_row][@local.target_field]` Chris proposed in [debug_delimiters](https://docs.google.com/spreadsheets/d/1jr04BjxLRzKD2QTBoCq0iEF2-JcZ3EINbWsLgs3ZoDM/edit#gid=1276574703) does work at the moment

PR #1602, opened 2022-10-28
test: add data list parser tests

chrismclarke:
PR Checklist

## Description

Add spec tests for data_list specific code. These were drafted while debugging some other issues, so thought would be a good time to add (will still need to add future tests for template and default parsers)

## Review Notes
- The data_list parser extends the default parser, and so doesn't actually handle many extra features. It would be good to review the features it does handle and see if still relevant (not sure `notification_schedule` is used in any data_lists)

- Run the tests via `yarn workspace scripts test:watch`. Any single test can be run standalone by converting `it(...)` to `fit(...)`

## Git Issues

Closes #

## Screenshots/Videos
**Specific data_list capabilities added to test spec**
image


jfmcquade (approved):
Nice to have. Will be good to build upon and expand test coverage of parsers

PR #1603, opened 2022-10-28
Feat/feedback sidebar

chrismclarke:
PR Checklist

## Updates 2022-11-08
**Dev Notes**
I realised I'm a bit limited in what I can do regarding what is shown in the suggest changes mode as this actually is managed by one of the gdrive templates (had previously been requested to do so). So for now I've just updated to include the original text in the box

I've also added a small hardcoded set of instructions, improved the nav enable/disable handling, and added the option to delete feedback prior to syncing

**Author Notes**
The current [feedback templates](https://docs.google.com/spreadsheets/d/1IsjWr7NcPspsUWstdFKzTRD6iKmNxCutipQ5a0vvLsU/edit#gid=700972401) are marked as debug , so they won't be included in za/tz versions unless changed

There is currently no means within the current templates to enable feedback mode, suggest adding a button somewhere that navigates to a URL with the feedback-url sidebar enabled, e.g. `template/home_screen(sidebar:feedback)`

[Parenting App (2).webm](https://user-images.githubusercontent.com/10515065/200660475-b72987b7-edba-475f-98bb-a2d44a9d4343.webm)

## Description

- Integrate the previous feedback debug page into the sidebar view
- Add support for manipulating content page width
- Add toggle to enable/disable feedback-mode click interactions
- Add toggle to enable/disable navigation events whilst in feedback mode (e.g. to add feedback onto button which would otherwise navigate to another page)

## Review Notes
- Video shows default width as 400px (with 50px increments), but now changed to 480px (with 160px increments), to align with comments from #1585 

## Git Issues

Closes #

## Screenshots/Videos

[Parenting App.webm](https://user-images.githubusercontent.com/10515065/198421747-c19cd46c-3ac5-428b-a4d8-270c47fd4d89.webm)


chrismclarke:
Note - broken build due to issue on main branch (resolved via #1652)

EsmeeIDEMS (changes requested):
- When navigating to a different page the page width jumps back to its default 480px. Similarly, when you untick the "Enabled" tickbox and navigate to a different page it is ticked again. 

[screen-capture (12).webm](https://user-images.githubusercontent.com/74557272/203808905-2dca0468-9221-4fb5-9291-adb33e125ce3.webm)

- I think it's potentially confusing that feedback can also be given on the feedback/suggest changes pop-ups. Especially in the suggest changes pop-up it is natural to select some parts of the text to delete/adapt, and so the feedback menu appears. 

[screen-capture (13).webm](https://user-images.githubusercontent.com/74557272/203810675-0a38151c-3060-46d4-b0f9-2c502e93a1a0.webm)

chrismclarke:
Just to try quickly revive this one.

Thanks for the comments @EsmeeIDEMS , I've fixed the issue where settings didn't persist across navigation and also made it so you can't add feedback on the feedback popup. I've also updated the feedback metadata to include app theme and skin.

Let me know if all looks ok from your side and then hopefully this will be good to go (can always add improvements once people have started using)

image

chrismclarke:
@EsmeeIDEMS 

Just to follow-up on what was mentioned on the call last week:

If you want to include the name of the author on the feedback template then probably the easiest option is just to include it in the [feature_feedback](https://docs.google.com/spreadsheets/d/1IsjWr7NcPspsUWstdFKzTRD6iKmNxCutipQ5a0vvLsU/edit#gid=700972401) templates. The only caveat is the `emit : feedback : some_feedback` script expects a single string entry for the feedback, so you might need an intermediate variable like `feedback: @global.feedback_author_name - @local.feedback_text`

chrismclarke:
@EsmeeIDEMS 
I was able to take a quick look and figure out a way to fully disable all actions (not only navigation), so I think that should fix the main issue you raised (see video below)

Otherwise I think combo boxes make sense to stay how they are, because in feedback mode you probably will want to be able to access the list of values and also because there isn't an easy way to change that behaviour

[Parenting App (3).webm](https://user-images.githubusercontent.com/10515065/213366270-905970f5-27d1-44b8-a2f0-42a7cbed8f0d.webm)

src/app/feature/feedback/routeGuards.ts
```diff
@@ -0,0 +1,16 @@
+import { Injectable } from "@angular/core";
+import { CanActivate } from "@angular/router";
+
+@Injectable({ providedIn: "root" })
+export class RouterDisableGuard implements CanActivate {
```
chrismclarke, line 5: Adding router guards seemed to be the best way I could find to disable navigation. They can be applied/toggled to all routes, to either block or allow navigation to that route

src/app/shared/components/template/services/template-field.service.ts
```diff
@@ -14,16 +15,34 @@ export class TemplateFieldService extends AsyncServiceBase {
...
     this.registerInitFunction(this.initialise);
+    this.registerTemplateActionHandlers();
   }
 
   private async initialise() {
     await this.ensureAsyncServicesReady([this.dbService, this.translateService]);
     this.ensureSyncServicesReady([this.localStorageService]);
   }
+
+  private registerTemplateActionHandlers() {
```
chrismclarke, line 31: It wasn't strictly necessary to move the field actions to their parent service (and use the new registry system to register), as I later realised I could simply override all global action registry entries which take precedence over anything defined in the template-action service. However now it's here it feels to me to be a more sensible place to be and helps tidy the template-action registry code a little

jfmcquade (approved):
Looks good. 

I haven't added any feedback to the code, but good to have a look and get an idea of how it works.

Issue #1598, opened 2022-10-27
[BUG] Filter operations don't work when argument contains variables


esmeetewinkel:
**Describe the bug**
1. I don't seem to be able to filter for a column name that is set through a field. e.g. `filter: @fields.some_field`

2. An item pipe of the form `filter: @item.some_column == @local.some_local` does not seem to work. See [example_items_pipe_2](https://docs.google.com/spreadsheets/d/1n82prWaYx-HWKW0ibxv-sO_Vuw9WxvuNLRJ_jI_WXtk/edit#gid=1871410965)

**Use case (for 1)**
The field `workshop_path` can have value `individual` or `together`. Also `individual` and `together` are column names in a data list (with boolean entries). I would like to have a `begin_items` row, with value this data list, and parameter list `filter: @fields.workshop_path`. This does not work.

chrismclarke:
A few notes on this issue mentioned in #1607

chrismclarke:
Trying to unpick this issue (and more a note to self).

The main problem is the way that the filter operation that only appears as a single row entry needs to be applied to every row generated after applying the template to the item list.

This means that the dynamic context is evaluated in 2 separate actions, step 1 is the template evaluator which handles `@local`, `@field` etc. The nature of the template evaluator means that the partial evaluation replaces as an unquoted string (normally to be rendered), i.e.

`@item.first_name=== @local.other_field` -> ` @item.some_field === some_value`

When it comes time to evaluate the item which is done using a separate evaluation system, the `appStringEvaluator` performs more string replacement, which doesn't help to evaluate the expression.
`@item.first_name=== some_value` -> ` Ada=== some_value`

The possible workaround could be:
1. Add a workaround to force template evaluator to retain quoted strings `@item.some_field === "some_value"` and update the `AppStringEvaluator` to do similar before evaluating ` "Ada"=== "some_value"` -> `false`

2. Bypass the template evaluator to avoid partial string substitution and use a different system to evaluate as a single expression with full context value `@item.first_name=== @local.other_field` , `{item:first_name:'Ada', local:other_field:'some_value'}`

PR #1607, opened 2022-10-28
feat: add support for item shuffle

### LINKED ISSUE #1606: [FEATURE] Runtime operation sort randomly
**What?**
I would like to be able to pull an item randomly from a data list.

**Why?**
For certain tasks I want to select one at random (among those with highest priority). 

**How?**
I thought this could be a data_pipe operation applied on an items environment (added to the existing ["Runtime Operations"](https://docs.google.com/document/d/1Pk2gYNhtu68XJo-U-h6RY8ZD8gOPaCnLOlCft4IjLY4/edit#heading=h.6uv9domhy4kr)). Suggestions: 
-  (preferred) Built into `sort`: When `sort` finds items with the same value then it orders them randomly. So the parameter list would be: `sort: priority; limit: 1`
- Alternatively, we could add a new operation `sort_randomly` - but then it's not so clear how to author the parameter list entry: `sort: priority; filter: [those of lowest priority]; sort_randomly`

chrismclarke:
PR Checklist

## Description

- Add a `shuffle` operation that can be used to randomise the order of an item list when displaying
- Add [example_items_operations](https://docs.google.com/spreadsheets/d/1n82prWaYx-HWKW0ibxv-sO_Vuw9WxvuNLRJ_jI_WXtk/edit#gid=579616705) to capture main runtime item operations

## Review Notes
- From the dev side changes are minimal. We already had an array shuffle method so I've just included as the _shuffle_ operator. So I'd say no code review required, just check with authoring whether _shuffle_ works as an operation name or if anything else is preferred.

- I could see quite a few sheets related to items, although I couldn't find any that contained a list of all current operations so felt it made sense to add (but feel free to move/modify)

## Git Issues

Closes #1606

## Screenshots/Videos

[Parenting App (1).webm](https://user-images.githubusercontent.com/10515065/198692690-a1a38e9d-9ae7-46d3-873d-1405325f5fea.webm)


EsmeeIDEMS (changes requested):
`shuffle` seems to work correctly when used on its own, and I'm fine with the terminology. We didn't have a sheet with all operations so thanks for creating.

When used together with filter and limit though, it doesn't seem to work: in the example you created items that don't satisfy `field_value >2` do show up. 
image

image

chrismclarke:
> When used together with filter and limit though, it doesn't seem to work: in the example you created items that don't satisfy `field_value >2` do show up. image
> 

Whoops, that's on me - in the frontend we need to refer to the items with `@item` syntax.

The reason the syntax is different filtering in backend and frontend is because the backend only has one context (the sheet row), and so all the variables of the row are populated to access directly (which relates to the issue flagged previously on sheet row naming), whereas frontend item _should_ the looped item context alongside globals, locals and other field variables.

The reason I say _should_ is because currently it's only receiving the item, hence the issue raised in #1598. So when we address that it would be good to also clarify/document these issues

EsmeeIDEMS (approved):
Functional test passed

Issue #1246, opened 2022-02-17
[FEATURE] Dynamic list variables


esmeetewinkel:
**What?**
- [ ] **dynamic data list entries**
Currently the entries in every column of a data list are static, similar to global constants. Previously proposed syntax was to add a prefix `fields::` to a column name to indicate it is dynamic. The entries in this column would be used as the "field default"
- [ ] **(dynamic) list variables**
Currently variables cannot be interpreted as lists. We'd like to have this, and functionality to remove from/add to/clear a list.

**Why?**
Currently I'm using workarounds involving fields, but the proposed would be more powerful and neater.

chrismclarke:
Discussed on call 2022-02-18
No consensus on best way to proceed. Agreed that prefixes aren't ideal as less human-readable, and also that could conflict with the `text::lang` syntax. 
Links back to another issue of how to do things like specify data types for fields (e.g. `Set_field<number>: my_number: 5`) - much of which is currently interpreted or workedaround via `+` operators, and additionally operations on existing data such as `update_field` (particularly is field can be used to save a list.

Possibly may want means of adding a write-layer on top of existing datalists so that a user could override
`@data.workshop.w_self_care.completion_level` via a `set_data | workshop.w_self_care.completion_level` action or similar, and then would just need a syntax to indicate in the data list that the column is writeable (e.g. `completion_level :: default` or `$completion_level` or similar).

PR #1609, opened 2022-10-28
chore: tidy task-specific code

chrismclarke:
PR Checklist

## Description

As PR #1592 has already been merged a few specific changes which would have been requested as part of PR re-review. Namely:
- move task-specific code to own function in carousel component
- split `getHighlightedTaskGroupIndex` calls to two functions within component calls 
- minor bug fixes for cases where data_lists do not exist
- tidy a few function names and comments

Note 1 - When it comes to comments intellisense will usually provide autocomplete and info boxes for comments that start `/** ` (and then a line break), so that is most of the changes there

Note 2- I started trying to work through the `task.service.ts` but I think the coupling bewteen data_lists and specific fields is a bit tricky to unpick and will likely require larger, breaking changes - so I've mostly left alone for the time being but would be good to revisit in the future.

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade (approved):
Thanks for this. Good to have another pair of eyes on some of these methods, these changes have helped to clarify what's going on. 

I agree that we should revisit the task service before too long. Perhaps the context in which to do so would be when implementing a system of dynamic data lists that can be altered at runtime (see issue #1246), if that is still a goal. This functionality could replace that in the current implementation of tasks whereby the task data lists contain references to field names which must be looked up and altered independently of the data lists themselves

PR #1616, opened 2022-10-31
feat: added 'button' style param to task-card

### LINKED ISSUE #1608: Button with card style vs task card
**Details**
I was wondering if the button with style "card" (used in the drawer) should really just be a variant of the task card. 

Currently, since they are "buttons", they don't show that they've been completed when the task has been completed. This feels a bit counterintuitive, since the task cards for individual tasks look very similar (might have the same title and click-action) but these  **do** get completed. 

This counterintuitive behaviour is particularly present at the moment, since the things-to-do doesn't currently update unless the app is refreshed - and so the completed task will stay there, with nothing visual indicating that it has already been done.  

**Screenshots and videos**
image

[screen-capture (5).webm](https://user-images.githubusercontent.com/74557272/198698179-58dc264d-bec2-401f-893e-18023674ebcf.webm)

jfmcquade:
PR Checklist

## Description

Adds a new style parameter option to the `task-card` component: `style: button`. Makes the task card display as a narrow button with the image appearing small and on the right. A toggle can also be provided in the child rows of the component. This is visually equivalent to the existing `button` component with `style: card`, but, as outlined in Issue #1608, the `task-card` component supports displaying the completion status of its related task, e.g. a green tick.

## Git Issues

Closes #1608 

## Screenshots/Videos

[comp_task_card](http://localhost:4200/template/comp_task_card) updated with examples of new style.

Screenshot 2022-11-01 at 12 52 22

Screenshot 2022-11-01 at 12 52 55

### DIFF
--- a/src/app/shared/components/template/components/task-card/task-card.component.html
+++ b/src/app/shared/components/template/components/task-card/task-card.component.html
@@ -30,6 +30,17 @@
       </plh-template-component>
     </ng-container>
     <ng-container *ngIf="title">
+      <ng-container *ngIf="isButton">
+        <div class="button-content">
+          <plh-template-component
+            *ngFor="let childRow of _row.rows | filterDisplayComponent; trackBy: trackByRow"
+            [row]="childRow"
+            [parent]="parent"
+            [attr.data-rowname]="_row.name"
+          >
+          </plh-template-component>
+        </div>
+      </ng-container>
       <div class="text-wrapper">
         <div *ngIf="title" class="title-wrapper">
           <h1>
--- a/src/app/shared/components/template/components/task-card/task-card.component.scss
+++ b/src/app/shared/components/template/components/task-card/task-card.component.scss
@@ -36,7 +36,7 @@
   }
 
   &.button {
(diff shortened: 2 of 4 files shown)


EsmeeIDEMS (commented):
I am a bit confused here - the `completed_field` parameter does not show up in the template `comp_task_card`, is it just that it's out of date? So how are the components in this template aware of whether they are completed?

jfmcquade:
> 



> I am a bit confused here - the `completed_field` parameter does not show up in the template `comp_task_card`, is it just that it's out of date? So how are the components in this template aware of whether they are completed?

You're right, the [comp_task_card](https://docs.google.com/spreadsheets/d/1_RY8HbsATOAZU2EDXY5R9OOE_YBEbW0EjTP5zxYP2aA/edit#gid=569531329) template was out of date. I've updated it, along with a minor tweak to the code to properly handle the "completed" status on cards for individual tasks.
I've also added styling for a task-card with `style: button` in the case where an image is not provided

EsmeeIDEMS (approved):
Functional test passed

PR #1629, opened 2022-11-02
fix: drawer component always expands fully

### LINKED ISSUE #1628: [BUG] Drawer doesn't open fully
**Describe the bug**
After releasing the TZ app to the play store, I noticed that the drawer doesn't open fully. I didn't find a way to make the issue disappear - closing and reopening the app didn't resolve it. I did also notice the issue when testing locally on my pc today erratically. 

**App version**
v0.16.7 plh_tz (Play Store App)

**Screen video**

https://user-images.githubusercontent.com/74557272/199365035-d4a47e61-e83e-4705-837f-e90d534f618d.mp4

jfmcquade:
PR Checklist

## Description

Fixes a bug whereby the drawer component would sometimes not fully open (Issue #1628).

This bug was due to the "open height" for the drawer being calculated on initialisation of the component, which would sometimes happen before the drawer element and its children had finished loading. This PR forces the open height to be recalculated each time the drawer is expanded, ensuring that the value is up to date with the DOM. This has the additional benefit of allowing for the drawer's contents to change dynamically between being opened, should this functionality be required.

## Git Issues

Closes #1628

## Screenshots/Videos

Demonstrating the drawer working as expected in the template [home_screen_modular](https://docs.google.com/spreadsheets/d/1Whi-zledVspIAd_hmAiItv2OxWnmtOVs8PEuCl0Eu8U/edit#gid=1771109624). As the bug was intermittent, this doesn't demonstrate that it is fixed, but it does show that the new method is functional.

https://user-images.githubusercontent.com/64838927/199494388-f65f90fe-341a-4526-9ecc-555d0ca13e82.mov

### DIFF
--- a/src/app/shared/components/template/components/drawer/drawer.component.ts
+++ b/src/app/shared/components/template/components/drawer/drawer.component.ts
@@ -8,16 +8,10 @@ import { TemplateBaseComponent } from "../base";
   styleUrls: ["./drawer.component.scss"],
 })
 export class TmplDrawerComponent extends TemplateBaseComponent implements OnInit {
-  @ViewChild("drawer", { read: ElementRef, static: false }) drawer: ElementRef;
+  @ViewChild("drawer") drawer: ElementRef;
 
   isOpen = false;
   style: string;
-
-  get openHeight() {
-    const drawer = this.drawer.nativeElement;
-    const handle = drawer.children[0];
-    return drawer.offsetHeight - handle.offsetHeight - 10;
-  }
 
   ngOnInit() {
     this.getParams();
@@ -35,8 +29,14 @@ export class TmplDrawerComponent extends TemplateBaseComponent implements OnInit
       this.isOpen = false;
     } else {
       drawer.style.transition = "transform .3s ease-out";
-      drawer.style.transform = `translateY(${-this.openHeight}px)`;
+      drawer.style.transform = `translateY(${-this.getOpenHeight()}px)`;
       this.isOpen = true;
     }
+  }
+
+  private getOpenHeight() {
+    const drawer = this.drawer.nativeElement;
+    const handle = drawer.children[0];
+    return drawer.offsetHeight - handle.offsetHeight - 10;
   }
 }
--- a/src/app/shared/components/template/components/drawer/drawer.component.ts
+++ b/src/app/shared/components/template/components/drawer/drawer.component.ts
@@ -1,4 +1,4 @@
-import { Component, ElementRef, ViewChild, AfterViewInit, OnInit } from "@angular/core";
+import { Component, ElementRef, ViewChild, OnInit } from "@angular/core";
 import { getStringParamFromTemplateRow } from "src/app/shared/utils";
 import { TemplateBaseComponent } from "../base";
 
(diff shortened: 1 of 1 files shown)


src/app/shared/components/template/components/drawer/drawer.component.ts
```diff
@@ -7,22 +7,20 @@ import { TemplateBaseComponent } from "../base";
...
+export class TmplDrawerComponent extends TemplateBaseComponent implements OnInit {
+  @ViewChild("drawer", { read: ElementRef, static: false }) drawer: ElementRef;
 
   isOpen = false;
-  openHeight = 0;
   style: string;
 
-  ngOnInit() {
-    this.getParams();
+  get openHeight() {
```
chrismclarke, line 16: nit(non-blocking)
I don't see this accessed from the component html so should it be marked as private? also does it need a getter or would it work fine as a regular method? Generally getters are more associated with lookups or alisasing to quickly access things from the frontend component, so whilst the code still works fine it does make me think that openHeight exists in the template somewhere.

src/app/shared/components/template/components/drawer/drawer.component.ts
```diff
@@ -7,22 +7,20 @@ import { TemplateBaseComponent } from "../base";
   templateUrl: "./drawer.component.html",
   styleUrls: ["./drawer.component.scss"],
 })
-export class TmplDrawerComponent extends TemplateBaseComponent implements OnInit, AfterViewInit {
-  @ViewChild("drawer", { read: ElementRef }) drawer: ElementRef;
+export class TmplDrawerComponent extends TemplateBaseComponent implements OnInit {
+  @ViewChild("drawer", { read: ElementRef, static: false }) drawer: ElementRef;
```
chrismclarke, line 11: question
I've never used the 'read' property on viewchild, what does it do in this example?
Also given that the draw is only read during toggle do we need to mark as static: false? (again, out of interest, haven't tested myself)

chrismclarke (approved):
minor (non-blocking) comments but otherwise looks good to me

src/app/shared/components/template/components/drawer/drawer.component.ts
jfmcquade: It seems the "read" property is used to specify what to return in the case that the selector has some ambiguity. The ambiguity could be the element is an Angular component, in which case `@ViewChild()` could return the component instance or the ElementRef instance (explained [here](https://stackoverflow.com/a/37476195)), or "read" can be used to select a specific provider token (explained [here](https://blog.logrocket.com/understanding-the-viewchild-and-viewchildren-decorators-in-angular-10/#:~:text=modify%20embedded%20views.-,Using%20the%20read%20property,-The%20read%20property), I'm not sure I fully understand without having a reason to use it like this). 

Either way, I think it's superfluous here so I've removed it.

And I think you're right about `static: false`. I added it when I was still trying to get the element's height in `ngAfterViewInit()`, hoping that it would mean that the value would update after all the `*ngFor` elements had rendered, but this was not the case. In fact it's the default value anyway since Angular 9 apparently.

chrismclarke: Nice! yeah that sorta makes sense, and also makes sense that it isn't required for this use-case anymore

Issue #1372, opened 2022-05-12
[FEATURE] - Add global app strings to translation pipeline


chrismclarke:
**What?**
The app has a number of hardcoded strings for things like buttons, toast messages etc. 

Most of these currently sit in `packages\data-models\app-config\globals.ts` and used in the app via an exported `APP_STRINGS` property available from `APP_CONSTANTS` in `src/app/data`

These should ideally be integrated into existing translation system so that they can be translated in the same way global strings are translated, and populated dynamically depending on the current app language

**How?**
- [ ] Add methods to translation extract/compile scripts that also include the app config globals
- [ ] Extract any additional strings from codebase that would be good to translate
- [ ] Replace methods that call direct `APP_STRINGS.some_value` with a function call that also translates the string while returning (likely use existing translate service or introduce a pipe)
- [ ] Consider any better way to reduce amount of import/re-export of values 
(currently `packages\data-models\app-config\globals.ts` -> `packages\data-models\constants.ts` -> `src/app/data` -> components)

**Other details**
_Add any extra details that you think may be useful._

EsmeeIDEMS:
Some pages I can think of that contain strings that are not coming from spreadsheets:

https://plh-global.web.app/tour
https://plh-global.web.app/template
https://plh-global.web.app/campaigns
https://plh-global.web.app/notifications
Further to that:

toast message when feedback is submitted
default text and title for notifications

EsmeeIDEMS:
- "sections" in the task card
- "Active" in the task card 
(@jfmcquade I've added these here, not sure if you want to resolve this separately)

PR #1630, opened 2022-11-02
feat: override strings in task-card from templates

jfmcquade:
PR Checklist

## Description

Adds 2 optional parameters to the `task_card` component: `highlighted_text` and `progress_units_name`. If provided, the values replace the strings that were previously hardcoded into the component: "Active" and "sections" respectively.

This enables these strings to be included in the translation system. These parameters should be specified, with the values "Active" and "sections", for all instances of the `task_card` component currently in the templates that display a task group (i.e. a workshop in our current usage), along with taking any other steps to ensure that the values of these parameters are included in the translation system. 

## Git Issues

Closes #

## Screenshots/Videos

Demonstration of custom values in [feature_task_card_params](https://docs.google.com/spreadsheets/d/1_RY8HbsATOAZU2EDXY5R9OOE_YBEbW0EjTP5zxYP2aA/edit#gid=1039832506)
Screenshot 2022-11-02 at 14 01 56

### DIFF
--- a/src/app/shared/components/template/components/task-card/task-card.component.ts
+++ b/src/app/shared/components/template/components/task-card/task-card.component.ts
@@ -13,7 +13,8 @@ import { TemplateFieldService } from "../../services/template-field.service";
 export class TmplTaskCardComponent extends TemplateBaseComponent implements OnInit {
   style: string;
   highlighted: boolean;
-  highlightedText = "Active";
+  highlightedText: string;
+  subtasksText: string;
   progressStatus: IProgressStatus = "notStarted";
   taskGroupId: string | null;
   taskGroupDataList: string | null;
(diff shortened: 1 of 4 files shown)


EsmeeIDEMS (approved):
Functional test passed. Happy with this for now, as it resolves the problems with translation. Ultimately we should lift the defaults out of the code as well but we've tracked this in issue #1372 already. 
image

src/app/shared/components/template/components/task-card/task-card.component.ts
```diff
@@ -13,7 +13,8 @@ import { TemplateFieldService } from "../../services/template-field.service";
 export class TmplTaskCardComponent extends TemplateBaseComponent implements OnInit {
   style: string;
   highlighted: boolean;
-  highlightedText = "Active";
+  highlightedText: string;
+  subtasksText: string;
```
chrismclarke, line 17: nit(non-blocking)
I often try to avoid 'sub' as a prefix as it tends not to be a very extensible keyword (e.g. sub-subtasks). I find 'child' seems to work a bit better as devs reasonably understand the children of one component/element can then become parents of others, and children of children usually work fine defined just as 'child'. But that's just personal preference, no huge issue either way

chrismclarke: actually in second thought does it even need reference to the subtasks? would it make more sense as progressText or some other keyword that better defines what it's actually targeting? Is this just the word being used to describe the group of items? I'm sure there could be something a bit more specific e.g. progressGroupName or similar? Probably worth checking with authoring

chrismclarke (approved):
Looks fine to me, minor (non-blocking) comment on naming but otherwise looks good

chrismclarke:
`progress_units_name` - I like it!

PR #1638, opened 2022-11-04
fix: deployments without task groups do not error on initialisation

jfmcquade:
PR Checklist

## Description

The task service was causing a breaking error for builds of deployments that did not make use of the task system. Specifically, the task service currently uses a hardcoded value, `workshop_tasks`, to look up the data list of task groups that can be highlighted. If this data list was not present for a given build, the app would error on initialising the core services.

This PR offers a workaround by not fully initialising the task service unless the `workshop_tasks` data list is found. There is further discussion to be had about the potential for generalising the "task" system in a way that makes sense across deployments.

## Git Issues

Closes #

## Screenshots/Videos

The bug as it previously presented:
image-1

### DIFF
--- a/src/app/shared/services/task/task.service.ts
+++ b/src/app/shared/services/task/task.service.ts
@@ -11,8 +11,8 @@ export class TaskService {
   highlightedTaskFieldName = "_task_highlighted_group_id";
   taskGroupsListName = "workshop_tasks";
 
-  taskGroups: any[];
-  taskGroupsHashmap: Record<string, any>;
+  taskGroups: any[] = [];
+  taskGroupsHashmap: Record<string, any> = {};
   constructor(
     private templateFieldService: TemplateFieldService,
     private appDataService: AppDataService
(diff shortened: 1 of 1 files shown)


src/app/shared/services/task/task.service.ts
```diff
@@ -17,14 +20,17 @@ export class TaskService {
 
   async init() {
     await this.getListOfTaskGroups();
-    this.evaluateHighlightedTaskGroup();
+    if (this.taskGroups.length) {
```
chrismclarke, line 23: nit(blocking)
Can you add `>0`? I know zero is falsy but still just feels a bit wrong to me.
Also I think no harm initialising both the taskGroups and taskGroupsHashmap variables (as `[]` and `{}`) respectively, just to tidy up in cases where issues happen during init

EsmeeIDEMS (approved):
Thanks. I'm able to sync EFM content on this branch and am no longer getting the error (in red). I am still getting the warning about not finding `workshop_tasks`, though (in yellow, see screenshot).

image

chrismclarke (approved):
Thanks for making the changes

PR #1641, opened 2022-11-07
Content/2022 11 07

EsmeeIDEMS:
PR Checklist

## Description
PR updates PLH content

### DIFF
--- a/packages/app-data/assets/index.ts
+++ b/packages/app-data/assets/index.ts
@@ -3170,8 +3170,8 @@ export const ASSETS_CONTENTS_LIST = {
     md5Checksum: "681910eca8201998e670a0f3caffa9e7",
     themeVariations: {
       theme_professional: {
-        size_kb: 12.7,
-        md5Checksum: "340dc0cd22ef431e6d09f9af4b643640",
+        size_kb: 27.9,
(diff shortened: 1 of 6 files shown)


packages/app-data/assets/index.ts
```diff
@@ -3170,8 +3170,8 @@ export const ASSETS_CONTENTS_LIST = {
     md5Checksum: "681910eca8201998e670a0f3caffa9e7",
     themeVariations: {
       theme_professional: {
-        size_kb: 12.7,
-        md5Checksum: "340dc0cd22ef431e6d09f9af4b643640",
+        size_kb: 27.9,
```
EsmeeIDEMS, line 3173: @jfmcquade I added one image and the professional theme asset size more than doubled - does that make sense to you?

jfmcquade: This is just the size of that single image, namely the professional variation of `plh_images/workshops/w_self_care/guide_2/reward.svg` – each one has a separate entry in this `index.ts` file.

 And looking at the changed image, the new one is significantly more complicated so it being double the size sounds about right:
Screenshot 2022-11-08 at 10 28 48

~30kb is very small anyway, one of the advantages of using svg files. A suitably scaled png or jpg might be more like 500kb.

So yes, all good.

PR #1654, opened 2022-11-10
spike: add component docs generator

chrismclarke:
PR Checklist

## Updates 2022-11-24
Ready for review

@EsmeeIDEMS 
- The generated docs will be included in the main docs site after this PR is merged (no easy way to preview beforehand but shouldn't be too important in any case). See comment thread below for latest screenshot with example of the button component docs for a clearer idea.

- As well as adding support for the generated docs both me and Johnny have done a little bit of housekeeping on the button component. I don't expect there to be any knock-ons and have checked the `comp_button` template to see all looks as expected, but I'd appreciate if you could also quickly check that template from the preview link just to make sure it also matches how you remember

@jfmcquade 
If you want to view any of the minor code updates you'll have to skip past all the generated component docs in the changes (in hindsight I should have just waited to add the generated docs as a follow-up)

The only _real_ code changes are the ones seen outside the docs/generated folder
image

## Description
Test generating component docs via compodoc to assess its merits. I've added the package with some rough configuration (limited mostly just to template components) and yarn script to run, generate docs and open in web preview. I've omitted the generated outputs from the PR just to keep the main code changes more visible

**Run locally**
```
yarn install
yarn docs:generate
```
_Note_ this could take some time to generate. Up till now output files not committed into branch

My initial feeling is that it populates quite a lot, possibly too much to be much use for non-developers. That being said it might be easier to adjust the components slightly to expose in a better way the interfaces/params most useful to content-developers and encourage us to improve our codebase at the same time. I haven't had chance to test much

**Discussion points**
- Are they useful? How much work would it take to make more useful?
- How best to integrate into existing docs?
- How would this compare to something more curated like storybook?

**Possible next steps**
- Improve outputs
- Integrate into docs
- Add to CI

## Git Issues

Closes #

## Screenshots/Videos
**Generated docs demo**

[frontend documentation.webm](https://user-images.githubusercontent.com/10515065/200983533-cd0b3756-7fc8-4aa3-9e1a-4f829bfa99a8.webm)


jfmcquade:
Thanks for looking into this. I've had a play around and I have some (not very decisive) thoughts. The way I see it there are 2 options for making use of auto-generated docs:

### Option 1: Use auto-generated docs as an intermediary between the code (devs) and the docs (docs contributors)
- The aim would be to provide a resource from which "contributors" without in-depth technical knowledge could write documentation that is more accessible than these auto-generated docs. For this purpose, I agree that there is probably too much information included in the generated docs by default.
- I've pushed a commit where I annotate the button component ts file in order to make its generated docs more readable. For properties that relate directly to template row paramaters, I've added `TEMPLATE PARAMATER: <name of corresponding parameter>`, as a suggestion of how we could flag which properties are relevant to content authors. With this goal in mind, I've also added `@ignore` tags to properties that aren't needed so that don't appear in the docs. However in the latter case, I feel like this clarity in the docs probably isn't worth peppering the code with too much additional information like this. Cleaning up and annotating the code is of course no bad thing – it could be a good task for a new developer, or someone wanting to learn their way around the code. 
  https://user-images.githubusercontent.com/64838927/203544863-3564e5ee-465b-4ead-88ff-64ea4aa9a207.mov
- For this purpose, I don't necessarily see much added value in using these auto-generated docs. The impetus to clean-up, standardize and annotate existing code would be a positive. But as part of a multi-step process that starts with the codebase and ends with author-readable documentation, I think it might be simpler for devs to just add the parameters for a given component to a Google Doc, for someone (possibly themselves) to flesh out. If a change was made to a component (e.g. a new parameter added), then it wouldn't be reflected anywhere automatically, but if the docs already existed then it would be very simple for a dev to add the new parameter to the existing markdown files immediately.

### Option 2: Use autogenerated docs to replace our current docs.
- As our docs aim to be readable by less technical content authors, they couldn't be replaced by the auto-generated component docs alone. However, compodocs does support including a README for each component. This is achieved by putting a conventionally named markdown file in the same folder as the component's ts file. Image. These md files could be identical to those used in our current docs. Downsides to this are that it mixes docs and functional files together in the codebase, and makes it more difficult for someone who wants to work just on docs.
  Screenshot 2022-11-23 at 11 46 49
- Compodocs also supports https://compodoc.app/guides/tips-and-tricks.html#additional-documentation adding "additional documentation", which could be used to contain all our non-component related documentation (e.g. Authors Quickstart etc)
- I haven't properly looked into storybook https://storybook.js.org/ It looks like it could be really useful at some point to demonstrate what is possible in the app UI. But presumably it would take more work to set up and maintain – and possibly wouldn't be the place for our additional docs detailing dev set-up and authoring techniques? Perhaps it could be though


As you can tell I'm a bit on the fence about whether using auto-generated docs would be a good option. I'm inclined to think that with our current resources, a simpler effective option would be to use google docs as an intermediary for devs to sketch out documentation that can be built out and added to the official docs separately. But auto-generated docs could still be part of that process, and I can see other benefits such as cleaning up the codebase and having an up-to-date resource to refer to. Having docs like these could have been useful for me as a dev when I was getting to know my way around the codebase, but I'm not sure they would have made things much easier than looking at the code itself and working on a few components.

In terms of Google Docs, I've made a template for component documentation [here](https://docs.google.com/document/d/1EIG2vfyELRJqYRzU3XvpLahYMwkVJ2I-6Fn46YPKbLk/edit#).
[This Google Workspace app](https://workspace.google.com/u/1/marketplace/app/docs_to_markdown/700168918607) can convert the document to markdown, and seems to work quite well. Although images still need to be handled manually to an extent.

chrismclarke:
Thanks for the detailed and thoughtful responses. I can definitely see where you're coming from on both accounts - really cool that you were able to find ways to make the generated docs more readable/clean, agree that it would be a good impetus to tidy up the codebase but also that it could be a fair amount of work that doesn't get us much closer to easy-to-read documentation for authors.

The readme option is a really interesting one - I'm not so worried about things being nested within the components codebase as I think it will make sense to hoist all the components out of the nested template folders anyway and just have a standalone `components` workspace (e.g. packages/components or just components). So having additional files for each wouldn't be a big issue, but it definitely changes the technical requirements of the potential documentation authors.
 
Sidenote - my only hesitation on separating out the components workspace right now is it will add a little extra complication with things like the start script (which would need extra hooks to watch for changes both within the main codebase and the components package), and would be better solved through improving our monorepo tooling (using something like nx or turborepo... probably nx)... but I digress...

It also makes me wonder if we do decide to go the route of authoring documentation readme files for each component what the benefit/tradeoff would be of using something like [storybook](https://storybook.js.org/) instead of compodoc. The nice thing about storybook is you can author quite interactive docs, so the author could potentially see the rendered component and just change the fields we want to show them how to change and observer the result. The obvious downside again is technical know-how of the author and time to setup infrastructure for it (probably a couple dev days).

A lot to think about for sure, might be good to roll into a few meeting when we have a bit more breathing space.

chrismclarke:
As another quick follow-up I looked at the button component code and properties you had ignored from the docs, and actually most can just be removed - all the code related to scale factors was legacy and doesn't appear in the frontend HTML anywhere anymore. Likewise the innerHTML looks like it's always created from the value, but not used in the component (no idea original intent, but all appears to work fine when I remove).

I've also told the generator to ignore lifecycle and private methods, and marked some of the base template component methods with `@ignore`. So now I think it actually looks pretty interpretable
 
127 0 0 1_8000_generated_components_TmplButtonComponent html (3)

chrismclarke:
I've also exported the documentation and added bindings to the main docs site to automatically open in a new tab from the menu under **External -> Compodoc Generated**

image

chrismclarke:
I think this should now be ready for merge. I'll update the main description with notes about how to view and potential knock-ons for button component, but otherwise I'd say something we can just slowly build on

jfmcquade (approved):
Looks good! Will be nice to have and build upon

EsmeeIDEMS (approved):
Functional test passed. 

- Button is working fine for me, I don't notice any changes. 
- I'm not able to get this to run locally by running `yarn install` followed by `yarn docs:generate`, however I'm able to navigate to the page via the mkdocs - seems to work well. 
image

chrismclarke:
Thanks both
@EsmeeIDEMS - apologies, I renamed the script to `yarn compodoc:generate` (to build) `yarn compodoc:serve` (to run locally) but forgot to update the notes. But great you were able to get it running within the main docs, as I'd say moving forward that's a more sensible place to look at. The live version is also deployed now, so can be linked to from  https://idemsinternational.github.io/parenting-app-ui/

@jfmcquade - another thing I forgot to mention is the generate script above isn't hooked into any CI triggers, so for now we will just run manually whenever we want to update the developer docs (in the future can trigger when source code changes within the components folders, but again might make more sense to wait until we've tidied up the template components folder structure a bit).

PR #1657, opened 2022-11-10
fix: default theme when switching deployment; skin's default theme

jfmcquade:
PR Checklist

## Description

Fixes 2 minor bugs relating to themes and skins:
1. Handles the case of a developer switching to a deployment which does not support the currently active theme. The theme will now be set to that deployment's default theme in such cases, where previously no theme was set, causing visual bugs.
2. Previously, when setting the active theme to one which was not the current skin's default, the theme would reset to the current skin's default whenever the app was reloaded. Now the chosen theme persists on reload.

Also removes some unused code from the style variables file.

## Git Issues

Closes #


src/app/feature/theme/services/theme.service.ts
```diff
@@ -21,12 +21,12 @@ export class ThemeService {
...
   init() {
-    // Retrieve the last active theme with default fallback
+    // Retrieve the last active theme and apply it, with default fallback
     const currentThemeName = this.getCurrentTheme() ?? this.defaultThemeName;
-    this.setTheme(currentThemeName);
+    this.setTheme(currentThemeName, true);
   }
 
-  public setTheme(themeName: string) {
+  public setTheme(themeName: string, isInit = false) {
```
chrismclarke, line 29: nit(non-blocking)
I'm not quite sure what the `isInit` parameter does. Could you rename it to something more explicit or add an `@param` comment? Or perhaps it might make more sense to include the logic in the init method itself? E.g. attempt to set the theme, and if not successful set default theme during init (assuming we have a `getTheme` or similar that can be used to check if the theme has been set)

chrismclarke (changes requested):
Just to check, will this still be needed after #1664?

jfmcquade:
> Just to check, will this still be needed after #1664?

Yes, this is still necessary as it relates to general behaviour of the theme service on initialisation, not any dependencies between the theme service and other services.

I've refactored the code after your comment from the last review, so I think this is good to merge.

EsmeeIDEMS (approved):
I'm happy for the changes as described to go in, however I'm not sure how to do a functional test.

What's an app "reload"? I tried refreshing the browser tab when running locally and quitting + re-running `yarn start` on the master branch, but in neither case I reproduce the issue you describe in 2.

chrismclarke (approved):
thanks for making the changes, code looks fine to me although I also haven't run a functional test on the skin changing

jfmcquade:
>

jfmcquade:
> I'm happy for the changes as described to go in, however I'm not sure how to do a functional test.
> 
> What's an app "reload"? I tried refreshing the browser tab when running locally and quitting + re-running `yarn start` on the master branch, but in neither case I reproduce the issue you describe in 2.

I think I could have been clearer about testing, here's how to replicate each of the 2 issues in the PR description. In each case, the issue should be present on the master branch, but fixed on this feature branch:

1. 
- With the deployment set to `plh_global`, run the app and change the theme to "professional" (can be done on the page http://localhost:4200/theme)
- Now change the deployment to one that does not support the "professional" theme, e.g. `early_family_math`.
- If you have master checked out, then on running the app you should notice that the colours are all wrong. Effectively no theme is set, and so some of the colours are ion's default colours.
  
- If you have this feature branch checked out, `fix/skin-and-theme-bugs`, then the theming should be the default EFM theming, as the theme "default" should be applied.

2.
The issue can be observed in this video:
- The skin is set to `weekly_workshop`
- The theme is changed to `professional`. This is evidenced by the background colour changing from light blue to white
- Upon reloading the page, the theme resets to `default`, evidenced by the background colour changing back to light blue.

https://user-images.githubusercontent.com/64838927/204014971-93e99c01-b11c-4937-a13e-5177cf670c39.mov

On this feature branch, the active theme persists after reloading

EsmeeIDEMS (approved):
Functional test passed.

PR #1659, opened 2022-11-10
Feat/crash sourcemaps

chrismclarke:
PR Checklist

## TODO
Currently I have a rough proof-of-concept working to take in a stack message and produce a mapped error trace, however it is quite limited in that it only works with a custom local build of the repo at the version of the app the time the error was made. 
This work is quite quickly expanding massively in scope, so before continuing it would be worth taking a step back to consider a few things. Namely:

- [ ] Is including sourcemaps in build an option? This would by far the easiest option populate the correct code path in the crash message removing any need to retrofit sourcemaps, but would also add ~10MB to build size (source maps are always going to be larger than the codebase they are mapping.

- [ ] Should we consider alternate logging systems that have better in-built sourcemap support (I know sentry allows submitting sourcemaps at build time which it will use to populate better messages). There might be some self-hosted options (maybe https://glitchtip.com/)

- [ ] If we want to proceed with our own custom retrofit we'll need better scripting to populate sourcemap files to github during build time, and ideally link all our builds to formal github releases (not just manual action triggers). The sourcemap will need to be linked both to the exact git commit sha as well as any other variables set during build process (e.g. deployment config)

## Description
We want to be able to retroactively apply sourcemaps to stack traces receiving in crashlytics, to pinpoint the actual parts of the code flagging errors. This PR aims to achieve by doing the following:

- Add script that can read in a error stack and retrofit sourcemaps
- Add webview version and navigator useragent to stack traces

**Author notes**
The easiest way to debug crashlytics errors would be to include the sourcemap files in the build itself, and use to log a human-readable error. The problem with this is that it would (currently) add about 10MB to the core build size

**General notes**
The process of retrofitting sourcemaps to errors turns out to be incredibly finicky. The first challenge is that sourcemaps are non-human-readable by design and so requiring quite a lot of low-level tooling to extract.

The second challenge is that sourcemaps are only relevant against an exact version of the codebase, i.e. specific git commit. So generating sourcemaps for old stack traces means trying to create a custom build of the codebase each time at the exact commit used when creating the release. Github automatically populates release assets when generating a formal release, however it appears most of our releases currently are generated on demand via github actions instead of creating release from the main page.

**Testing notes**
1. Identify an error with stack. E.g. [stack](https://console.firebase.google.com/u/0/project/plh-teens-app1/crashlytics/app/android:international.idems.plh_teens_tz/issues/7b676a835e209d921ad6eaded0dfdbed?time=last-seven-days&sessionEventKey=636CE4F203C300014D1B6FC93687124F_1742799109224659634)

2. Determine what commit was used at time of build (for link above its in plh_tz branch, version 0.16.8)
image

Assumed this commit to be e32d194731fcf0ef808c3405572afc583d2ef70c (that was the version bump commit)

3. Checkout a branch at the commit associated with the build that contained the error

4. Create a production build with sourcemaps from that branch
```sh
yarn ng build --configuration=production --source-map=true
```
5. Check if the sourcemap does in fact contain the `main.js` file sourcemap hashed at the version referenced in the error stack (in this case `main.d1af5c4fe9575d20.js`), i.e. does file exist `www\main.e618d4ecec9a7372.js.map`
If not likely the wrong commit checked out, however could also be related to other global configuration variables changing build

6. Assuming file exists, revert back the branch with sourcemap-extract scripts (sourcemap will stay in the www folder)

7. Run the script
```
yarn scripts stacktrace extract
```
This will prompt pasting error stacktrace into a text editor and will try to retrofit sourcemap

## Git Issues

Closes #

## Screenshots/Videos

Example of a retrofitted error, created custom and built with sourcemap
image

image


chrismclarke:
closing in favour of #1667

Issue #1640, opened 2022-11-07
[BUG] Notifications require refresh to display correctly


EsmeeIDEMS:
**Describe the bug**
Notifications require a refresh before they are scheduled correctly. The problem seems to be that when the `activation_condition_list`, `deactivation_condition_list` and `schedule.day_of_week` column entries depend on fields, it takes a re-render for these computations to be done correctly. I was wondering if maybe these initialisation actions were not done in the right order.

**App version**
v0.16.7 

**To Reproduce**
Reset app, select either workshop or modular skin, open notifications overview page, re-render to see a lot of them disappear and the schedule change of others. 

**Template Name**
[nf_schedule_in_week_messages
](https://docs.google.com/spreadsheets/d/1JXb0lzz8Hztl6K4_3zaUpYrpq9XJjTO7YRPAsx03F-4/edit#gid=513586624)
<img width="601" alt="image" src="https://user-images.githubusercontent.com/74557272/200337277-e0130097-8205-4d19-a834-58176e403403.png">

**Screenshots and videos**
For workshop skin (so `@fields.iwm_as_notifications == TRUE`): 

- Immediately after app reset
<img width="181" alt="image" src="https://user-images.githubusercontent.com/74557272/200337793-eb5b95f6-a925-438f-b4da-aa4901822acd.png">

- After refreshing the browser tab
<img width="177" alt="image" src="https://user-images.githubusercontent.com/74557272/200337985-ecbcdb60-362e-4641-9f69-00739742c754.png">

jfmcquade:
Adding my diagnosis of the issue that was previously on Slack:

In the example sheet attached to the issue, the notification schedule rows depend on fields, the values of which aren't set at the point of the campaign service being initialised (wherein the notifications are scheduled). The values are set by the template-process-service, which is initialised after the campaign-service in app.component.ts.

Specifically, the field workshop_number is set in the template timing_calculations, which is one of the "startup templates" (i.e. it has a value for process_on_start, and is therefore processed by the template-process-service on initialisation). When the campaign service is initialised and these notifications are scheduled, the field workshop_number is undefined and so the deactivation conditions that depend on its value being equal to 1 do not pass, and the notifications are not deactivated.

chrismclarke:
[...]

PR #1664, opened 2022-11-13
Refactor: Core service intialisation methods

chrismclarke:
PR Checklist

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade:
This is great to have. I believe everything is working as expected, although I haven't performed extensive functional tests. I've tried to do a thorough check of the points you raise in the PR description.

### Services not built on bases
The following services are not built on either the sync or async base service. There may be good reason for this, but I've included them all here for the sake of completeness.

- app_feedback
- app_notification_interaction
- app_user
- contact_field
- tables
- feedback
- template-asset
- context-menu
- audio.recording
- audio
- error-handler
- ipc
- push-notification
- rapidpro-notification
- performance-monitor

### Sync:
The following services extend the Sync base service. Unless stated, they do call their init methods in their constructors.

- theme
- template-nav (does not call init method in constructor)
- template
- template-action (does not call init method in constructor)
- template-process (does not call init method in constructor – contains comment that explicitly says that it shouldn't)
- template-row (does not call init method in constructor – contains comment that explicitly says that it shouldn't)
- settings (does not call init method in constructor)
- analytics
- app-config
- auth
- app-data
- event (does not call init method in constructor)
- lifecycle-actions (does not call init method in constructor)
- local-storage (does not call init method in constructor)
- server
- skin

### Async
The following services extend the Async base service. Unless stated, they do register their init methods in their constructors.

- campaign
- tour
- template-calc
- template-field
- template-translate
- template-variables
- global (doesn't register init method in constructor)
- app-events
- crashlytics
- data-evaluation
- db-sync
- db
- local-notification-interaction
- local-notification
- task-action
- task
- userMeta

### Checking child service ready status
For checking your 4th bullet point, I checked every service that imported another service to see whether the imported services are checked by the parent. For the following services, they imported one or more services that they did not check the "ready" status of. The unchecked services appear under the parent service:
- campaign
  - TemplateActionService
  - AppDataService

- tour
  - TemplateTranslateService

- template-action
  - TemplateProcessService

- settings
  - LocalStorageService

- auth
  - appConfig

- lifecycle-actions
  - TemplateActionService


### Further points
When I reset the app via the "Reset App" menu button, the initialisation reliably takes around 7 seconds. The vast majority of this time is taken by the campaign service. I've checked on the master branch, and whilst there is no precise way to measure the time taken, timing manually suggests that it also takes around 7 seconds, so I don't think there's a problem with this PR. 

When reloading the app by reloading the browser window, the initialisation reliably takes around 1.2 seconds on my machine.
Screenshot 2022-11-16 at 17 01 33


These changes do not appear to have resolved issue #1640. My diagnosis of that issue, now added to the issue page, is that it is dues to a circular dependency between the `template-process-service` and the `campaign-service` – the former is not currently imported by the latter. I've ran out of time today, but I can look further into this – it needn't stop this PR from being reviewed and merged.

chrismclarke:
Thanks for taking such a detailed look @jfmcquade 

I think laying everything out in the way you have is really useful, and also a bit of a stark warning about our longer term maintainability of the repo (so many services handled in so many different ways). But hopefully this should improve handling of the truly 'global' services (i.e. that we want to make sure are fully initialised before the user starts interacting with any templates)

The list has also helped me go through item-by-item and double check, I've added a few comments to your and updated/refactored/ignored/removed as required so now I feel a lot better about the overall list (super helpful, thanks again).

EsmeeIDEMS (approved):
Functional test passed. @chrismclarke Could you resolve the conflicts before merging?

PR #1673, opened 2022-11-30
Perf/campaign notifications

chrismclarke:
PR Checklist

## TODO
- [x] builds on top of #1664, should merge after

## Description
Currently initial startup takes a significant amount of time, with the main culprit being the campaigns service. 

There are 2 main reasons for this:

1) Campaigns schedule and force the reprocessing of notifications after each campaign is scheduled, so in the case of multiple campaigns we repeatedly evaluate all pending notifications. This comes with a delay because scheduled notifications can't be edited, instead we have to remove all notifications and reschedule all again, which also has to manage interaction with Java I/O on devices. This can happen 10-20 times on startup and is addressed in this PR.

2) The way in which most often all campaign rows need to be reprocessed after actions triggered because no easy way to track whether an action such as setting a variable impacts on other campaigns. This only involves comparing data entries and has some level of caching so shouldn't have too great an impact (hard to measure), but could be a source of inefficiency. This is not addressed in this PR, as will be linked to planned improvements of data_groups

**Main Changes**
- Defer all notification scheduling to happen off the main thread (campaigns don't need to wait for notifications to be scheduled before continuing to process)
- Debounce notification scheduling so that multiple calls to re-evaluate will be bundled into a single operation

Combined these should ensure initial campaign processing only triggers a single batch of notification scheduling at the very end of processing, which also is performed in a non-blocking way.

I've tried benchmarking the performance difference and seems really variable so not entirely convinced by the timing results, but should at least prevent some of the slowest init cases

**Additional Changes**
- Add ability to track appConfig change diffs. This allows the notification service to only reprocess notification defaults when the specific notification app config fields update (removes duplicate init methods triggered if initial config is changed by a skin that doesn't actually contain any notification-specific overrides).

## Dev Notes
Now that appConfig changes are observable I'd suggest we make the main appConfig$ private and refactor existing methods to watch changes only, as likely they will be triggering on any config changes even when properties used have not. This should probably be a follow-up task/issue (beyond scope of this PR)

## Review Notes
Would be good to check that the expected notifications are still scheduled on first init

## Git Issues

Closes #

## Screenshots/Videos

**Before** - init takes around 3200ms for first load and 800ms for subsequent on my local machine
image

**After** - init takes around 1600ms for first load and 800ms for subsequent on my local machine
image


jfmcquade (changes requested):
BUG
It appears that with each successive reload of the app, all the scheduled notifications are doubled up. This video shows starting with a freshly reset app, and how the scheduled notifications appear for subsequent reloads:

https://user-images.githubusercontent.com/64838927/206242811-b0ccb33c-d367-4581-ba90-9064e81369c5.mov

jfmcquade (approved):
The notifications seem to be working well now. So I'm happy for this to be merged.

In terms of improvements to startup times, I'm getting similar results to you: around 1900ms for first load and around 500ms for subsequent inits.

I like the functionality to track change diffs to the appConfig, I've made an issue (#1689) to roll out the changes to all services that consume appConfig.

I'm still getting the bug outlined in #1640, but I can roll out a hacky fix once this PR is merged.

chrismclarke:
Thanks for the review and creating the follow-up issues @jfmcquade , will merge now

Issue #1615, opened 2022-10-30
[FEATURE] Detect JS Incompatibilities


chrismclarke:
**What?**
Mentioned in missing qr code bug email thread. The JS and CSS code used in the app won't always be compatible on older browsers so it would be good to try and check/flag cases where this happens

**Why?**
Incompatibility issues are likely to be thrown as silent errors which could make the app behave unexepectedly

**How?**
Quite a few packages exist e.g [npm search](https://www.npmjs.com/search?ranking=popularity&q=keywords%3Afeature%20detection), or could try to manually implement known code for core features (e.g. supports local storage, es6 syntax etc.)

https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection
https://featurejs.com/

**Other details**
Would also be good to check android version and provide tips to fix compatibility (i.e. update google chrome app for android 7+ or android webkit webview for 5.1-6)

The checks should ideally be run before the main app is loaded, so likely via script in the `index.html` page

PR #1679, opened 2022-12-02
fix: add string matchAll polyfill

### LINKED ISSUE #1660: [Crashlytics] Uncaught (in promise): TypeError: et.matchAll is not a function TypeError: et.matchAll is not a function
**Describe the bug**
[Crashlytics link](https://console.firebase.google.com/u/0/project/plh-teens-app1/crashlytics/app/android:international.idems.plh_teens_tz/issues/7b676a835e209d921ad6eaded0dfdbed?time=last-seven-days&sessionEventKey=636CE4F203C300014D1B6FC93687124F_1742799109224659634)
```
Non-fatal Exception: com.getcapacitor.community.firebasecrashlytics.JavaScriptException: Uncaught (in promise): TypeError: et.matchAll is not a function
TypeError: et.matchAll is not a function
    at Me (http://localhost/main.d1af5c4fe9575d20.js:1:1420815)
    at we.parseTemplatedString (http://localhost/main.d1af5c4fe9575d20.js:1:1423195)
    at we.parse (http://localhost/main.d1af5c4fe9575d20.js:1:1422959)
    at Object.entries.forEach (http://localhost/main.d1af5c4fe9575d20.js:1:1422841)
    at Array.forEach (<anonymous>)
```

This suggests the `matchAll` method is not supported on the user's device. Matchall was to chrome 73 in March 2019
https://caniuse.com/mdn-javascript_builtins_string_matchall

Code used in delimters scripts (created by chris)
image

The issue can be resolved in any of:
1. Refactor code that uses matchall
2. Add checks and prompt in-app for user to update their browser
3. Manually ensure users have updated browsers (chrome on Android 7, Webkit webview earlier)
4. Change build settings to target more older browsers (our current build settings only cover browsers with more than 0.5% market share, which as of now is chrome 92+ (as defined in our browserslist file)
https://browsersl.ist/#q=%3E+0.5%25%2Clast+2+versions%2CFirefox+ESR%2Cnot+dead%2Cnot+IE+9-11

**Screenshots and videos**
image

chrismclarke:
PR Checklist

## Description
This fixes a bug for older devices that do not support the `matchAll` function (as noted in #1660)

Whilst this will fix the bug, for reasons noted in #1615 it isn't actually a very good solution, as there's enough other issues with older browsers that fixing this alone is unlikely to achieve very much. But it does at least give a good example for how we can add specific support in the future is similar issues arise

The final implementation is very minimal, so review more just for reference/interest

## Dev Notes
Support is added by including a link to the relevant polyfill in the polyfills.ts file. It uses the same polyfills that angular exposes in the `core-js` package (no install needed, part of angular deps). 

## Review Notes
This is particularly tricky to test as it involves running a legacy browser that does not support `matchAll` for strings, namely one from https://caniuse.com/mdn-javascript_builtins_string_matchall (e.g. chrome 72 or lower, internet explorer), add script to reference the matchAll function in the main app component (couldn't figure out what template would reference otherwise), and serving as a local production build.

These are the steps I took to test (repeat only if interested)
1. Download a legacy version of chrome. It has to be lower than chrome 72 which unfortunately discounts all version at https://chromium.cypress.io/ 

I actually cross-checked the user id that created the original crashlytics log with the webviewversion stored in the db (definitely overkill)
image

The version retrieved from the user for google-chrome has to be mapped to a corresponding version for chromium. This can be done with https://omahaproxy.appspot.com/

image

Finally that version number can be used to download the archive (or at least the version that closest matches). E.g. direct link for that chrome version: https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Win_x64/520847/ 

(Did I mention this was all a bit overkill???)

2.  You will also need to add a short script to the main app component to call the string matchAll function, e.g. 
image

3. Create a new production build of the app (serving will not be supported in the older browser), i.e. `yarn build`

4. Serve the production build folder, e.g. `npx serve www`

## Git Issues

Closes #1660

## Screenshots/Videos

**Before polyfill**
image

image

**After polyfill**
image

### DIFF
--- a/src/polyfills.ts
+++ b/src/polyfills.ts
@@ -18,6 +18,9 @@
  * BROWSER POLYFILLS
  */
 
+/** Add support for String.prototype.matchAll in older browsers */
+import "core-js/es/string/match-all";
+
 /** IE10 and IE11 requires the following for NgClass support on SVG elements */
 // import 'classlist.js';  // Run `npm install --save classlist.js`.
 


jfmcquade (approved):
Looks good.

Helpful for me to see how to add a polyfill. I haven't gone through the testing steps to reproduce the bug, but I'm happy to trust your testing! Thanks for showing your working, could be useful to refer back to in the future

EsmeeIDEMS (approved):
Read through the PR notes, looks good. I've not repeated the steps.

PR #1690, opened 2022-12-09
fix: hack: reschedule campaign notifications after processing startup templates

### LINKED ISSUE #1640: [BUG] Notifications require refresh to display correctly

jfmcquade:
PR Checklist

## Description

A workaround to fix issue #1640.

## Dev notes

This workaround handles the cyclic dependency between the `template-process-service` and the `campaign-service` by forcing the latter to reinitialise after the former has finished processing the startup templates. The performance load of reinitialising the campaign service should be significantly lower after #1673.

## Git Issues

Closes #1640

## Screenshots/Videos

New behaviour shown in video. See issue for previous behaviour.

https://user-images.githubusercontent.com/64838927/206680541-5a5261c0-1d8a-4f13-9ad9-a4ddace124cf.mov

### DIFF
--- a/src/app/shared/components/template/services/instance/template-process.service.ts
+++ b/src/app/shared/components/template/services/instance/template-process.service.ts
@@ -6,6 +6,7 @@ import { SyncServiceBase } from "src/app/shared/services/syncService.base";
 import { TemplateContainerComponent } from "../../template-container.component";
 import { TemplateNavService } from "../template-nav.service";
 import { TemplateService } from "../template.service";
+import { CampaignService } from "src/app/feature/campaign/campaign.service";
 
 /**
  * The template process service is a slightly hacky wrapper around the template container component so that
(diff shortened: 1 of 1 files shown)


chrismclarke (approved):
Thanks @jfmcquade - agreed would be nicer to have a way for campaigns to be aware themselves of the need to re-init but seems like a fair enough short term solution. Appreciate it being clearly marked as a `hack`

PR #1691, opened 2022-12-09
chore: add push to master branch trigger for sourcemaps-upload github action

jfmcquade:
PR Checklist

## Description

Pushing any branch was causing the `sourcemap-upload` action to fail, as [here](https://github.com/IDEMSInternational/parenting-app-ui/actions/runs/3656503088), with the error `No event triggers defined in 'on'`.

This PR adds a trigger to the github action. I've assumed that we want the action to trigger on pushes to master, but I'm happy to be corrected.

### DIFF
--- a/.github/workflows/sourcemaps-upload.yml
+++ b/.github/workflows/sourcemaps-upload.yml
@@ -6,10 +6,7 @@
 
 # NOTE - currently only used for debugging as methods are integrated into build scripts
 name: Sourcemaps Upload
-on:
-  push:
-    branches:
-      - master
+on: workflow_dispatch
 jobs:
   sourcemaps_upload:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/sourcemaps-upload.yml
```diff
@@ -6,10 +6,10 @@
...
-  # push:
-  #   branches:
-  #     - feat/glitchtip-error-logs
+on:
+  push:
+    branches:
+      - master
 jobs:
   sourcemaps_upload:
     runs-on: ubuntu-latest
```
jfmcquade, line 15: Just noticed the comment on line 21, should something else be changed?

chrismclarke:
Thanks @jfmcquade - yes the problem was the floating `on` which should have been commented out. I pushed hotfix 0230a76fb1b603f45521a6f642906b6481e1ca5c so this PR is no longer required, thanks for checking though

PR #1693, opened 2022-12-13
fix: service timeout

chrismclarke:
PR Checklist

## Description

While testing on a particularly slow emulator device I noticed some of the services where timing out on init which then broke things entirely. I'm now second-guessing whether we actually want services to be allowed to fail silently, but for now I've just upped the timeout delay from 10s to 60s 

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade (approved):
This seems good for now. 

As I'm late to reviewing this PR, it will have to be resolved with #1697

PR #1694, opened 2022-12-16
core: angular 15 updates

chrismclarke:
PR Checklist

## Description

- Core 3rd party deps update, angular 13->15 plus minor fixes (see comments inline)
- Github action updates for runtime version (Node 16) and caching

I don't expect any of the changes to make much difference beyond keeping up-to-date with security and features, a couple minor improvements of note:

- Error logs should be a bit more concise (stack traces no longer include angular internal calls, e.g. zone.js)
- Potential support for new tooling to make builds slightly fatser (tested and not fully compatible yet, so maybe in the future)

https://blog.angular.io/angular-v14-is-now-available-391a6db736af
https://blog.angular.io/angular-v15-is-now-available-df7be7f2f4c8

## Review Notes
I've fixed any minor breaking changes I could identify from the angular update notes and did a manual check of a few workshop pages. All looks as expected to me but will add visual preview test in case. 

@jfmcquade Angular was unhappy with the slider.scss import so I've changed to slider.min.css - looking at the carousel component template all looks as expected to me but would be good if you could also take a quick check.

@EsmeeIDEMS Looking through component visual diffs all the main changes seem to be due to a change in the combo-box, it appears to fill backgrounds whiteish instead of transparent. Not quite sure why, looks reasonable to me but could be reverted if desired
image

## Git Issues

Closes #

## Screenshots/Videos


angular.json
```diff
@@ -50,7 +50,20 @@
               }
             ],
             "scripts": ["./node_modules/intro.js/intro.js"],
-            "allowedCommonJsDependencies": ["lottie-web", "firebase/app", "@firebase/auth"]
+            "allowedCommonJsDependencies": [
```
chrismclarke, line 53: Changes here are just to supress warnings in the console about 3rd party libraries that use different build systems to angular

src/app/feature/campaign/pages/campaign-debug/components/campaign-variables-editor.ts
```diff
@@ -34,7 +34,7 @@ import { TemplateFieldService } from "src/app/shared/components/template/service
         </ion-label>
         <ion-datetime
           #firstLaunch
-          (ionChange)="setAppLaunchData(firstLaunch.value, appDay.value)"
+          (ionChange)="setAppLaunchData($any(firstLaunch.value), appDay.value)"
```
chrismclarke, line 37: minor type issue because datetime can output string or string[] depending on configuration. So just cast to `any` type (as we know the output is string format)

src/app/shared/components/template/template-component.ts
```diff
@@ -154,8 +150,7 @@ export class TemplateComponent implements OnInit, AfterContentInit, ITemplateRow
     row: FlowTypes.TemplateRow
   ) {
     const viewContainerRef = this.tmplComponentHost.viewContainerRef;
-    const factory = this.componentFactoryResolver.resolveComponentFactory(component);
-    const componentRef = viewContainerRef.createComponent(factory);
+    const componentRef = viewContainerRef.createComponent(component);
```
chrismclarke, line 153: minor refactor required for how angular handles dynamic components. Tested working (by loading any template page which will in turn load dynamic components)

src/global.scss
```diff
@@ -37,7 +37,7 @@
 
 /* Styles for 3rd prty libraries */
 @import "~nouislider/distribute/nouislider.min.css";
-@import "~swiper/scss";
+@import "~swiper/swiper.min.css";
```
chrismclarke, line 40: Minor import update as mentioned @jfmcquade  - I can't see any difference on the carousel component which uses swiper

jfmcquade: Yep, seems to work correctly

jfmcquade (approved):
Thanks, looks good. 

Everything seems to be working for me running locally. Running my first `yarn start` from this branch didn't make the app available on localhost. I ran `yarn build` first and then `yarn start` again, at which point it worked.

Regarding the combo-box visual changes, I've pushed a fix to preserve the previous style. The issue was that a CSS custom property was being referred to without its `--` prefix, specifically:
```
var(combo-box-background-no-value, transparent);
```
Previously, the behaviour must have been to use the fallback value in the case that the first value was invalid, but it would seem that the new behaviour is to treat the whole CSS rule as invalid in this case and not apply it:
Screenshot 2022-12-19 at 10 23 12

PR #1697, opened 2022-12-20
chore: update rxjs 6 to 7

chrismclarke:
PR Checklist

## Description
Upcoming work on dynamic data includes adding [rxdb](https://rxdb.info/) which is dependent on rxjs 7. Upgrade v6 -> v7 in preparation

- Update rxjs 6 to 7
- Handle breaking changes ([changelog](https://rxjs.dev/6-to-7-change-summary), only affected found was timeoutWith)
- Code tidying (standardize componentDestroyed$ name, remove unused imports)

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade (approved):
Nice.

I noticed that we used `toPromise()` in a few other places, which is deprecated, although not removed, in rxjs 7. I'm still getting my head around reactive programming and rxjs, so I thought I'd try and refactor these instances to use the new `firstValueFrom()` and `lastValueFrom()` functions. I've added a commit to this PR

src/app/shared/components/template/services/instance/template-action.service.ts
```diff
@@ -147,7 +147,9 @@ export class TemplateActionService extends SyncServiceBase {
       log_groupEnd();
     }
     // resolve once full queue processed
-    await this.actionsQueueProcessing$.pipe(takeWhile((v) => v === true)).toPromise();
+    await lastValueFrom(this.actionsQueueProcessing$.pipe(takeWhile((v) => v === true)), {
+      defaultValue: "No action queue",
```
chrismclarke, line 151: nit(non-blocking)
Is there a reason for passing a default value?

chrismclarke:
Thanks @jfmcquade 
The changes you added seem reasonable to me, the only one I would be unsure about is the `actionsQueueProcessing$` subscription as I'm not quite sure what the role of the default value is. Did you have a chance to functionally test this part (i.e. actions still triggering and processing as expected)?

chrismclarke:
I've run a few quick checks manually and I think should be fine to merge so going to go ahead

src/app/shared/components/template/services/instance/template-action.service.ts
jfmcquade: From my testing, if no value was provided for `defaultValue` then triggering an action would cause an `EmptyError` error to be thrown.

As outlined [here](https://rxjs.dev/deprecations/to-promise#lastvaluefrom), `lastValueFrom` differs slightly from `toPromise` in that when the observable completes without emitting a single value, the latter resolves successfully with value `undefined`, whereas the former rejects with `EmptyError`.

Without a default value, actions still appear to be triggered correctly, but the following error is thrown and logged to the console:
Screenshot 2023-01-04 at 16 13 08

PR #1698, opened 2022-12-21
Feat/dynamic data items

chrismclarke:
PR Checklist

## Description

As outlined in [RFC](https://docs.google.com/document/d/1saH0_PzkbLTwO0fea4lPtuLnyYga30W_uhiO9CEcEgg/edit#heading=h.nd960lmg5uuf), add support for a new way to interact with data lists, so that users can store edits directly on top of lists.

This creates a new `data_items` component, which observes changes to a given data list and can update a user's entry on the list via the `set_item` action.

Full changes as

**Core**
- [x] Review options for reactive databases
- [x] Implement in-memory write-layer db to store user edits to flow data
- [x] Add mechanism to persist write-layer to disk and retrieve on start of new session
- [x] Implement additional in-memory merged db to retain full state of sheets + writes and allow data change subscription
- [x] Add mechanisms to populate merged db on load
- [x] Refactor memory/persistence layers to be more interoperable

**Frontend**
- [x] Add frontend bindings to display live data in data list
- [x] Add authoring method to display live data lists
- [x] Add authoring method to trigger updates to live data list item
- [x] Add authoring methods to trigger bulk list changes (partial support)

**QA**
- [x] Add testing infrastructure and unit tests
- [x] Add example component demo [sheet](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit?usp=sharing)

## Future Tasks
Ideally after merging any feedback/follow-ups should be added to a new issue for tracking

- [ ] Add methods to perform operations on groups of rows or collections
- [ ] Add methods to store data_query result in local variables (e.g. the length of items in a filtered data_list)
- [ ] Add authoring options to trigger data updates outside of list item (not easily possible)
- [ ] Add support for merged/computed lists
- [ ] Add documentation (#1725)
- [ ] Add QA checks for data to ensure ids exist for all rows in data lists
- [ ] Add resolution for cases where rows are removed from original data list but user still has persisted changes
- [ ] Add QA check to ignore merging rows which have since been deleted from original data list

## Review Notes
See examplesheet  [component_data_items](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit?usp=sharing)

## Dev Notes
Unit tests can be run independently of full (broken) set of tests via
```
yarn ng test --include="src/app/shared/services/dynamic-data"
```

## Git Issues

Closes #

## Screenshots/Videos
**Demo template** - changes are made on top of the corresponding data list and persist across device reload

[Parenting App (1).webm](https://user-images.githubusercontent.com/10515065/212188264-e580877c-256f-4a76-bd10-48c1ca75da2a.webm)

component demo [sheet](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit?usp=sharing)

image

image


EsmeeIDEMS (changes requested):
`begin_data_items` does not seem to work when the data list to loop over is set through a local variable rather than directly in the value of this row. For example, rows 15-20 (example 2b) of [comp_data_items](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=845703892)

image

results in 
image

src/app/shared/services/dynamic-data/dynamic-data.service.ts
```diff
@@ -0,0 +1,196 @@
...
+   * This allows users to query and subscribe to data changes in an efficient way
+   */
+  private db: ReactiveMemoryAdapater;
+
+  /**
+   * A separate cache stores user edits flow data, initially in memory
+   * for faster writes but mirrored to a persisted IndexedDB regularly
+   *
+   * This means that user changes will persist app reloads and also be retained
+   * if the underlying data changes so long as compatible (e.g. data list ids still exist)
```
jfmcquade, line 37: This is a good feature, but could it cause issues when app updates introduce data_lists that are compatible with existing user overwrites, but aren't intended to be? I'm not sure I can think of an example, but flagging here more out of interest

src/app/shared/services/dynamic-data/dynamic-data.service.ts
```diff
@@ -0,0 +1,196 @@
...
+ * The dynamic data service handles the process of loading data from sheets, storing
+ * in memory as a reactive data object, and enabling a persisted write-layer on top
+ * for storing user changes
+ */
+export class DynamicDataService extends AsyncServiceBase {
+  /**
+   * The main DB stores in-memory the rows of any requested sheets, merged with user writes
+   *
+   * Each flow is represented in its own collection, and populated as requested.
+   * This allows users to query and subscribe to data changes in an efficient way
```
jfmcquade, line 28: I'm just learning about NoSQL (I'm currently working on an app using Firestore). I've been surprised by the fact that you can't query across multiple collections. Storing each flow in its own collection seems the neatest approach here, and I know it is always possible to combine data from multiple reads if we ever wanted to query multiple data_lists

jfmcquade (approved):
Very impressive! I didn't fully understand the plan in the RFC at first, but I think I can make sense of it now having seen the code – very cool.

I've added a few comments, more for my own benefit than anything else

EsmeeIDEMS (approved):
Thanks @chrismclarke - functional test passed

src/app/shared/components/template/components/data-items/data-items.component.ts
```diff
@@ -0,0 +1,172 @@
...
+    await processor.processContainerTemplateRows();
+    return processor.renderedRows;
+  }
+
+  /**
+   * If datalist referenced as @data.some_list it will already be parsed, so extract
+   * name from raw values.
+   * Alternatively any list provided as a string value can be returned directly
+   * */
+  private hackGetRawDataList(row: FlowTypes.TemplateRow) {
```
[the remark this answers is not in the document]
chrismclarke: updated, thanks

src/app/shared/components/template/services/template-variables.service.ts
```diff
@@ -365,11 +372,11 @@ export class TemplateVariablesService extends AsyncServiceBase {
         parsedValue = evaluateJSExpression(expression, thisCtxt, globalFunctions, globalConstants);
         break;
       case "item":
-        log("evaluate item", evaluator, context);
-        try {
+        // only attempt to evaluate items if context passed, otherwise leave as original unparsed string
```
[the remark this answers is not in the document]
chrismclarke: Yes that is correct, we have both the item data list and template begin/end data_items rows, and create a loop for each data list row (termed `itemContext`), and evaluate the template rows replacing references to `@item` with that context variable.

The catch here is because data_items are processed after being rendered, we need to make sure the parser ignores references to `@item` and leaves for the data-items component to handle with the appropriate context.

This is different than the original begin/end items, which has a hardcoded intervention in the parser to process directly instead of delegating to the child component (retrieving the item context itself). The benefit to this approach is that we don't need to render an additional container component for the data_items, the downside is we would have to re-render/reprocess at the template level any time the list changes (to force full reprocess), instead of just letting the items render once and manage their own change detection (which is what data_items do)

src/app/shared/services/dynamic-data/dynamic-data.service.ts
chrismclarke: Firestore does expose a method for [collectionGroup queries](https://firebase.google.com/docs/firestore/query-data/queries#collection-group-query), however usually you need to define a custom index in advance and support is somewhat limited (mostly cases where multiple collections have subcollections or documents with the same name)

The nice thing about using an in-memory database is that if we later decide we want to change how it is used we don't have schema-breaking changes to worry about (because nothing is persisted). It's also for that reason that the caching layer actually stores all documents in a single table, indexed by flow_type and flow_name, and a single `data` column that stores the full json of any changes to persist onto that flow

src/app/shared/services/dynamic-data/dynamic-data.service.ts
chrismclarke: Agreed, and right now I don't have a specific resolution code or tests to cover those cases. 

Currently, we just have a fairly basic deep merge, where the persisted saved data is merged on top of the sheets. It would make sense to customise the functionality to ensure that the data_list row exists before trying to merge on top and not just leaving dangling rows.

I've added a placeholder test and will update the todo above to cover it.

PR #1700, opened 2023-01-03
fix: unit test infrastructure

chrismclarke:
PR Checklist

## Description

Unit tests have been broken for a long time, and as such we haven't been making any use of within the app. This is quite problematic particularly when trying to work on larger frontend features in a systematic way

This PR fixes the core dependencies (in line with ng 15 updates), so that we can at least write/run/test spec files for unit testing frontend code. It also fixes compiler-breaking issues (still have many failed tests for other reasons)

## Review Notes
Just to check the frontend tests load and execute. You should see a mixture of failed and passed tests
```
yarn install
```
```
yarn test:unit
```

## Git Issues

Closes #

## Screenshots/Videos

**Test suite running - with various failed tests**
image


jfmcquade (commented):
Looks good. 

I can confirm that running `yarn test:unit` seems to work as expected, opening the test suite UI in Chrome, with the same results as in your screenshot

chrismclarke:
Thanks, yeah I'd definitely be keen to start building out a bit more our frontend test suites as I definitely feel the codebase is starting to get a bit harder to maintain and that there could be a bunch of longer-term time savers doing things this way 

Angular has some reasonable documentation on unit testing below, in case you were interested in helping to try fix some of the old/broken tests and want to begin adding some more
https://angular.io/guide/testing

PR #1701, opened 2023-01-04
feat: add support for template action registry

chrismclarke:
PR Checklist

## Description

Currently, any actions triggered from the template-action service need to first be registered from within the action service. This is somewhat an anti-pattern, as it means the action service needs to import every other service it will interact with (e.g. template, field, tour, analytics, theme etc.), and quickly leads to confusing dependency issues. To make matters more confusing, the action service is instantiated per template container, resulting in a large number of references to the same global services.

This PR tries to reverse this pattern by creating a separate template action registry which can be used to register action handlers from within the child service they originate from. This provides an alternative way to register actions without import into the action service. 

The main motivation for this is to allow easier decoupling/registration of specific dynamic data action handlers planned for use in #1698

## Dev Notes
You'll notice I've only started by refactoring the `start_tour` action out of the template-action service, but in the future could probably move around half of the action handlers to their own services (anything that calls global services can be refactored, but more work would be required for things like `set_local` as that makes use of the specific context of the instantiate template-action service, and so each template could try to intercept the wrong actions)

## Review Notes
As the only refactored action is `start_tour` simply checking that tours do start should suffice

Unit tests have been added and can be run via
```
yarn ng test --include="src\app\shared\components\template\services\instance" 
```

## Git Issues

Closes #

## Screenshots/Videos
**Unit Tests**
image

### DIFF
--- a/src/app/feature/tour/tour.service.ts
+++ b/src/app/feature/tour/tour.service.ts
@@ -7,6 +7,7 @@ import { FlowTypes } from "data-models";
 import { TemplateFieldService } from "src/app/shared/components/template/services/template-field.service";
 import { AppDataService } from "src/app/shared/services/data/app-data.service";
 import { AsyncServiceBase } from "src/app/shared/services/asyncService.base";
+import { TemplateActionRegistry } from "src/app/shared/components/template/services/instance/template-action.registry";
 
 @Injectable({
   providedIn: "root",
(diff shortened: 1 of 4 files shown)


jfmcquade (approved):
This looks good to me, and I think I understand the motivation. It's neat for services to register their own template actions.

The `start_tour` action is working as expected.

One possible downside, from a dev perspective, is that currently the template-action service file contains a complete list of all available template actions. Refactoring some of the actions to not appear in this list means that no such list would exist. There is the list of `action_id`s in the `TemplateRowAction` interface in `data-models/flowTypes.ts`, but this doesn't include the actions themselves, i.e. logic that each action ID triggers. Although the `action_id` should be descriptive enough for this to be inferred. 
Do you think it would be worth me adding a page to the docs (in the Authors section) that lists all of the available actions, with their ID and their effects?

chrismclarke:
> This looks good to me, and I think I understand the motivation. It's neat for services to register their own template actions.

Thanks @jfmcquade, yeah this is the exact reason, I had thought otherwise of some sort of emit or pub/sub callback system but just seems easier for services to be in charge of their own code. 

> One possible downside, from a dev perspective, is that currently the template-action service file contains a complete list of all available template actions. Refactoring some of the actions to not appear in this list means that no such list would exist, as far as I'm aware. 

I had thought about this, and felt the same that keeping the centralised type definitions should suffice for now. 

> Do you think it would be worth me adding a page to the docs (in the Authors section) that lists all of the available actions, with their ID and their effects?

Would be really nice to have for sure, although I'd be tempted to try and couple documentation with the code somewhat and use the docs generator (if possible). I.e. creating a typescript object that is strongly typed to include all actions alongside description, e.g.
```
start_tour: {
   service: TourService,
   args: [tour_name]
   description: 'start a tour'
}
```
Although not sure how easy it would be to output something like that nice in a table using doc gens

PR #1702, opened 2023-01-04
chore: configure angular to not share usage data

jfmcquade:
PR Checklist

## Description

Since we updated Angular to version 15 in PR #1694, when running `yarn start` I would receive a prompt asking whether to share anonymous usage data with the Angular Team.
Screenshot 2023-01-04 at 10 05 53 copy

This prompt would disappear without allowing time for a response, but not providing a response would prevent the `yarn start` processes from completing successfully.

This PR commits the changes enacted by responding `N` to the prompt, so it should not be triggered in future.

(Similar to this issue: https://stackoverflow.com/questions/56355499/stop-angular-cli-asking-for-collecting-analytics-when-i-use-ng-build)

## Testing

On a branch without this change (e.g. master), run `yarn start`. On my machine, the process does not complete: the terminal hangs on this message and the app is not be available on localhost
Screenshot 2023-01-04 at 09 43 16

Now checkout this feature branch and run `yarn start`. The app should be available on localhost as expected.

## Git Issues

Closes #

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -193,7 +193,8 @@
   },
   "cli": {
     "defaultCollection": "@ionic/angular-toolkit",
-    "packageManager": "yarn"
+    "packageManager": "yarn",
+    "analytics": false
   },
   "schematics": {
     "@ionic/angular-toolkit:component": {


EsmeeIDEMS (approved):
Functional test passed. I'm experiencing the described issue on master but the app is working normally on this branch. 

I did notice two unfamiliar warnings in the terminal when running `yarn start` on this branch: 
image

chrismclarke (approved):
Thanks both, yes I have the same update in my local branch so nice to merge in centrally

As for the new warnings, will add to my v. low priority backlog to look at, shouldn't be a real issue though as we don't actively try to support non-chrome browsers at this point.

PR #1704, opened 2023-01-05
feat: add local deployment and sync mode

chrismclarke:
PR Checklist

## Description

Adds experimental support for running the app with data from local folder xlsx files instead of gdrive sync. This can reduce time to reflect changes in the app down from around 40-60s to 15-20s (running on my machine, comparing with baselines in #1400)

The main motivation for this is to improve the current development workflow when working on new features that require authoring, however in the future we could further refine to function as a fully independent authoring environment

Currently only changes to sheets are supported (not assets or translations), and is designed mostly to work with a specific `local` deployment. 

## Review Notes
There shouldn't be much scope for knock-ons, a couple of the folder structures have changed (notably the app reads from the local app assets folded instead of the app-data package folder), however additional script has been added to replicate content from one to the other so should be fine. Would be good to double-check plh content still syncs as expected after next merge (testing locally all looks fine to me)

Otherwise the main thing to test is the local development workflow:

Set the local deployment
```
yarn scripts deployment set local
```

Start the angular server alongside file watcher (utility script created to handle both tasks)
```
yarn start:local
```

Make changes to any sheets found in the `.idems_app\deployments\local\sheets` folder. You should be able to create or modify any templates in the folder (but not assets or translations). Changes made should just be saved locally (except the one demo.xlsx file)

## TODO (future dev)

**Authoring**
- [x] Add documentation
- [ ] Improve the demo.xlsx sheet to be more of a quick-start guide or easier template to work from

**Dev**
- [ ] Add support for assets and translations
- [ ] See if other options will speed up ts-node compiler (see links below, can also consider multithreading via pm2 or similar)
- [ ] Add means to sync back to google drive deployment

Links
- `https://github.com/TypeStrong/ts-node/issues/754`
- https://typestrong.org/ts-node/docs/swc
- https://www.typescriptlang.org/docs/handbook/esm-node.html
- https://github.com/alangpierce/sucrase
- https://github.com/folke/esbuild-runner

## Git Issues

Closes #

## Screenshots/Videos

**Develop using local spreadsheets**
Direct edits to xlsx sheets considerably faster to reflect update than when syncing from gdrive. Changes reflect in 15-20s on local machine

[Untitled_ Jan 9 2023 4_55 PM.webm](https://user-images.githubusercontent.com/10515065/211437948-96dc4d3d-8c21-4474-b579-9d1af17710a1.webm)

127 0 0 1_8000_authors_local-sheets_


jfmcquade (approved):
The local development workflow works for me, with reload times of around 10-15 seconds, so definitely a time saver when appropriate.

I can see the immediate use of this when working on new features, and how it could be useful down the line to build the app quickly from local files. As you mentioned on Slack, one use case could be demos and workshops.

chrismclarke:
I've added documentation and captured the notes from the meeting so I think this is now good to know.

@EsmeeIDEMS - it would be great to see if you can also run locally following the information in the documentation.

As for the TODO tasks, I'm not sure whether we want to create any specific low-priority issues now (or even just a single local-dev improvements issue) or just wait for things to arise organically in the future, I'll leave it to others to decide.

PR #1715, opened 2023-01-11
fix: removed margins around hidden nested elements

### LINKED ISSUE #1709: [BUG] Hidden components have nonzero height
**Describe the bug**
Hidden items seem to take some space on the page. Therefore, deeply nested items (through a sequence of [pair](https://docs.google.com/spreadsheets/d/1ZkAXSGDgCQrDMUbyVVgJrdMUTnuZOicvecP6qytA9sY/edit#gid=1931878555)s) in a workshop activity appear lower and lower on the page each time the nav buttons are clicked.

**App version**
v0.16.9

**Template Name**
For example, see [w_money_read_2_temp](https://docs.google.com/spreadsheets/d/18oKsHQ0DZCjpp3yMvesOfkFH9_qERX3HgxnzSV3fIJY/edit#gid=609273714)

**Screenshots and videos**
image

[screen-capture (17).webm](https://user-images.githubusercontent.com/74557272/211362835-b67ed810-7dea-45f0-bc17-1e901a50811a.webm)

**Additional comments**
- @chrismclarke said we should probably deal with this specific case in a different way anyway, rather than nesting pairs the items in the story should be coming from a list. 
- Potentially related: Issue of blank text having nonzero height https://github.com/jfmcquade/ae-soils-app/issues/19

### LINKED ISSUE #1710: [BUG] Display group introduces extra LHS padding
**Describe the bug**
Some display group styles seem to introduce extra padding on the LHS of the screen. This causes the components inside it to be off-centre. 

**App version**
v0.16.9

**Template Names**
- [x] 1. [feature_display_group](https://docs.google.com/spreadsheets/d/1ARbNRGDer5vj9qSpRMZFrMkYifGkH3TLtDVp72YbaqU/edit#gid=1231713195)
image

- [ ] 2. For example, see [w_money_read_2_temp](https://docs.google.com/spreadsheets/d/18oKsHQ0DZCjpp3yMvesOfkFH9_qERX3HgxnzSV3fIJY/edit#gid=609273714) - In this case I'm not sure if the padding is introduced by the display group or by the animated section...
image

jfmcquade:
PR Checklist

## Description

Fixes issue where nested animated sections would have excess margin on the left hand side (see section 2. of issue #1710) and where nested animated sections that are hidden would have nonzero height (issue #1709)

I have reviewed the visual tests – besides the usual false-positives, there are a few cases of genuine difference. I believe these reflect the expected behaviour that is now being applied by the corrected style rules. See screenshots below. There may be other examples in the app that are not captured in the visual tests.

## More detail/dev notes

Template rows that are nested inside another have a top margin applied, to ensure that there is spacing between elements. The first child doesn't get a margin added, so if there is only one child then no margin should be added.
Likewise, display groups with the default `row` style parameter apply a margin on the left hand side of each of their child elements, excluding their first child, so if a display group has only one child then it should display centred. 

Our CSS should apply a left margin to every _non-hidden_ child element but the first. But the code previously did not correctly exclude hidden elements.

In the case of the `workshop_activity` slideshows referenced in the issues, a `display_group` (with name `pages_activity`) is being used with multiple children, where all but one child is hidden at a time. The hidden elements erroneously had margins applied to the top and left hand sides, causing the recorded visual bugs.

## Git Issues

Closes #1709 
Closes #1710 (second part)

## Screenshots/Videos

The template [w_money_read_2_temp](https://docs.google.com/spreadsheets/d/18oKsHQ0DZCjpp3yMvesOfkFH9_qERX3HgxnzSV3fIJY/edit#gid=609273714) displaying correctly, in contrast to behaviour logged in issues.

https://user-images.githubusercontent.com/64838927/212067112-dc46e72f-9da4-4e22-b297-7867478ebc94.mov

### Points of difference shown in visual test screenshots:
[debug_evaluate_hidden](https://docs.google.com/spreadsheets/d/1A159jPS8rSDnA2GDUTda6xeC_Kyu3oYk_-KbuhzMqmQ/edit#gid=2021973417)
Screenshot 2023-01-12 at 12 55 57

[debug_negation](https://docs.google.com/spreadsheets/d/1A159jPS8rSDnA2GDUTda6xeC_Kyu3oYk_-KbuhzMqmQ/edit#gid=1844496920)
Screenshot 2023-01-12 at 12 55 18

[w_example_tools](https://docs.google.com/spreadsheets/d/1TTyjMWDzA2BbiYTcqZdSGEmPKI1P7kMPkd0Ylw_Hlew/edit#gid=2101655363)
Screenshot 2023-01-12 at 12 53 31


jfmcquade:
> My only nit/minor comment would be if the CSS would be a bit more readable if we used the [first-of-type](https://developer.mozilla.org/en-US/docs/Web/CSS/:first-of-type) selector to provide a general rule for all children and then override for first? But not super important as the comments are pretty clear to describe what is going on.

That's how the CSS was written before. Agreed that it appears to be a more readable solution, but in fact it doesn't work because the `:first-of-type` selector doesn't consider attributes.
So whilst we want to select the first _non-hidden_ element to override its styles, the following selector just selects the first child of type `plh-template-component`:
```
plh-template-container > plh-template-component:not([data-hidden="true"]):first-of-type
```
Hence we have to use the method suggested [here](https://stackoverflow.com/questions/29422894/css-select-first-element-found-with-attribute/29422984#29422984).

I may have misunderstood and you're suggesting another way to accomplish it...

chrismclarke:
> That's how the CSS was written before. Agreed that it appears to be a more readable solution, but in fact it doesn't work because the `:first-of-type` selector doesn't consider attributes. So whilst we want to select the first _non-hidden_ element to override its styles, the following selector just selects the first child of type `plh-template-component`, which may be hidden:

Oh cool, good to know (makes a lot of sense)

EsmeeIDEMS (approved):
Functional test passed, I think this is a massive improvement to the look of some pages. Thanks!

PR #1718, opened 2023-01-12
style: swap colours of parent points' clicked/unclicked state, and add border, in professional theme

### LINKED ISSUE #1708: ParentPoint box design improvements in professional theme
**Details**
(Suggested by the graphic designer): To improve the look of the ParentPoint box component in the Modular Skin, we should 
- flip clicked/unclicked colours (so that "unclicked" is similar to the background and "clicked" jumps out)
- white version should have a grey border all around, just like the workshop cards. Currently it merges into the background since it doesn't have a top border

**Screenshots and videos**
image

workshop card example with border all around
image

jfmcquade:
PR Checklist

## Description

In the Professional skin (applied by the Modular skin), the colours of the parent point boxes when clicked and unclicked has been swapped in order to highlight the "clicked" state, in line with the designer's comments.
Also, a subtle border has been added to the white parent point box in order to separate it from the white background of this skin, drawing it more in line with the look of the task card component (see issue #1708).

Technically, this PR adds a new value available to all skins, `--ion-color-shade` (used here for the borders of the parent point box and task card), which is an almost fully transparent black, which may have other applications in the future.

## Git Issues

Closes #1708  

## Screenshots/Videos

Side-by-side comparison, before and after
Screenshot 2023-01-12 at 16 52 33


src/theme/themes/utils/generate-theme.scss
```diff
@@ -23,6 +23,7 @@
     "color-red": #eb445a,
     "color-green": #92cb44,
     "color-gray-light": $gray,
+    "color-shade": rgba(black, 0.07),
```
chrismclarke, line 26: nit(non-blocking)
I don't think this colour is defined in such a way that it is very clear when/where it's used. It's not really a named color in the same way that others are, and also when we talk about color shades we usually are referring to the shade of a colour (e.g. light blue), instead of what you are trying to describe as 'the color you should use when you want to provide subtle contrast against a white background). 

I'd probably suggest either you change it to `color-shadow` and use the same color in the `default-box-shadow` below, or leave as a hardcoded color in the templates that are using them. 

Alternatively we could provide a monochrome pallete and have the lightest value as something like `color-black-50` which may or may not line up to something similar.

chrismclarke (approved):
One minor comment inline, but in either case I'd say good to go

jfmcquade:
Your comment definitely makes sense, I've opted to leave the colour as a hard-coded value

PR #1730, opened 2023-01-18
refactor: prepare script

chrismclarke:
PR Checklist

## Description

We have a script that automatically tries to track changes to the `yarn.lock` file and automatically run a yarn install to update dependencies so that developers/authors don't need to constantly remember. For the most part this works fine, however I've noticed a couple times this process can get interrupted and leave the yarn install in a bad state (mostly happens when I manually intervene not realising it's running in the background when changing branches in vscode)

So this is a minor refactor so that the auto-install check happens when running the app via `yarn start`, instead of an automatic hook triggered when checking out a branch. This should prevent excessive installs when jumping between branches, and also show in the console output more clearly (vscode runs as a background process if changing branches form their UI)

## Review Notes
This branch includes yarn.lock changes, so hopefully if syncing it will not auto-install when you first checkout, but will auto-install when you first run a `yarn start` command (and not on subsequent start commands). Good if you can confirm it works locally (my only hesitation is whether the hooks will process correctly on first checkout, given the checkout code contains the changed hooks)

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade (approved):
This seems like a good idea. I haven't been aware of times when I've checked out a branch while the script is running, but I have sometimes thought that yarn install has been run unnecessarily when I've been switching between multiple branches to look at the code.

It seems to work as expected from my testing – I can confirm that the install scripts are not run on the first checkout of this feature branch, so the hooks seem to be being processed correctly and immediately.

chrismclarke:
Thanks @jfmcquade, yeah that was the same issue I was having so hopefully should be smoother moving forwards.

PR #1737, opened 2023-01-20
chore: create 'perinatal' deployment

jfmcquade:
PR Checklist

## Description

Creates a new deployment, "perinatal", including:
- Deployment config
  - Including DSN for error logging (linking to [this glitchtip project](https://app.glitchtip.com/idems/issues?project=2443)) and gdrive folder IDs
- Firebase hosting setup for web preview (URL: https://perinatal.web.app/)

A branch exists for working on the content of the Perinatal App, `deployment_dev/perinatal`. It's my understanding that the deployment config should be added to the PLH app master branch also, as is the case for the Early Family Math app etc.


EsmeeIDEMS (changes requested):
I don't seem to be able to select `perinatal` from the list of deployments when on this branch (whereas from your description I think I should be?)

image

I think this might be the same issue as on commit https://github.com/IDEMSInternational/parenting-app-ui/commit/a779ca1b6bf645f127fbe73a0456f909a7a622c7 on the `development_dev/wash` branch

chrismclarke:
Thanks @jfmcquade , I have the same issue as @EsmeeIDEMS, I can see the reason is because you have assigned the `error_logging.dsn` as an override however the default config doesn't provide a default `error_logging` object so you would need to provide as `error_logging: {dsn: '...'}` instead of direct.

I've updated and will test locally

jfmcquade:
That was my oversight, thanks for fixing Chris. This should be ready to merge now.

I've also updated the `deployment_dev/perinatal` branch, which had the same issue, so that we needn't wait for these changes to be merged into that branch via master.

chrismclarke (approved):
Thanks @jfmcquade , follow-up issue created as mentioned for gdrive folder access issue https://github.com/IDEMSInternational/parenting-app-ui/issues/1746

EsmeeIDEMS (approved):
Functional test passed.

PR #1756, opened 2023-01-30
feat: add support for deployment workflows

chrismclarke:
PR Checklist

- [x] - Latest `master` branch merged
- [x] - PR title descriptive (can be used in release notes)

## Breaking Changes
Use of `yarn scripts deployment set ...` has been deprecated. Use `yarn workflow deployment set ...` instead

## Description

As part of trying to unblock #1755, extracts the code related to adding support for deployment workflow to standalone PR for ease of merge. The reason for moving code to a deployment, is to allow for more complex scripting when setting deployment (such as future syncing git repo content and clearing angular cache on deployment change).

Main changes:

- Add support for workflows to receive args, e.g. `yarn workflow my_workflow arg1 arg2`
- Add support for workflows to define child workflows, e.g `yarn workflow parent_workflow child_workflow arg1 arg2`
- Update documentation and CI references to replace references to `yarn scripts deployment set...` -> `yarn workflow deployment set ...`

**TODO (follow-up PR)**
- [ ] Migrate code related to deployment create/compile
- [ ] Update existing deployments to use flat folder structures

## Review Notes
This PR is aimed to have no impact on the main codebase, beyond having to specify `yarn workflow deployment` instead of `yarn scripts deployment`. Simple review to confirm can switch deployments and operate as expected.

## Git Issues

Closes #

## Screenshots/Videos
**Console warning message given when calling legacy script**
image

PR #1764, opened 2023-01-31
fix: set deployment after config update bug

jfmcquade:
PR Checklist

## Description

I noticed a bug after #1756 was merged, whereby the CLI would enter an infinite loop – I believe this was when trying to set the deployment after the current deployment's config had been updated. See video below. This PR should fix this issue.

## Git Issues

Closes #

## Screenshots/Videos

https://user-images.githubusercontent.com/64838927/215776024-9f9837e8-b98c-4dca-9a91-bd6d8b3d68e1.mov


chrismclarke (approved):
Thanks @jfmcquade , the script is actually removed in #1763 but good to fix here until that is merged

PR #1769, opened 2023-02-03
feat: add seo service

chrismclarke:
PR Checklist

## Description

- Add a service to handle update of any SEO metadata
- Add placeholders for future SEO properties
- Remove default `title` from main page HTML ("Parenting App" for all deployments)
- Set page title from deployment

## Review Notes
The only property being set is the document title, which is what the browser tab uses for a name. So can just test that on a couple different deployments (should be same as app header title)

## Author Notes
I expect we will also want to include an authoring option to change the icon that appears in the browser bar, this would be best as a follow-up issue as we will likely need to update the deployments a little so better after #1763 sorted.
We may also want to provide best-practice info for authors, e.g.
https://www.emergeinteractive.com/insights/detail/the-essentials-of-favicons/ 
(tl;dr - a square svg should work fine in most browsers, or otherwise a square png up to 192x192)

## Dev Notes
I've also included additional SEO metadata fields for opengraph (facebook) and twitter as these are the most commonly used. If we decide in the future that we want web deployments to be more SEO friendly we could think of strategies of how these might get updated as we navigate different templates, so for now they are really just a placeholder. Examples of the tags in action can be seen at https://metatags.io 

Additionally depending on the service retrieving the data there is a good chance the bot will not wait long enough for the full spa to load and populate the tags from the SEO service, so in those cases we would want to use a service like https://prerender.io to add an intermediate caching layer between bots and web deployment

## Git Issues

Closes #

## Screenshots/Videos
**ParentApp populates title (tab name) from deployment config**
image

**Early Family Math deployment sets a different title**
image

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -34,6 +34,7 @@ import { IAppConfig } from "./shared/model";
 import { TaskService } from "./shared/services/task/task.service";
 import { AsyncServiceBase } from "./shared/services/asyncService.base";
 import { SyncServiceBase } from "./shared/services/syncService.base";
+import { SeoService } from "./shared/services/seo/seo.service";
 
 @Component({
   selector: "app-root",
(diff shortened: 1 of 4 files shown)


esmeetewinkel (approved):
Functional test passed and follow-up issue about browser tab icon created. https://github.com/IDEMSInternational/parenting-app-ui/issues/1773

jfmcquade (approved):
Looks good. Nice to have the additional SEO features ready if we want them. I've not looked into og/twitter metatags before so interesting to see

PR #1771, opened 2023-02-03
feat: add docs spellcheck lint and action

chrismclarke:
PR Checklist

- [x] - Latest `master` branch merged
- [x] - PR title descriptive (can be used in release notes)

## Description

- Add tool to spellcheck documentation (cspell)
- Add github action to automatically run spellcheck whenever documentation files change

## Review Notes
I haven't corrected all the spelling, it would be great if somebody else could help make the required fixes.
Run the spellchecker locally via
```
yarn lint:docs
```
This will show a list of all potential typos, e.g.
image

Any words that aren't recognised but spelt correctly can be added to the list of words in `cspell.config.yml`, e.g.
```yml
"words":
  [
    "idems",
    "quickstart",
    "glitchtip",
```

Any lines that should specifically be ignored should include a comment to disable
```yml
<!-- cspell:disable-next-line -->
E.g. `1ja6lzbphZaxnVv5mpQ4YHnn2qmxMiEBW`
```
Similarly sections of files or entire files can be disabled as required, discussed in more detail in [CSpell Docs](http://cspell.org/configuration/document-settings/#enable--disable-checking-sections-of-code)

## Git Issues

Closes #

PR #1772, opened 2023-02-03
Fix/unit tests

chrismclarke:
PR Checklist

## Description
The build test action is currently failing on #1771, due to some unit tests that were broken in #1763. The broken tests weren't detected due to a quirk where the unit tests weren't being run if a preview deployment is being created (had assumed the deployment preview would run the same checks that the standalone build test does, however that is not the case)

This PR fixes the underlying issue with the tests, namely how due to some updates in ts-node and deployment configs the tests can't add mock functions for anything that is exported (needs to be refactored either into a class or nested in another constant)
`https://github.com/jasmine/jasmine/issues/1414`

It also adds an extra set of linting rules to catch issues with tests (namely if focused tests are set with keywords like `fit`, which omits the rest of the testbase). It also adds a lint check to the build-test ci pipeline and enforces it even when a deployment preview is being used

## Review Notes
The only real changes are two functions which we sometimes need to mock in our tests, namely `getActiveDeployment` which now is wrapped as `ActiveDeployment.get` and `logError` which is now wrapped as `Logger.error`. So most the file changes are just making those replacements

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,8 @@
 {
   "root": true,
   "ignorePatterns": ["projects/**/*"],
+  "plugins": ["jasmine"],
+  "env": { "jasmine": true },
   "overrides": [
     {
       "files": ["*.ts"],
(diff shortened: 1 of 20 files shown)


jfmcquade (approved):
Looks good. A good workaround for the quirks of the testing suite

chrismclarke:
Yeah it sort of defeats the point of testing when you have to modify the code to support the testing infrastructure, but does seem to be somewhat unavoidable for this specific issue (beyond migrating to jest, which I'm sure would still have some issues)

PR #1776, opened 2023-02-05
refactor: app-data contents json

chrismclarke:
PR Checklist

## Description

Part of the never-ending refactor nesting dolls, this time a minor update to reformat contents sheets as json instead of .ts files and to move the visual test config to its own folder. This is to help tidy up pr #1775

There shouldn't be any real changes (apart from the latest content being synced in this branch as well). Will run the visual-test just to make sure it still runs as planned.

Main changes
- Output contents files as json instead of .ts for app-data workspace
- Move app-data index into src code (can import jsons directly from app-data instead of relying on app-data to export ts)
- Move visual test config out of app-data (not really sure why it was there in the first place)
- Remove all .ts dependencies from app-data (in preparation for git repos which will only store json data)

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade (commented):
Looks good. As well as the practical requirements, it seems more intuitive to have the index files as .json.

When running a copy within a sync, I get the following errors. I think it's returned by prettier via the `runPrettierCodeTidy()` method in `packages/scripts/src/commands/app-data/copy.ts`. Now that there are no .ts files in `/app-data`, perhaps we can remove the method call?
Screenshot 2023-02-06 at 12 11 49
Screenshot 2023-02-06 at 12 17 04

chrismclarke:
Thanks for double-checking all this @jfmcquade - yes I noticed the issue after the fact too and you're right it's the prettier call which is no longer needed. Seeing as the error doesn't kill the process should be fine to wait for removal in #1775, or can add a quick hotfix later in the week if that PR gets held up

PR #1778, opened 2023-02-07
Update ParentApp icon

esmeetewinkel:
PR Checklist

## Description
PR updates ParentApp android icon.


jfmcquade:
I ran the app locally in an emulator and found that the launcher icon background doesn't seem to be applied correctly – the background is solid white rather than using the gradient. I can't see any problems with the files generated so I'm not sure what the issue is. I've added a `test - appetize` tag, so we can see how the launcher icon looks in the appetize emulator.

Also, round launcher icon cropped too closely around the image in my opinions. I have altered the `icon-foreground.png` image to resize the central artwork and increase the border around it, which I think looks better (I've uploaded the updated version of the file to the [relevant gdrive folder](https://drive.google.com/drive/u/1/folders/1uTxOtNgra_tLHGD_ho7H6ahRC3YU_25V)).

jfmcquade (approved):
I've made a commit and the app launcher icon now looks correct on the updated appetize preview link, including the gradient background. So this should be good to merge.

PR #1780, opened 2023-02-08
feat: add support for deployment encrypted files

chrismclarke:
PR Checklist

## Description
Add supports for encrypted files within a designated deployment folder (`./encrypted`), and workflow methods to encrypt and decrypt data

An example use-case for this is including service accounts with a deployment that can automatically sync a specific set of google sheets

- [x] `deployment encrypt` workflow
- [ ] `deployment decrypt` workflow
- [x] Utilities to help process of generating new encrypted folder and private key
- [ ] Documentation  
- [ ] Refactor legacy config decrypt methods

**To Discuss**
- File naming conventions (couldn't find standards so opting for files ending `.crypt` (previously `.enc`))
- Clarify potential use cases or questions

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/data-models/workflows/deployment.workflows.ts
+++ b/packages/data-models/workflows/deployment.workflows.ts
@@ -46,6 +46,15 @@ const workflows: IDeploymentWorkflows = {
           },
         ],
       },
+      encrypt: {
+        label: "Encrypt deployment config",
+        steps: [
+          {
+            name: "encrypt",
+            function: async ({ tasks }) => tasks.encryption.encrypt(),
+          },
+        ],
+      },
     },
   },
 };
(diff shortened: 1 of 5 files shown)


esmeetewinkel (approved):
I was able to decrypt the super secret message (why did I enjoy this). 

image

I'm also able to encrypt and decrypt my own message. One comment on encrypting: if the `encrypted` subfolder already exists but the `private.key` file is missing, then for some reason running `yarn workflow deployment encrypt` still generates an encrypted message but the `private.key` file does not show up.

chrismclarke:
Quick note on failing tests - this is due to some backend-specific code being in files accessed by frontend (but not actually used), resulting in build issues. This has been fixed in PR #1775 so suggest waiting until that has been merged, or duplicating changes here

https://github.com/IDEMSInternational/parenting-app-ui/pull/1775/files#diff-7515343f225d94adac010fb864f917428fcdf36b68ded7eadbe147db11599e61

chrismclarke:
@esmeetewinkel 
> One comment on encrypting: if the encrypted subfolder already exists but the private.key file is missing, then for some reason running yarn workflow deployment encrypt still generates an encrypted message but the private.key

That's actually expected, you only ever need a public key to encrypt (private key to decrypt). So in theory anyone can add encrypted files to deployments, but only people with the private key can decrypt them. I tried adding a bit to the readme to clarify, but appreciate still not very intuitive

image

Feel free to update if you think it could be clearer

PR #1783, opened 2023-02-08
fix: feedback template path

### LINKED ISSUE #1782: [BUG] Not able to submit feedback
**Describe the bug**
The feedback system seems to be broken: 
1. When navigation is switched off the buttons on the Feedback pop up don't work. I'm guessing this is a consequence of disabling all actions (since this is technically a template), see https://github.com/IDEMSInternational/parenting-app-ui/pull/1603.
2. When navigation is switched on, these buttons do work but still no new (pending nor sent) database entry appears when trying to submit feedback. 

**App version**
v0.16.11

**To Reproduce**
1. Go to https://plh-global.web.app/template/home_screen(sidebar:feedback)
2. With navigation switched on or off, add some feedback and click "submit"

**Screenshots and videos**
Screen video on master branch v0.16.11

[screen-capture (30).webm](https://user-images.githubusercontent.com/74557272/217643640-0137ff00-f6cd-4db0-9d54-484b76927387.webm)

N.B. on the video you can see a lot of warnings in the console, but these appear independently of using the feedback sidebar. So I've tracked this in a separate issue https://github.com/IDEMSInternational/parenting-app-ui/issues/1781

chrismclarke:
PR Checklist

## Description

Fixes issue where feedback not submitting due to an error that is thrown while generating the feedback metadata (specifically the path to the nested template). 

## Git Issues

Closes #1782

## Screenshots/Videos

[ParentApp.webm](https://user-images.githubusercontent.com/10515065/217657004-18d726d0-4c99-4b8c-b93d-da5947c00596.webm)


esmeetewinkel (changes requested):
Thanks! This fixes the second issue mentioned in https://github.com/IDEMSInternational/parenting-app-ui/issues/1782 

Do you have any suggestions of how to fix the first issue? i.e. _When navigation is switched off the buttons on the Feedback pop up don't work._ When launching the feedback pop-up you're automatically unticking the "feedback enabled" tickbox to ensure one cannot give feedback on the feedback pop-up. Would it be possible to add there to also tick the "navigation" tickbox so that the submit button on the feedback pop-up works?

chrismclarke:
Thanks @esmeetewinkel , I think I must have glossed over the first issue.

The trouble with pop-ups is that we actually treat them as navigation events (when you click a popup you'll notice the navbar at the top change). This makes it easier for us to track state in cases such as clicking a link in a popup to navigate to a new page, and then returning with expectation popup is still there. 

How important is it to allow these while navigation disabled? I'm sure we could write some sort of workaround code although might be a little messy to do.

chrismclarke:
@esmeetewinkel 
I've hopefully resolved the first issue now, you were right that disabling all the actions caused issues for the actual feedback submission. I've now excluded the feedback-specific actions (as well as set_local) from the blocked list so things should submit as expected. I've also updated the feedback documentation which was quite outdated

I had to do a little bit of extra code tidying, I don't expect it to have knock-ons anywhere else but looping in @jfmcquade for quick code review

esmeetewinkel (approved):
This fixes both issues mentioned in https://github.com/IDEMSInternational/parenting-app-ui/issues/1782.

However, this re-introduces the issue of being able to give feedback on the feedback pop-up (https://github.com/IDEMSInternational/parenting-app-ui/pull/1603#pullrequestreview-1193256830).  

[screen-capture (32).webm](https://user-images.githubusercontent.com/74557272/218420285-47d1b997-c79a-46bf-b418-8871f34ce58b.webm)

I'll merge here and track this issue separately (because I need the fix to https://github.com/IDEMSInternational/parenting-app-ui/issues/1782 asap).

jfmcquade (approved):
Looks good. Having dedicated `feedback` actions makes sense I think

chrismclarke:
The game of whack-a-mole continues....
(not sure if you get the reference @esmeetewinkel !)

PR #1784, opened 2023-02-08
Feat/hidden sourcemaps

### LINKED ISSUE #1781: [BUG] Warning Unexpected token '<', "<!DOCTYPE "... is not valid JSON
**Describe the bug**
When opening a web preview of the app that is on v0.16.11 (https://plh-global.web.app/ or https://plh-tz.web.app/) I get >20 "warnings" in the console
```
DevTools failed to load source map: 
Could not parse content for https://plh-global.web.app/9989.7ac1e4675bf5d07c.js.map: 
Unexpected token '<', "<!DOCTYPE "... is not valid JSON
```
This doesn't seem to happen on https://plh-za.web.app/ which is on v0.16.5 so I imagine it was introduced later. Also, this doesn't happen when I'm running the app locally.

**App version**
v0.16.11

**Screenshots and videos**
image

chrismclarke:
PR Checklist

## Description
As mentioned in the issue notes, production deployments have a number of console warnings related to the fact that the original build included sourcemaps for mapping error logs to correct code files, but were deleted after upload to glitchtip (to reduce final build size).

 This PR addresses by adding a custom build configuration that includes sourcemaps but marks as hidden (so browser does not look for them). These sourcemaps can be uploaded to glitchtip and deleted without the browser complaining

It additionally updates the test-preview action to include sourcemaps for easier debugging of preview deployments.

## Review Notes
The easiest way to test will be to see if the warnings disappear on global after the deployment merged into master. Alternatively can mimic the build process locally via `yarn build --configuration "production,glitchtip"` and then run locally via `npx serve`

NOTE - I've added a test-preview tag to test the additional update that includes full sourcemaps in preview versions, so that console logs show the correct sourcefile. This is different from how the production deployment versions which require checking glitchtip

## Git Issues

Closes #1781

## Screenshots/Videos


jfmcquade (approved):
When I run `yarn build --configuration "production,glitchtip"` and `npx serve`, I do still have the following sourcemap-related warnings in the console:
Screenshot 2023-02-10 at 11 11 53

Likewise the preview build generated by the `test - preview` flag has the same warnings:
Screenshot 2023-02-10 at 11 15 55

This doesn't seem like an important issue, particularly if the many warnings detailed in the issue have been suppressed.

chrismclarke:
@jfmcquade 
Looks like those warnings are coming from google extensions not our code. I have one of the same for adblock, guessing the other is another extension. If you follow the link you can see which.
image


The map warnings currently in the plh-global deployment which should be gone with this one look like
image

esmeetewinkel:
Just had a look on the web preview of master branch, the warnings are gone indeed.

image

PR #1789, opened 2023-02-10
feat: deployment git import

chrismclarke:
PR Checklist

## Description

Further parts of #1755 - this time to support use of git in workflows and importing of deployments from remote repo.
This only handles the git import process, future PR required for managing content and syncing

## Review Notes
Test import process for an external repo. I've created a public repo with deployment config at `https://github.com/c-squared-dev/idems-app-c2dev-demo`, I'd be interested to see if you can export and use as expected 
```
yarn workflow deployment import https://github.com/c-squared-dev/idems-app-c2dev-demo
```
Any feedback on process welcome

(Note - sync scripts won't work for repo as I haven't shared access to the gdrive sheets)

I also manually tested importing from a repo with invalid configuration (it shows a message to say config is invalid, but will allow import if confirmed so user can fix config manually), but feel free if you want to create your own repo for testing anything else.

The import URL should be a github repo, although should work if ending `/idems-app-c2dev-demo` or `/idems-app-c2dev-demo.git` (anything supported by `git clone` operation)

There are also minor changes to the scripts that set deployments, so would be good just to quickly check you can revert to the previous deployment via set and sync as usual. There are no other changes to the way any sheets are processed or frontend code, so frontend functional testing not required

## Dev Notes
The big change here is using [simple-git](https://www.npmjs.com/package/simple-git) as a wrapper around git operations (as opposed to directly calling user's git command line). This makes it a lot easier to add type-checking and handle programmatic operation

There are some additional pre-emptive change to the deployment model, mostly just code tidying to remove the optional `?` from fields which will be populated by the default config (so yes optional as an override, but not optional for the generated output). There is also a placeholder `git` deployment field to allow the testing repo to parse (will be used in future pr)

The only other changes are to how the various deployment scripts are triggered - I don't expect these to have any real impact (should speed up things slightly calling scripts directly instead of through spawned processes)

## Git Issues

Closes #

## Screenshots/Videos
**Example outputs from import process**
image


esmeetewinkel (changes requested):
- I named your deployment `chris_dep` but I'm still seeing "c2dev_demo" in some places (e.g. when setting my deployment and in the file `activeDeployment.json`)

image

- When running `yarn start` I'm seeing the app header from the external config, but I'm not seeing the content from the external repo (just double checking that this was the expected behaviour)

image

- I'm able to move back and forth between this deployment and others as expected.

chrismclarke:
Thanks @esmeetewinkel 

> I named your deployment chris_dep but I'm still seeing "c2dev_demo" in some places (e.g. when setting my deployment and in the file activeDeployment.json)

Yeah this is a bit of a challenge I haven't quite figured out. Essentially _names_ exist in 2 ways, the name of the folder a deployment sits in and the _name_ given to the deployment itself. In an ideal world these 2 would just always be the same, however people move things and rename things and there's not really any guarantees. The added challenge we have is that the workflows don't know the actual name of a deployment until it has loaded it (it needs to compile the `config.ts` file first), and so often just reads the names of folders to provide an initial reference until updating.

So I don't think I really have a plan/proposal/solution - but would be good to think about a bit more.

> When running yarn start I'm seeing the app header from the external config, but I'm not seeing the content from the external repo (just double checking that this was the expected behaviour)

Yes that is expected for now. I did do some work in the linked PR so that when you load a deployment it should also load all the content for it, however it started to get a bit big so for now this just imports and you would need to manually run the sync workflow to populate content (which will fail if you don't have access to my google drive folder)

> I'm able to move back and forth between this deployment and others as expected.
Thanks for checking, yeah I was hoping this PR could be fairly basic/non-breaking to make it easier to get merged in

chrismclarke:
@esmeetewinkel @jfmcquade 
Just to quickly check-in on this one. Following the call last week I think the consensus was not to worry too much about the deployment/folder name issue for now - so I think this is good to merge pending any code-review feedback or other issues raised

jfmcquade (approved):
> Following the call last week I think the consensus was not to worry too much about the deployment/folder name issue for now
That's how I remember the conversation going too

> so I think this is good to merge pending any code-review feedback or other issues raised
Looks good to merge. I've done a quick functional test and had a look over the code, all looks god. I like the use of simple-git. And requiring all fields of the `IDeploymentConfig` makes senst.

PR #1795, opened 2023-02-17
feat: handle translated theme assets

### LINKED ISSUE #1798: [FEATURE] Handle theme and language assets
From #1526 
> - We want translation-specific assets and theme-specific assets to be as independent as possible
> - Currently, if a translated version of an asset exists, this will be used regardless of theme-specific versions
>   - This is fine for our current use case, as we only have images for the second ("professional") theme, which do not have translated versions – all of the assets for the professional theme sit inside a "global" subfolder
>   - But more work is needed to handle cases where translated, theme-specific assets exist, i.e. to handle language-code subfolders inside the professional theme folder.

jfmcquade:
PR Checklist

## Description

This PR adds the functionality to override assets with a variation relating to the current theme and language, including a combination thereof (e.g. a translated version of an asset that differs based on theme).

If a variation of the requested asset for the current theme and language does not exist, then a different version of the asset will be returned as a fallback. The order of priority for these fallbacks is:
1. current theme, current language
2. default theme, current language
3. current theme, default language
4. default theme, default language

## Testing

Checkout the feature branch and run `yarn workflow sync_assets`. The assets should be downloaded and populated to folders as usual. The contents file at `packages/app-data/assets/contents.json` now has a different format: asset language variations are nested under their respective theme variation. E.g.
Screenshot 2023-02-20 at 18 13 48

The template [debug_theme_language_assets](https://docs.google.com/spreadsheets/d/1pdhgtkQR4aVT9exigJrhUQu9TUtSEzLIpEqK47LYJNY/edit#gid=1422144003) shows the functionality of applying asset overrides based on theme and language settings. The source assets used for the example can be found in their respective folders (e.g. [plh_assets/theme_professional/tz_sw/quality_assurance/](https://drive.google.com/drive/u/1/folders/1uu2K2i2bgL-Gv_PlrJNx6C__UQRAqu4p)). We could add more debug assets in order to functionally test the fallback functionality and priority.

## Git Issues

Closes #1798 

## Screenshots/Videos

New assets summary:
Screenshot 2023-02-20 at 14 46 34

The [debug_theme_language_assets](https://docs.google.com/spreadsheets/d/1pdhgtkQR4aVT9exigJrhUQu9TUtSEzLIpEqK47LYJNY/edit#gid=1422144003) template demonstrating the functionality of applying asset overrides based on theme and language settings.

https://user-images.githubusercontent.com/64838927/220179353-c36ce0f4-5dd0-4a83-87eb-479ad4ff2de1.mov


packages/scripts/src/commands/app-data/postProcess/assets.ts
```diff
@@ -238,66 +255,58 @@ export class AssetsPostProcessor {
...
+        themeName = getThemeNameFromThemeAssetFolderName(baseFolder);
+        [baseFolder, ...nestedPaths] = nestedPath.split("/");
+        nestedPath = nestedPaths.join("/");
+      }
+
+      if (baseFolder === globalFolder || isCountryLanguageCode(baseFolder)) {
+        entries.complete[nestedPath] ??= {} as any;
+        entries.complete[nestedPath].themeVariations ??= {};
+        entries.complete[nestedPath].themeVariations[themeName] ??= {};
+        entries.complete[nestedPath].themeVariations[themeName][baseFolder] = assetEntry;
```
jfmcquade, line 282: Is there a better way to do this (set a property on a nested object that may not already exist)?

src/app/shared/components/template/services/template-asset.service.ts
```diff
@@ -25,53 +26,53 @@ export class TemplateAssetService extends AsyncServiceBase {
...
   /**
-   * Retrieve the path to translated version of an asset path for the current language.
-   * Fallsback to original path if does not exist
+   * Retrieve the path to a variation of an asset for the current language and theme.
+   * It is possible that such a variation does not exist, in which case the path to a
+   * different version of the asset will be returned as a fallback.
+   * The order of priority for these fallbacks is:
+   * 1. current theme, current language
+   * 2. default theme, current language
+   * 3. current theme, default language
```
chrismclarke, line 36: Thanks for the clear description, the reasoning makes sense to me also

src/app/shared/components/template/services/template-asset.service.ts
```diff
@@ -25,53 +26,53 @@ export class TemplateAssetService extends AsyncServiceBase {
...
+    // in their chosen language if any translation exists.
+    else if (assetEntry?.themeVariations?.[DEFAULT_THEME_NAME]?.[currentLanguageCode]) {
+      relativePathToOverride = `${currentLanguageCode}/${assetName}`;
     }
-    if (assetEntry?.themeVariations?.[themeName]) {
-      return this.convertPLHRelativePathToAssetPath(
-        `theme_${themeName}/${ASSETS_GLOBAL_FOLDER_NAME}/${assetName}`
-      );
+    // Otherwise use a the global version of the asset for the current theme, if it exists
+    else if (assetEntry?.themeVariations?.[currentThemeName]?.[ASSETS_GLOBAL_FOLDER_NAME]) {
```
chrismclarke, line 67: nit(non-blocking)
There's probably a nicer way to avoid so many if statements (e.g. creating an array with a list of mappings to check, stopping at the first one it finds), however the comments make the functionality clear so not a huge issue

src/app/shared/components/template/services/template-asset.service.ts
```diff
@@ -25,53 +26,53 @@ export class TemplateAssetService extends AsyncServiceBase {
...
+      relativePathToOverride = `${themePath}${ASSETS_GLOBAL_FOLDER_NAME}/${assetName}`;
     }
-    return this.convertPLHRelativePathToAssetPath(`${ASSETS_GLOBAL_FOLDER_NAME}/${assetName}`);
+    return this.convertPLHRelativePathToAssetPath(relativePathToOverride);
   }
 
   private cleanAssetName(value: string) {
     // remove prefix slash
-    if (value.startsWith("/")) value = value.replace("/", "");
+    if (value.startsWith("/")) value = value.substring(1);
```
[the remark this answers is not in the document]
chrismclarke: Well, next time the app is sluggish that's one fewer place to look!

packages/scripts/src/commands/app-data/postProcess/assets.ts
```diff
@@ -238,66 +255,58 @@ export class AssetsPostProcessor {
...
 
 interface IAssetEntriesByType {
-  /** Assets listed in global folder with reference to language or theme overrides */
-  global: IAssetEntryHashmap;
-  translations: { [languageCode: string]: IAssetEntryHashmap };
-  themeVariations: { [themeName: string]: IAssetEntryHashmap };
-  /** Assets that appear in translation folders but have no corresponding globals */
-  missing: IAssetEntryHashmap;
+  /** Assets that have a global in the default theme, including their respective overrides */
+  complete: IAssetEntryHashmap;
```
chrismclarke, line 309: nit(non-blocking)
I might suggest distinguish `tracked` / `untracked` instead of `complete` / `orphaned` as slightly clear naming, but not a strong opinion

packages/scripts/src/commands/app-data/postProcess/assets.ts
chrismclarke: Not using vanilla JS, but a while back I created a utility `setNestedProperty` function for something similar. 
I've just updated and added some spec tests in #1812 if you want to check it out and see if it will work for your purpose (feel free to add to the tests if there's anything extra you want to include)

chrismclarke (changes requested):
Thanks @jfmcquade, really nice to see this feature coming together. Thanks also for adding the additional tests, definitely makes it easier to follow what should be happening and follow more of the potential use-case-scenarios. I also like the updated asset breakdown

I've added a few comments inline, as for larger discussion items:

1. I think I remember on the call agreeing it would be good to allow a file system with just top-level assets which automatically get populated as global theme/default lang. Does this pr add support? if so would be good to add a test (possibly separating core/theme/lang tests a bit to use different mock folder structures)

2. I'm less concerned about the output json as these are mostly just for minor tracking/debugging and the additional json size for nesting not too significant.

jfmcquade:
Thanks @chrismclarke, I've made various code changes as per your inline comments.
 
> 1. I think I remember on the call agreeing it would be good to allow a file system with just top-level assets which automatically get populated as global theme/default lang. Does this pr add support? if so would be good to add a test (possibly separating core/theme/lang tests a bit to use different mock folder structures)

As of this PR, "top-level" assets must still be inside a folder named "global", so that behaviour is unchanged. I have added a simple test to demonstrate that providing only a "global" version of an asset (i.e. not including any theme/lang versions) does not cause any issues

> 2. I'm less concerned about the output json as these are mostly just for minor tracking/debugging and the additional json size for nesting not too significant. So I'd say probably stick with whatever is easiest for now, although if you did decide to refactor it might be worth as you say prefixing `theme_` folders and also probably `lang_` folders too (feels a bit off that authors are basically blocked from using any folder names that could be mistaken for ab_cd lang codes). I guess any changes would also depend on whether we want to also impose such conventions on the authoring and/or have a slight mismatch between folder names and json entries.

This makes sense, I've left the structure of the contents file as it is described in the PR description. It may be that in working on asset packs we revisit this structure, which I think is another reason to avoid making any changes that affect authors for now.

chrismclarke:
merge into #1874

PR #1803, opened 2023-02-21
Feat/in app update

jfmcquade:
PR Checklist

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos


src/app/app.component.ts
```diff
@@ -96,6 +98,7 @@ export class AppComponent {
       // ensure deployment field set correctly for use in any startup services or templates
       localStorage.setItem(this.appFields.DEPLOYMENT_NAME, this.DEPLOYMENT_NAME);
       localStorage.setItem(this.appFields.APP_VERSION, this.APP_VERSION);
+      await this.checkForAppUpdatesAndApply();
```
chrismclarke, line 101: nit(blocking)
Would be better to move all init code to service itself. If it is actually required to run before anything else then you can put it first in the list of `blocking` services below (although I expect this is actually something that can be deferred to background and run async)

src/app/shared/services/app-update/app-update.service.ts
```diff
@@ -0,0 +1,39 @@
+import { Injectable } from "@angular/core";
+import { AppUpdate, AppUpdateInfo, AppUpdateAvailability } from "@capawesome/capacitor-app-update";
+
+@Injectable({
+  providedIn: "root",
+})
+export class AppUpdateService {
```
chrismclarke, line 7: nit(blocking)
Would be good for the to extend our patter of sync/async service and define any init methods like other services.

src/app/app.component.ts
```diff
@@ -283,4 +287,10 @@ export class AppComponent {
       }
     }
   }
+
+  private async checkForAppUpdatesAndApply() {
+    if (Capacitor.isNativePlatform()) {
+      await this.appUpdateService.checkForUpdatesAndAttemptImmediateUpdate();
```
chrismclarke, line 293: nit(non-blocking)
Would be nice to have some sort of AppConfig that can define whether to opt in to this feature, and strategy to use. E.g. `AppUpdatePrompt: {enabled: true, forceUserUpdate: false}

I'm also not sure how the UI of the feature works, but if we simply get a notification that an update is available it may be worth also providing an optional template to display when update ready

src/app/shared/services/app-update/app-update.service.ts
```diff
@@ -0,0 +1,39 @@
...
+    } catch (error) {
+      console.error("[App Update]", error);
+    }
+  }
+
+  async getAppUpdateInfo() {
+    return (await AppUpdate.getAppUpdateInfo()) as AppUpdateInfo;
+  }
+
+  async performImmediateUpdate() {
```
chrismclarke, line 28: nit(non-blocking)
If we're not planning on accessing these methods from other services probably no need to duplicate all the methods as public. If we move all init/handling logic to this service then probably can just rely on one or two functions

chrismclarke (changes requested):
Find a few code request changes inline

chrismclarke:
As for testing, here is the strategy 
0. Remove local app-data assets to make build size smaller
```
rm -rf 
```

1. Create local build and open in android studio
```
yarn build
```
Also manually deleted folders from `www\assets\app_data\assets` to reduce size for easier upload 

2. Sync to android and open in android studio
```
npx cap sync
npx cap open
```

3. Create a debug build from android studio menu (assumes google.json populated locally)
_Build -> Build Bundles / APKs -> Build Bundle_

It should prompt a message on complete to locate the bundle, e.g. `...\parenting-app-ui\android\app\build\outputs\bundle\debug`

4. Upload to internal app sharing (link unique to user)
https://play.google.com/console/internal-app-sharing

5. Run local emulator with device that has access to play store

4. Increase version
```
yarn version
```

6. Repeat steps

chrismclarke (changes requested):
Thanks @jmcquade, appreciate the painful iteration/testing process. I think it's nearly there, barring some minor restructuring to the actions handling (see comments inline)

It would also be incredibly useful it you could include a documentation page on this feature, mostly just copying over the keypoints you've already outlined in the PR notes

jfmcquade:
Thanks @chrismclarke. I've moved all of the logic to now be triggered by template actions, and added documentation.

It should be ready for more testing if you approve of the code. I'm happy to do the testing myself

chrismclarke (commented):
Thanks for all the updates @jfmcquade , the documentation is really useful to have and code feels well-organised.

My only concern with placing the service as an eager/blocking one is I'm not sure how capacitor handles the casee where internet connection is limited - I can imagine a situation where that could prevent the rest of the app from loading while trying to check for updates.

So I've refactored the code to have a sync init and only check for updates once the user explicitly triggers the action. The action itself will be blocking, so if authors want to prevent start until check (can't imagine why) then they can include as the first startup action. Otherwise I'd probably recommend making it the last startup action (may be worth adding to the docs if we do figure out the behaviour/reccomendation).

I've updated the code in 653103bc, alongside minor docs update. I haven't had a chance to functional test yet, so let me know if you do (otherwise I'll try find some time later)

jfmcquade:
Nice one, thanks @chrismclarke.

> My only concern with placing the service as an eager/blocking one is I'm not sure how capacitor handles the casee where internet connection is limited - I can imagine a situation where that could prevent the rest of the app from loading while trying to check for updates.
> 

That makes sense, perhaps I should add some extra error handling in that case – e.g. wrap the `checkForUpdates()` logic in a try/catch block? 

> So I've refactored the code to have a sync init and only check for updates once the user explicitly triggers the action. The action itself will be blocking, so if authors want to prevent start until check (can't imagine why) then they can include as the first startup action. Otherwise I'd probably recommend making it the last startup action (may be worth adding to the docs if we do figure out the behaviour/reccomendation).
> 
Sounds good, I thought I had a reason for keeping the check in the init logic and outside of the template actions, but I'm not sure what it was now. 
Regarding the order of start up actions, I was thinking about user experience when I placed the update as the first action: I thought it could be a bit jarring for a user to make their way through the introductory screens and the tour and be ready to start using the app, only to be prompted to update it. Although I can see that opening the app for the first time and immediately being prompted to update could also be off-putting (it may also delay the start-up, which I know is bad UX), so I can see how reasons for both. Either way, the template actions make it an authoring question! @esmeetewinkel let us know your opinion when you get around to reviewing this PR.

> I haven't had a chance to functional test yet, so let me know if you do (otherwise I'll try find some time later)

I'm off this afternoon, but happy to test on Monday.

chrismclarke:
> I thought it could be a bit jarring for a user to make their way through the introductory screens and the tour and be ready to start using the app, only to be prompted to update it

Everything about in-app updates is jarring, that's why I'd strongly suggest sticking to using the `prompt` mechanism only (the force action is far more intrusive and doesn't even really make sense given how it can still be bypassed). Similarly, I suggest keeping it as the last action so that it doesn't block the user from actually using the app, and easier for them to reject the prompt if all they want to do is use the app (updates will still happen in the background through the play store, so more of a friendly reminder)

esmeetewinkel (commented):
> I expect it should work for another user to test just following step 5-6 (and maybe 8) to use the build files I created.

I tried doing the shortcut version of the review that @chrismclarke suggested but the links don't work for me...
index

@jfmcquade Do you want the full test? I don't have android studio installed - or I could install it, maybe it's useful for other things in the future.

jfmcquade:
I suspect that the functionality can't be tested without uploading files from your own Google account, since step 8 won't trigger the app to be aware of the update.

Either way, I have run the tests again on the latest version of the code, and have updated the videos linked to in the PR description. 

Do we think that these functional tests are sufficient? I am aware that we haven't tested on multiple devices, and I may have failed to capture some possible user journeys. However, I think that as the core of the app update functionality is handled by Android's (presumably robust) native feature, and that our error handling should mean that no error prevents the app from being used, I would be reasonably confident that we can include this feature without further testing.

esmeetewinkel (approved):
I've watched the updated videos, thanks @jfmcquade for those. I'm happy to go ahead and merge this.

chrismclarke:
Sounds good to me!

PR #1809, opened 2023-02-23
Update PLH config

esmeetewinkel:
PR Checklist

## Description
This PR updates the PLH config so that 

1. Change default title for notifications
2. Change the default skin to "modular"
3. Change typo `weely_workshop.ts`

@jfmcquade I'm not able to get 2. to work, could you double check this PR and let me know what I'm doing wrong? And can you confirm I'm not breaking anything by changing the file name 3.


jfmcquade:
@esmeetewinkel There's nothing wrong with your code changes, but I've noticed 3 issues associated with changing the default skin for the plh deployment. I hope to have some time to return to this today and make some further progress, but I'm documenting here in case I don't.

#### 1. There is a bug in the deployment config compilation scripts that means that changes to a parent deployment config file do not trickle down to the deployment's children.
I'm working on this now but haven't fixed it yet. A workaround is to delete the child config.json file, e.g. `.idems_app/deployments/plh_global/config.json`, and then set the deployment again to recompile the json to reflect the changes made to the deployment's parent's config.

#### 2. With the config file compiled correctly, I found another bug relating to switching the active skin at runtime: changing the active skin does not change the active theme to the new skin's default theme.
This means that I was seeing the Modular skin with the Default styling and images when first launching the app. I have fixed this bug with a commit to this branch, however this now means that the active theme will recent to the active skin's default theme whenever the app is relaunched. As we don't currently have a use case for mixing a skin with its non-default theme, I think we can solve this problem independently.

#### 3. If the user has already launched the app, then deleting Local Storage is required to reset the skin to the deployment's default.
We use local storage to save the user's chosen skin so that it persists after relaunching the app. Currently, upon app launch, local storage is checked to see if the user has already set an active skin. If the user's saved active skin is "available" to the current deployment, then that skin remains active, but if the skin is not "available", or if the user has no skin previously set, then the active skin is set to the deployment's default. This logic means that currently, for users who are already using the app with the Weekly Workshop skin enabled, updating the app will not cause their active skin to change to the deployment's default. Some thought is required as to how to handle this use case alongside skins persisting after app relaunch.

PR #1830, opened 2023-03-07
Fix: infinite errors thrown when entering the app through a notification

### LINKED ISSUE #1805: [BUG] Infinitely many errors when entering the app through a notification
**Describe the bug**
When entering the app through a notification the console shows an unlimited number of errors (quickly climbing up to >1000). These seem to have to do with a missing title. This issue happens both for notifications that have an explicitly defined "title" in the sheet and those that don't (i.e. are using the default title "New message from PLH"). 

```
ERROR TypeError: Cannot read properties of undefined (reading 'title')
```

image

**App version**
v0.16.11

**To Reproduce**
- navigate to https://plh-global.web.app/notifications
- send yourself a notification
- click the notification to enter the app
- observe the errors in the console. 

**Screenshots and videos**

[screen-capture (37).webm](https://user-images.githubusercontent.com/74557272/220416485-ce1344c5-84c7-4b3d-9b02-936c0a629465.webm)

**Additional comments**
I don't know whether this behaviour has any knock-ons of breaking anything in the app that a user would notice. If it doesn't then this issue is low priority.

jfmcquade:
PR Checklist

## Description

Interacting with a notification whilst on the `notifications` page of the app would cause an error to be thrown repeatedly and indefinitely. This was, I think, only an issue when running in browser, and only applies when the user has the notifications page open (which is only available in dev mode), so is not an important issue.

This PR provides a workaround so that sent notifications appear correctly when interacting with them from the `/notifications` page.

## Dev notes

I believe this is just an issue on web and that this is a workaround.

Previously, interacting with a notification caused the notification to be saved to the database, however the `notification_meta` was not saved, as the `sessionNotifications$` observable did not update to trigger the `notification_meta` being saved for all session notifications. This meant that the `notification_meta` could not be read off by the `notification-debug-interacted-row` component, throwing the errors.
In the `local-notification-service`, there's a note that the "triggers don't appear to be working on web", which I believe is the cause of the issue (line 363 should mean that the `notification_meta` is saved at the next notification defer cycle, i.e. every 5000ms or on refresh)
Screenshot 2023-03-07 at 13 29 02

Therefore this PR offers a workaround by saving the `notification_meta` at the time of interacting with the notification.

## Git Issues

Closes #1805 

## Screenshots/Videos

The notifications page showing an interacted notification under "Sent Notifications" immediately.
(My screen recorder suppresses notifications, so interacting with the notification happens off-screen in this recording, after 1 second)

https://user-images.githubusercontent.com/64838927/223440091-d611aec2-3a50-466b-8f3f-cc3d0b70747d.mov

### DIFF
--- a/src/app/feature/notification/pages/notifications-debug/components/notification-debug-interacted-row.component.ts
+++ b/src/app/feature/notification/pages/notifications-debug/components/notification-debug-interacted-row.component.ts
@@ -10,9 +10,9 @@ import { ILocalNotificationInteractionDB } from "src/app/shared/services/notific
           <div style="flex:1">
             <div>
               <div>
-                {{ notification.notification_meta.title }}
+                {{ notification.notification_meta?.title }}
               </div>
-              <div class="info-text">{{ notification.notification_meta.id }}</div>
+              <div class="info-text">{{ notification.notification_meta?.id }}</div>
             </div>
           </div>
           <div>
@@ -31,7 +31,7 @@ import { ILocalNotificationInteractionDB } from "src/app/shared/services/notific
         <div class="row-content">
           <div style="flex: 1; margin-right: 8px">
             <div class="divider"></div>
-            <div>{{ notification.notification_meta.text }}</div>
+            <div>{{ notification.notification_meta?.text }}</div>
           </div>
           <div style="text-align: right">
             <div class="divider"></div>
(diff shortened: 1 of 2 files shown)


esmeetewinkel (approved):
Functional test passed.

PR #1832, opened 2023-03-07
Update WASH App android assets

esmeetewinkel:
PR Checklist

## Description
PR updates WASH App android assets.


esmeetewinkel:
@jfmcquade I think the generated splash screen images look a lot better. The icon could maybe have a bit less padding - feel free to crop as you see fit as you're testing

jfmcquade (approved):
The launcher icon wasn't generating correctly for me in Android Studio (it was displaying the old ParentApp icon), so I added the foreground/background image variants, and that worked. I've updated the docs to indicate that including these is mandatory.

All looks good to me,  I think the padding around the icon is good.

Screenshot 2023-03-09 at 13 52 23

PR #1834, opened 2023-03-09
feat: add support for ci sha tracking in glitchtip

chrismclarke:
PR Checklist

## Description
Error logging uses uploaded sourcemaps to create human-interpretable stack traces of errors, identifying where in the source code production errors originate from. These sourcemaps are automatically generated on release, and so the ability to create these error logs requires that the codebase used to upload the sourcemaps is identical to the codebase in the release.

This can result in broken sourcemaps in cases where new web releases are created independently of android app release, as we now have to versions of sourcemaps for the same app version, one which works with the android release and one with latest web release. 

This PR fixes by attaching the current git commit SHA to sourcemaps, so that all releases create a unique set of sourcemaps independent of the actual deployment version.

Additional changes
- Update deployment configs so that plh only uses global logging config for global build (previously default for all until override)

## Review Notes
I've included screenshots to show the process working for linking github commit sha with build and error logging. The only check remaining is whether the functions also work as part of the actions triggered when deployments are created. I tried to manually test using the web preview alongside a manual sourcemaps upload, however getting the codebases on the two actions to match is a pain so I ended up giving up.

So once merged it would be good to check future error logs to see if new errors are recorded as expected

## Git Issues

Closes #

## Screenshots/Videos
**Glitchtip tracking multiple deployment versions using commit SHA**
image

**Example action now also stores a local artifact (for 30 days) for use in debugging**
image

### DIFF
--- a/.github/workflows/deployment-hosting.yml
+++ b/.github/workflows/deployment-hosting.yml
@@ -23,7 +23,7 @@ jobs:
           then echo "DEPLOYMENT_NAME=plh_global" >> $GITHUB_ENV;          \
           else echo "DEPLOYMENT_NAME=${GITHUB_REF##*/}" >> $GITHUB_ENV;   \
           fi;
-          echo "SHA_SHORT=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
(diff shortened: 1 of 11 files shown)


jfmcquade (approved):
Looks good, we can check the results once we have a few errors come in

PR #1842, opened 2023-03-14
Update ParentApp icon and splash screen

esmeetewinkel:
PR Checklist

## Description
PR updates ParentApp's android assets, i.e. app icon and splash screen.


esmeetewinkel:
@jfmcquade Could you just confirm this is working well for you and the image sizes look sensible?

jfmcquade:
You can see the app icon if you drag up the app draw from near the bottom of the screen:
Screenshot 2023-03-14 at 16 08 17

You can then click and hold the icon to drag it to the home screen to see how it looks there:
Screenshot 2023-03-14 at 16 08 51

It looks good to me, and I like the splash screen too:
Screenshot 2023-03-14 at 16 10 05

However, I'm getting an issue with the appetize preview: the app doesn't load, it just displays a blank screen after the splash screen. Is it working for you?
Screenshot 2023-03-14 at 16 06 20

esmeetewinkel:
Thanks for explaining how to add the icon to the home screen! 

I am experiencing the same issue as you, on most device / android version combination I'm just getting a blank screen when trying to open the app. I've not used Appetize lately so I'm not sure when this might have broken... 

FYI (I don't think this is something we need to resolve): I found one other combination on which the app **does** work, Pixel 7 / Android 13. This seems to use the "icon" image both as an app icon and as a splash screen:
image

PR #1844, opened 2023-03-14
fix: standardise newline char in parsed json outputs

### LINKED ISSUE #1838: \r\n vs \n
**Details**
I'm seeing lots of `\r\n` in the files on the WASH App deployment. Normally I would have expected to see `\n` here, I have not noticed `\r` before. 
image
Normally I would have expected to see `\n` here. This is not a problem in the app, the content appears as usual. 

However, in Crowdin these appear as `\r` (the `\n` gets interpreted as a new line). 
image

This can be confusing to translators and adds unnecessary complexity to their task. 

**Hypothesis** 
Most of the files that contain these have been authored by INNODEMS or Elena, and the original files that don't display the issue were authored by me. So I was wondering if it is maybe depending on your computer or the way you achieve a "new line" in a Google sheets? I'm using CTRL+Enter, Elena is using ALT+Enter, but I don't think that makes a difference?

jfmcquade:
PR Checklist

## Description

As documented in issue #1838, some strings that appear in the json files generated from the templates would contain `\r\n` (CRLF) newline characters, as opposed to the standard `\n` (LF). This causes an issue when the strings are sent to be translated as "\r" characters would appear, unparsed, in the strings which could cause confusion for the translators.

This PR enforces the `\n` newline character by replacing all instances of `\r\n` with `\n` in all the outputted app-data sheets.

## Testing notes

1. Checkout `master` and pull the latest changes. 
2. Checkout `fix/standardise-newlines-in-parser`
3. Run `yarn workflow sync_sheets`
4. In VSCode's Source Control, you should see only changes that have been made to content since the last sync to master, alongside the newline character changes enforced by this PR (you may need to stage and un-stage all changes to remove erroneous file differences). At the time of writing, those changes are as shown in the screenshot below. The only file from PLH that previously had `\r\n` line endings is `packages/app-data/sheets/data_list/survey_question.json`. You should see by comparing the differences that these have been replaced with `\n`.

Screenshot 2023-03-14 at 16 26 55

It would be sensible to check that this content still appears in the app as expected.

The sheets for the WASH App contain many more examples of the `\r\n` character (and it was these that originally prompted the issue to be raised). However, as this feature branch was created from the master branch, it is more difficult to compare the results of the updated AppDataConverter on the WASH sheets. I have tested this and can confirm that the only changes appear to be the correct substitution of `\n` for `\r\n` in those files.

## Git Issues

Closes #1838

## Screenshots/Videos

The changes made by the new AppDataConverter to the file referenced in the original issue:
Screenshot 2023-03-14 at 15 52 05


packages/scripts/src/commands/app-data/convert/utils/app-data-string.utils.ts
```diff
@@ -122,3 +122,8 @@ export function extractDynamicDependencies(dynamicFields: FlowTypes.TemplateRow[
   });
   return dynamicDependencies;
 }
+
+// Standardised newline characters within a string (i.e. replace "\r\n" and "\r" with "\n")
```
esmeetewinkel, line 126: I haven't found any cases of `\r` not followed by `\n` . Do you expect these will show up at some point and if so are you sure we want to replace these?

esmeetewinkel (approved):
Functional test passed.

packages/scripts/src/commands/app-data/convert/utils/app-data-string.utils.ts
chrismclarke: Agreed, `\r` doesn't exist in the wild without `\n` (see my [comment](https://github.com/IDEMSInternational/parenting-app-ui/issues/1838#issuecomment-1469751567) on the linked issue), so just need the single replacement

chrismclarke (approved):
minor(non-blocking) nit from Esmee's comment, but otherwise looks good to me

PR #1848, opened 2023-03-16
feat: add radio_group_grid component

### LINKED ISSUE #1014: [BUG] Radio button size improvements
**Describe the bug**
- [ ] The size of the buttons in a radio group changes depending on whether or not the parameter options_per_row is given. Desired behaviour: A set of _n_ radio buttons that fit on one row should not change if parameter options_per_row is set to _n_. 
- [ ] Radio buttons should increase in size to fit the text. 

**App version**
v0.11.2

**To Reproduce**
Open debug template. 

**Debug Template**
[debug_radio_group_size](https://docs.google.com/spreadsheets/d/1pSFFOv9B9x0oUSuSJDp6RNePJ1XSP6yVSjXA8jzBclQ/edit#gid=481972108)

**Screenshots and videos**
image

chrismclarke:
PR Checklist

## Description

- Add new `radio_buttons_grid` component to display radio groups items in a grid
- Add interactive [component demo template](https://docs.google.com/spreadsheets/d/1yHLQHxZQuxs34gBU-ny95ZWs9e8-HybC4IAeFxwpcVw/edit#gid=569531329)
- Minor code tidying on older radio group component

Started as a fix for #1014, however the code for the existing radio groups is quite complex and so likely to bring in a lot of breaking changes trying to refactor/tidy up

So for now I've created a new component which we can use to slowly migrate content as required and address the immediate issue

## Git Issues

Closes #

## Review Notes

A few specific Qs
@jfmcquade 
- Interested to know what you think of the format I've used for the generated docs. I was hoping the parameter_list would get inlined definitions but needs an extra click, not sure if you've found a cleaner solution anywhere? (would need to run `yarn compodoc:serve` locally to view, haven't exported as it makes the diffs huge)
- Do the styles look correct? I've tried to copy your theme-based style variables so just to check if I've picked the right variables

@esmeetewinkel 
- Does the name `radio_group_grid` make sense? Or should it be `radio_grid` or anything else?
- I've not exposed many authoring options, I'm hoping the three included should cover a good number of scenarios but let me know if you think any others are required
- I had planned it to be mostly as an (optional) replacement for the original `radio-group` component, but I haven't included any of the custom style options (e.g. checked) because the code is a bit hard to follow and I don't know what is still needed or not. So let me know if there are specific things as needed if you do plan to migrate any existing content. I'll be keeping the existing `radio-group` component for the foreseeable future.
- Do you think this is sufficient to close #1014

## Screenshots/Videos

[ParentApp.webm](https://user-images.githubusercontent.com/10515065/225863275-463236a0-b6cb-4e7f-b795-9aed32529606.webm)

**Author parameters stored in standalone data list and shown interactively**

[component demo template](https://docs.google.com/spreadsheets/d/1yHLQHxZQuxs34gBU-ny95ZWs9e8-HybC4IAeFxwpcVw/edit#gid=569531329)

image

image


chrismclarke:
Thanks again @esmeetewinkel - I've now marked it ready for review. To address the points you've raised:
0. Renamed to `radio_button_grid` as suggested

1. For now the way to constrain heights is by constraining widths, but there should be good options to do so (see demos above). 

I could try to expose some sort of item_height parameter, although it can often lead to bits of confusion for the layout if trying to limit both width and height of the grid items but the two often rely on each other (e.g. want to keep within a max height, but that would mean the items are not wide enough if they still want to display the full content). But let me know if things still prove problematic and I can add some sort of `image_height` property to use

2. I've made the text slightly bigger but left as black for now. I'll possibly leave it to @jfmcquade to follow-up and add better way to change with themes

jfmcquade (approved):
Very nice! The interactive component demo is fun (and useful) – it could be good to implement similar demos for similar components with specific style parameters.

I like the format for the docs, I see what you mean about the extra click but I can't find a way around it. I suppose that it does add a bit of complexity for a layperson using the docs, but I think we agreed that, as they exist currently, we wouldn't expect someone to refer to the generated docs without a bit of an explanation anyway. 
The `IRadioButtonGridParams` type is definitely nice to have within the component code file anyway, and it's nice to see the constituent types in the docs (i.e. `answer_list: IAnswerListItem[]`).
I've made some minor edits to some of the comments – the multiple line comments come out on one line in the docs, so I added some punctuation for clarity.

Re the theme-based style variables, I think you've got the format right and used the existing `--radio-group-background-selected` correctly. I've added the 2 new variables (`--radio-button-font-size` and `--radio-button-font-color`) to the `$variable-overrides` maps that live in the theme definition files (`src/theme/themes/default.scss` and `src/theme/themes/professional.scss`). Each map contains a complete list of all the theme-based style variables, even those that are not being overridden by the given theme, for reference. I could probably handle this in a better way at some point. Also it's worth saying that up until now, these variables only apply colours and gradients, as I never got around to including all of the non-colour variables in the theme system (they currently still live in `src/theme/variables.scss`), but the format for doing so would be exactly the same.

chrismclarke:
Thanks for the thorough once-over and tidying @jfmcquade , I'll hand-over to @esmeetewinkel to see if things work well enough from an authoring perspective.

As for more component demos, might be worth a quick discussion at some point to see what would be a higher priority between markdown documentation, compodoc and demo component templates. But hopefully good to at least have an example of each type.

esmeetewinkel (commented):
This is working well. One minor comment: You implemented that the text becomes **bold** when a cell is selected.  In some cases this causes the text to wrap on more lines when it is selected, which could, if this radio button is the largest one hence it determines the size of all buttons in the grid, result in a bit of awkward scaling when trying out various options. I think we should leave the font fixed (either regular or bold) to prevent this. 

[screen-capture (41).webm](https://user-images.githubusercontent.com/74557272/228002664-8c06115b-5080-4be8-80c1-06b0179483be.webm)

chrismclarke:
> This is working well. One minor comment: You implemented that the text becomes **bold** when a cell is selected. In some cases this causes the text to wrap on more lines when it is selected, which could, if this radio button is the largest one hence it determines the size of all buttons in the grid, result in a bit of awkward scaling when trying out various options. I think we should leave the font fixed (either regular or bold) to prevent this.

Thanks @esmeetewinkel , really nice attention to detail I hadn't thought about
I know there are some ways you can magically pre-size the text so that it fills the same space whether bold or not, however possibly overkill for now so I've removed.

Although in case of interest in the future @jfmcquade - there's a handful of solutions in these threads:
- https://css-tricks.com/bold-on-hover-without-the-layout-shift/
- https://stackoverflow.com/questions/5687035/css-bolding-some-text-without-changing-its-containers-size
- https://stackoverflow.com/questions/556153/inline-elements-shifting-when-made-bold-on-hover

esmeetewinkel (approved):

PR #1850, opened 2023-03-16
docs: update component docs

chrismclarke:
PR Checklist

## Description

- Update generated component docs via `yarn compodoc:generate`

## Review Notes
Looks like a lot of changes, but I think that's just because each page contains a sidebar of components (think so it can show the dom tree in the UI preview and mark which are custom components), and so because we have added a couple of new ones all pages get updated.

Feels a bit inefficient, but as they're auto-generated docs I don't think worth investigating much further

## Git Issues

Closes #

## Screenshots/Videos


chrismclarke:
Only generated docs so going to merge without review, although have tagged you @jfmcquade in case you are interested in the pr note mentioned

PR #1862, opened 2023-03-22
Core/android api 33

chrismclarke:
PR Checklist

## Description
According to recent updates play store submissions need to target android API 33 (Android 13) by November
https://support.google.com/googleplay/android-developer/answer/11926878

- Bump Android api target 32->33 (Android 13)
- Update capacitorJS dependencies

## Review Notes
Should be tested on a device running Android 13 (assuming users might use). Particularly focus on native android features such as splash screens and notifications. Seems like Appetize only goes to 11 on the nexus 5 device but has 13 on the Pixel 7

I can see potential issue already flagged with notifications with a PR to fix in `https://github.com/ionic-team/capacitor-plugins/pull/1189`, but it appears that it will only be released as part of CapacitorJS v5 which is currently only in draft
https://github.com/ionic-team/capacitor-plugins/releases/tag/%40capacitor%2Flocal-notifications%405.0.0-alpha.1

So if not working as expected we should ~wait until stable release of Capacitor 5~ [Update to Capacitor 5](https://capacitorjs.com/docs/updating/5-0), update this PR and test again

## Git Issues

Closes #

## Screenshots/Videos


chrismclarke:
@jfmcquade 
Would you be able to test Android 13 locally?

jfmcquade:
I've been trying to test this natively on my phone running Android 13. I've been building an app bundle with Android Studio and deploying to [internal testing](https://play.google.com/console/u/1/internal-app-sharing?pli=1). I've been using the `plh_tz` deployment, as this should be fully featured (the debug deployment does not, for example, have the notifications testing page).

I've found that the Google Authentication (which calls `FirebaseAuthentication.signInWithGoogle()`) does not work for me. See video below

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/2c96e9c5-e7e9-4187-9eb6-57cbac752334

I tried to update the library we use, [@capacitor-firebase/authentication](https://www.npmjs.com/package/@capacitor-firebase/authentication?activeTab=versions) to the latest version (`1.3.0` -> `5.1.0`). However, Android Studio/gradle fails to build the bundle when I do so. 
Build error
I assume this error is because the latest version of the library only supports Capacitor 5 (hence the new naming convention), see [changelog here](https://github.com/capawesome-team/capacitor-firebase/blob/main/packages/authentication/CHANGELOG.md).

I built a version of the app for the `plh_tz` deployment with `enforceLogin` disabled, in order to test the other native features. When running this version on my phone, the notifications debug page is empty and doesn't allow me to test notifications. This may well be due to the [issue referenced](https://github.com/ionic-team/capacitor-plugins/pull/1189) in the PR notes (I didn't receive any permissions requests when installing and opening the app). 

notifications page native

This can be replicated using the appetize build for this PR. [Configured to run on Pixel 7 android 13.0](https://appetize.io/app/wbt5facrtrcczv7wcbbnhh386r?device=pixel7&osVersion=13.0&scale=75), the notifications page doesn't allow debugging notifications:

Screenshot 2023-08-21 at 17 25 47

Whereas the [Pixel 4/android 12.0 version](https://appetize.io/app/wbt5facrtrcczv7wcbbnhh386r?device=pixel4&osVersion=12.0&scale=75) works as expected:

notifications appetize

So it does seem like we can't target Android 13/API version 33 until we have updated to Capacitor 5, as in #2038. Apologies for forgetting about this issue until now.

chrismclarke:
Thanks for the thorough testing, yes let's prioritise getting Capacitor upgraded first then
Interested to know, do the notifications and login features work on your device when we haven't updated the target sdk? I.E if we were to merge would this introduce a new regression issue or just fail to address the issue of supporting Android 13 devices?

chrismclarke:
closing in favor of #2046

PR #1873, opened 2023-04-04
feat(remote assets) – supabase integration; placeholder services

jfmcquade:
PR Checklist

## Description

- Adds opt-in supabase integration for deployments
  - The supabase project can be accessed [here](https://app.supabase.com/projects) (will need to accept email invitation to join Org first)
- New deployment `plh_asset_packs` for building and testing functionality
  - Includes an encrypted supabaseConfig file for connecting to supabase
- Adds 2 placeholder services. These services contain WIP methods to handle remote assets.
  - `remote-asset` service
    - Sketches out a method for downloading assets and populating them to local file storage
      - New template action to trigger this (initially useful for debugging), `asset_pack: download`, see template [debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit#gid=1422144003)
  - `file-manager` service for managing files on the device
    - Method to save a downloaded blob to local file storage
    - Preliminary method for generating src URI for local file

### Proposed supabase storage structure

- Organization: IDEMS
  - Project: plh (or broader still, e.g. "IDEMS Apps")
    - Storage
      - Bucket: plh_tz (deployment name)
        - Folder: asset-packs
          - Subfolder: theme_professional
            - Subfolder: language_tz_en
      - Bucket: e.g. wash

### Naming

We've been using "remote assets" to refer to the feature internally, but "asset packs" might be a more intelligible name for authors in general (although it doesn't encompass the fact that the assets are remotely hosted)?

#### Generating asset packs (TODO)
- [ ] Download assets from gdrive and bundle them together in "asset packs"
  - [ ] Generate manifest
- [ ] Upload bundle and manifest to supabase

#### Hosting asset packs (IN PROGRESS)
- [x] Set up supabase
- [x] Define structure of supabase Storage

#### Download asset packs (IN PROGRESS)
- Not needed for MVP?
  - [ ] Check config and download relevant manifest of expected files
  - [ ] Generate manifest of local files
  - [ ] Compare these to determine what files need to be downloaded 
- [x] Download required files from supabase to device
- [ ] Populate app with assets from downloaded bundle

## Git Issues

Closes #

## Screenshots/Videos


chrismclarke:
Looks like a solid start. I think to help avoid this PR growing too large/hard-to-review I might suggest freezing the scope pretty much where you have it (adding placeholder services/config) and try to get merged in, continuing follow-up work (download/sync) in another branch/PR - I think you may have already started on it

Let me know if there's anything you need to change on this, or if you think in a reasonable state for review as-is

.idems_app/deployments/plh_asset_packs/config.ts
```diff
@@ -0,0 +1,12 @@
+import { extendDeploymentConfig } from "scripts";
+import { supabaseConfig } from "./encrypted/supabaseConfig";
+
+const config = extendDeploymentConfig({ name: "plh_asset_packs", parent: "plh" });
+// override app constants here
+config.app_config.APP_HEADER_DEFAULTS.title = "plh_asset_packs";
+config.app_config.APP_SIDEMENU_DEFAULTS.title = "plh_asset_packs";
+
+config.app_config.ASSET_PACKS.enabled = true
+config.app_config.ASSET_PACKS.supabase = supabaseConfig
```
chrismclarke, line 10: nit(non-blocking)

As supabase would likely be used across multiple features in the future probably better to have its own config property higher up. If we really want to link here, could use `ASSET_PACKS.provider = 'supabase'` or similar if we expected to have support for multiple providers. But for now probably just stick to one and keep simple.

.idems_app/deployments/plh_asset_packs/config.ts
```diff
@@ -0,0 +1,12 @@
+import { extendDeploymentConfig } from "scripts";
+import { supabaseConfig } from "./encrypted/supabaseConfig";
```
chrismclarke, line 2: nit(non-blocking)

Really nice use of the encrypted configs, makes a lot of sense to pass the config into the deployment in this way, ensuring that the deployment will only compile if available.
It might also make sense to refactor the firebase config to behave in the same way (future issue)

The only issue with this is that we won't be able to swap to the deployment to decrypt the config initially, as you can only load a deployment that passes compilation. So we will need a workaround to either attempt decrypt before load, or make the import dynamic (e.g https://mariusschulz.com/blog/dynamic-import-expressions-in-typescript)

Can discuss on dev call

packages/data-models/appConfig.ts
```diff
@@ -192,6 +192,16 @@ const APP_UPDATES = {
   app_update_downloaded_field: "_app_update_downloaded",
 };
 
+type ISupabaseConfig = {
```
chrismclarke, line 195: nit(blocking)
As mentioned in other comment, better if we can move this to main deployment model instead

src/app/shared/services/file-manager/file-manager.service.ts
```diff
@@ -0,0 +1,51 @@
+import { Injectable } from "@angular/core";
+import { Directory, Filesystem } from "@capacitor/filesystem";
+import { Capacitor } from "@capacitor/core";
+import write_blob from "capacitor-blob-writer";
+
+@Injectable({
+  providedIn: "root",
+})
+// TODO: Refactor to use sync/asyncService base. Methods could also be moved to remote-asset service
+export class FileManagerService {
```
chrismclarke, line 10: nit(blocking)
This should extends either a sync or async base (as commented)

src/app/shared/services/remote-asset/remote-asset.service.ts
```diff
@@ -0,0 +1,109 @@
...
+import { AppConfigService } from "../app-config/app-config.service";
+import { IAppConfig } from "../../model";
+import { createClient, SupabaseClient } from "@supabase/supabase-js";
+import { FileManagerService } from "../file-manager/file-manager.service";
+import { TemplateActionRegistry } from "../../components/template/services/instance/template-action.registry";
+
+@Injectable({
+  providedIn: "root",
+})
+export class RemoteAssetService extends SyncServiceBase {
```
chrismclarke, line 12: nit(non-blocking)
I expect lots of work will be needed on this in the future so no need to focus on this for now. But as a rough outline makes sense to me, good to see remote asset management separate to file management services

supabase/config.toml
```diff
@@ -0,0 +1,72 @@
+# A string used to distinguish different Supabase projects on the same host. Defaults to the working
```
chrismclarke, line 1: nit(non-blocking)
I'm guessing the config is used as part of supabase cli? Does it need to be a top-level folder here? Also does it need to be a toml file or can we use ts/js?

chrismclarke (changes requested):
Really nice start, as mentioned in previous comment I think it would be good to focus on getting this merged in before building out the service methods much further. I've included a few minor comments inline, once resolved I think should be good to go

supabase/config.toml
jfmcquade: That's right. I've found the supabase docs to be pretty good for what they cover, but they are often missing examples or explanations of alternative ways of doing things. But it does seem like the file has to be .toml and that it has to be in a `supabase` folder at the top-level. It is automatically generated by running [supabase init](https://supabase.com/docs/reference/cli/supabase-init), and other supabase functionality fails if the .toml file is not found at that location. Further docs on the config file [here](https://supabase.com/docs/reference/cli/config).

Perhaps it would be desirable to not commit this folder, and have authors and devs run `supabase init` themselves as part of the set-up? Although if we end up extending supabase functionality and changing parts of the config from the default, then we would want the file to be committed.

chrismclarke: Given that this is only for interacting with CLI I don't expect authors will need to do anything with this folder. Nothing in this folder will be available to the runtime app, and is just for running a local server. I expect even developers running the app locally may choose to simply connect to an online dev project instead of running their own postgres server locally.

So I'd say probably safe to ignore and add developer docs if we find we want to support people running cli locally in the future

chrismclarke (approved):
Thanks @jfmcquade 
Following the call I've quickly added 91fb1f3 to the branch to move the supabase config to top level and update services accordingly. 

I also made some minor config changes to improve the error logging (which may become redundant) and fix an issue I had locally where the app config `ASSET_PACKS` property wasn't populating as expected (wasn't merging the default values, not sure why but will save debugging for another day)

Logging update - log now shows warning, deployment still compiles with warning
image

I can't start testing any of the functionality as I will need you to share the deployment decryption key, but only needed for future work (also I can just populate manually if I want to be less lazy)

But overall I think a nice base to have in and build from, good work!

PR #1875, opened 2023-04-05
fix: parent point item scaling

### LINKED ISSUE #1181: [BUG] ParentPoints page tile scaling
**Describe the bug**
The tiles on the ParentPoints page don't scale sensibly. 
- Currently it seems that the font size decreases infinitely. Instead there should be a minimal font size.
- Currently it seems that the minimum width for a tile is very small. Rather than keeping on shrinking the tiles to fit two side-by-side, the default behaviour should be to fall back to just having a single column. 
- Currently, once the app does decide to display all tiles in a single column, the tiles still take up only a small proportion of the width of the screen. Instead the tiles should possibly use the full screen width, to improve legibility. 

**App version**
v0.13.0

**Screen video**

https://user-images.githubusercontent.com/74557272/149977524-d478567c-7a3a-47dd-afd6-925b18294231.mp4

**Additional context**
This relates to issue #1179. Since this page is very central to the app I'm creating this issue dedicated to just improving this page.

chrismclarke:
PR Checklist

## Description

- Remove all scaling from parent box items. This is undesirable as only impacts devices under 420px and can make writing unreadable
- Prevent parent box items from overlapping

I've also removed the `- [ ] - Latest master branch merged` from pr checklist as no longer relevant and keep forgetting to remove

## Git Issues

Closes #1181

## Screenshots/Videos

[ParentApp (1).webm](https://user-images.githubusercontent.com/10515065/230062708-6e18dd91-b9ac-4ebb-994b-2686c95adceb.webm)


esmeetewinkel (changes requested):
Thanks, this resolves the issue of tiles overlapping and image/text shrinking. 

In some "reasonable" screen dimensions the images overlap the top-left star icon, the top-right question mark icon and the top border of the box. 
image

I think we could prevent the first two by not allowing the box to become this narrow, or all three by displaying the image a bit smaller. For reference, on the current master branch these images appear a lot smaller 

image

chrismclarke:
Increased the min width a bit (150px -> 175px), here's a screenshot at its most compressed

image

chrismclarke:
For reference - templates using `parent_points` style

Note - current changes only impact templatename specifically named `parent_points`
image

chrismclarke:
@esmeetewinkel - I can't remember if anything else is needed on this, I think the screenshot with named parent_point style was more for future reference of templates we may wish to tidy up (and not directly impacted by this PR), although I could be wrong on that...

esmeetewinkel (approved):
All good, thanks!

PR #1880, opened 2023-04-08
feat: add display grid component

### LINKED ISSUE #1877: [FEATURE] Display grid
**What?**
A `display_group` style, or its own component `display_grid`, that displays the components inside it in a grid. Would be good if some properties are exposed to the spreadsheet author, as for the `radio_button_grid`. This will make most sense when these components are similar (e.g. all buttons, all images of similar ratios).

**Why?**
Currently we can do this through the `parent_points` display group style, but this is quite specific and hard coded so it would be good to have a general way to deal with this.

chrismclarke:
PR Checklist

## Description
**Main Changes**
- Add a new `display_grid` component used to show items in a grid layout
- Add component demo [template](https://docs.google.com/spreadsheets/d/1ul3lknh0YGXtRrCjyqF7Wdq9OaQERhfTZTetVPYO_2M/edit?usp=sharing)

## Follow-up Tasks
- [ ] Add component documentation

## Review Notes
Exposes similar options to `radio_grid`, with additional border configuration and without ability to pass an answer_list. Instead child items are rendered according to the template, with `begin_data_grid` and `end_data_grid` tags used to wrap the items (see examples in component demo)

## Author Notes
In the future we may want to refactor some of the other custom grid templates like parents points to simply be templates nested within a display grid component (particularly as a means of testing whether the new component has enough flexibility to do so). But not a high priority

## Dev Notes
I've also included package updates for eslint and related packages as they haven't been updated in a while (no breaking changes expected). I've also included an opt-in rule to sort typescript literal types and imports alphabetically, and imposed on template components to make the code a little tidier and less prone to conflict.

I had thought about enabling the rules globally, however would require changes to a significant number of files and many of which have reasonable reason to not have things in alphabetical order, so instead will likely enable the rules just for template components once we have migrated to a standalone workspace

## Git Issues
Closes #1877

## Screenshots/Videos
**Demo component [template](https://docs.google.com/spreadsheets/d/1ul3lknh0YGXtRrCjyqF7Wdq9OaQERhfTZTetVPYO_2M/edit?usp=sharing)**

[ParentApp (4).webm](https://user-images.githubusercontent.com/10515065/230743359-25c16fad-99cf-4f05-b166-a111e7725e1f.webm)


esmeetewinkel (approved):
This is great, thanks. One question: Are you imagining the padding between the item border and its content to come from the content or from the display grid? Currently

- Text appears with hardly any padding on the left, quite a lot on the top and bottom. I think this is due to the built-in padding of the text component. How I would get the text not to touch the left item border?  
image

- Image padding seems to come from tmpl-image-container and looks reasonable. 
image

jfmcquade (approved):
Looks good. Another nice demo.

In response to Esmee's comment above, I would think it would be nice to offer a parameter to control padding around each item.

Re the dev notes, that all sounds good

chrismclarke:
> This is great, thanks. One question: Are you imagining the padding between the item border and its content to come from the content or from the display grid? Currently
> 
> * Text appears with hardly any padding on the left, quite a lot on the top and bottom. I think this is due to the built-in padding of the text component. How I would get the text not to touch the left item border?
> 
> image
> 
> * Image padding seems to come from tmpl-image-container and looks reasonable.
> 
> image

Really good question, this is always a bit of a tricky one because the more opinionated the component the higher the risk of accidentally overlapping multiple styles. E.g. applying a standard padding to items will then have the text top-bottom padding added in addition. Similarly for images, which I expect will never want grid padding applied.

So I'd say for now probably safer to keep the grid unopinionated, and instead make it easier to add more styles to things like display groups to do things like add padding, and/or horizontally/vertically align text inside.

Can discuss further on the call if useful

PR #1882, opened 2023-04-12
feat: add support for named deployment decrypt

chrismclarke:
PR Checklist

## Description
Currently deployments support use of encrypted files, however if these files are required in the config itself we have a catch-22 where a config will only load if decrypted files exist but files will only decrypt once the config is loaded.

This aims to fix in the following ways:
- Add an example of using import from encrypted file with fallback in case file does not exist
- Enforce deployment to be decrypted during set command (will decrypt after setting)
- Add support for deployment name in encrypt/decrypt command to allow decryption of non-active deployment

## Review Notes
- Ensure active deployment set as any that is not the `encryption_demo`
- Run `yarn workflow deployment set encryption_demo`

This should decrypt the set the deployment, decrypt, and update the config with decrypted values

If missing the private key for the encryption_demo deployment a copy can be found here (link expires in one week):
https://onetimesecret.com/secret/knr56pxznossi8oqf4x2uxaq58gdkni

## Git Issues
Issue originally identified in review of #1873

## Screenshots/Videos
Updated config with notes
image

Example config loaded with encrypted values populated to `error_logging`
image


chrismclarke:
@jfmcquade 
Following the call I've fixed the auto-decrypt following deployment set, so now all should be working as expected to decrypt the config whenever setting a deployment that has encrypted values.

I also double-checked what happens in cases where the encrypted value is updated, and as checking depends on file mtime of both encrypted/decrypted version syncing a new encrypted version should ensure the values are updated next time deployment is set

chrismclarke:
Thanks @jfmcquade , I think the example you raise is probably sufficiently enough an edge-case that we can proceed without (expect further work needed in the future more generally, like recompiling when parent config changed).

I'm going to go ahead and merge as I don' think any additional testing is needed, but feel free to let me know @esmeetewinkel if you have any questions around this

esmeetewinkel:
@chrismclarke I'm struggling to set my deployment to `plh_tz` (normally I'm using `plh_global`) and I'm guessing it's related to this PR

image

The link to the private key of encryption demo is not visible to me (although not yet a week old)
image

I've tried deleting the `config.json` of both `plh_tz` and `encryption_demo`.

chrismclarke:
Very strange, so you currently have the encryption_demo deployment loaded and it won't let you change back to plh_tz?
What is the error you see?

I'll share the private key on mattermost, otherwise deleting the `.idems_app\deployments\activeDeployment.json` file should help revert things if somehow the active deployment got into a bad state

esmeetewinkel:
No. I had `plh_global` loaded and could not change to `plh_tz` nor `encryption_demo` because both were showing as encrypted and I did not have access to the private key. Using the private key you sent me on mattermost I'm now able to access `plh_tz` nor `encryption_demo` deployments. Thanks!

PR #1893, opened 2023-04-21
feat(server): deployment databases

chrismclarke:
PR Checklist

- [x] PR title descriptive (can be used in release notes)

## Description
Currently, all server API operations write to a single database, such as syncing the list of user contact fields. This is not ideal for multiple reasons, such as data tables get long, potentially sensitive information is available across projects, and also data may be lost in cases where a user across multiple apps retains the same ID (as in #1865)

This PR addresses this issue in the following ways

**App**
- [x] Add deployment config setting to allow a deployment to specify target db for api operations
- [x] Add interceptor to pass target db with all api requests

**Server**
- [x] Update core dependencies 
- [x] Add middleware to toggle active db connection per request
- [x] Add scripts to automatically create dbs on demand
- [x] Update all api endpoints to support specifying deployment db name
- [ ] Consider refactoring all endpoints to extend a common base to reduce boilerplate code

**Testing**
- [x] Add lint checks for API workspace as part of Github Actions
- [x] Add docs to support manual testing and development
- [x] Add debug deployment to test server db implementation
- [x] Add demo video to show how to manully test api endpoints
- [ ] Create E2E infrastructure to allow testing of db methods
- [ ] Add tests for core use-cases

**Deployment**
- [x] Deploy to server (will need to be done manually)

## Review Notes
A debug deployment has been created to test implementation
```
yarn workflow deployment set debug_db
```

Run both server api and main app in separate terminals
Assumes api configured as per **API Development** section of the documentation added in this PR
```
yarn workspace api start
```
```
yarn start
```

Once the server is running requests can be sent either from a frontend template (e.g. `http://localhost:4200/template/user_code`, or using the interactive api http://localhost:3000

**Advanced use cases (for manual or automated tests)**
- Triggered hooks, e.g. contacts field `raw` and `_updated` columns gets populated as expected
- Multiple requests for different databases come in at the same time
- Default settings continue to work as expected (i.e. all deployments sync correctly to plh db unless config updated)

## Dev Notes
This PR ended up having to be quite big due to structural changes required to allow NestJS to interact with multiple databases as part of a single service. By default NestJS expects you to register a database connection during initialisation, and this same connection is injected into the context of all requests (using decorators such as `InjectModel`). As this is only ever done once, it prevents us from dynamically switching to target operations onto a different DB depending on the incoming request. 

[...]

PR #1902, opened 2023-05-03
fix: workflow-options

chrismclarke:
PR Checklist

## Description

While working on #1893 discovered content-watch mode no longer working as expected (possibly due to dependency upgrades or workflow name changes). Options set at the top level `sync` command were not filtering down to child commands

## Review Notes
Running `yarn workflow sync --content-watch` should now listen for keep connection alive and update on content updates

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade (approved):
Looks good, and `yarn workflow sync --content-watch` seems to be working for me from a quick functional test

PR #1910, opened 2023-05-09
Feat/flow generator

chrismclarke:
PR Checklist

## Description

- [x] Add support for `generator` flow type as defined in [RFC](https://docs.google.com/document/d/1cK_Mk3nTZIKxux8bygKvujUFkVENMOE_xIgM6w9PlRw/edit#)
- [x] Add parser tests
- [ ] Add example sheets
- [ ] Add documentation (should migrate across RFC content)

**Additional Changes**
As generator flows produce additional flows to populate to main list when parsing (as opposed to just modifying themselves), additional methods were required for populating these. A similar method already existed for handling this case in data pipes, so this was refactored into a general method used by both data_pipe and generator flow types. Otherwise all other flow type parsers should remain unchanged

## Review Notes
The PR contains parser changes and has updated cache version accordingly. Recommend testing when content recently updated to make it easier to detect any knock-on changes. It only changes expected should be generated flows populated.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/data-models/flowTypes.ts
+++ b/packages/data-models/flowTypes.ts
@@ -14,6 +14,8 @@ export namespace FlowTypes {
     | "data_list"
     // data_pipes are used to modify or generate new data_lists via processing methods
     | "data_pipe"
+    // generators can create any other flow type using a source datalist and flow skeleton
+    | "generator"
     // global data provides data to other modules, without namespacing (all top-level)
     | "global"
     | "template"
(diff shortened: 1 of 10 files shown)


esmeetewinkel (approved):
Functional test passed. I've added some examples for generating data lists and calling that in items loops the generated templates, all working as expected. 

This is great!

Issue #1912, opened 2023-05-10
[FEATURE] In-app PDF viewer


esmeetewinkel:
**What?**
A template component that can display assets in `.pdf` format. Desired functionalities 
- [ ] Display pdf
- [ ] Zoom (ideally pinch, +/- buttons also OK, or both)
- [ ] Scroll
- [ ] #1951

**Why?**
"Parenting for Respectability" manuals for facilitators exist in `.pdf` format and are intended to be disseminated through the app. Current manual is ~130 pages but will be subdivided into smaller files of ~8 pages.

esmeetewinkel:
SLACK thread: https://chat.idems.international/all/pl/mkyxmgscr3fj8x5ip4we6w4jee

esmeetewinkel:
Closed by #1913

PR #1913, opened 2023-05-10
feat: pdf viewer component

jfmcquade:
PR Checklist

## Description

Adds a pdf viewer component, with the type identifier `pdf`.

Component demo sheet is here: [comp_pdf](https://docs.google.com/spreadsheets/d/1VXM9zYgrsZIB4h8slC9mg3P9C-jsXuPScNaml-1qaOk/edit#gid=569531329). NB, the [example pdf file](https://drive.google.com/file/d/14INz6c0AVPgApAsVh1ZLPHP3s-XWKIOM/view?usp=sharing) should be removed from the repo before merging to avoid the unnecessary extra package size.

## Known Issues
- The pdf component is not currently customisable
  - I've stripped the pdf viewer down to its basic functionality, but it does support a lot of options that could be exposed to template authors in future if desired.
- The pdf component has a set height
  - It is difficult to adjust the total height of the pdf viewer element dynamically, so for the time being this is hardcoded to "80%"
- Increase in package size
  - The size of the [pdf viewer library](https://github.com/stephanrauh/ngx-extended-pdf-viewer) is not insubstantial
  - The suggestion [here](https://pdfviewer.net/extended-pdf-viewer/intro) is that it is somewhere between 0.5-3MB, however I'm not sure what the actual figure is for our implementation

## Git Issues

Makes a start on #1912 (meets the first 3 requirements). As discussed on Mattermost, opening a file within the app is more involved, but should have some shared logic with the [remote assets](https://github.com/IDEMSInternational/parenting-app-ui/milestone/4) feature.

## Screenshots/Videos
[comp_pdf](https://docs.google.com/spreadsheets/d/1VXM9zYgrsZIB4h8slC9mg3P9C-jsXuPScNaml-1qaOk/edit#gid=569531329) template
Screenshot 2023-05-10 at 16 47 06

Screenshot 2023-05-10 at 16 47 14

Screen recording of pdf component running in local emulation, including pinch-to-zoom:

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/cdf985b6-23fb-4c05-8a2c-6acd8a3f1696


esmeetewinkel (commented):
Thanks @jfmcquade for getting this in so quickly. I added a second dummy pdf to the comp_pdf template for testing, and it seems that the component is a bit buggy when there is multiple pdf rows on the page:

image

I'm now seeing the dummy.pdf file in the first `pdf` component, and no file at all in the second `pdf` component
image
image

Not the most common use case, I imagine we generally won't have two pdf viewers on the same page

package.json
```diff
@@ -90,6 +90,7 @@
     "marked": "^2.0.1",
     "ng2-nouislider": "^1.8.3",
     "ngx-color": "^6.1.0",
+    "ngx-extended-pdf-viewer": "17.0.0-alpha.4",
```
chrismclarke, line 93: nit(non-blocking)
Any particular reason to use v17 which is in alpha as opposed to v16 stable?

src/app/shared/components/template/components/pdf/pdf.component.html
```diff
@@ -0,0 +1,23 @@
+<div class="pdf-viewer">
+  <ngx-extended-pdf-viewer
+    [src]="_row.value || src | plhAsset"
```
chrismclarke, line 3: nit(blocking)
I'm not sure of the order of operations here, probably better to resolve the source in the component ts and just have a single possible value

src/global.scss
```diff
@@ -167,3 +167,12 @@ $tour-next-button-background: var(
     border: none;
   }
 }
+
+ngx-extended-pdf-viewer {
```
chrismclarke, line 171: nit(blocking)
I see some styles applied at the component level and some at the global level - do we need these both? Can we put everything in the component styles or do we need to override some over style applied somewhere?

chrismclarke (changes requested):
Nice to see this implemented so quickly. See a few mostly minor comments inline, a couple other considerations:

**External URLs** - see no reason to try support these right now, the app should be fully functional offline so if we want external URLs will want to combine with remote asset system to download locally anyway. If really required we could use a cloud function to help bypass cors (e.g. https://github.com/Rob--W/cors-anywhere), however wouldn't recommend for now

**Parameters** - not sure if we really want to use both row value and parameter to define source, makes sense to just keep it as value (which would always be required) and then leave parameters for any optional overrides we choose to expose in the future

**Demo** - given there's a web preview for this branch would be nice to include the component demo with it.

**Size** - we have the `yarn analyse` command which builds the app along with webpack bundle analyzer to show sizes. Looks like the core is coming in at around 350kb which is reasonable but then the assets folder you've copied over is a massive 40mb! I'd probably work to reduce that before we consider merging

image

angular.json
```diff
@@ -35,6 +35,11 @@
                 "glob": "**/*.svg",
                 "input": "node_modules/ionicons/dist/ionicons/svg",
                 "output": "./svg"
+              },
+              {
+                "glob": "**/*",
+                "input": "node_modules/ngx-extended-pdf-viewer/assets/",
```
chrismclarke, line 41: Do we need all these assets? The folder is a massive 40MB and has many duplicate and unnecessary files (e.g. both .js and .min.js files, map files etc.)

chrismclarke: Also I noticed you can define the assets folder name in the using the `assetsFolder` option (https://pdfviewer.net/extended-pdf-viewer/options) - might be nice to bundle all these into their own folder instead of the main assets in case of any future filename conflicts

src/global.scss
jfmcquade: These styles are not applied when put in the component styles file unfortunately. I'm not entirely sure why, but the `!important` flag is also required for them to take effect, about which there is a small note [here](https://pdfviewer.net/extended-pdf-viewer/smartphone), under "_Caveat_"

jfmcquade:
@esmeetewinkel Good find, it's just a limitation of the pdf viewer library that we can't currently have 2 pdf components on the same page. There is a [workaround](https://pdfviewer.net/extended-pdf-viewer/side-by-side), but I don't think it's worth worrying about unless we have a specific use case. I've added a note to the docs and the PR description.

Thanks, @chrismclarke, I've addressed the points in your review, and added some documentation.

As for the package size, I that I got an error in my browser that caused the pdf not to display if I didn't include the `*-es5.min.js` files from the library's assets folder – you set them to be ignored in #1918 but I reversed that with [this commit](https://github.com/IDEMSInternational/parenting-app-ui/pull/1913/commits/911b7035ead194ca0dc3681228c6fe3a18d656fa).
ignored

Can you see a way around this whilst offering support for common browsers? Otherwise it means that the total size added to the core code by this component is around 8MB (~350MB of code from ngx-extended-pdf-viewer plus ~7.5MB copied to the `comp-pdf` assets folder), which more than doubles the existing core size. This does seem like a lot to add for a single component (which we don't have an immediate use for). It may be that we want to avoid merging this PR until we have a system for deployments to opt-in/out to including certain components, either manually or automatically. That being said, 8MB doesn't seem objectively like a huge amount to add.

chrismclarke:
@jfmcquade - what browser are you using?
I still think probably worth dropping the es5 files as shouldn't be required for any app users with a reasonably up-to-date version of chrome (which has been our main target group so far)

jfmcquade:
@chrismclarke Ah, I found the problem, quite interesting. I'm using the latest version of Chrome, but by copying over the `needsES5()` code, I found that my browser was failing the `isIOs13OrBelow()`. Turns out I was always using the iPhone SE device option in Chrome DevTools 🤦

Changing this or closing DevTools fixes the issue. I wasn't previously aware that changing the device did anything other that set the page dimensions to match that device's screen, and I usually use the iPhone SE option as a default small screen size (I'll use an Android phone from now on).
Screenshot 2023-05-18 at 13 31 24

In terms of running our own check of whether ES5 is required, it's quite a chunk of code (see below, also requires the [trusted-types](https://www.npmjs.com/package/trusted-types) library), but it could be useful as a general utility if we don't have anything else similar, what do you think?

```
  private async needsES5(): Promise<boolean> {
    if (typeof window === 'undefined') {
      // server-side rendering
      return false;
    }
    const isIE = !!(<any>window).MSInputMethodContext && !!(<any>document).documentMode;
    const isEdge = /Edge\/\d./i.test(navigator.userAgent);
    const isIOs13OrBelow = this.iOSVersionRequiresES5();
    let needsES5 = typeof ReadableStream === 'undefined' || typeof Promise['allSettled'] === 'undefined';
    if (needsES5 || isIE || isEdge || isIOs13OrBelow) {
      return true;
    }
    return !(await this.supportsOptionalChaining());
  }

  private iOSVersionRequiresES5(): boolean {
    if (typeof window === 'undefined') {
      // server-side rendering
      return false;
    }
    const match = navigator.appVersion.match(/OS (\d+)_(\d+)_?(\d+)?/);
    if (match !== undefined && match !== null) {
      return parseInt(match[1], 10) < 14;
    }

    return false;
  }

  private supportsOptionalChaining(): Promise<boolean> {
    return new Promise((resolve) => {
      const support = (<any>window).supportsOptionalChaining;
      support !== undefined ? resolve(support) : resolve(this.addScriptOpChainingSupport());
    });
  }

  private addScriptOpChainingSupport(): Promise<boolean> {
    return new Promise((resolve) => {
      const script = this.createScriptElement(pdfDefaultOptions.assetsFolder + '/op-chaining-support.js');
      script.onload = () => {
        script.remove();
        resolve((<any>window).supportsOptionalChaining as boolean);
      };
      script.onerror = () => {
        script.remove();
        (<any>window).supportsOptionalChaining = false;
        resolve(false);
      };

      document.body.appendChild(script);
    });
  }

  private createScriptElement(sourcePath: string): HTMLScriptElement {
    const script = document.createElement('script');
    script.async = true;
    script.type = 'text/javascript';
    const ttWindow = window as unknown as TrustedTypesWindow;
    if (ttWindow.trustedTypes) {
      const sanitizer = ttWindow.trustedTypes.createPolicy('foo', {
        createScriptURL: (input) => input,
      });
      script.src = sanitizer.createScriptURL(this.location.normalize(sourcePath)) as any;
    } else {
      script.src = this.location.normalize(sourcePath);
    }

    return script;
  }
```

chrismclarke:
> @chrismclarke Ah, I found the problem. I'm using the latest version of Chrome, but by copying over the `needsES5()` code, I found that my browser was failing the `isIOs13OrBelow()`. Turns out I was always using the iPhone SE device option in Chrome DevTools 🤦
> 
Ah nice, yes it's a bit of a weird quirk, the device manager can be used to spoof user-agent and a bunch of other properties that are sometimes (but not always) used for OS detection, so sometimes the ios devices will detect and other times won't.

> In terms of running our own check of whether ES5 is required, it's quite a chunk of code (see below, also requires the [trusted-types](https://www.npmjs.com/package/trusted-types) library), but it could be useful as a general utility if we don't have anything else similar, what do you think?
> 
The majority of the custom code is to check for optional chaining, which seems to stem from the stack overflow discussion here: 
https://stackoverflow.com/questions/66288922/check-if-optional-chaining-is-supported

I'd probably drop all the dynamic script code in favor of the accepted simpler script. This should hopefully also mean not needing trusted-types. I'd also probably drop the ios/edge/ie checks (maybe leave a comment to this thread), as really we're focused on chrome-based mobile browsers (would never get as far as rendering a component in IE, would likely have a host of other issues in ios and I expect edge now supports same since move to chromium)

jfmcquade:
> I'd probably drop all the dynamic script code in favor of the accepted simpler script. This should hopefully also mean not needing trusted-types. I'd also probably drop the ios/edge/ie checks (maybe leave a comment to this thread), as really we're focused on chrome-based mobile browsers (would never get as far as rendering a component in IE, would likely have a host of other issues in ios and I expect edge now supports same since move to chromium)

Thanks for the pointers, @chrismclarke, I can't re-request your review but it should be ready

src/global.scss
chrismclarke: I think a more elegant solution might just be to remove the shadow dom encapsulation so deep selectors not required. I'll test locally and add a commit

chrismclarke (approved):
Looks great! I think the legacy check looks sensible to me and overall code in a good shape

I've added a minor commit to bump the bundle size estimation (6abd8ba) - by my calculations the feature weights in at about 4.5mb total, which feels fairly reasonable given the scope of the component (huge amount of work goes into rendering pdfs). 

I've also refactored the global styles to work from within the component (removing view encapsulation) (c03e977), but otherwise I'd say good to go

esmeetewinkel (approved):
Functional test passed, happy with the error message set through the parameter to make it translatable.

PR #1874, opened 2023-04-05
Feat/assets theme lang alt

chrismclarke:
PR Checklist

- [ ] - Latest `master` branch merged
- [ ] - PR title descriptive (can be used in release notes)

## Follow-up tasks (to create issue)
- [ ] Update documentation
- [ ] Sync content (will update assets contents)

## Description
Builds from #1795

> This PR adds the functionality to override assets with a variation relating to the current theme and language, including a combination thereof (e.g. a translated version of an asset that differs based on theme).
> 
> If a variation of the requested asset for the current theme and language does not exist, then a different version of the asset will be returned as a fallback. The order of priority for these fallbacks is:
> 1. current theme, current language
> 2. default theme, current language
> 3. current theme, default language
> 4. default theme, default language

**Main Changes**
- Treat assets in top-level folders as global, even if no named global folder
- Provide more flexibility for theme and language folder nesting.  Now language or theme folders can be placed in any directories and used to provide overrides at the current level. E.g. to override for both theme and lang

**Source**
`global/nested/img.jpg`

**Before**
`theme_professional/tz_ke/nested/img.jpg`

**After**
`theme_professional/tz_ke/nested/img.jpg`
`nested/tz_ke/theme_professional/img.jpg`
`nested/theme_professional/tz_ke/img.jpg`
`tz_ke/theme_professional/nested/img.jpg`

Or any path that will resolve to the same filename (ignoring global prefix) `nested/img.jpg`, once the theme and language segments of the path have been removed

**Additional Changes**
- Refactor tests to use more explicit per-test file and folder structures
- Track relative filepaths in override contents to support more varieties of folder structure

## Review Notes
Will need to sync content to populate assets in updated format and view changes

## Dev Notes
Formatting of asset contents slightly changed, now default theme/language not nested, `themeVariations` renamed to `overrides`, and additional filepath included for assets that sit in a location not identical to the entry name (e.g. language/global/override assets)

image

## Git Issues
Closes #1795
Closes #1798

## Screenshots/Videos
**Demo**

[ParentApp (2).webm](https://user-images.githubusercontent.com/10515065/230375577-dc3a39db-9d22-4c2c-b08c-1baa56649208.webm)

PR #1920, opened 2023-05-16
fix: log of asset size breakdown

jfmcquade:
PR Checklist

## Description

Fixes an issue where the asset sizes logged when running `yarn workflow sync` were incorrect. I believe this should have been picked up when testing #1874. 

## Testing

1. Checkout master, and set the active deployment to plh_global – `yarn workflow deployment set plh_global`
2. Run `yarn workflow sync_assets` and note the log of the asset size breakdown. It should be similar to the following screenshot. Note that, as well as "theme_default theme", there is an extra entry for "default theme" which will always display 1MB.
    Screenshot 2023-05-16 at 12 54 39

3. Now checkout the feature branch, `git checkout fix/asset-size-breakdown`
4. Run `yarn workflow sync_assets` again, and you should see the following. Note that the default global assets are now being included correctly in the total size.
    Screenshot 2023-05-16 at 12 54 06

## Git Issues

Closes #

## Screenshots/Videos
The asset size breakdown logs for the `plh_global` deployment:

Before
Screenshot 2023-05-16 at 12 54 39

After
Screenshot 2023-05-16 at 12 54 06

### DIFF
--- a/packages/scripts/src/commands/app-data/postProcess/assets.ts
+++ b/packages/scripts/src/commands/app-data/postProcess/assets.ts
@@ -164,10 +164,13 @@ export class AssetsPostProcessor {
 
   private checkTotalAssetSize(sourceAssets: IAssetEntriesByType) {
     let totalSize = 0;
-    let themeAndLanguageSizes = { default: { total: 0, global: 0 } };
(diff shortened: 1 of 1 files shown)


chrismclarke (approved):
Nice catch, looks good to me

Issue #1316, opened 2022-04-05
[FEATURE] - Remote asset sync


chrismclarke:
**What?**
The current app bundle is approaching size limits that surpass play store limit (150MB) and forces users to download a large number of assets that they may never need (e.g. translated audios)

We need to improve on this system by:
1. Explore use of [android asset delivery](https://developer.android.com/guide/playcore/asset-delivery) as a means to distribute bundles larger than 150MB if required. 

2. Implement a system that allows users to download resources direct from a server and integrate into the app

**How?**
For task (1) that should be a more simple task of exploring the docs to see what is possible and propose how we might integrate into existing pipelines. This is not a preferred option but may be something quick that can be achieved in the short term. Longer term it may still have some use to distinguish between code that could be updated OTA and/or assets that can be optimised to not repeatedly download

For task (2) likely will involve
- Tracking external download URLs for assets (either from google drive or copied to another s3 storage/server equivalent)
- Possible server updates to allow static file storage (e.g. minio)
- Scripts update to summarise _asset packs_ that may need to be downloaded to support a specific deployment/language
- Capacitor File plugin use to detect what files a user has downloaded and provide means to download and extract assets from the server as required
- UI pages to manage download
- Possible future updates to allow sharing of asset packs via Bluetooth or similar app sharing means

**Other details**
_Add any extra details that you think may be useful._

esmeetewinkel:
Timeline: needed before June 2023 for **ParentApp for Teens**

jfmcquade:
Notes taken from #1989 
## TODO/needs discussion
### `asset_pack`s as manifests for downloading files
- When determining which files to download from the remote server, an `asset_pack` is used as a manifest
- It was agreed that for an MVP, we don't necessarily need to worry about managing asset packs from an author's perspective
- Currently this PR supports downloading a complete set of assets that is explicitly specified in an `asset_pack` json file
	- This is triggered through a template action, e.g. `asset_pack: download: asset_pack_name`, where an `asset_pack_name.json` exists in the appropriate folder in supabase storage
	<img width="795" alt="Screenshot 2023-08-08 at 13 16 20" src="https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/d4b37096-7c19-431a-861f-b1534b049f38">

- We need to think about how these asset packs are generated and consumed
	- The asset pack generation could work something like this:
[...]

PR #1923, opened 2023-05-22
feat(remote assets) – add download methods; handle displaying assets from supabase on web and native

jfmcquade:
PR Checklist

## Description

Adds download methods for handling remote assets.

When remote assets are enabled (supabase enabled and ASSET_PACKS.enabled), assets sources can be handled in a new way, dependent on platform:
- On web, the src is an external URL (supabase CDN)
- On native, the src is a path to the downloaded file within the local filesystem

Debug template: [debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit#gid=1422144003)

Also removes the deployment folder for the `plh_asset_packs` deployment. This folder is no longer needed, as the asset packs/remote asset feautures can be tested with the new debug deployment. Follows from [here](https://github.com/IDEMSInternational/parenting-app-ui/pull/1955).

## Dev Notes
- Supabase buckets which are public do not support the SDK download method – and that download method does not support download progress. So I've used methods from the [picsa service](https://github.com/e-picsa/picsa-apps/blob/main/libs/shared/src/services/native/storage-service.ts) to download from a supabase url as if it was an arbitrary url.
- I think the code can be a bit confusing to read in terms of tracking the differences between filepath/relativePath/URI/URL/assetName etc. Let me know if you can see ways to make this clearer/more consistent.

## Git Issues

Work towards #1316 

## Screenshots/Videos

Web demo ("downloaded" asset points to external supabase URL)

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/dc80ba72-7eb8-4f23-b9d1-384a2686bde2

Emulator demo (downloaded asset is read from internal filesystem)

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/6e074a82-4392-4c9b-8086-949133e7ad73


src/app/shared/components/template/services/template-asset.service.ts
```diff
@@ -74,6 +78,25 @@ export class TemplateAssetService extends AsyncServiceBase {
     return value;
   }
 
+  private getFilePath(assetName, contentsEntry: IContentsEntryMinimal | Partial<IAssetEntry>) {
+    if (environment.deploymentConfig.supabase.enabled) {
```
jfmcquade, line 82: As well as `environment.deploymentConfig.supabase.enabled`, this should also be checking `appConfig.ASSET_PACKS.enabled`.
The separation makes sense as we move forward I think (we may want to expand supabase integration soon to cover functionality other than remote assets, e.g. auth).
So do we check both (requiring a subscription to `appConfigService.appConfig$`)? We're unlikely to want to change remoteAssetsEnabled at runtime, so it could just be another setting in environment.deploymentConfig?

packages/data-models/appConfig.ts
```diff
@@ -195,7 +195,7 @@ const APP_UPDATES = {
 const ASSET_PACKS = {
   enabled: false,
   bucketName: "deployment_name",
```
chrismclarke, line 197: nit(non-blocking)
Probably better to leave this as an empty string so it can pass type-checks but also trigger a future warning if user has not specified

```suggestion
  bucketName: "",
```

packages/scripts/src/utils/file-utils.ts
```diff
@@ -173,6 +173,12 @@ export interface IContentsEntry {
   modifiedTime: string;
   md5Checksum: string;
   localPath?: string;
+  /** specific path to file when not the same as relativePath, e.g. asset overrides */
+  filePath?: string;
+  /** External URL to display remote assets when running in browser */
+  url?: string;
+  /** Path to file in local device filesystem, populated after asset download */
+  cachedFilepath?: string;
```
chrismclarke, line 181: nit(blocking)
This contents entry is used quite a few other places, if extending for the remote-asset feature probably better to create a new interface entirely (could extend if useful or not)

Also I'm not sure if I fully understand the reason to have filePath or cachedFilePath properties (would they not both  be covered by localPath?)

packages/data-models/deployment.model.ts
```diff
@@ -127,8 +127,13 @@ interface IContentsEntry {
...
   modifiedTime: string;
   md5Checksum: string;
-  /** specific path to file when same as relativePath, e.g. asset overrides */
+  localPath?: string;
+  /** specific path to file when not the same as relativePath, e.g. asset overrides */
   filePath?: string;
+  /** External URL to display remote assets when running in browser */
+  url?: string;
+  /** Path to file in local device filesystem, populated after asset download */
+  cachedFilepath?: string;
```
chrismclarke, line 136: nit(blocking)
Same comment as file-utils, recommend creating interface in remote-assets feature code instead of changing here

src/app/shared/components/template/services/template-asset.service.ts
chrismclarke: nit(blocking)
Agreed asset pack enabling doesn't make much sense to toggle during runtime so let's assume no subscriptions required. 

I think better than relying on checking supabase setting here we should probably expose a method to update/alter the `assetsContentList` and do so following download (linked to remote-asset system), so that this service doesn't need any awareness of supabase or cached file paths

chrismclarke (changes requested):
Great to see this moving forwards. I think given the number of moving pieces here it might take a couple iterations to get it right. I've tried to give a rough once over to all the code so far and added various comments in-line. Let me know if you think anything wants further discussion

Also a couple additional comments:

*Download Method*
Make sense to use http sevice instead of sdk, gives us more flexibility anyways for working with different providers. The only issue you'll see inline is how to ensure waiting for request completion within action

*File Manifest*
Appreciate this needs a bit more thought, but probably best as a follow-up as the PR already quite substantial. I have a feeling it will be better to avoid hardcoded json files and maybe use the existing dynamicData service utility to start with the json list as a base and add a write-layer on top to track downloaded and/or progress. The benefit with using rxdb would also simplify subscriptions to download progress

*Additional Thoughts*
As updating download progress within a button component is a bit tricky it might be nice to create a new component that minimally subscribes to the remote asset service and shows download progress?

packages/scripts/src/utils/file-utils.ts
jfmcquade: Good point. Instead of adding these extra properties, I've made use of the existing `filePath` property (introduced in [this PR](https://github.com/IDEMSInternational/parenting-app-ui/pull/1874/files#diff-37887cafe4009287e88b297e43b425cb7f04342b1f04689cc6e6310b27b76902)). 

There is a divergence between `IContentsEntry` as defined in `packages/scripts/src/utils/file-utils.ts` (node environment) and `packages/data-models/deployment.model.ts` (runtime environment): the former has the `localPath` property, which refers to the path to the file in the dev's local repo (I think), whereas the latter has the `filePath` property, which gives the path to the file within the app bundle folders, to be read at runtime to get the actual path to the file (this path contains any theme and language segments, unlike the `assetName`).
It's a bit confusing, but as these are different properties I think it makes sense to keep them separate – but then should `IContentsEntry` have both properties in both locations even though each property will only be used in its respective environment?

In terms of this PR, I'm now making use of the `filePath` property to store _either_ the external supabase URL or the internal Android filesystem path (e.g. `file://` etc), depending on the runtime platform. Then the template asset service doesn't need to be aware of the platform and reads the same property regardless. But it is a bit confusing, especially as you might expect `localPath` to store the internal Android filepath, but as mentioned, this refers to the local path on the dev's machine when running content syncs etc.

jfmcquade:
Thanks @chrismclarke, I've acted on your comments, and responded to one inline.

I really like the idea of using the dynamic data service (rxdb) to handle updating the asset contents list. As you say, this seems to be a good use case – adding user writes over the top of a core JSON. The dynamic data service will need a bit of tweaking to handle contents files (it currently only handles flows). I think it makes sense to leave that for the next incremental PR, along with looking a bit more at handling the manifests for downloading the files generally.

packages/scripts/src/utils/file-utils.ts
chrismclarke: Ah, yeah that's a little on the messy side, no reason to have both `filePath` and `localPath` but don't really like the idea of refactoring as part of this PR. I still expect that longer term we will want to simply extend either of those definitions to add fields to track remote status (e.g. `download_progress`, `download_location` etc.), but fine to keep as-is for now

chrismclarke (approved):
Not sure whether it was part of the refactor, but for me the demo is no longer working. But otherwise happy if you want to just merge and make incremental progress as I can see the main blocking issues have all been addressed

jfmcquade:
> Not sure whether it was part of the refactor, but for me the demo is no longer working. But otherwise happy if you want to just merge and make incremental progress as I can see the main blocking issues have all been addressed

I think this was due to the [debug sheet](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit#gid=1422144003) moving to the dedicated content repo, which didn't have the same filepaths to the debug assets. This is fixed in [this PR](https://github.com/IDEMSInternational/app-debug-content/pull/18) on the content repo, so after that's merged then the demo should work again on this remote assets PR.

There's also the issue that supabase project "pauses" after a week of inactivity. Have you found a way to handle this other than manually logging into the console regularly?

jfmcquade:
> Not sure whether it was part of the refactor, but for me the demo is no longer working. But otherwise happy if you want to just merge and make incremental progress as I can see the main blocking issues have all been addressed

The demo is now working again, using the debug deployment, after [this PR](https://github.com/IDEMSInternational/app-debug-content/pull/18) was merged on the debug deployment repo. I'm going to merge this PR now to continue working on the remote assets features.

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/07f8e864-cc73-4a6b-b2c8-01ad4761c607

In order to update the debug deployment and test this feature:
> Run yarn workflow deployment set debug to refresh the remote content (i.e. pull the latest from the main branch of the debug repo). You should get the message in the attached screenshot. You need to add a private.key file to decrypt the config: this is the same as is required by the plh_asset_packs deployment, so if you already have the file .idems_app/deployments/plh_asset_packs/encrypted/private.key, you can copy it over to .idems_app/deployments/debug/encrypted/private.key. Otherwise, let me know and I will send you the private.key file. After populating that file, run yarn workflow deployment set debug to decrypt the relevant files.
> Run `yarn workflow sync` to sync the latest content (not required if [this PR](https://github.com/IDEMSInternational/app-debug-content/pull/20) has been merged)

chrismclarke:
> There's also the issue that supabase project "pauses" after a week of inactivity. Have you found a way to handle this other than manually logging into the console regularly?

What is defined as _activity_? Would it be possible to create a basic edge function and use something like uptimerobot to ping it once a day?

chrismclarke:
@jfmcquade 
What's the status on this PR - I approved a while back but keep getting pinged with updates and start to re-review before realising I already have... 

From this iteration I noticed the debug template not working
```
error-handler.service.ts:63 ERROR Error: Uncaught (in promise): Error: No handler for action
asset_pack : download
Error: No handler for action
asset_pack : download
```
So if you are planning to continue working here could you please unassign me and convert to draft? Otherwise I'd say better to merge what is ready and fix up demo issues in a follow-up, as otherwise additional commits added could be merged without review

jfmcquade:
I've merged now, I meant to do so straight after [this PR on the debug repo](https://github.com/IDEMSInternational/app-debug-content/pull/18) was merged.

> From this iteration I noticed the debug template not working
> 
> ```
> error-handler.service.ts:63 ERROR Error: Uncaught (in promise): Error: No handler for action
> asset_pack : download
> Error: No handler for action
> asset_pack : download
> ```

That is the error that will be displayed currently when the active deployment does not set `config.supabase.enabled = true` (the action handlers are only registered in the case that this is true). I was getting this error too, but I noticed that my `.idems_app/deployments/activeDeployment.json` wasn't reflecting the supabase settings defined in `.idems_app/deployments/debug/config.ts`, so forced a recompile by deleting `.idems_app/deployments/debug/config.json` and re-setting the deployment, which fixed the issue. I suspect this is an issue with the deployment compilation, where decrypting the encrypted folder doesn't prompt the config to recompile since the config file itself has not changed. Definitely feels like an edge case, but maybe worth noting in case it crops up again.

PR #1941, opened 2023-06-09
Feat: share

### LINKED ISSUE #1925: [FEATURE] Share to WhatsApp
**What?**
Ability to select text (essential) and files (images, pdfs - desirable) in the app and share to WhatsApp. 

**Why?**
Requirement from Facilitator App Malaysia deployment. 

**How?**
Assuming this is only needed to select known text in certain contexts (as opposed to adding the ability to select arbitrary text within the app), I think it would make sense to add the functionality to the select_text component ([demo template here](https://docs.google.com/spreadsheets/d/1MX-sy4RLy7xPkl0vHZ0lfp-kzGWyohViWjNhyGn7Vts/edit#gid=845703892), [web preview demo here](https://plh-global.web.app/template/comp_select_text)). It should be straightforward to do so using [capacitor's share API](https://github.com/ionic-team/capacitor-plugins/tree/main/share) ([docs](https://capacitorjs.com/docs/apis/share#shareoptions)). This would bring up Android's native share menu, meaning the user would manually have to choose to share via whatsapp and select the intended whatsapp group

See also #1951 about sharing files

jfmcquade:
PR Checklist

## Description

Adds an action, `share`, which launch Android's native share modal, sharing the content supplied as an argument to the action.

Adding the ability to share files should be reasonably straightforward to add once we have a system for interacting with the local filesystem, incoming with [Remote Assets](https://github.com/IDEMSInternational/parenting-app-ui/milestone/4).

Whilst the free [capacitor-share plugin](https://capacitorjs.com/docs/apis/share) exposes options to change the title of the share modal (dialogTitle), supplying this option doesn't seem to have any effect, at least when running in my emulator. I've not exposed this option to authoring for now, but could investigate further if desired.

## Testing

The share functionality is not fully supported on web, as the Web Share API is not fully supported. But most browsers are covered, see https://caniuse.com/web-share. Trying to share on an unsupported browser should generate the following error in the console:
Screenshot 2023-06-09 at 11 37 06

The debug sheets sit in a [remote repo](https://github.com/IDEMSInternational/app-debug-content), and as we don't currently have working github actions to build a preview link from content in a remote repo, I've manually added a preview to appetize [here](https://appetize.io/app/ftiirdwmvpevntoohmc6jo2z3e?device=pixel4&osVersion=12.0&scale=75).

## Git Issues

Closes #1925 

The [feature_share](https://docs.google.com/spreadsheets/d/1VJFxbuUtKLiw9HucShQalalKcVpOKjryOzfKQlBHE_o/edit#gid=960980632) template:
Screenshot 2023-07-04 at 17 07 21

Demo running in local emulator:

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/ac15a2ec-b225-4507-ba7f-7d281e0eb116

Demo running on web (Safari on macOS):

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/60af2330-1d91-4c22-b0bb-91bd39128803


esmeetewinkel (approved):
Functional test passed. Does the word "Copied" that pops up after something was copied come from Android or from our app code? Assuming the first - but asking in case we need to consider it for translation.

[screen-capture (47).webm](https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/b35381c2-a7d1-42fc-b9f8-684ababe9289)

jfmcquade:
> Functional test passed. Does the word "Copied" that pops up after something was copied come from Android or from our app code? Assuming the first - but asking in case we need to consider it for translation.

That's native Android so we don't need to worry about it

src/app/shared/services/share/share.service.ts
```diff
@@ -0,0 +1,53 @@
...
+  private initialise() {
+    this.registerTemplateActionHandlers();
+  }
+
+  private registerTemplateActionHandlers() {
+    this.templateActionRegistry.register({
+      share: async ({ args }) => {
+        const [actionId, ...shareArgs] = args;
+        const childActions = {
+          text: async () => {
```
chrismclarke, line 24: nit(non-blocking)
Is there a reason to split the text/url functionality? It might make more sense to just keep as a single `share` action and allow parameters the exact same as the capacitor share api (and underlying web-share api), as this simplifies the process of documentation and supporting any future features

src/app/shared/services/share/share.service.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  async share(options: ShareOptions) {
+    const { value: canShare } = await Share.canShare();
+    if (canShare) {
+      try {
+        const { activityType } = await Share.share(options);
+        console.log("[SHARE] Content shared to", activityType);
+      } catch {
+        console.log("[SHARE] Share cancelled");
```
chrismclarke, line 49: nit(blocking)
It would be good to track errors like this to better understand how often and the reason sharing fails. I'd recommend either leaving the error to bubble, or explicitly calling the logError method from the error logging service

src/app/shared/services/share/share.service.ts
```diff
@@ -0,0 +1,53 @@
...
+        const [actionId, ...shareArgs] = args;
+        const childActions = {
+          text: async () => {
+            const [text] = shareArgs;
+            await this.share({ text: String(text) });
+          },
+          url: async () => {
+            const [url] = shareArgs;
+            await this.share({ url });
+          },
```
chrismclarke, line 31: nit(non-blocking) - Do we also plan to add support for sharing files? Might be useful to consider how to convert from the plh asset path to local file for sharing and including within the main methods

chrismclarke (changes requested):
Thanks, see replies to your inline comments, I think I'd be happy to merge

> See info at https://caniuse.com/web-share, supported by chrome on windows and android, and safari on mac or ios - so pretty reasonable support I'd say (almost 90% of global users)

Good point, my go-to browsers are Chrome and Firefox (both on Mac), so I'm in that 10%, but I've now tested the web functionality on Safari

src/app/shared/services/share/share.service.ts
jfmcquade: Yes, that's the plan. It should be reasonably straightforward to use a method introduced by the remote assets work to handle that conversion, so I think it'd make sense to add the file sharing feature after #1923 is merged at least. I've updated #1951 accordingly.

src/app/shared/services/share/share.service.ts
jfmcquade: That sounds sensible, but I don't think we have a way to support key/value parameters being passed to an action currently. We interpret the arguments as an ordered list – I think trying to pass in key/value pairs might reach the limits of our list separator syntax, but I could well be wrong. Maybe something to be considered alongside working towards a universal syntax between the app and chatbot authoring, I have an RFC on the back burner.

jfmcquade:
Thanks, see replies to your inline comments, I think I'd be happy to merge

> See info at https://caniuse.com/web-share, supported by chrome on windows and android, and safari on mac or ios - so pretty reasonable support I'd say (almost 90% of global users)

Good point, my go-to browsers are Chrome and Firefox (both on Mac), so I'm in that 10%, but I've now tested the web functionality on Safari

src/app/shared/services/share/share.service.ts
esmeetewinkel: From my authoring point of view it would make sense if `share` just expects a string (which could be text or URL) - but I appreciate I don't know the full picture

esmeetewinkel:
Functional test passed on Chrome and Chrome for Android.

image          
image

src/app/shared/services/share/share.service.ts
chrismclarke: I think every action can decide how to process their own args, so for example we could read the share action arg and process depending on whether authored as a single string or as a parameter list
```
click | share | url: https://some-url.com; text: Here is my text;
```
We would just need to parse the second arg as a param list

Whereas
```
click | share | https://some-url.com
```
could imply a default of `url: https://some-url.com` and still remain functional.

I think there should already be a utility function somewhere to parse string parameter lists, although might need a little extra testing to see how best to handle case where only a string included

chrismclarke (approved):
Code looks good to me as do the comments. I might suggest resolving the authoring convention ahead of merge to save changing afterwards, but otherwise should be good to go

jfmcquade:
Having looked into this a bit further, I can't see any difference in the way that the shareable values are handled in my Android emulator when they are supplied to the Capacitor Share API as a `url` vs when they are supplied as `text`. I've simplified the syntax to remove the child actions, so now the syntax is just, for example, `click | share: Hello`, and the shared value is always passed to the API as `text`.

Whilst there is a difference in the way URLs and text are handled when supplied to the Capacitor Share API, this difference does not seem to manifest on Android. I suggest we merge this PR and return to the share functionality when we have the need for more control over the handling of the shared data, at which point we can add support for passing the action a parameter list of the form `url: https://some-url.com; text: Here is my text;`, via a local variable.

esmeetewinkel (approved):
Functional test passed

PR #1944, opened 2023-06-09
Chore: add web preview for debug deployment 

jfmcquade:
PR Checklist

## Description

With the contents for the debug deployment now in [its own repo](https://github.com/IDEMSInternational/app-debug-content), the debug components are no longer available in the [plh_global web preview](https://plh-global.web.app/). This PR configures a new Firebase hosting site for a web preview of the debug deployment, available at [https://idems-debug.web.app/](https://idems-debug.web.app/).

Additionally, the components documentation files typically link to the live preview of the respective component demo. These links were broken and have been updated to point to the debug site.

Whilst we will eventually enable git actions on content repos that can generate web previews, and likely move away from Firebase hosting, the changes in this PR are necessary in the interim.

## Screenshots/Videos

Screenshot 2023-06-09 at 12 46 33

### DIFF
--- a/.firebaserc
+++ b/.firebaserc
@@ -28,6 +28,9 @@
         ],
         "pfr": [
           "parenting-for-respectability"
+        ],
+        "debug": [
+          "idems-debug"
         ]
       }
     }
--- a/documentation/docs/components/audio.md
+++ b/documentation/docs/components/audio.md
@@ -9,7 +9,7 @@
 ![](images/audio.png)
 
 [Google Sheet Demo](https://docs.google.com/spreadsheets/d/1ARbNRGDer5vj9qSpRMZFrMkYifGkH3TLtDVp72YbaqU/edit#gid=551506513)   
-[Live Preview Demo](https://plh-global.web.app/template/feature_audio)
+[Live Preview Demo](https://idems-debug.web.app/template/feature_audio)
 
 ## Parameters
 
--- a/documentation/docs/components/button.md
+++ b/documentation/docs/components/button.md
@@ -10,7 +10,7 @@
 ![](images/button.png)
 
 [Google Sheet Demo](https://docs.google.com/spreadsheets/d/1OmgZICjM5EMT1KgLOU_ovDljRF_SPlpQAgkKWPrNX0s/edit#gid=569531329)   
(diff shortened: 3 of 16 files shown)


esmeetewinkel (approved):
Thanks, read through the notes and makes sense to me. Not sure if there's anything to test other than confirming that the weblink https://idems-debug.web.app/ works as expected.

jfmcquade:
> Thanks, read through the notes and makes sense to me. Not sure if there's anything to test other than confirming that the weblink https://idems-debug.web.app/ works as expected.

I agree, I don't think there's a reason not to merge

chrismclarke:
Minor nit - So does that mean the plh-global.web.app site is no longer required? We might want to take down the hosting site to avoid people using and seeing outdated information

esmeetewinkel:
> Minor nit - So does that mean the plh-global.web.app site is no longer required? We might want to take down the hosting site to avoid people using and seeing outdated information

Good point. I am still using this to generate a preview of the `plh_global` deployment, which has been used for giving feedback on content pre-play store release (but has no corresponding Play Store listing), whereas I try to use https://plh-tz.web.app/ and https://plh-za.web.app/ only to show exactly what's in the Play Store at that moment (each having a corresponding Play Store listing). 

It of course doesn't make sense that the master branch preview is used to preview a PLH deployment, so good to think it through. I expect the use case of someone needing to review content pre-play store release will continue to exist. The PR previews serve this purpose partially, downside being that 
- it might be more convenient for reviewers to bookmark a fixed hyperlink, rather than being sent a new hyperlink for each bit of content
- it regularly takes >2 weeks for the reviews to be done, by which time the PR preview is already deleted
- given that the the review cycle can be long, I normally continue working on other bits of content while something is outstanding for review. It's not clear to me whether it would be best to create a new PR for this (although the previous one is unmerged), or to continue committing to the branch of the outstanding content PR.

chrismclarke:
Maybe we can add to the next agenda, would indeed make sense to think of what makes a sensible workflow per content repo (instead of focusing on the main repo). It might make sense to have multiple urls to deploy to alongside dynamic (e.g. plh-global alongside plh-global-staging or plh-global-preview or something similar)

We can also modify the pr preview action if useful to keep available for a maximum of 30 days (7 days is just the default which we've never changed)

PR #1945, opened 2023-06-09
chore: move debug skin out of plh deployment folder

jfmcquade:
PR Checklist

## Description

Moves the `debug` skin out of the `plh` deployment folder. The skin has been moved to the `debug` deployment folder, which now lives in a separate repo, so there is a separate PR in that repo [here](https://github.com/IDEMSInternational/app-debug-content/pull/14).

Skins could also live at the top level, available to all deployments, see closed PR #1930. However, we discussed that more thought is needed in terms of the nested structure of deployments and how skins fit into that – see the [Hierarchy of deployments](https://github.com/IDEMSInternational/parenting-app-ui/milestone/11) milestone. It was decided that for now, it wouldn't make sense for our `weekly_workshop` and `modular` skins to live outside of the `plh` deployment, as their names are only descriptive in reference to that deployment (and the overrides applied at the template authoring level).


esmeetewinkel (approved):
Makes sense, thanks

chrismclarke:
Looks good to me

PR #1948, opened 2023-06-12
fix: skin can be set by manually changing _app_skin field

jfmcquade:
PR Checklist

## Description

Fixes a bug whereby the skin could not be set by manually changing the relevant field (`rp-contact-field._app_skin`), despite the target skin appearing in the deployment config as "available". As observed in [this comment](https://github.com/IDEMSInternational/app-debug-content/pull/14#pullrequestreview-1474818604).

## Dev notes
This issue was due to the skin service taking the static value of `appConfigService.APP_CONFIG` in order to set the available skins. `appConfigService.APP_CONFIG` is initially just the default `APP_CONFIG`, so depending on when the skin service was initialised, it could not have access to the list of available skins supplied by the deployment config. I've refactored so that the skin service subscribes to changes from the `appConfigService` in the same way as many other dependent services do.

## Screenshots/Videos

Successfully setting the skin to `debug` manually on the debug deployment. Contrasted with the video in [this comment](https://github.com/IDEMSInternational/app-debug-content/pull/14#pullrequestreview-1474818604).

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/f6c4021b-6f79-43ee-9bb1-d115c8e00865


jfmcquade:
> This does not imply that the PR is not a fix - I would guess I'm just not managing to produce an app out of these two branches correctly...

Hmm yes I suspect that's the case. If the config file at `.idems_app/deployments/debug/config.ts` doesn't contain the 2 lines referring to skins, as added on [this PR branch](https://github.com/IDEMSInternational/app-debug-content/pull/14), then you will still get the issue as shown in [your video](https://github.com/IDEMSInternational/app-debug-content/pull/14#pullrequestreview-1474818604). 

I'm still confused by the functionality of the deployment folder, configured to track its remote content repo, nested inside of the local code repo folder. It makes sense to me in theory, but I'm surprised that checking out a branch from the command line within the deployment repo folder doesn't seem to correctly update the content. 

I managed to update the contents of my deployment folder to reflect the `chore/move-debug-skin-to-debug-deployment-folder` branch of the content repo by editing the code that pulls from the content repo when running `yarn workflow sync` to pull from that branch (line 33 of `packages/scripts/src/tasks/providers/git.ts` for reference). This obviously isn't how we want to handle it, but that is something we could expose to a `yarn workflow` command if necessary (e.g. `yarn workflow repo checkout <branch-name>`). But I would have thought that the same thing could be accomplished by checking out/pulling a given branch from the command line from inside the deployment folder. As it is, there definitely seems something wrong: in the screenshot below, you can see that I apparently have the `main` branch of the content repo checked out, but the files are in the state of the `chore/move-debug-skin-to-debug-deployment-folder` branch.

Screenshot 2023-06-14 at 11 22 04

I think we should discuss this with @chrismclarke next week. From some searching, managing repos nested inside others [can be complicated](https://stackoverflow.com/questions/47008290/git-how-to-make-outer-repository-and-embedded-repository-work-as-common-standal) – using [git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) is one way to handle repos nested inside others, but I don't think that's what we've got set up, and I think it may be overly complicated.

src/app/shared/services/skin/skin.service.ts
```diff
@@ -35,23 +37,23 @@ export class SkinService extends SyncServiceBase {
       this.appConfigService,
       this.themeService,
       this.templateService,
+      this.appConfigService,
     ]);
-    const skinsConfig = this.appConfigService.APP_CONFIG.APP_SKINS;
-    this.availableSkins = arrayToHashmap(skinsConfig.available, "name");
+    this.subscribeToAppConfigChanges();
```
chrismclarke, line 42: nit(non-blocking)
As we are making changes to the service it might be worthwhile refactoring to register the action handler here also and remove from the `template-action.service`. This would also require adding similar handlers for `emit` functionality as the rest of actions (i.e. a way to register emit handler), or re-authoring to replace emit `set_skin` with a pure `skin set` action

chrismclarke (approved):
One minor non-blocking comment from me but otherwise happy if working as expected

src/app/shared/services/skin/skin.service.ts
jfmcquade: Good suggestions. 
Perhaps it's worth some brief discussion about which option would be preferred. Personally I think I prefer the `skin: set` syntax to `emit: set_skin`. But this would be a breaking change for some existing templates, just 2 from ParentApp, plus our debug sheets and a couple more across different deployments.
Screenshot 2023-06-20 at 15 30 38

There are also other actions that use the `emit: ` pattern that would be candidates for being refactored. I'm thinking in particular of `set_theme`, although my main reason for thinking of this over `set_language` is that the latter has its own service, which shouldn't necessarily hold any significance from an authoring perspective
Screenshot 2023-06-20 at 15 40 03

So maybe the simplest option would be to add a way to register an emit handler, as you suggest, unless we think it's important to move towards authoring actions in the `skin: set` style

esmeetewinkel (approved):
Functional test passed

image

src/app/shared/services/skin/skin.service.ts
chrismclarke: Yeah I think now that we have a cleaner system for registering actions the whole `emit` system seems a little unnecessary. 

As an intermediate we could add a small script to the parser to deprecate and rewrite emit statements whilst waiting for authors to update (like how the default parser manages `DEPRECATED_FIELD_NAMES`). Only thing to resolve would be what prefix to use as an action handler`force_reprocess`, `force_reload` etc.... maybe something like `template reload` or `page reload`?

Not super important, but might be good to make a low priority issue to follow-up now this PR has been merged

esmeetewinkel: https://github.com/IDEMSInternational/parenting-app-ui/issues/1963

PR #1952, opened 2023-06-13
chore: configure web preview for plh_facilitator deployment

jfmcquade:
PR Checklist

## Description

Adds a web preview for the `plh_facilitator` deployment, available here: https://plh-facilitator.web.app

See #1944 for a brief discussion of new web previews in the period of moving deployment contents to their own repos.

After merging, the `plh_facilitator` web preview can be updated from the CLI according to the following instructions:
1. Set the active deployment to `plh_facilitator` (if not previously imported, run `yarn workflow deployment import https://github.com/IDEMSInternational/plh-facilitator-app-content`
2. Run `yarn workflow sync` to sync the latest content
3. Run `yarn build` to build the app the `www` folder
4. Run `firebase deploy --only hosting:plh_facilitator` to deploy the local build to Firebase
    - You'll need to install Firebase tools and log in if you haven't done so previously. Run `npm install -g firebase-tools`, then `firebase login`

## Screenshots/Videos

https://plh-facilitator.web.app
Screenshot 2023-06-13 at 16 38 52

### DIFF
--- a/.firebaserc
+++ b/.firebaserc
@@ -31,6 +31,9 @@
         ],
         "debug": [
           "idems-debug"
+        ],
+        "plh_facilitator": [
+          "plh-facilitator"
         ]
       }
     }
--- a/firebase.json
+++ b/firebase.json
@@ -98,6 +98,17 @@
           "destination": "/index.html"
         }
       ]
+    },
+    {
+      "target": "plh-facilitator",
+      "public": "www",
+      "ignore": ["firebase.json", "**/.*", "**/node_modules/**"],
+      "rewrites": [
+        {
+          "source": "**",
+          "destination": "/index.html"
+        }
+      ]
     }
   ]
 }


esmeetewinkel (approved):
Thanks! Just to confirm, does `firebase deploy --only hosting:plh_facilitator` generate the preview from the two branches (code and content) that are open at the moment of running, or does it take master + main by default?

esmeetewinkel:
@jfmcquade  I had to change a hyphen for an underscore for this to work on my end, should I commit that change? 

image

jfmcquade:
> Thanks! Just to confirm, does firebase deploy --only hosting:plh_facilitator generate the preview from the two branches (code and content) that are open at the moment of running, or does it take master + main by default?

It generates the preview from the `www` folder that sits at the route of the project locally, which is not committed to the remote repo. This folder is populated when you run `yarn build`, which builds the app using the current configuration – that is, it will use the currently active branch, the currently active deployment, and the content as it appears in the deployment subfolder inside the code repo. So if, for example, changes had been made to the remote content repo that hadn't been pulled into the `.idems_app/deployments/plh_facilitator` folder in your local code repo, then those changes would not be represented, you'd have to do a `yarn workflow sync` first.
Maybe it's clearest to say that the build of the app generated by `yarn build` will match what you see when you run `yarn start` (some background settings will be different such as environment variables to connect to the live versions of the servers etc.). So it draws on the content and code in the same way as `yarn start`, using the current branch and content.

> @jfmcquade I had to change a hyphen for an underscore for this to work on my end, should I commit that change?

Oops thanks, well spotted. I've made another PR here: #1953

PR #1955, opened 2023-06-15
Chore: clean up deployments

esmeetewinkel:
PR Checklist

## Description
1. PR removes deployments that are no longer in use.
@jfmcquade Could you confirm whether we still need `plh_asset_packs` and `plh_debug`? I'm not using them. 

2. Proposal to have three dev deployments
   - debug: App deployment for debugging, documentation and demonstrating complex authoring. 
   Sits in separate repo https://github.com/IDEMSInternational/app-debug-content
   - minimal: A minimal collection of sheets that are relevant to any app. Can be used as a starting point whenever creating a new app (please copy over rather than using this deployment directly). 
   No config file set up, as this is just meant for copying when creating a new app, in which case you can just run `yarn workflow deployment create` to create a config (see [documentation](https://idemsinternational.github.io/parenting-app-ui/developers/deployments/)).
   - sandbox: App deployment that can be used as a playground for experimenting with the authoring system. Feel free to make changes even if that means running the risk of breaking things! 
   Will be worked on from a branch `deployment_dev/sandbox` rather than a separate repo, so that people can start using this  while we're ironing out some issues with the repo separation. 

image


jfmcquade (approved):
Nice to have these cleaned up, thanks.

The 3 dev deployments seem very sensible to me, as we discussed on Mattermost.

In terms of `plh_asset_packs` and `plh_debug`, we could have a broader conversation about where to test experimental deployment features. 

- `plh_debug` was used to test skin functionality before it was merged, and then in-app update functionality before it was merged. These tests are both defunct now as those features are used in live deployments. And seeing as we have the `debug` deployment, `plh_debug` can be deleted. From now on, we should probably use the `debug` deployment as an example of how to configure various opt-in deployment features, for example in-app updates, and the encryption functionality that was tested on the `encryption_demo` deployment, deleted in this PR.

- `plh_asset_packs` was created to test the WIP asset-pack/remote asset functionality, it's included in the repo even though the feature isn't finished because building blocks for that functionality have been merged in over various PRs. Whilst the `debug` deployment does seem like the natural home for testing features, and for enabling opt-in features as an example of how to do so, there are some cases where we might want to have a test deployment that is not the `debug` deployment. I'm thinking of cases where we're working on code on a branch and want to make changes to a deployment in ways that it would break were that deployment to use the code on master. For example, when working on asset-packs I may want to test different folder structures for the assets in a deployment that would be incompatible with the way assets are handled by the code on the main branch. In this case, making changes to the `debug` deployment might mean that it gets broken for people working on other branches, and I think we don't want the `debug` deployment to be broken so that it can be used for debugging and as a working example of various features. So I would say it would make sense to keep `plh_asset_packs` deployment for now, and once the feature is completed and merged, we can delete it and move that deployment's functionality into the `debug` deployment (which includes using an `encryption` folder for sensitive information as mentioned in the point above).

PR #1970, opened 2023-06-26
feat: author task card background colour

### LINKED ISSUE #1933: [FEATURE] Task card parameter for background colour
**What?**
An additional parameter for the task card to set its background colour (default = white). 

In this particular case the designer wants to use variations of the theme colours, `color-primary-200` and `color-secondary-200`, so it would be sufficient to refer to these - but I could easily imagine a request to go beyond the theme colours. We might need to think through what's the best syntax (all other components using theme colours are legacy and there's not been careful thought about naming). 

image

**Why?**
Requested by ParentApp for Teens deployment. 

**Other details**
[Figma frame 583](https://www.figma.com/file/OeuqW9ahqwRR5aXbblMHK0/new-skin-human?type=design&node-id=912-237&t=ZlmIA6pmgUXz9wbM-0)

jfmcquade:
PR Checklist

## Description

As is often the case, implementing the minimal functionality required for the desired use case (#1933) is straightforward, but it prompts broader questions about authoring methods more generally.

- New parameter for task card component: `variant`
  - Intended to replace `style`, although `style` is still supported
  - Supported syntaxes:
    - `variant: portrait, background-secondary`
    - `variant: portrait,background-secondary`
    - `variant: portrait background-secondary`
- New values for `variant` property
  - `background-secondary`
  - `background-primary`

## Testing notes

Now that the debug templates exist in a separate repo, changes made to them cannot be included in this PR. See [comp_task_card](https://docs.google.com/spreadsheets/d/1_RY8HbsATOAZU2EDXY5R9OOE_YBEbW0EjTP5zxYP2aA/edit#gid=569531329) in the debug repo, the changes can be pulled with a `yarn content sync`

Screenshot 2023-06-29 at 17 21 02

## Git Issues

Closes #1933 

## Screenshots/Videos

[comp_task_card](https://docs.google.com/spreadsheets/d/1_RY8HbsATOAZU2EDXY5R9OOE_YBEbW0EjTP5zxYP2aA/edit#gid=569531329)

Screenshot 2023-06-26 at 15 43 15
Screenshot 2023-06-26 at 15 43 23


src/app/shared/components/template/components/task-card/task-card.component.ts
```diff
@@ -1,16 +1,91 @@
...
+}
+
 @Component({
   selector: "plh-task-card",
   templateUrl: "./task-card.component.html",
   styleUrls: ["./task-card.component.scss"],
 })
 export class TmplTaskCardComponent extends TemplateBaseComponent implements OnInit {
+  /** Authoring parameters */
+  public parameter_list: ITaskCardParams;
```
jfmcquade, line 87: This isn't actually used, but included here in order to link to the type in the generated docs. Alternatively to keeping the separate properties, we could store all the relevant properties in a `parameter_list` object and read them off, as in `radio-button-grid.component.ts`, however this removes the use of `getStringParamFromTemplateRow()` etc, which doesn't seem desirable. Eventually, perhaps we would want to handle the `get<type>ParamFromTemplateRow()` in the parser itself and remove the reliance on these utility functions, which would allow neater organisation of a component's properties within a `parameter_list` object.

esmeetewinkel:
Would be great if you could request me for review again after you've made the changes discussed in the meeting on 27/06

src/app/shared/components/template/components/task-card/task-card.component.ts
```diff
@@ -1,16 +1,91 @@
...
+  /**
+   * A list of named style variants of the component, separated by spaces or commas.
+   * Use interchangeably with "variant_list". Default "landscape"
+   * */
+  variant: string;
+  /**
+   * A list of named style variants of the component, separated by spaces or commas.
+   * Use interchangeably with "variant". Default "landscape"
+   * */
+  variant_list: string;
```
chrismclarke, line 55: nit(blocking)
Can't remember exactly what we decided on the call, but I can't really think of any reason why we need to support list format for variants, as all variants are simple strings so a list really still just a string like `portrait background-500`. The only potential would be loading a whole list of styles from a data_list, however this feels unnecessarily complex to me so recommend removing all list references

src/app/shared/components/template/components/task-card/task-card.component.scss
```diff
@@ -69,6 +69,14 @@
     }
   }
 
+  &.background-secondary {
```
chrismclarke, line 72: nit(blocking)
It would be good to avoid potential confusion with things like global classes. I'd probably recommend we use data attributes which can match multiple strings, e.g. `[data-variant~='background-secondary']`

src/app/shared/components/template/components/task-card/task-card.component.ts
```diff
@@ -51,8 +126,13 @@ export class TmplTaskCardComponent extends TemplateBaseComponent implements OnIn
     this.title = getStringParamFromTemplateRow(this._row, "title", null);
     this.subtitle = getStringParamFromTemplateRow(this._row, "subtitle", null);
     this.image = getStringParamFromTemplateRow(this._row, "image", null);
+
+    // Use "variant" param to determine styles, including support for legacy "style" param
     this.style = getStringParamFromTemplateRow(this._row, "style", "landscape");
+    const variant = getVariantParamFromTemplateRow(this._row, "");
+    this.style = `${this.style} ${variant}`;
```
chrismclarke, line 133: nit(non-blocking)
I think if we treat variant the same as legacy style we might get ourselves a bit muddled down the road. I'd probably recommend keeping a separate `variant` or `style_variant` property and adding to specific css rules as required

src/app/shared/utils/utils.ts
```diff
@@ -207,6 +207,25 @@ export function getBooleanParamFromTemplateRow(
...
 }
 
+/** Return a string of space-separated values for the "variant" or "variant_list" parameters  */
+export function getVariantParamFromTemplateRow(
+  row: FlowTypes.TemplateRow,
+  _default: string
+): string {
+  const params = row.parameter_list || {};
+  let variantValue = getParamFromTemplateRow(row, "variant", "");
+  if (params.hasOwnProperty("variant_list")) {
```
chrismclarke, line 217: nit(non-blocking)
Assuming we drop the _list syntax I think this can all be removed

chrismclarke (changes requested):
Nice feature to have. I've added a couple minor comments for this implementation that it would be good to tidy up.

i'm also having a few second thoughts moving forwards on the syntax. I think this `variant` form should be fine, although I'm also wondering if we might want to be scoping the styles to specific components within the authoring, and whether we want any _global_ variants

I feel like what we need is a distinction between globally available variants (e.g. `background-secondary`) and component-specific variants (e.g. `portrait`). Or maybe this is the difference between _styles_ and _variants_ ? 

With that distinction made it would also be nice if authors new at what level styles were being updated. E.g.  It might be nice to suggest an author can use `background-primary` on any component and have the background change colour. In that case we would want to apply variants to all components at the template container levels and define the style variants in global css. A bit like tailwind class styles

Then for specific component-level styles we could have something like `variant_task: portrait background-secondary` that applies to a specific element within the component (ideally with the #task) id

But I don't want to totally derail this PR, so might just be something to consider and discuss later... or perhaps could replace the existing style_list implementation

jfmcquade:
Thanks, I believe I've addressed your comments. I think it makes sense to merge this PR before working towards further improvements to the style system, as this feature is needed imminently. But I definitely think there's further improvements to be made.

To continue the discussion:

> I feel like what we need is a distinction between globally available variants (e.g. `background-secondary`) and component-specific variants (e.g. `portrait`). Or maybe this is the difference between _styles_ and _variants_ ?
> 
> With that distinction made it would also be nice if authors new at what level styles were being updated. E.g. It might be nice to suggest an author can use `background-primary` on any component and have the background change colour. In that case we would want to apply variants to all components at the template container levels and define the style variants in global css. A bit like tailwind class styles

I like this idea. I agree that there is a fundamental difference between variants like "background-secondary-200" and variants like "button" and "portrait". And whilst there's bound to be some grey areas, I do think it might make sense to think of the former as "styles" (setting a specified property of the component to have a specified visual attribute) and the latter as "variants" (fundamental variations in the way a component displays, e.g. layout (what else?)).

So we'd end up with a tailwind-like system of "style" properties that could be applied across elements throughout the app, and a separate system of "variants"

I also think, as you mention, that these global "styles" could replace the existing "style_list", although that might warrant some discussion. As already discussed, having authors write pure CSS doesn't work in general, as they are not able to target particular elements within a component. So having a layer of interpretation would seem sensible to me. In the case of colours, we shouldn't expect authors to write `background: var(--ion-color-primary-200)`, so some level of abstraction seems inevitable. Although as @esmeetewinkel pointed out, we do use `style_list` sometimes to make changes that wouldn't otherwise be possible currently, such as setting, e.g. `flex: 2`, or `max-height: 300px`, and there may be good reason to keep "style_list" as it is for applying somewhat emergency overrides.

> Then for specific component-level styles we could have something like `variant_task: portrait background-secondary` that applies to a specific element within the component (ideally with the #task) id

I'm not sure I understand this – wouldn't the fact that the variant applies to a task card be determined by the fact that (component-specific) variants would only exist relative to a specific component/could only be applied at the component level? Or are you thinking of targeting sub-elements within a component, in order to apply globally existing styles, e.g. to make the title of the task card component the secondary colour, you might add `title-secondary` at the component level, i.e. somewhere within the template row that instantiates the task card.

I was thinking that for the global styles/variants, there would be some component-specific code that determined how to apply them. So in the case of the functionality of this PR, the global style might just be referred to with "background-secondary-200", which could be applied to any component, but the component itself would determine what element to apply the "background" properties to – in this case, the div with the class "card-wrapper".


I've created an issue here to track this discussion: #1971

src/app/shared/components/template/components/task-card/task-card.component.ts
```diff
@@ -51,8 +119,14 @@ export class TmplTaskCardComponent extends TemplateBaseComponent implements OnIn
     this.title = getStringParamFromTemplateRow(this._row, "title", null);
     this.subtitle = getStringParamFromTemplateRow(this._row, "subtitle", null);
     this.image = getStringParamFromTemplateRow(this._row, "image", null);
+
+    // Use "variant" param to determine styles, including support for legacy "style" param
     this.style = getStringParamFromTemplateRow(this._row, "style", "landscape");
-    this.isButton = this.style.includes("button");
+    this.variant = getStringParamFromTemplateRow(this._row, "variant", "landscape")
+      .replace(",", " ")
+      .replace("  ", " ");
```
jfmcquade, line 127: If we don't want to support authors supplying a comma-separated of variants, we could remove these, otherwise if we expand this `variant` feature across components it could be worth wrapping them inside a `getStringParamFromTemplateRow()` util function

chrismclarke: That script will only replace the first instance of a comma, you'll likely need regex e.g. `.replace(/,/g,' ')`, or maybe more simple to just use `.split(',').join(' ')`

src/app/shared/components/template/components/task-card/task-card.component.scss
```diff
@@ -69,11 +73,11 @@
     }
   }
 
-  &.background-secondary {
+  &[data-variant~="background-secondary"] {
     background-color: var(--ion-color-secondary-200);
   }
 
-  &.background-primary {
+  &[data-variant~="background-primary"] {
```
chrismclarke, line 80: nit(non-blocking) - could this now be a global style applied to any selector? E.g. `*[data-variant~="background-primary"]`... Although possibly less important if we think we're more likely to handle this through styles instead of variants

src/app/shared/components/template/components/task-card/task-card.component.ts
```diff
@@ -101,6 +93,7 @@ export class TmplTaskCardComponent extends TemplateBaseComponent implements OnIn
   completedIcon: string;
   inProgressIcon: string;
   isButton: boolean;
+  variant: string;
```
chrismclarke, line 96: nit(blocking)
I think moving forward really important we define the variants in the code so that we are aware of what is available. Although the authored variant could be a list container multiple, e.g. `portrait background-secondary` I think for now still fine to just add type definitions as if single string `variant: portrait | background-secondary`

chrismclarke (changes requested):
Think we're very nearly there, I've moved my general comments to the new issue thread (thanks for creating) and just added a couple minor nit-picks for this PR. Overall looking useful though

src/app/shared/components/template/components/task-card/task-card.component.scss
jfmcquade: I've enacted this. I did also consider changing the variant name to "background-primary-200", but I think that might imply a control over styles that doesn't exist yet (e.g. choosing other colour values) – whilst this is still a component _variant_, one of the handful that are available for the task card, "background-primary" makes sense to me as a name.

chrismclarke: Makes sense, will just need to be careful in the future if we do decide to deprecate/migrate the syntax for this case

chrismclarke (approved):
All looking good from my side, I'll pass to @esmeetewinkel for any further functional testing

esmeetewinkel (approved):
Functional test passed. 

image

PR #1981, opened 2023-07-13
chore: excluded ngx-extended-pdf-viewer assets from build

### LINKED ISSUE #1940: Exclude PDF viewer component from plh_tz deployment
**Details**
It was suggested in a 23/05/23 meeting that a workaround for now would be to make a change on the deployment/plh_tz branch post-merge to prevent that deployment's app bundle from surpassing the 150MB threshold: manually remove the code from angular.json that copies the ngx-extended-pdf-viewer assets folder into the app bundle.

To be actioned when we next do a content sync on this deployment (because merging master into this branch gives lots of content merge conflicts).

jfmcquade:
PR Checklist

## Description

As a temporary workaround to minimise the app bundle size for the plh_tz deployment, it was agreed in a 23/05/23 meeting that the assets required by the pdf viewer library used in the pdf viewer component introduced in #1918 should be excluded from the build on the plh_tz deployment branches. This means that the pdf viewer component will not work on builds created from these branches, which is acceptable in the short term as this deployment does not make use of that component.

## Git Issues

Closes #1940

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -35,12 +35,6 @@
                 "glob": "**/*.svg",
                 "input": "node_modules/ionicons/dist/ionicons/svg",
                 "output": "./svg"
-              },
-              {
-                "glob": "{*.min.js,op-chaining-support.js,cmaps/*,images/*,locale/en-GB/*,locale/locale.properties,standard_fonts/**}",
-                "input": "node_modules/ngx-extended-pdf-viewer/assets/",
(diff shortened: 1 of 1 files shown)


chrismclarke (approved):
Makes sense to me, much longer term would of course be nice to be able to look through all templates, detect what components actually used and streamline build more (e.g. also removing odk assets if not using), but that would be really complex to do so for now happy to just manually configure

Issue #1979, opened 2023-07-06
[FEATURE] Parser QC check for missing assets


chrismclarke:
**What?**
Some means to flag during compilation any assets present in templates that are missing from the build (or remote asset packs)

**Why?**
Changing folder structures or typos could easily break asset links without authors being fully aware, and similarly apps could be bundled with assets they don't actually use. 

**How?**
Implementation may be a bit tricky as asset references appear in multiple places and not defined in consistent way. Recommended approach should include checks for:
- data_list columns ending `_asset`
- template value column for single-word entries that include potential 3-4 character filename extensions
- parameter list columns ending `_asset`

This won't cover all existing use-cases, but should at least give authors a means of refactoring if they do want to ensure all assets are checked (e.g. updating data list columns and reading from data lists to provide assets where useful). We may need to manually update some components if we want to impose parameter list asset naming conventions.

**Other details**
WiP regex to match asset files:
`^[A-Z0-9/\\_-]*\.[A-Z0-9_-]{3,4}$` 
[regexr.com/7gkb6](https://regexr.com/7gkb6)

Issue #1971, opened 2023-06-29
[FEATURE] - Expand style/variant system, allowing for global styles


jfmcquade:
#1970 introduces the `variant` parameter to the task card component, as a proposed replacement for the previous `style` parameter. 
We want to role out a general styling system across the app whereby authors can adjust the visual properties of components in a sensible and predictable way.

See discussion on #1970.

Also of relevance are the [style system RFC](https://docs.google.com/document/d/1mJHmjgkSeZ2vnUH_gAZyHQWEw_kR3fxLmui4duuWhHQ/edit?pli=1#heading=h.nd960lmg5uuf) and #1468

chrismclarke:
Just as a quick summary/response to comments raised in PRs

**Styles vs Variants**
I think I'd see variants as collective changes that can't be done by CSS alone, e.g. adding/removing component elements or reordering them. They should be defined at a component level as as strongly typed list available to `the parameter_list`

Ideally styles should encompass css changes, both in a raw sense (e.g. `margin-top: 12px`) and via common utilities (e.g. `background-primary-200`). They are defined at a global level and by default should be available on the container element for all components

**Component Selectors**
As well as being able to apply styles at the component container level, it would also be good to expose accessors to modify styles of multi-element components. E.g. a task card might have a _title_ element within it that the user might want to apply custom styles to. Ideally this should be done in a reasonably systematic way, deciding on a case-by-case bassi what component selectors to expose (or as required)

**Style_List**
Currently style_list used to apply raw css, however it might be better long term to treat more like a tailwind set of utility classes. In practice this could look like a mix of utility/raw/scoped styles, e.g. a task card:
```
background-primary;
margin-top: 12px;
title_style: color-secondary-500;
title_style: margin-left: 12px;
```
Utility classes could be identified as single (key-only) properties, scoped styles via the `_style` suffix, and everything else treated as raw css. 

If we wanted to go one step further we could even consider removing the raw styles and just using utility classes extended from an existing system (e.g. [tailwind](https://tailwindcss.com/), cf. [daisyUI](https://daisyui.com/docs/colors/))

What do you think @jfmcquade @esmeetewinkel ?

jfmcquade:
That all sounds sensible to me and captures the discussion, I think the difference between styles and variants as you've identified it makes a lot of sense.

[...]

PR #1984, opened 2023-07-17
feat: new 'card-portrait' variant for button component

### LINKED ISSUE #1934: [FEATURE] Activity card with toggle
**What?**
A button that looks similar to the "task card" to be used for activities of a repeatable nature. Component parameters: image / text (or we could use the value column for this) / toggle_field / toggle_icon_true (e.g. filled heart) / toggle_icon_false (e.g. heart outline). If either the toggle_icon_true or toggle_icon_false is missing, a regular toggle will be used as a default. If the toggle field parameter is not supplied, the toggle is hidden and the component functions as a regular "button". Not sure if it should be called button, card or tile.

image

[Figma frame 578](https://www.figma.com/file/OeuqW9ahqwRR5aXbblMHK0/new-skin-human?type=design&node-id=901-1445&t=k4b08wSoR6U7F1rR-0)

**Why?**
To be used for ParentApp for Teens deployment. In this particular use case:
- A user would be expected to access an activity more than once, and so the completion progress-bar (like in the task card) doesn't make sense
- We expect a large databank of activities, so to be able to navigate more easily there is a need for a mechanism to mark as "my favourite" 

**Other details**
If we make the definition of this component general enough I think we can use this as well for the two white buttons displayed in [Figma frame 580](https://www.figma.com/file/OeuqW9ahqwRR5aXbblMHK0/new-skin-human?type=design&node-id=901-2032&t=ZlmIA6pmgUXz9wbM-0).
image

jfmcquade:
PR Checklist

## Description

- Adds a new variant for the button component, `card-portrait`
  - Adds an additional parameter related to this variant: `image_asset` (NB, this uses the new `_asset` suffix convention as suggested in #1979)
- Adds a new variant for the toggle bar component, `icon`
  - Adds additional params related to this variant: `icon_true_asset` and `icon_false_asset`

This continues the "variant" convention introduced in #1970 . We may want to alter this convention according to changes we decide to make in relation to #1971 (e.g. "variant" may end up being a column rather than a component parameter). However, the convention followed in this PR would still represent a step in the right direction, and this component variant is needed urgently.

It seemed sensible to me to leave the inclusion of the toggle up to the authoring, as opposed to embedding it within the button component itself. This allows the author to access the value of the toggle bar directly, without needing to expose another parameter of the button, and allows them to set actions on the toggle bar directly. As a result of this decision, the icon representation of the toggle is a variant of the toggle bar component itself.

## Testing

In order to view the updates to the relevant debug sheets, with the `debug` deployment active, run `yarn workflow sync`. 
These changes are included in [this PR](https://github.com/IDEMSInternational/app-debug-content/pull/23) on the deployment repo.

It may be that the styling of the card portrait button variant needs some tweaks in response to interaction with actual images, and how the buttons look in a carousel.

## Dev notes

- In general, the button component makes use of an `ion-button` component
  - However, this doesn't allow for the customisation we want here: we can't set the height of the ionic element or make it portrait, and it only supports images in one of its designated "slots"
  - Therefore, this PR effectively renders a completely different component in the case of `variant: card-portrait`. This arguably leads to messy code, but hopefully it's still simple for authors, who don't need to understand the difference between using the ionic component and not.
  - It was discussed whether this component should be a variant of button or a variant of task-card, and it was agreed that fundamentally the difference lies in whether or not the component instance has an associated task. The component required for this use case is not related to tasks, and so it should be a variant of a button.

- I've employed the new convention of defining an interface that represents a component's parameters at the top of the component file (good for documentation, and code clarity in general). Let me know if this looks sensible.
  - Ultimately, we may want to rethink the util functions "getStringParamFromTemplateRow()" etc., as well as the manually written "getParams()" method that exists on all components that take parameters: should there be a way of just defining the params interface and feeding that into a util function to extract all the necessary params into properties of the appropriate type?

## Git Issues

Closes #1934 

### Updates to [comp_button](https://docs.google.com/spreadsheets/d/1OmgZICjM5EMT1KgLOU_ovDljRF_SPlpQAgkKWPrNX0s/edit#gid=569531329)
Screenshot 2023-07-18 at 16 26 37

Screenshot 2023-07-18 at 16 27 39

### Updates to [comp_toggle_bar](https://docs.google.com/spreadsheets/d/17SbJITgQSdNs3PNusDJrkbmCLDutOH9jN_qiea8r0zo/edit#gid=569531329)
Screenshot 2023-07-18 at 16 44 52

Screenshot 2023-07-18 at 16 45 36

### DIFF
--- a/src/app/shared/components/template/components/button/button.component.html
+++ b/src/app/shared/components/template/components/button/button.component.html
@@ -1,42 +1,46 @@
-<ion-button
-  *ngIf="!variantIsCardPortrait"
-  [class]="'full standard medium' + ' ' + style + ' ' + buttonAlign"
-  [disabled]="disabled"
-  (click)="triggerActions('click')"
-  [attr.data-param-style]="style"
-  [attr.data-variant]="variant"
->
-  <ion-icon *ngIf="icon" slot="start" src="{{ icon | plhAsset }}"></ion-icon>
-  <span *ngIf="_row.value" [class]="'left ' + textAlign" [innerHTML]="_row.value | markdown">
-  </span>
-  <span *ngIf="_row.rows" class="children">
-    <plh-template-component
-      *ngFor="let childRow of _row.rows | filterDisplayComponent; trackBy: trackByRow"
-      class="child"
-      [row]="childRow"
-      [parent]="parent"
-      [attr.data-rowname]="_row.name"
-    ></plh-template-component>
-  </span>
-</ion-button>
+<ng-container [ngSwitch]="variantMap.cardPortrait">
+  <!-- Default variant -->
+  <ion-button
+    *ngSwitchDefault
+    [class]="'full standard medium' + ' ' + style + ' ' + buttonAlign"
+    [disabled]="disabled"
+    (click)="triggerActions('click')"
+    [attr.data-param-style]="style"
+    [attr.data-variant]="variant"
+  >
+    <ion-icon *ngIf="icon" slot="start" src="{{ icon | plhAsset }}"></ion-icon>
+    <span *ngIf="_row.value" [class]="'left ' + textAlign" [innerHTML]="_row.value | markdown">
+    </span>
+    <span *ngIf="_row.rows" class="children">
+      <plh-template-component
+        *ngFor="let childRow of _row.rows | filterDisplayComponent; trackBy: trackByRow"
+        class="child"
+        [row]="childRow"
+        [parent]="parent"
+        [attr.data-rowname]="_row.name"
+      ></plh-template-component>
+    </span>
+  </ion-button>
 
-<div
-  class="button-container"
-  *ngIf="variantIsCardPortrait"
-  (click)="triggerActions('click')"
-  [attr.data-variant]="variant"
->
-  <img *ngIf="image" src="{{ image | plhAsset }}" />
-  <div *ngIf="_row.value" [class]="'button-text ' + textAlign">
-    {{ _row.value }}
+  <!-- "card-portrait" variant. This variant is not achievable using an ion-button -->
+  <div
+    *ngSwitchCase="true"
+    class="button-container"
+    (click)="triggerActions('click')"
+    [attr.data-variant]="variant"
+  >
+    <img *ngIf="image" src="{{ image | plhAsset }}" />
+    <div *ngIf="_row.value" [class]="'button-text ' + textAlign">
+      {{ _row.value }}
+    </div>
+    <span *ngIf="_row.rows" class="children">
+      <plh-template-component
+        *ngFor="let childRow of _row.rows | filterDisplayComponent; trackBy: trackByRow"
+        class="child"
+        [row]="childRow"
+        [parent]="parent"
+        [attr.data-rowname]="_row.name"
+      ></plh-template-component>
+    </span>
   </div>
-  <span *ngIf="_row.rows" class="children">
-    <plh-template-component
-      *ngFor="let childRow of _row.rows | filterDisplayComponent; trackBy: trackByRow"
-      class="child"
-      [row]="childRow"
-      [parent]="parent"
-      [attr.data-rowname]="_row.name"
-    ></plh-template-component>
-  </span>
-</div>
+</ng-container>
--- a/src/app/shared/components/template/components/button/button.component.html
+++ b/src/app/shared/components/template/components/button/button.component.html
@@ -1,5 +1,5 @@
 <ion-button
-  *ngIf="!isCardPortrait"
+  *ngIf="!variantIsCardPortrait"
   [class]="'full standard medium' + ' ' + style + ' ' + buttonAlign"
   [disabled]="disabled"
   (click)="triggerActions('click')"
@@ -22,11 +22,11 @@
 
 <div
   class="button-container"
-  *ngIf="isCardPortrait"
+  *ngIf="variantIsCardPortrait"
   (click)="triggerActions('click')"
   [attr.data-variant]="variant"
 >
-  <img src="{{ image | plhAsset }}" />
+  <img *ngIf="image" src="{{ image | plhAsset }}" />
   <div *ngIf="_row.value" [class]="'button-text ' + textAlign">
     {{ _row.value }}
   </div>
--- a/src/app/shared/components/template/components/button/button.component.html
+++ b/src/app/shared/components/template/components/button/button.component.html
@@ -1,11 +1,14 @@
 <ion-button
+  *ngIf="!isCardPortrait"
   [class]="'full standard medium' + ' ' + style + ' ' + buttonAlign"
   [disabled]="disabled"
   (click)="triggerActions('click')"
   [attr.data-param-style]="style"
+  [attr.data-variant]="variant"
 >
   <ion-icon *ngIf="icon" slot="start" src="{{ icon | plhAsset }}"></ion-icon>
-  <span *ngIf="_row.value" [class]="textAlign" [innerHTML]="_row.value | markdown"> </span>
+  <span *ngIf="_row.value" [class]="'left ' + textAlign" [innerHTML]="_row.value | markdown">
+  </span>
   <span *ngIf="_row.rows" class="children">
     <plh-template-component
       *ngFor="let childRow of _row.rows | filterDisplayComponent; trackBy: trackByRow"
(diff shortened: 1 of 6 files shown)


src/app/shared/components/template/components/button/button.component.html
```diff
@@ -1,11 +1,14 @@
 <ion-button
+  *ngIf="!variantIsCardPortrait"
```
chrismclarke, line 2: nit(non-blocking)
Would be good to add comments to separate out the different variant visualisations, particularly noting why the Portrait is so standalone/custom

chrismclarke: nit(blocking)
It would be better to use an ngswitch statement to define specific case behaviour and then default in case we chose to have more custom variants in the future (save lots of !isVariantA && !isVariantB etc.)

src/app/shared/components/template/components/button/button.component.html
```diff
@@ -16,3 +19,24 @@
     ></plh-template-component>
   </span>
 </ion-button>
+
+<div
+  class="button-container"
```
chrismclarke, line 24: nit(non-blocking) probably good to include the variant name in the class to make css more explicit and then nest all child styles within the `.variant-button {....} ` top style

src/app/shared/components/template/components/button/button.component.scss
```diff
@@ -67,6 +68,52 @@ ion-button[data-param-style~="card"] {
...
+    background-color: var(--ion-color-gray-light);
+  }
+
+  img {
+    width: 85%;
+  }
+
+  .button-text {
+    width: 100%;
+    padding: 1rem 1rem 0.2rem;
```
chrismclarke, line 97: nit(blocking) 0.2rem seems pretty arbitrary, better to avoid fine-tuning so much on one component and use numbers more common throughout (for consistency across components). Also given 1rem is 16px on most displays 0.2rem would be 3.2px which doesn't make much sense

src/app/shared/components/template/components/button/button.component.scss
```diff
@@ -67,6 +68,52 @@ ion-button[data-param-style~="card"] {
   }
 }
 
+.button-container[data-variant~="card-portrait"] {
```
chrismclarke, line 71: style scoping nice and clean, thanks for doing that

src/app/shared/components/template/components/button/button.component.scss
```diff
@@ -67,6 +68,52 @@ ion-button[data-param-style~="card"] {
...
+  border: 1px solid var(--ion-color-gray-light);
+  border-radius: var(--ion-border-radius-secondary);
+  min-height: var(--buttons-short-height);
+  filter: drop-shadow(var(--ion-default-box-shadow));
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: 1.5rem 0.5rem 0.5rem;
+  // TODO: make a variable, shared with task card
+  width: 165px;
```
chrismclarke, line 85: nit(non-blocking) another fairly non-standard size, given most sizing will scale in increments of 16px (or sometimes 8px/4px) 160px might make a more sense

src/app/shared/components/template/components/button/button.component.ts
```diff
@@ -14,25 +42,22 @@ import { TemplateBaseComponent } from "../base";
...
   /** TEMPLATE PARAMETER: "button_align" */
-  buttonAlign: "left" | "centre" | "right" = "centre";
+  buttonAlign: IButtonParams["buttonAlign"] = "centre";
   /** TEMPLATE PARAMETER: "icon". The path to an icon asset */
-  icon: string;
+  icon: IButtonParams["icon"];
+  /** TEMPLATE PARAMETER: "image_asset". The path to an image asset */
+  image: IButtonParams["image"];
+  /** @ignore */
+  variantIsCardPortrait: boolean = false;
```
chrismclarke, line 60: nit(blocking) adding extra variables to track individual variants might not scale very well. Probably better to include a local `variantMap` variable that splits the string, populates as `{portrait: true}` along with any other variants and use as a lookup in the code

src/app/shared/components/template/components/toggle-bar/toggle-bar.html
```diff
@@ -0,0 +1,20 @@
+<div
+  class="container"
+  [class]="position"
+  [attr.data-param-style]="style"
+  [attr.data-variant]="variant"
+>
+  <div
+    *ngIf="!variantIsCardPortrait"
```
chrismclarke, line 8: nit(blocking) many of the same comments from the button component may apply here

src/app/shared/components/template/components/button/button.component.ts
```diff
@@ -5,6 +5,34 @@ import {
 } from "src/app/shared/utils";
 import { TemplateBaseComponent } from "../base";
 
+interface IButtonParams {
```
chrismclarke, line 8: nice to have all the params so clearly outlined, thanks for that

src/app/shared/components/template/components/button/button.component.ts
```diff
@@ -14,25 +42,22 @@ import { TemplateBaseComponent } from "../base";
...
-    | "tall"
-    | "standard"
-    | "alternative" = "information";
+  /** TEMPLATE PARAMETER: "variant" */
+  variant: IButtonParams["variant"] = "information";
+  /** TEMPLATE PARAMETER: "style". Legacy, use "variant" instead. */
+  style: IButtonParams["style"];
   /** TEMPLATE PARAMETER: "disabled". If true, button is disabled and greyed out */
-  disabled: boolean = false;
+  disabled: IButtonParams["disabled"] = false;
```
chrismclarke, line 50: nit(non-blocking)
it might be easier to use a single `params` variable typed with the definition above to ensure all parameters are covered in the code

chrismclarke (changes requested):
See inline code comments, will add general comments separately

chrismclarke:
Thanks @jfmcquade , quite a lot to go through here so apologies if my review is a little rushed. I think the much more detailed type definitions and styled scoping for the changes is a really nice move forward. Agree that deviating from ion-components within a single template component is quite messy (hard to capture all the ionic styling), I might be tempted to revisit what is possible with their slots system but otherwise we might want to think longer term about moving to a more flexible styling system.

Also agreed now that we are formalising more of the parameters would be good to remove some of the code duplication found always extracting one-by-one. I'm not currently aware of any tooling that will allow use to read from our type definitions and make available at runtime to cast code to the right type, but there may be something somewhere worth looking into. Otherwise, assuming we provide default values for all parameters we could use a function that takes the input json and compares against the default parameter and casts to same type (e.g. if default { param1:'hello", param2:false, param3:100} could then also try to map any param overrides)

src/app/shared/components/template/components/button/button.component.html
```diff
@@ -1,18 +1,50 @@
...
-    <plh-template-component
-      *ngFor="let childRow of _row.rows | filterDisplayComponent; trackBy: trackByRow"
-      class="child"
-      [row]="childRow"
-      [parent]="parent"
-      [attr.data-rowname]="_row.name"
-    ></plh-template-component>
-  </span>
-</ion-button>
+<ng-container [ngSwitch]="variantMap.cardPortrait">
```
jfmcquade, line 1: I couldn't think of a cleverer way of using `ngSwitch` with the `variantMap` than this, which would require additional nested `ngSwitch` statements if we add support for additional elementary variants. But let me know if I'm missing something

jfmcquade:
Thanks, @chrismclarke. I believe I've affected your suggestions. 

I've made follow-up issues for the larger structural changes: #1992 and #1993. Let me know if you think they correctly capture what they should.

esmeetewinkel (approved):
Functional test passed

src/app/shared/components/template/components/button/button.component.html
chrismclarke: Ah, yeah I see the issue. Really we want a slightly more sophisticated system where as part of the variant mapping system we identify specific base template (with multiple html templates defined either in one file or multiple).

I'd probably recommend tidying up at some point by adding another variable that tracks `templateBase` or similar, but not really urgent (maybe just next time working on the component if there's more variants to add that have structural changes)

chrismclarke:
Thanks @jfmcquade , the changes look really good and the issues capture nicely the larger scale points raised. Nice one

PR #1987, opened 2023-07-20
Added slider component description

cabrinenyona:
## Description

Added slider component description

## Screenshots/Videos

image


documentation/docs/components/slider.md
```diff
@@ -1,4 +1,6 @@
 # Slider
+
+A slider refers to a user interface element or control that allows users to adjust and set a value along a visual scale. It typically consists of a horizontal or vertical bar with a movable thumb or handle that users can drag or slide to select a specific value within a predefined range. The slider is often used to enable users to fine-tune settings, make selections, or adjust numerical values.
```
esmeetewinkel, line 3:
```suggestion
A slider allows users to adjust and set a value along a visual scale. It consists of a horizontal bar with a movable handle that users can drag (or "slide") to select a specific value within a predefined range. Sliders are often used to enable users to fine-tune settings, make selections or adjust numerical values. 
```

esmeetewinkel (approved):
Looks good, thanks! I've edited the text slightly. 
- I don't think there's need to mention that it's a user interface element (so we can cut some words), because every display component is a UI element. So we just need to make sure to mention that in the section where we describe what display components are. 
- I think the description can be specific to what this component is able to do at the moment. So you don't need to mention what a slider "typically" does, you can just say what our slider does. Also, I've removed "vertical", because our implementation of the slider actually only has a horizontal bar (and I thought it could be confusing suggesting here that we also have a version with a vertical bar). When we add more options (parameters) to the slider we can always edit this description so that it stays up to date.

esmeetewinkel (approved):

PR #1988, opened 2023-07-21
chore: remove unused plh_tz encryption folder

jfmcquade:
PR Checklist

## Description

@ChrisMarsh82 noticed that there was a private key committed to the repo. I don't believe that it represents a security issue, as this private key is not actually used anywhere and it doesn't match any of those that are actually in use. Therefore I don't think it's necessary to completely remove it from the git history. It was likely created when testing the encrypted deployments functionality introduced in #1780, and committed unintentionally.


esmeetewinkel (approved):

chrismclarke:
Yes I think we can considered that key as revoked for all intents and purposes, looks good to me

PR #1989, opened 2023-07-21
feat[remote assets] New flow type, 'asset_pack'; Use dynamic data to store downloaded asset info

jfmcquade:
PR Checklist

### New features
- Adds a new flow type, `asset_pack`, which corresponds to a contents file/manifest of assets, structurally identical to `data_list`
- The remote asset service now uses the "dynamic data" feature (introduced in #1698) to store user writes over the core asset contents manifest, to be consumed by the template-asset service when displaying assets

### Additional changes
- Adds a method to the dynamic data service, `resetFlow()`, which removes the user writes to a given flow, resetting it to its original state
	- Removing the user writes on the core asset pack is exposed via a template action `asset_pack: reset`, which is useful for testing the asset pack download functionality
- Handle case where supabase and remote assets are not enabled in the environment/deployment config: log a warning to the console when trying to use `asset_pack` actions in this case

## Testing
The template [debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit#gid=1422144003) can be used to test core functionality. 
Screenshot 2023-10-20 at 16 33 42

Clicking the first button to run the the `asset_pack: download: debug_asset_pack_1;` action triggers a download of an asset pack with the name `debug_asset_pack_1` from supabase, found in a folder named `asset_packs` within a bucket named `debug` (these last two values are set in the deployment config).
Screenshot 2023-10-20 at 16 35 19

See screenshots/videos below demonstrating the functionality.

### asset_packs
This new flow type is intended to serve 2 main purposes:
1. Provide something like a lookup object in order for the template asset service to get the location of files to be displayed (as it does currently with the assets `contents.json` file)
    - Currently an `asset_pack` is converted from a `data_list`-like "rows" array to an object resembling the pre-existing `contents.json` for this purpose, although perhaps it would be good for the `asset_pack` format to be directly readable by the template asset service
2. Act as a manifest of files contained within an asset pack. This may be a "core" set of assets (those included in the core app bundle) or an arbitrary "asset pack" comprising a collection of assets
    - These manifest files are intended to be used to specify which files to download from the remote server when an asset pack download is triggered. A complete "asset pack" that gets uploaded to supabase would consist of a collection of assets and a manifest file that catalogues their paths and relevant metadata.
    - They could also be compared to each other in order to determine a subset of files which may need to be downloaded – for example, if some but not all files in an asset pack have been updated since the user downloaded the asset pack.
  - Question: Is it right to use one format for both of these purposes?

- Upon download of a file, the `filePath` property of the `asset_pack` file entry is updated to contain either the path to the downloaded file in local storage (on native platforms) or the remote URL (on web). Whichever value is provided, this is read by the template asset service at runtime in order to display the relevant file.
  - This means that the `filePath` property, used for lookup by the template asset service, may be one of 3 types:
    1. The path to the file within the app bundle, relative to the base assets folder
        - Used for "core" assets
    2. The path to the file within the device's local storage
        - Used for remote assets (after download) on native devices
    3. The URL to the file on the remote servers (supabase)
        - Used for remote assets (after mock "download") on web
  - This means that the template asset service can refer to the same property of each asset entry when looking up the path that should be used to fetch the asset. But potentially it's confusing. Question: would it be clearer to use different properties for these purposes?

### Dynamic Data 
- The dynamic data service is used to store the updates to the core asset_pack as user writes
- I've added a new method that resets a specified flow to its original state, removing any user writes. This is useful for testing. However, deleting the data doesn't cause the `query$` observable to update
	- Even a `force_reload` action doesn't cause the asset path to update, a manual refresh in the browser is needed
	- This is OK for now, as currently the only use for this method is for testing this functionality, but it would be good to figure out

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/32188353-c827-4e1f-88b0-48a9528a03a9

## Git Issues

Closes #

## Screenshots/Videos

The [debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit#gid=1422144003) template running in browser, getting public URLs for the relevant asset files (NB, the `asset_pack: reset` action requires a manual browser refresh, see Dev notes above).

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/e3e56acd-ae8a-4688-bb94-29c55af0e131

The [debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit#gid=1422144003) template running in an emulator, saving the relevant asset files to local storage. Run using the command `ionic cap run android -l --external`. (NB, the `asset_pack: reset` action requires the app to be manually restarted to take effect, not shown in this clip).

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/7cfe97e1-dff6-40a1-b4de-48ef9b63527b


packages/data-models/deployment.model.ts
```diff
@@ -146,9 +146,11 @@ interface IContentsEntry {
   relativePath: string;
   size_kb: number;
   modifiedTime: string;
-  md5Checksum: string;
+  md5Checksum?: string;
```
chrismclarke, line 149: nit(blocking)
This interface has a note that it duplicates the interface used in scripts, however the changes added makes this no longer apply.
To avoid confusion it would be better to create a new interface `IAssetContentsEntry` that extends.

Additionally would you be able to confirm why md5Checksum may now no longer be included? Seems very useful to have

packages/scripts/src/commands/app-data/postProcess/sheets.ts
```diff
@@ -45,6 +45,7 @@ class SheetsPostProcessor {
     fs.emptyDirSync(appSheetsFolder);
     // Handle Copy
     this.sheetsCopyFiles(sourceSheetsFolder, appSheetsFolder);
+    this.hackGenerateCoreAssetContentsDataList(app_data.output_path, appSheetsFolder);
```
chrismclarke, line 48: nit(blocking)
- Would be good to refer to any methods related to asset_pack sheets with the term `assetPack` to disambiguate from assets
- As the asset pack sheet will need to be updated whenever assets are synced really it would be better to find a way to include this method as part of asset post-processing
- Currently the postProcess methods has support for tests, so would be good to include some specific tests related to this feature to cover core behaviour

src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
```diff
@@ -118,8 +118,14 @@ export class PersistedMemoryAdapter {
...
     if (this.get(flow_type, flow_name)) {
+      const idsToDelete = [];
+      const rowHash = this.state[flow_type][flow_name];
+      for (const row_id of Object.keys(rowHash)) {
+        idsToDelete.push(`${flow_type}__${flow_name}__${row_id}`);
+      }
       delete this.state[flow_type][flow_name];
-      this.persistStateToDB();
+      // Persist deletes to DB
+      this.collection.bulkRemove(idsToDelete);
```
chrismclarke, line 128: nit(non-blocking)
I'm struggling to follow this a bit, could you clarify why we no longer need the `persistStateToDB` method call following delete and what the extra code does in its place?

src/app/shared/services/dynamic-data/adapters/reactiveMemory.ts
```diff
@@ -128,8 +128,12 @@ export class ReactiveMemoryAdapater {
       const updatedDoc = await existingDoc.atomicPatch(data);
       return updatedDoc.toMutableJSON();
     } else {
-      await collection.insert(update);
+      await collection.insert(data);
```
chrismclarke, line 131: nice catch!

src/app/shared/services/dynamic-data/adapters/reactiveMemory.ts
```diff
@@ -128,8 +128,12 @@ export class ReactiveMemoryAdapater {
...
       return update;
```
chrismclarke, line 132: nit(non-blocking)
nice catch on the insert, would probably be good to also return the updated doc as mutableJson (not that the return seems to be used anywhere)

src/app/shared/services/file-manager/file-manager.service.ts
```diff
@@ -75,4 +76,20 @@ export class FileManagerService extends SyncServiceBase {
       encoding: Encoding.UTF8,
     });
   }
+
+  /** @returns filepath and file size metadata from a file saved in local storage */
+  async generateAssetContentsEntry(relativePath: string): Promise<Partial<IAssetEntry>> {
```
chrismclarke, line 81: nit(non-blocking)
Shouldn't all the metadata be available from the asset pack to begin with? Do we need to be generating (particularly if it means no option for an md5 hash)

src/app/shared/services/dynamic-data/dynamic-data.service.ts
```diff
@@ -109,16 +109,16 @@ export class DynamicDataService extends AsyncServiceBase {
...
   ) {
     if (update) {
       const { collectionName } = await this.ensureCollection(flow_type, flow_name);
       const existingDoc = await this.db.getDoc<any>(collectionName, row_id);
-
       if (existingDoc) {
         const data = existingDoc.toMutableJSON();
         update = deepMergeObjects(data, update);
-      } else {
+      } else if (!options.upsert) {
```
chrismclarke, line 121: nit(non-blocking)
I think we can probably allow the default behaviour of `update` should be to upsert, remove the option and error (and if required include code to create doc if db update doesn't work)

src/app/shared/services/dynamic-data/dynamic-data.service.ts
```diff
@@ -147,6 +147,18 @@ export class DynamicDataService extends AsyncServiceBase {
     return { collectionName };
   }
 
+  /** Remove user writes on a flow to return it to its original state */
+  public async resetOverwrites(flow_type: FlowTypes.FlowType, flow_name: string) {
```
chrismclarke, line 151: nit(non-blocking)
nice method to have, although I can imagine some confusion between clearCache and resetOverwrites methods (in fact the remote-asset services uses its own `clearCache` method to call this reset method). I think probably best to combine into a single `reset` method. Currently I can't think of any use case where we would want to delete and not return to original state, but could be an optional parameter/option if required.

nit(non-blocking)
Should move public method above private

chrismclarke (changes requested):
Thanks Johnny, at this point I'm mostly just trying to look and code and haven't moved on to functional testing, so apologies if the review seems more nit-picky. Will be happy to go more in depth to functionality for second pass.

src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
jfmcquade: The function `persistStateToDB()` didn't handle deletions because it only performed a bulk _upsert_. I've now updated that with [this commit](https://github.com/IDEMSInternational/parenting-app-ui/pull/1989/commits/76f1217894d65c9a79cd677f95567ece98e00a7a) so that `persistStateToDB()` should handle deletes in general. I'm not sure that I've found the simplest solution though.

src/app/shared/utils/utils.ts
```diff
@@ -326,6 +326,17 @@ export function deepDiffObjects<T extends Object, U extends Object>(a: T, b: U)
   return diff(a, b) as RecursivePartial<T | U>;
 }
 
+/** @returns the keys from the first object that are not present in the second object. */
+export function shallowDiffObjectKeys<T extends Object, U extends Object>(a: T, b: U) {
```
chrismclarke, line 330: nit(blocking) - I don't think the name is very indicative that it only returns list of keys that appear in A but not in B (i.e. removed from A) - would be better to show both added and removed. I'll add a quick commit to update

chrismclarke (commented):
Thanks for the updates, I've reviewed and left comments for the backend code. I think the main issue is that the current implementation puts a dependency on sheets to know the state of assets to populate the core asset pack which is a bit awkward - but should be a quick fix by simply generating the asset_pack sheet at runtime instead.

As for the frontend, I've gone over all the code beyond the remote-asset service and all looking good. I haven't funtionally tested the remote-asset service itself, would it be possible to start scaffolding out some tests to make the process easier? Otherwise can do more functional testing when completed (see a lot of outstanding TODOs in the code)

src/app/shared/utils/utils.ts
chrismclarke: Updated with 926f438

src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
```diff
@@ -138,15 +138,35 @@ export class PersistedMemoryAdapter {
       }
     }
     await this.collection.bulkUpsert(updates);
+
+    // Handle deletions (check for flows that exist in DB but not in state, and remove from DB)
+    const db = await this.mapDBToObject({});
+    const idsToDelete = [];
+    for (const flow_type of Object.keys(db)) {
+      const { deleted } = compareObjectKeys(db[flow_type], this.state[flow_type]);
```
chrismclarke, line 146: @jfmcquade 
Could you confirm I've got this the right way round? The previous logic confused me a little

jfmcquade:
Thanks for the review, I've made a few changes.
- The `core_assets` `asset_pack` is now generated at runtime, as you suggested in your last review.
	- To accomplish this, I've added a new method to the app-data service, `addFlow()`, which adds a flow to the service's contents and cache at runtime. This was required because the dynamic-data service expects to find a pre-existing sheet in order to add writes on top of it – we can't trivially write a sheet to the existing app-data file structure at runtime, so this seemed to be a sensible way to achieve this, but perhaps there are better approaches.
- I've added some notes and screen recordings on functional testing to the PR description. I've also moved some of my comments from the PR description to this issue, which are mainly for me to return to when working on the next remote assets PR. 
- Testing
	- The general unit tests (`yarn test:unit`) are broken, I think by these errors thrown by some of the ODK packages.
	- Screenshot 2023-10-20 at 17 00 59

	- But I was able to run individual unit test files using `ng test --include="src/app/shared/services/remote-asset/remote-asset.service.spec.ts"` 
	- I'd be happy to spend some time writing some unit tests for the new services. It's a bit complicated because the methods

chrismclarke (approved):
I haven't rechecked any of the old code but going through the latest commits looks good to me. 

I've added a small housekeeping commit bc61cc6, hopefully seems sensible to you. The only real change is a small refactor to the new `addFlow` method to rename more explicitly `addRuntimeFlowToContents`, remove the flow subtype (as I don't think it really does anything at this point), and merge some of the duplicate code into a single method.

chrismclarke:
> * I've added some notes and screen recordings on functional testing to the PR description. I've also moved some of my comments from the PR description to this issue: [[FEATURE] - Remote asset sync #1316](https://github.com/IDEMSInternational/parenting-app-ui/issues/1316), which are mainly for me to return to when working on the next remote assets PR.

Sounds good, yeah be nice to get this in and have a slightly cleaner slate to start with for future work

> * Testing
>   * The general unit tests (`yarn test:unit`) are broken, I think by these errors thrown by some of the ODK packages.
>   * But I was able to run individual unit test files using `ng test --include="src/app/shared/services/remote-asset/remote-asset.service.spec.ts"`

I mentioned in another PR, been a known issue for a while and quite a number of reasons (see #2080). Would be nice to get full test suite passing to integrate into CI, but yes short term workaround is to use the `--include` flag as stated (might be handy to add comment to test code as a reminder).

>   * I'd be happy to spend some time writing some unit tests for the new services. It's a bit complicated because the methods call lots of other services or make external requests. I'd appreciate some pointers, perhaps just what to prioritise writing tests for

Yeah always takes a bit of extra consideration, the main options you have are:
1. Create mock implementations for the required services and inject instead of real. Some of these already exist, e.g. `MockLocalStorageService` and `MockErrorHandlerService` to replace the corresponding services. Example use in tests in `src\app\shared\services\data\app-data-variable.service.spec.ts`

2. Extend the class for test cases and replace/mock methods as required. There some examples of this in `src\app\shared\services\data\app-data.service.spec.ts`

3. Use spies/stubs to replace functions. I usually find this option as the best-sounding in theory but hardest to implement in practice (due to quirks with how jasmine defines imports). But there's at least some example in `src\app\shared\components\template\services\instance\template-action.registry.spec.ts`

When working on the tests feel free to reach out if you have any more concrete examples of what exactly you want to achieve

Issue #1884, opened 2023-04-13
[BUG] Deployment configs do not always recompile when they should


jfmcquade:
## Describe the bug
Making changes to a deployment config should cause the config to recompile when setting the active deployment. Whilst this works as expected when editing the deployment's `config.ts` file directly, sometimes changes made to other files have effects on the config and should also therefore trigger a recompile. There are two main cases where changes made to other files do not trigger the deployment config to recompile as it should, which can cause knock-on issues: 

- [ ] 1. Changes made to a deployment's parent deployment config (first identified in #1809)

The following bugs are addressed by #1885, whereby a dev can manually bump the `DEPLOYMENT_CONFIG_VERSION` number (this feature was already available but less prominent)
- [x] 2.1. Changes made to `packages/data-models/deployment.model.ts`, which should affect all deployment configs
- [x] 2.2. Changes made to `packages/data-models/appConfig.ts`, which should affect all deployment configs, as above

### 1.
1. Choose a pair of deployments that are parent and child, e.g. `plh` and `plh_global`.
2. Run `yarn workflow deployment set plh_global` to ensure the child's config is compiled
3. Make a change to one of the values assigned in the parent deployment's config file, e.g. in `.idems_app/deployments/plh/config.ts`, change `config.app_config.APP_LANGUAGES.default = "gb_en"` to `config.app_config.APP_LANGUAGES.default = "test"`
4. Run `yarn workflow deployment set plh_global`. This will trigger the `plh` config to recompile, but these changes will not be reflected in the compiled child's config. You can see this by inspecting `.idems_app/deployments/plh_global/config.json`

This bug is now resolved by #1885:
> ### 2.
> After [#1873](https://github.com/IDEMSInternational/parenting-app-ui/pull/1873) was merged (which adds a `supabase: {}` property to the top level of deployment configs), the bug can be observed when switching to a deployment that already has a compiled config.json file:
> 1. Checkout master 
> 5. Choose a deployment with a compiled contents.json file, and check out that deployment. E.g. if the file `.idems_app/deployments/plh_global/config.json` exists, then you can run `yarn workflow deployment set plh_global`
> 6. Inspect the file `.idems_app/deployments/activeDeployment.json` and note that there is no entry for "supabase". If you start the app with `yarn start`, the following error will be logged in the console:
[...]

PR #1994, opened 2023-07-27
feat: exclude uncompressed audio assets from plh deployments

jfmcquade:
PR Checklist

## Description

In order to reduce the size of the app bundle to under the 150MB Play Store limit, which the `plh_tz` deployment was in danger of exceeding, it was decided that we could compress the audio files that are included in the app. 

All audio files for the `global` and `tz_sw` languages have now been compressed, reducing their size by around 60%, and reducing the size of all assets included in the `plh_tz` deployment to around 98MB. We agreed that any loss in audio quality was very difficult to detect (see [this Mattermost thread](https://chat.idems.international/all/pl/ux8jhtkhcjfu88yyzbh7qyr19a) for discussion and description of the compression process).

In order to keep the original uncompressed audio files, in case there are any issues with the compressed versions and potentially to be included in future high-quality asset packs, I've put the original uncompressed files in folders called "uncompressed", which can be found in the original location of the files. The compressed files are to be found at the original locations of the uncompressed files, meaning that no authoring changes are required to use the compressed audio files in the app. See the folder [plh_assets > tz_sw > plh_audio > relax](https://drive.google.com/drive/u/1/folders/1ruwIDKzmxkDRlmvJL40443Q7pSzQO2Rd) for an example.

In order to exclude the original, uncompressed audio files from the build `plh` deployments, this PR adds an asset filter function to exclude any file whose path contains the substring "uncompressed".

We may want to revisit this approach in the future, but it should be sufficient for meeting our requirements for the imminent `plh_tz` release.

## Testing

With this feature branch checked out, set the active deployment to `plh_tz`. You may need to delete the file `.idems_app/deployments/plh/config.json` before setting the deployment to `plh_tz` in order to ensure that the config recompiles to include the changes made to the parent deployment config (issue #1884). Then run `yarn workflow sync`. You should see that the asset size summary matches that in the screenshot below.

It would be good to quickly check that the changes I have made to the asset folders in Google Drive look right: namely the subfolders of [plh_assets > tz_sw > plh_audio](https://drive.google.com/drive/u/1/folders/1HyNE8dBo4RoaqNh6NIquN-AHBFAPCE0J) and [plh_assets > global > plh_audio](https://drive.google.com/drive/u/1/folders/1IfXCZ4Cv4gCDayhdYhITPDJpiXxr2CzP). It would also be good to run some spot checks to ensure that the correct audio assets are being populated to the app correctly.

## Git Issues

Closes #

## Screenshots/Videos

The assets size summary generated for the `plh_tz` deployment when syncing assets:

Screenshot 2023-07-26 at 17 23 56

### DIFF
--- a/.idems_app/deployments/plh/config.ts
+++ b/.idems_app/deployments/plh/config.ts
@@ -22,6 +22,8 @@ config.translations = {
   source_strings_path: "packages/app-data/translations_source/to_translate",
 };
 
+config.app_data.assets_filter_function = (fileEntry) => !fileEntry.relativePath.includes("uncompressed")
+
 config.app_config.APP_LANGUAGES.default = "gb_en";
 config.app_config.APP_SIDEMENU_DEFAULTS.title = "ParentApp";
 config.app_config.APP_HEADER_DEFAULTS.title = "ParentApp";
--- a/.idems_app/deployments/plh_tz/config.ts
+++ b/.idems_app/deployments/plh_tz/config.ts
@@ -8,6 +8,7 @@ config.app_data!.sheets_filter_function = (flow) =>
     flow.flow_subtype!
   );
 config.translations!.filter_language_codes = ["tz_en", "tz_sw"];
+config.app_data.assets_filter_function = (fileEntry) => !fileEntry.relativePath.includes("uncompressed")
 
 // Override constants
 config.app_config!.APP_LANGUAGES!.default = "tz_sw";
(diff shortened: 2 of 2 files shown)


esmeetewinkel (approved):
Thanks for this and thanks for the testing notes. I had to remove also the subdeployment config 
`.idems_app/deployments/plh_tz/config.json` to make it work.

image

chrismclarke:
Really cool to see such a massive size reduction, can definitely imagine in the future adding workflows to support media optimisation but for now still a lot to be gained by the manual work. Nicely done!

PR #1995, opened 2023-07-27
Feature/error if user does not have gdrive folder access

### LINKED ISSUE #1746: [FEATURE] Add console error if user does not have gdrive folder access
**What?**
Mentioned in #1737, when a user tries to sync from a gdrive folder that they do not have access to the scripts simply return empty data. Would be better to display a message that makes clear the reason there is no data

**Why?**
Make it easier to prevent/troubleshoot access issues

**How?**
Unsure what checks exist in the sdk, a naive approach could be just to log a generic warning/message if the number of templates or assets is 0.

**Other details**
May have different implementation if moving ahead with #1724, but likely would still want something to communicate to the user if there might be permissions access issues

kwAsant:
PR Checklist

## Description

It fixes the error in which the program doesn't address when the user doesn't have access to gdrive

## Git Issues

Closes #1746 

## Screenshots/Videos

image


esmeetewinkel:
Thanks for working on this! Is there an easy way to test this?

kwAsant:
I think the easiest way to test it is to have your access to the google drive folder removed, and then run 'yarn workplace sync'.

jfmcquade:
> Is there an easy way to test this?

I think the easiest way might be to change the `assets_folder_id` in the active deployment to point to a folder that you don't have access to. I can create such a folder and share the ID with you for testing. I think first I'll review the code and then tag you to let you know it's ready for functional testing

packages/@idemsInternational/gdrive-tools/src/commands/download.ts
```diff
@@ -95,6 +95,13 @@ export class GDriveDownloader {
     await this.setupGdrive();
     console.log(chalk.yellow("Retrieving list of files"));
     const serverFiles = await this.listGdriveFilesRecursively(folderId);
+    if (Object.keys(serverFiles).length === 0) {
+      console.log(
+        chalk.red("No files found. Failure to find files may result in an error.") +
+          "\n" +
+          chalk.red("Check if you are authorised to access the Google Drive folders.")
```
jfmcquade, line 102: I think this is a good place to log an error/warning.

I might phrase the message a bit differently:
"No asset files were found. Please ensure your Google account is authorised to access to relevant asset folder."

packages/scripts/src/commands/app-data/postProcess/assets.ts
```diff
@@ -63,6 +64,15 @@ export class AssetsPostProcessor {
     fs.ensureDirSync(appAssetsFolder);
     // Generate a list of all deployment assets, merge with list of assets from parent
     const sourceAssets = generateFolderFlatMap(sourceAssetsFolder, { includeLocalPath: true });
+    if (Object.keys(sourceAssets).length === 0 && sourceAssets.constructor === Object) {
```
jfmcquade, line 67: I don't think this message actually ever logs, so this section can be deleted

packages/scripts/src/commands/app-data/postProcess/assets.ts
```diff
@@ -95,16 +105,30 @@ export class AssetsPostProcessor {
...
-    fs.writeFileSync(contentsTarget, JSON.stringify(sortJsonKeys(assetEntries), null, 2));
-    const missingTarget = path.resolve(appAssetsFolder, "untracked-assets.json");
-    if (fs.existsSync(missingTarget)) fs.removeSync(missingTarget);
-    if (Object.keys(missingEntries).length > 0) {
-      logWarning({
-        msg1: "Assets override found without corresponding entry",
-        msg2: Object.keys(missingEntries).join("\n"),
-      });
-      fs.writeFileSync(missingTarget, JSON.stringify(sortJsonKeys(missingEntries), null, 2));
+    try {
```
jfmcquade, line 108: I think better than a `try/catch` block might be to handle the specific case of the assets folder not existing (i.e. the case where there are no assets).

Instead of `try/catch`, you could use an `if` statement along the lines of:
```
if(fs.existsSync(appAssetsFolder)) {
...all the logic in the 'try' block
} else {
...log our error
}
```

packages/scripts/src/commands/app-data/postProcess/assets.ts
```diff
@@ -95,16 +105,30 @@ export class AssetsPostProcessor {
...
+      console.log(chalk.red("ERROR: the program was stopped due to a runtime error."));
+      if (!fs.existsSync(appAssetsFolder)) {
+        console.log(
+          chalk.red(
+            "The folder '" +
+              appAssetsFolder +
+              "' does not exist. Check you have access to the relevant Google Drive folders."
+          )
+        );
+        process.exit(1); // exit the program if it cant find the folder!
```
jfmcquade, line 130: I think we don't want to exit the process in this case, as we should let the user sync the sheets even if the assets sync has failed

packages/scripts/src/commands/app-data/postProcess/assets.ts
```diff
@@ -95,16 +105,30 @@ export class AssetsPostProcessor {
...
+      if (fs.existsSync(missingTarget)) fs.removeSync(missingTarget);
+      if (Object.keys(missingEntries).length > 0) {
+        logWarning({
+          msg1: "Assets override found without corresponding entry",
+          msg2: Object.keys(missingEntries).join("\n"),
+        });
+        fs.writeFileSync(missingTarget, JSON.stringify(sortJsonKeys(missingEntries), null, 2));
+      }
+    } catch {
+      console.log(chalk.red("ERROR: the program was stopped due to a runtime error."));
```
jfmcquade, line 121: I would say don't worry about catching and handling a general error here. If you remove the `try/catch` block then you can remove this line, and let me know if there are still any errors being thrown and we can handle them separately

packages/scripts/src/commands/app-data/postProcess/assets.ts
```diff
@@ -95,16 +105,30 @@ export class AssetsPostProcessor {
...
+      if (!fs.existsSync(appAssetsFolder)) {
+        console.log(
+          chalk.red(
+            "The folder '" +
+              appAssetsFolder +
+              "' does not exist. Check you have access to the relevant Google Drive folders."
```
jfmcquade, line 127: I think the user probably doesn't need to know the location of the folder, so we could just log something like:
"No assets were downloaded. Please ensure that your Google account is authorised to access to the relevant folder in Google Drive, as defined in the deployment config."

jfmcquade (changes requested):
Well done. A few comments inline

packages/scripts/src/commands/app-data/postProcess/assets.ts
```diff
@@ -95,16 +97,26 @@ export class AssetsPostProcessor {
...
-    const missingTarget = path.resolve(appAssetsFolder, "untracked-assets.json");
-    if (fs.existsSync(missingTarget)) fs.removeSync(missingTarget);
-    if (Object.keys(missingEntries).length > 0) {
-      logWarning({
-        msg1: "Assets override found without corresponding entry",
-        msg2: Object.keys(missingEntries).join("\n"),
-      });
-      fs.writeFileSync(missingTarget, JSON.stringify(sortJsonKeys(missingEntries), null, 2));
+    if (fs.existsSync(appAssetsFolder)) {
+      {
```
jfmcquade, line 101: Code tidy: there's an additional set of curly brackets around this block that should be removed

packages/scripts/src/commands/app-data/postProcess/assets.ts
```diff
@@ -63,6 +64,7 @@ export class AssetsPostProcessor {
     fs.ensureDirSync(appAssetsFolder);
     // Generate a list of all deployment assets, merge with list of assets from parent
     const sourceAssets = generateFolderFlatMap(sourceAssetsFolder, { includeLocalPath: true });
+
```
jfmcquade, line 67: Code tidy: delete this line. Not important but it would make the PR neater because this wouldn't show as a difference

jfmcquade (approved):
Good job, I've added some minor comments with suggestions to tidy up the code. I'll also add Chris C as a reviewer and we can see if he thinks this error logging is sensible

chrismclarke (approved):
Thanks @kwAsant 
Nice idea to check for empty folders, I tried to see if the google drive api could be used directly to check if a user has permissions but proving quite tricky so I think your solution is good for now.

I've done a small bit of tidying in 58ab31c to make the check run earlier in the scripts process (linked to the gdrive download script as opposed to asset postprocess, so that it will also check templates). I've also updated the logging message to include a direct link to the folder that a user can use to check/request permissions, e.g. screenshot

image

But otherwise I think should be good to go

jfmcquade:
> I think the easiest way might be to change the `assets_folder_id` in the active deployment to point to a folder that you don't have access to. I can create such a folder and share the ID with you for testing. I think first I'll review the code and then tag you to let you know it's ready for functional testing

@esmeetewinkel Sorry, I forgot to come back to this. Here's a folder ID that you can use for functional testing: 16fBOWPkpGe3TRv4pU_Dpqw7B0EJ7mG-0

PR #1864, opened 2023-03-22
ci: enable test-appetize for deployment/* branches

jfmcquade:
PR Checklist

- [x] - Latest `master` branch merged
- [x] - PR title descriptive (can be used in release notes)

## Description

Allows PRs made to "deployment" branches (e.g. `deployment/wash`) to trigger the "Test Appetize" github action and thereby deploy a preview Android build to appetize.io. This is achieved in the same way as it is when making a PR to the master branch, namely by adding the `test - appetize` label to the PR on github.

I've set up a new appetize account on IDEMS's behalf. Each app deployment has an associated app on appetize.io. I've invited @esmeetewinkel to the account, but not @chrismclarke as appetize says of invitations, "All of the user's existing and future apps will be added to the account", which is not desirable.

## Testing

The `test - appetize` label is on this PR, and the action is running correctly in the case of a PR to the master branch. After merging, we can test the functionality for a PR to a `deployment/*` branch.

## Dev notes

In order to handle the different deployments deploying to different appetize endpoints, I have set up [github environments](https://github.com/IDEMSInternational/parenting-app-ui/settings/environments) to store the respective secrets for each deployment. These environments relating to the app deployments could have further uses in the future – the pattern for dynamically setting the github actions environment based upon the branch name is contained in [test-appetize.yml](https://github.com/IDEMSInternational/parenting-app-ui/pull/1864/files#diff-125f09497de8e01f26e07df779ef6fcb6aa80a615a5c08d69b27313171268e6b). However these environments will likely be usurped once we have separate content repos.
NB: each APPETIZE_TOKEN secret is identical, and refers to the API token for the appetize account. I didn't want to change the general repo secret until this branch was pushed and tested, but once it is we can overwrite that repo-level secret with the new token and delete the secrets that exist on the environment level.

## Git Issues

Closes #1829

PR #2025, opened 2023-08-14
Refactor/gh actions

chrismclarke:
PR Checklist

## Description
During #2019 an additional line of code is required for all CI builds (to compile data-models folder before executing any scripts). This is a slight pain as most existing actions have had their build code copy-paste and require a large number of small changes. Additionally while trying to make changes the code is quite confusing and fragile due to subtle changes between files and methods (e.g. 2 separate methods for generating git SHAs and extracting deployment names)

This PR aims to improve the CI system by use of common templates (as created for deployment repos), specific changes:

- Add common web-build action to be used across all other actions. This includes logic to set deployment based on various conditions (e.g. target branch for PRs or user input for android deploy action)
- Add common android-build action (which also uses web-build) for basis of android actions. This can now also be adapted for deployment-repo templating
- Begin process or refactoring all actions to use reusable `web-build` or `android-build` actions
- Add concurrency groups to all actions to cancel in-progress actions when retriggered (e.g. new commit pushed to pr branch)

**Additional changes**

- Add improved yarn node_module caching (faster ci builds)
- Add support for all actions to target deployments based on branch
- Upgrade various actions to use node 18 (instead of 16) and updated 3rd party runners
- Add more structured comments to most actions
- Deprecate legacy `sourcemaps-upload` and `test-e2e` actions

## Review Notes
There could be potential knock-ons if any actions not migrated correctly, so would be good to test each of the actions modified work as expected. 

The key thing to test for is that actions use both the intended deployment and intended source code, as there are a lot of subtle cases of using different branches and deployment combinations. Will also want to confirm git SHAs are populating as expected

Current status of existing actions

**Migrated and tested (Chris)**
- [x] web-build (newly added)
- [x] test-build
- [x] test-preview
- [x] android-build
- [x] test-appetize
- [x] test-visual-compare

**Migrated (pending test)**
- [x] android-build-dispatch (newly added, manually triggered so will test post-merge)
- [x] android-release (migrated, manually triggered so will test post-merge)
- [x] deployment-hosting (migrated, to confirm best deployment to test on, i.e. one not in active use)
- [x] test-visual-generate (migrated, auto-runs on master post merge)

**Not Affected**
- [x] documentation-deploy
- [x] documentation-lint
- [x] release-drafter

**Deprecated**
- [x] sourcemaps-upload (have deprecated, upload step integrated into deployment-hosting action so only used in manual testing. Can be migrated in future if needed)
- [x] test-e2e (have deprecated, never fully implemented and will need revisit in future)

**Further Testing**
Would be good to test deployment-specific actions via PR to deployment branch post merge (ideally against a deployment that is not in production use)

## Git Issues

Closes #

## Screenshots/Videos


.github/workflows/test-appetize.yml
```diff
@@ -1,83 +1,46 @@
...
+      debug-apk: true
+      release-bundle: false
+   
   test_appetize:
-    needs: set_deployment_name
-    if: contains(github.event.pull_request.labels.*.name, 'Test - appetize')
+    needs: build
     runs-on: ubuntu-latest
-    environment: 
-      name: ${{ needs.set_deployment_name.outputs.DEPLOYMENT_NAME }}
```
jfmcquade, line 33: Previously we were setting the github environment to match the deployment name in order to use the relevant appetize secrets for each deployment, configured [here](https://github.com/IDEMSInternational/parenting-app-ui/settings/environments). As in #1864. 

Are these deployment-dependent `APPETIZE_TOKEN`s still supported? Or is the thought that this won't be needed once deployments exist outside of the code repo?

jfmcquade (commented):
Really nice to have this logic cleaned up, with the new pattern of extracting environment variables at the top. I haven't meticulously looked through every line, but it makes sense to me skimming through.

I have added one comment regarding the deployment-specific Appetize related secrets that we configured in #1864.

.github/workflows/test-appetize.yml
chrismclarke: nice catch! I think I naively saw `environment` next to `env` and thought it was a legacy/typo issue, had forgotten all about the deployment-dependent secret environments you set up.

Have reverted and added an additional comment accordingly

chrismclarke:
Thanks @jfmcquade - I've addressed the comment so hopefully should be good to go

jfmcquade (approved):
Looks good.

As mentioned in the PR description and discussed on the call, after merging it would be good to test deployment-specific actions via a PR to a `deployment/*` branch. @esmeetewinkel I think you said that the opportunity to test this should arise organically soon

PR #2028, opened 2023-08-16
Feat: component icon defaults

### LINKED ISSUE #2007: [BUG] Components with missing default icons
**Describe the bug**
Some components (such as the audio player) do not make sense without icons. Therefore, when the icons are not specified the default should be to populated these with icons that are not deployment-specific. We propose to use ionicons for these. Ideally a spreadsheet author can override these defaults through the parameter list, as done for instance in the accordion component. The overrides could be ionicons or paths to icons in the assets Google Drive. 

**Relevant components**
- audio player
I think that currently the play/pause icons comes from the Google drive (hardcoded path `plh_images/...`),  the rewind/fast forward icons sit in an `assets` folder directly in the code repo, and the mute/unmute icons come from ionicons. As a consequence, the play/pause icons are missing when the audio player is used on a deployment that doesn't contain the `plh_images` asset folder.
https://idems-debug.web.app/template/example_audio_player
- timer
https://idems-debug.web.app/template/comp_timer
- accordion 
https://idems-debug.web.app/template/example_accordion

**App version**
v0.16.17

### LINKED ISSUE #2034: Accordion component gives error about state
**Details**
In the [learn_play](https://docs.google.com/spreadsheets/d/1N8ObxOpDZv_dtzrYu16OyOfpkTK_HsuPAZPPXXCnZpk/edit#gid=1612329363) template in the `plh_tz` deployment, errors appear about the state of the accordion sections (which I believe means whether they are open or closed). Although these errors are being thrown, there is not really a noticeable bug in the app. 

**App version**
v0.16.18

**To reproduce**
https://plh-tz.web.app/template/learn_play

**Screenshots and videos**
image
image

jfmcquade:
PR Checklist

## Description

The `audio`, `timer`, and `workshop_accordion` components all use icons which had hardcoded paths to the source image, either referencing a file from the deployment asset folder (nested within a `plh_images` folder), or to a file that was directly included in the repo source code. Each of these components now makes use of [ion icons](https://ionic.io/ionicons), with a default icon chosen to match the original, and also allows the icon to be overridden by a custom image supplied at the template level.

Previously, these icons displayed as broken images for deployments which did not include them (and did so in a way that matched the original `plh_images/` folder structure), including the debug deployment. In these cases, the default icons should now display instead. The `parentapp_kids` deployment branches previously had their own additional commits that changed the hard-coded icon paths within the code. This will no longer be necessary after this PR.

There may be some minor visual differences in the component icons, noted below.

Also includes a quick fix for #2034 

## Testing notes

Checkout the PR feature branch and set the `debug` deployment to active. Run `yarn workflow sync` to sync the latest debug content. Then, examine the respective component/example templates for each component:
[comp_audio](https://docs.google.com/spreadsheets/d/1utD4Kr0WVjvDwnMbg2wYupaPQ0VY0H2csb86fyi4bPI/edit#gid=603128459)
[comp_timer](https://docs.google.com/spreadsheets/d/1jQAvELG5EXj6zgj0bBQmVg7Iaa-66OpemdZzgOh2PP0/edit#gid=569531329)
[example_workshops_accordion](https://docs.google.com/spreadsheets/d/1gfnWmr59DF0ZsuSuBh0K1o9fMqEg-z6Qa7eJNsiB-rQ/edit#gid=1595229908)

### audio component
New template: [comp_audio](https://docs.google.com/spreadsheets/d/1utD4Kr0WVjvDwnMbg2wYupaPQ0VY0H2csb86fyi4bPI/edit#gid=603128459)

#### Previous default (taken from [plh-global web preview](https://plh-global.web.app/template/box_audio))
Screenshot 2023-08-16 at 16 47 43

#### New default, and example with icon overrides for "sharp" ion icons (live debug template [comp_audio]
comp_audio

#### New parameters
The following parameters are now exposed to template authors, which can be used to override their respective icons:
- play_icon_asset
- pause_icon_asset
- forward_icon_asset
  - Only one asset is provided for both the forward/backward buttons, and is reflected horizontally as appropriate
- help_icon_asset

The "mute" icon and functionality has been removed, as this didn't seem appropriate on an audio player

#### Previous default (taken from [plh-global web preview](https://plh-global.web.app/template/box_timer))
Paused
Screenshot 2023-08-16 at 16 51 02
Playing
Screenshot 2023-08-16 at 16 51 07

#### New default (live debug template [comp_timer](https://docs.google.com/spreadsheets/d/1jQAvELG5EXj6zgj0bBQmVg7Iaa-66OpemdZzgOh2PP0/edit#gid=569531329))
Paused
Screenshot 2023-08-16 at 16 54 03
Playing
Screenshot 2023-08-16 at 16 54 08
NB, mainly due to the fact that I couldn't find an ion icon that neatly fitted around the "+1" text when the timer is playing to indicate adding another minute to the running time, I opted to remove the icon altogether when the timer is running, in favour of just displaying "+1". This is what the native timer app does on Android: there is no icon accompanying the "+1" button.

#### New parameters
The following parameters are now exposed to template authors, which can be used to override their respective icons:
- play_icon_asset
- pause_icon_asset
- reset_icon_asset

#### Previous default (rom [plh-global web preview](https://plh-global.web.app/template/weekly_workshops))
Screenshot 2023-08-16 at 17 16 22

#### New default (live [example_workshops_accordion](https://docs.google.com/spreadsheets/d/1gfnWmr59DF0ZsuSuBh0K1o9fMqEg-z6Qa7eJNsiB-rQ/edit#gid=1595229908) template)
Screenshot 2023-08-17 at 13 12 53

NB - I'm not sure if there are any live apps that actually use the `workshops_accordion` component, but if there are then in order to keep the same icons identical to how they currently appear, the icon paths should be passed in manually as parameters to each nested `accordion_section` component, i.e.:
```
completed_icon_asset: plh_images/icons/tick.svg;
in_progress_icon_asset: plh_images/icons/in_progress.svg;
disabled_icon_asset: plh_images/icons/temporarily_disabled.svg;
```
Otherwise, the new defaults will be used, which is probably acceptable. Here are the new default icons:

Screenshot 2023-08-16 at 17 23 47 In progress

Screenshot 2023-08-16 at 17 24 18 Completed

Screenshot 2023-08-16 at 17 24 41 Temporarily disabled

#### Further code tidying
- The code for the "workshop_accordion" component could be further tidied up and generalised, as it seems to have been made for a very specific use case (the weekly workshops page). However that isn't a priority at the moment
- The base "accordion" component, as seen in the template [example_accordion](https://docs.google.com/spreadsheets/d/1gfnWmr59DF0ZsuSuBh0K1o9fMqEg-z6Qa7eJNsiB-rQ/edit#gid=63099847), already uses ion icons and supports overriding the icons with a value relating the name of another ion icon. This could be reworked, but again, this doesn't feel like a priority
  Screenshot 2023-08-16 at 17 25 55

## Future steps

- It would be possible to programmatically change icons across the app between ionic variations (outline, fill, sharp) by exposing a single parameter that can be set at the theme level. 
  - Currently, we use a mixture of outline and solid icons depending on context, e.g. in the audio player component, so this feature could probably not be applied universally without further layers of complexity. But perhaps our current use of icons is not well thought through anyway.
- In future, it would be good for all icons to have a default then be able to take a value which can be _either_ an ion-icon name, _or_ a path to an arbitrary svg file

## Git Issues

Closes #2007 
Closes #2034 

## Screenshots/Videos


esmeetewinkel (approved):
This is great, thanks! Functional test passed. Minor comments below. 

> allows the icon to be overridden by a custom image supplied at the template level

> it would be good for all icons to have a default then be able to take a value which can be either an ion-icon name, or a path to an arbitrary svg file

I'm understanding that for now overrides need to be asset paths, but could also be ionicons in the future. Happy with that. 

**Audio player mute/unmute** and question mark
What did you do about the mute/unmute icon? I'd be happy to take that functionality out entirely since it doesn't make sense (you can just pause the audio). 
As for the question-mark-in-circle, I don't think it's in use at the moment, so we could just take it out. It could be sensible so I'm also happy to keep, but then the default should come from an ionicon.
image

**Play icon on round button (timer and audio)** 
I _know_ our graphic designer will hate that the triangle doesn't seem to be centred on the round button, since the distance of the corners of the triangle to the circle is not the same. Is there any solution we can propose for this? Adapt the design so that the triangle doesn't sit on a round button?

src/app/app.module.ts
```diff
@@ -68,6 +69,7 @@ export function lottiePlayerFactory() {
     Device,
     httpInterceptorProviders,
     { provide: ErrorHandler, useClass: ErrorHandlerService },
+    PLHAssetPipe,
```
chrismclarke, line 72: nit(blocking)
I'm confused why we are declaring the pipe as a provider in the main module, is this an import error or do we need for some reason (if so maybe a comment would help)

src/app/shared/components/template/components/audio/audio.component.ts
```diff
@@ -9,7 +9,39 @@ import { Howl } from "howler";
...
+// Will be overridden if user provides values for play_icon_asset, pause_icon_asset or forward_icon_asset
+const PLAY_ICON_DEFAULT = "play-outline";
+const PAUSE_ICON_DEFAULT = "pause-outline";
+const FORWARD_ICON_DEFAULT = "play-forward";
+
+interface IAudioParams {
+  /** TEMPLATE PARAMETER: "src". Will be overridden by a value passed as "value" to the component. */
+  src: string;
+  /** TEMPLATE PARAMETER: "play_icon_asset". The path to an svg to override the default play icon. Default icon is ion's "play-outline" */
+  playIcon: string;
```
chrismclarke, line 24: nit(non-blocking)
would it make life any simpler keeping template parameter names and local param names the second? I know we would lose a bit in naming convention, but probably less prone to typos that way?
But otherwise definitely like how you've kept all the template params together in a single interface

chrismclarke:
Overall looks pretty good to me, I've added a couple minor comments but nice to see the tidying up and being able to remove the assets.

As a quick note on the workshop accordion, Do we know longer have the linking line because it was directly drawn on the svg? If so it might be worth creating a div with a horizontal line that can float behind all the step images to recreate (assuming we are still using this component and want the connection between steps)

src/app/app.module.ts
jfmcquade: I wanted to use the pipe in the component files, e.g. [here](https://github.com/IDEMSInternational/parenting-app-ui/pull/2028/files#diff-0511da63615ec05d036268319070a77820fd7aa64b314c90113c8d23fb19718dR115) in the audio component file. Using the pipe just in the html would lead to complicated logic (we want to check if the author has provided an icon override path and only apply the asset pipe to the icon path property if they have, but we also want to use a different icon path depending on whether the audio is playing, in order to display the play/pause icon).

Previously, that file imported the `TemplateAssetService` itself, which is used by the `plhAsset` pipe, but importing the pipe seemed like a neater way to do it to avoid code duplication. 

In order to inject the pipe, it needs to be declared in a relevant module. I've now moved it to the [template module file](https://github.com/IDEMSInternational/parenting-app-ui/pull/2028/files#diff-2ad3d485b13847f8402e2242cd65f73e57242ce3f8c1c0672e62a2718a0f63caR48), which seems like a better place than the main module. 

Alternatively, we could avoid injecting the `plhAsset` pipe directly, which would mean that it wouldn't need to be declared as a provider. But in that case, we would still need to import and inject the template asset service into our component in order to pass it to the constructor of the pipe, something like this:
```
constructor(private templateAssetService: TemplateAssetService) {
  super();
  this.plhAssetPipe = new PLHAssetPipe(this.templateAssetService)
}
```

I'm not clear on what the best option would be, but I can see how we might want to potentially inject other pipes into components in the future, so having them [declared as providers in the template module](https://github.com/IDEMSInternational/parenting-app-ui/pull/2028/files#diff-2ad3d485b13847f8402e2242cd65f73e57242ce3f8c1c0672e62a2718a0f63caR48) doesn't seem unreasonable.

jfmcquade:
Thanks, both. I've addressed your comments.

@chrismclarke see my replies to your inline comments

@esmeetewinkel:
> I'm understanding that for now overrides need to be asset paths, but could also be ionicons in the future. Happy with that.

That's right, currently overrides can only be full paths to .svg files (will always be required in case that desired icon doesn't exist in ion icons), but it would be good to support override names of specific ion icons in future

## Audio player changes
- Removed mute button
- I've kept the "help"/tooltip button, but moved into corner where mute used to be, and exposed an icon override via a new `help_icon_asset` param
- Added some padding to bump the "play" icon to the right
  - Whilst this will mean that an author's "play" icon will appear off-centre if they use something that is not a triangle, in this rare case the user can edit their icon itself to ensure it still appears centred. The other option seemed to be to not apply this padding to the author's custom icon, but in that case we would be expecting the author to add their own padding to the triangle icon anyway, and I imagine this will be much more common than someone wanting to use a non-triangle play icon.
  comp_audio

src/app/shared/components/template/components/audio/audio.component.ts
chrismclarke: Sounds good to me, yeah happy to just try be consistent then with author `snake_case`, code `camelCase` and using same words.

chrismclarke (approved):

PR #2032, opened 2023-08-17
Feature/animated section styles

### LINKED ISSUE #1677: [FEATURE] Animated section styles
**What?**
Currently the default behaviour of an animated section is to fade in the next section. This could be moved to a parameter called "animation" (default: `fade_in`), which could also be changed to `none`. 

See also https://github.com/IDEMSInternational/parenting-app-ui/pull/1675

kwAsant:
PR Checklist

## Description

Allows for the option that there is no animation between slides. Originally, there was only the fade-in and fade-out options. However, this allows for none to occur at all.

## Git Issues

Closes #1677 

## Screenshots/Videos

Should see videos conveying the consequence of a no-fade animation happening for all three slides in the template plus one where there is a no-fade animation sandwiched between two fades:

https://user-images.githubusercontent.com/88288104/262080859-4e822854-68e7-4821-b0d3-50af744f9574.mov

Here we can see the transition between a fade and no-fade, and then a transition between no-fade and another no-fade.


src/app/shared/components/template/components/animated-slides/animated-slides.component.ts
```diff
@@ -3,8 +3,15 @@ import { PLHAnimations } from "src/app/shared/animations";
 import { getNumberParamFromTemplateRow, getStringParamFromTemplateRow } from "src/app/shared/utils";
 import { TemplateBaseComponent } from "../base";
 
+// create an enum that stores the values of the animation styles
+// at the moment (until later updated) there is two: fade and none
+export enum AnimationStyle {
```
jfmcquade, line 8: I don't think this is being used, so can be deleted

src/app/shared/components/template/components/animated-slides/animated-slides.component.ts
```diff
@@ -17,18 +24,37 @@ export class TmplAnimatedSlidesComponent extends TemplateBaseComponent implement
...
     this.getParams();
     this.runFade();
   }
 
+  // it appears that the purpose of getParams is to set the values of the variables
+  // and to set the following properties:
+  // skipText
+  // displayTimes
+  // style
+  // add another feature called slideParams which would store the duration and animation
```
jfmcquade, line 39: remove comments

src/app/shared/components/template/components/animated-slides/animated-slides.component.ts
```diff
@@ -17,18 +24,37 @@ export class TmplAnimatedSlidesComponent extends TemplateBaseComponent implement
...
   getParams() {
     this.skipText = getStringParamFromTemplateRow(this._row, "skip_text", "Skip");
+    console.log("***this.skipText***", this.skipText);
```
jfmcquade, line 42: remove all debugging `console.log()`s

jfmcquade (changes requested):
Looking good, I think using a `slideParams` array is sensible.

I've added some comments requesting to remove some logs and comments.

I've added some placeholder images to the [feat_animated_slides](https://docs.google.com/spreadsheets/d/15eQ1B7BFQKQVdAM1mD1qPa_vGrudfylYVVDOtPenkWI/edit#gid=569531329) template, so the example is a bit nicer to look at.

In terms of functionality, I think there's an issue when only one of the slides has `animation: none`. In this case, the slide in question appears as soon as the previous slide starts fading out. I think it should wait until the previous slide has finished its "out" animation before it appears. See video below:
https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/6acdb8e2-d144-485c-aa4a-30e5744bddb6

Also, I would recommend using Chrome for testing your local build, with devTools open and the "device toolbar" toggled on, so that the preview most closely resembles how the app will appear on phones.

src/app/shared/animations.ts
```diff
@@ -9,6 +9,11 @@ const fadeInOut = [
     transition("in => out", [animate("0.5s")]),
     // todo - could use :enter and :exit properties also (need to confirm doesn't break existing functionality)
   ]),
+  trigger("noFade", [
+    state("in", style({ opacity: 1 })),
+    state("out", style({ opacity: 0 })),
+    transition("* => in", [animate("0s")]), // this line is to allow for a prior fade to actually occur
```
jfmcquade, line 15: In order to wait for the previous slide's "out" animation to finish, I think potentially we want to wait 0.5s before our slide appears. 

See [these docs](https://angular.io/guide/animations#animation-metadata-duration-delay-and-easing), it may be possible to achieve by using the "delay" argument of `animate()`, along the lines of: `animate("0 0.5s")`

kwAsant: Addressed: created a new setting for animation. This should be better?

https://github.com/IDEMSInternational/parenting-app-ui/assets/88288104/70cee643-16d3-4cc2-b5a3-0bf53350ad7f

jfmcquade: That definitely looks better in my opinion. It would be worth testing how it looks when every slide has the parameter `animation: none`. Try updating the [feat_animated_slides](https://docs.google.com/spreadsheets/d/15eQ1B7BFQKQVdAM1mD1qPa_vGrudfylYVVDOtPenkWI/edit#gid=569531329) template and see what it looks like. I suspect that with all of them not animated, the gap will look too long, you may want to reduce it from 1.5s to 0.5s, for example

kwAsant: Shortened it, looks like this now:

https://github.com/IDEMSInternational/parenting-app-ui/assets/88288104/4e822854-68e7-4821-b0d3-50af744f9574

src/app/shared/animations.ts
```diff
@@ -9,6 +9,13 @@ const fadeInOut = [
     transition("in => out", [animate("0.5s")]),
     // todo - could use :enter and :exit properties also (need to confirm doesn't break existing functionality)
   ]),
+  trigger("noFade", [
+    state("in", style({ opacity: 1 })),
+    state("out", style({ opacity: 0 })),
+    transition("* => in", [
+      animate("0s 0.5s", style({ opacity: 0 })), // delay animation for 1.5s
```
jfmcquade, line 16: You could update this comment and make it a bit more descriptive, e.g.: "delay animation for 0.5s in order to allow previous animation to finish.

src/app/shared/animations.ts
```diff
@@ -9,6 +9,13 @@ const fadeInOut = [
     transition("in => out", [animate("0.5s")]),
     // todo - could use :enter and :exit properties also (need to confirm doesn't break existing functionality)
   ]),
+  trigger("noFade", [
+    state("in", style({ opacity: 1 })),
+    state("out", style({ opacity: 0 })),
```
jfmcquade, line 14: Do we still need these `state()` functions if we're not actually animating? You could try removing them and see if it still works. If not, we may be able to remove the `style()` functions, or at least the reference to opacity, as we're not interested in that property.

jfmcquade (changes requested):
A couple more minor comments. Plus it would be good to update the 2 videos in the PR description to reflect the up-to-date behaviour now that the edits have been made. After that, it should be good to go.

src/app/shared/animations.ts
kwAsant: Realised that I didn't need state(out) but did need state(in). The style is just for the disappearing, I believe.

src/app/shared/animations.ts
kwAsant: Added comment.

jfmcquade (approved):
LGTM

esmeetewinkel (approved):
Great! Functional test passed.

Issue #1266, opened 2022-03-08
[FEATURE] Data list overrides


esmeetewinkel:
**What?**

- [x] Longer term ideally we want a system where anything (including data lists) can be overwritten at runtime, however it will only make sense to implement once we have a single system that handles loading data (otherwise it means a lot of copy-paste for templates, data lists, and any other flow type). This is part of what is proposed with other issues related to integrating json instead of ts files

- [x] Additionally, because these datalists are statically defined we could also integrate a system that replaces files during compile so that only the relevant datalist ever makes it into the app. This would be quite easy to do once we've integrated pr #1210 (moved to #2081)

**Other details**
See also issue #1261

esmeetewinkel:
@chrismclarke Can we close this considering the work that went in on data items, or would you say this is something different?

chrismclarke:
I think it might still make sense to have long term so that data lists can be overridden in the same way that templates are (e.g. theme/language-dependent data lists, instead of relying on templates to know all the variations). So I'd say keep open for now, but low priority

jfmcquade:
@ChrisMarsh82 @esmeetewinkel This relates to your discussion

ChrisMarsh82:
@chrismclarke How difficult would this be to do? I was hoping for the facilitator to have all the same templates but have overwritten data_lists for each deployment.

chrismclarke:
I tried giving a quick go in #2040, although a few extra threads started to unravel and I didn't have time to finish. Will keep you updated as I progress (would guess about a half day's work if I can find the time)

chrismclarke:
As a quick update most of the backend code is in place, however the frontend code is a bit more tricky as most of the logic that handles dynamic variables (e.g. `@field.some_field`) is tightly coupled with the templating system and requires significant refactor to generalise. 

I've started work on the refactor and things seem to be mostly working, however will likely wait until more of the current backlog of PRs is merged as quite prone to conflict (particularly #2019, #2035 and #2041)

chrismclarke:
Close first point with #2040 and move second to #2081

PR #2040, opened 2023-08-19
Feat/datalist runtime override

chrismclarke:
PR Checklist

## Description
Add support for overriding data_lists at runtime with `override_target` and `override_condition` fields, in the same way as templates

**Main Changes**
- [x] Add backend code to support data_list overrides
- [x] Create backend spec test for data_list overrides
- [x] Create content test sheet for data_list overrides (or update existing if similar exist)
- [x] Add frontend code to support data_list overrides
- [x] Add frontend spec tests
- [x] Copy local debug sheets to shared drive
- [x] PR Demo
- [x] (Blocked) Builds on #2041 - wait for merge

**Additional changes**
- Add new methods for dynamic content evaluation to be used outside of templating system
- Add new methods for identifying dynamic content at runtime (outside of templating system)
- Tidy logging code to distinguish file-based and console-based logging
- Improve logging test integration
- Fix QC check for duplicate flows names within a given flow type
- Remove deprecated script code

## Review Notes
I've updated [example_overrides](https://docs.google.com/spreadsheets/d/1MpoH3BxhECZRmYM10HZ0pTOoe69FJ-fEW9FwzK-Q6yw/edit#gid=1745157248) sheet in debug repo to include an `example_list_override` template with 2 corresponding data lists (shown in demo below)

This will require a sync as parser methods have been updated, `yarn workflow sync`

## Author Notes
Override functionality is still implemented within individual flow types at the back-end (with template and data_list both calling shared method to implement). This means that the functionality is not available for flows that have been generated via data_pipe or generator flows, nor flow types such as globals or tours . We may wish to consider making a low-priority to address this in the future if we want similar override system on all flow types (may not be required if most authoring to be focused around sheets and lists)

There is also a limitation currently, that the only dynamic expressions supported in conditions are `@field` notation. This is due to the fact all other syntaxes (`@global`, `@local` etc.) are coupled tightly with the templating system logic and would require larger refactor to make available. For now just the field logic has been duplicated

## Dev Notes
In order to allow dynamic data evaluation outside of templating, a new `appDataEvaluator` service has been created. This currently only supports `@field` syntax and is mostly a duplicate of the template field service, but should form a base to extract additional methods in the future and eventually move all dynamic context handling out of the templating system. Service spec tests have also been created, and can be run directly via `yarn ng test --include="src\app\shared\services\data\app-data-variable.service.spec.ts"`

Additionally new methods have been added to identify a list of all dynamic variables used within a flow/sheet. Currently this is only used as a 2-part extract variables and evaluate method in the appDataEvaluator, however could be used in the future to extract more data during parsing (replace/combine with #1223) or to keep a list of realtime variable changes (alongside #1714).

Curiously in the example sheet the buttons don't align in the same way - likely due to some confusion from the ng:deep left selector (seems buttons by default align center but inner text aligns left and the 2 sets of rules can conflict in inconsistent ways - see output in demo video). Recommend to review

## Git Issues

This addresses the first part of #1266

## Screenshots/Videos
**Example sheet - items that can be updated by setting data list dynamically (based on override condition)**
image

image

[Debug App.webm](https://github.com/IDEMSInternational/parenting-app-ui/assets/10515065/532cdff4-3224-4bf1-836e-4c10e0725a99)

**Example inputs and outputs for new method and tests to extract list of context variables**
image

**Example output logs generated from duplicate flows**
image


chrismclarke:
Thanks for the review @jfmcquade , can confirm I'm also now seeing the error so can debug (guessing one of the incoming PRs changed some of the service mocking).

Would be nice to get the full ng test working (even if only partial coverage) so that we can include the tests against PRs, but probably a job for another day

chrismclarke:
Ng testrunner should now be passing @jfmcquade

packages/shared/src/models/templatedData/templatedData.spec.ts
```diff
@@ -118,8 +121,64 @@ describe("Templated Data Parsing", () => {
...
+    {
+      input: "hello @row.value_1 @row.value_2",
+      output: { row: { value_1: true, value_2: true } },
+    },
+    {
+      input: "hello @row.value_1.child_property",
+      output: { row: { value_1: true } },
+    },
+    {
+      input: "@unkonwn_prefix.value",
```
esmeetewinkel, line 143:
```suggestion
      input: "@unknown_prefix.value",
```

src/app/shared/services/data/app-data-variable.service.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+  it("evaluates expressions", async () => {
+    await service.set("field", "test_value", "2");
+    const evaluated = await service.evaluateExpression("@field.test_value < 3");
+    expect(evaluated).toEqual(true);
+  });
+
+  //   TODO - requires mock error logger pending refactor in open #2019
+
+  // it("logs error on unknown prefix", async () => {
+  //   await service.set("unkown" as any, "test", "value_1");
```
esmeetewinkel, line 79:
```suggestion
  //   await service.set("unknown" as any, "test", "value_1");
```

esmeetewinkel (approved):
Functional test passed. I noticed two typos in the code, which I've suggested corrections for. 

I'm understanding that this PR only closes the first part of the issue, and the second part will be a follow up PR, so I've edited the PR description so that the issue remains open.

chrismclarke:
Thanks both, I've added a couple small additions to the existing overrides documentation and will merge once tests passing
86ef2fdf07348e13a45e414c62bd49a769b352b5

@esmeetewinkel - I think we can probably go ahead and close the issue as the second point of overriding at compile time should be covered by multiple data sources proposal in #2081
I've also included this case in the updated documentation

PR #2041, opened 2023-08-19
fix: sync_local workflow

chrismclarke:
PR Checklist

## Description

Fix issue where the `yarn start:local` script would not correctly run the `sync_local` workflow - the arg to specify `--skip-download` for sheets and assets was not being passed on correctly, so was attempting to sync from google drive but fail as the local deployment used does not have a google drive id.

This appears to be a regression issue, hard to know exactly when/how introduced

## Dev Notes
~This will conflict with #2019, so depending on merge order changes should be adapted accordingly~
Updated

~Following merge of #2019 it may also be useful to add a set of spec tests to the new workflows package to try capture such regression issues in the future (e.g. basic tests just to make sure a set of nested test workflows can run and parse args as required)~ 
See https://github.com/IDEMSInternational/parenting-app-ui/issues/2059

## Git Issues

Closes #

## Screenshots/Videos


esmeetewinkel (approved):
I confirmed that this is broken on master when running `yarn start:local` or `yarn workflow sync_local`
image
image

On this PR branch, `yarn start:local` goes through correctly, producing a local preview including a change I made to a local sheet. It seems that `yarn workflow sync_local` "hangs" at "watching for changes to files..." - should this workflow be completing or is it more like `sync_watch` i.e. it keeps watching for file changes?
image

jfmcquade (approved):
When functionally testing, I first encountered this error, which was due to running `yarn start:local` with a deployment other than `local` set to active (in this case `debug`):

Screenshot 2023-08-23 at 16 54 47

After running `yarn workflow deployment set local`, I then received the following errors after running `yarn start:local`:
Screenshot 2023-08-23 at 17 00 22

This was resolved by running `yarn workflow sync` to populate the `contents.json` files.

If easy, it could be good to handle these errors, however as this functionality is only intended for devs it probably isn't required.

The functionality seems to be working as intended, with updates to the `demo.xlsx` reflected after causing a live reload

chrismclarke:
Thanks both

@esmeetewinkel 
> yarn workflow sync_local "hangs" at "watching for changes to files..." 
Indeed it behaves like the `sync_watch` scripts and will continue to watch for any changes until terminal closed

@jfmcquade 
Good suggestions, the documentation mentions local workflow needs to be specified but I can imagine that is something easy to forget to do. As for the missing files, I'm not sure whether this is due to changing deployment whilst already running or something that happens to all users running for the first time (haven't tried replicate myself, but will add both comments to follow-up issue #2062)

PR #2056, opened 2023-08-22
Correct nginx configuration to allow auto tls certificate renewals

istride:
# Problem

TLS certificate for production deployment was not being automatically renewed.

# Cause

The `nginx-certbot` container processes configuration templates and files when it starts up. The output of this process is stored in `/etc/nginx/conf.d` within the container. The Docker Compose configuration may have been interfering with this process by:

- Using configuration template files and writing output directly into `/etc/nginx/conf.d`
- Mounting a directory from the host to `/etc/nginx/conf.d` in the nginx container

# Proposed solution

To work with the nginx container's built-in startup processes. Templates are still processed but output is written to `/etc/nginx/user_conf.d`, which is where the container expects custom configuration files.

The volume mount to `/etc/nginx/conf.d` has been removed.


chrismclarke:
Code makes sense to me, although it might still be good to wait and make sure the auto-renew happens.

Looking at the apps-server site I can see latest cert renew was August 22nd (when the PR work was done), but can be assumed as manual renew as likely involved various additional steps (e.g. container restarts)


According to the docs the container should [check for renewal](https://github.com/JonasAlfredsson/docker-nginx-certbot/blob/master/docs/good_to_know.md#renewal-check-interval) every 8 days, although I think will only renew if the cert is within 1 month of expiry (so post October 20th). So I think we just see if auto-renew happens around then and merge on success. Let me add @jfmcquade to reviewers so that they can approve functional test

image

chrismclarke (approved):

jfmcquade:
Thanks, @chrismclarke. I'll functionally test by checking for certificate renewal after October 20th, and will come back and review this PR accordingly.

jfmcquade:
The certificate has currently *not* been renewed. However it seems possible that up to 8 days may need to elapse whilst we're within the 1 month from expiry window. I will check back in 8 days and report here.
Screenshot 2023-10-23 at 11 26 04

istride:
The logs for Let's Encrypt in the Nginx container show that the last time a renewal was attempted was on 17 October.

```
2023-10-17 02:00:33,343:DEBUG:certbot._internal.display.obj:Notifying user: Certificate not yet due for renewal; no action taken.
```

Logs can be found at `/var/log/letsencrypt`.

chrismclarke:
image
Looks like it was renewed on the 24th

PR #2080, opened 2023-09-04
Fix/ng tests

chrismclarke:
PR Checklist

## TODO

- [x] Builds on top of #2040, should merge after
- [ ] Fix or remove all component test specs to include basic row value if parsing parameter_list in ngOnInit
- [ ] Fix service injection issues, creating new mock services to inject where required
- [ ] Fix other test issues
- [ ] Add ng test to web-build action tests
- [ ] Add support for ng test coverage
- [ ] Consider migrating component tests to cypress or other runner (or keeping component tests but adding back e2e)
- [ ] Start to add more meaningful component spec tests (could be follow-up)

## Description
Angular has a component and service testing system which is included during new component and service creation, however only partially used for some of the newer services added to the platform. 

This PR aims to fix and remove legacy code to allow the testrunner to function, and include any working tests within existing github actions testing pipelines. 

https://angular.io/guide/testing

## Git Issues

Closes #

## Screenshots/Videos
**Significant number of test issues to fix**

image

image


chrismclarke:
@jfmcquade - #2040 now merged so diffs should be much clearer to see minimal changes required to add _row test data for an example component test in case of interest

Issue #2048, opened 2023-08-22
[FEATURE] Add support for managing assets with Canto DAM


jfmcquade:
**What?**
Deployments should be able to manage their assets using Canto.

RFC to follow

Issue #1520, opened 2022-09-29
[Docs] Document the processing order of sheets


jfmcquade:
This feels sensible to me, can see the function comment asserts that postprocessing methods should be called by flowtype but appears to be omitted. 

What I think would be most useful is trying to capture some documentation/flowchart on the processing order. I think it should actually be fairly straight-forward as the processing order is just something along the lines of

- Apply xlsx converter to extract flows from xlsx files.
- Apply default parser by flow, applying any `@default` or `@row` statements, convert _list and _collection columns, removes meta fields, handles translated fields, processes nested groups etc.
- Apply methods from extended parsers (e.g. data_list, data_pipe). These assign overrides per row, per flow, and across all processed flows

There is also a side flow that explains how caching and queues are used, but again reasonably straightforward (generally tracked by file md5 checksum, so that if a file contents are unchanged cached results used; exceptions possible, currently used for data_pipes which depend on child inputs so cache ignored)

If not creating now, might be good to ensure an issue is created (possibly linking to this comment for some reference)

_Originally posted by @chrismclarke in https://github.com/IDEMSInternational/parenting-app-ui/pull/1518#discussion_r982601174_

esmeetewinkel:
This might be something Ed could get involved in

PR #2088, opened 2023-10-01
feat: sheets multiple folder ids

### LINKED ISSUE #2081: [FEATURE] - Deployment multiple gdrive folders
**What?**
Add support for providing multiple multiple google drive template and asset folders to sync data from as part of deployment config

**Why?**
Make it easier to deploy mulitple apps with some common sheets

**How?**
Likely provide `config.google_drive.sheets_folder_ids` property (which could be used instead of `sheets_folder_id`)
Alternatively could consider more generic config property, such as 
```ts
data_sources: [
   { 
        provider:'google_drive',
        type:'template', 
        config:{id:'example_gdrive_id'}
    }
] 
```

**Other details**
- It would be more useful for users to be able to override individual flows detected in override folder even in cases where folder structures and filenames do not match (as the flow names do match), so will need to hold off merging/overriding flows until they have been extracted from sheets. No such problem for assets

- As this will require changes to various scripts may need to either wait post #2061 or try to avoid too many conflicts

chrismclarke:
PR Checklist

## Description
Adds support for specifying multiple google drive sheet and asset `folder_ids`, to allow syncing content from multiple data sources.

## Author Notes
All deployment configs should now specify an array of google drive `sheets_folder_ids` instead of a single `sheets_folder_id`, and an array of `assets_folder_ids` instead of a single `assets_folder_id`. A warning will be displayed if not the case, but should still continue to work as before (functionality deprecated but not removed).

While working on code it appears that `google_drive.sheets_filter_function` wasn't being properly passed to google drive sync scripts. I'm not sure if any deployments were using this, but would be good to ensure everything working as intended if they are. This function filters based on google drive entry (e.g. folder/file names), and shouldn't be confused with the `app_data.sheets_filter_function`, which is in use and filters based on the flow data.

## Review Notes
The changes are designed to be backwards-compatible, so if syncing content from an existing repo there should be no changes to sheet or asset files and contents beyond a warning to update the config. Likely it would be good to functionally test on a repo with fully synced data

A separate config has been created as the `debug_gdrive_multiple` which is included in this PR.
```sh
yarn workflow deployment set debug_gdrive_multiple
```
Once PR merged this config can be removed

image

The deployment contains a couple of the subfolders only from the debug google drive. When syncing it is expected that content from these subfolders should sync as expected. 
E.g. asset folders: [1xy5z6PQmsrgwLpCCs6HCCUY1-u2H5KYR](https://drive.google.com/drive/folders/1xy5z6PQmsrgwLpCCs6HCCUY1-u2H5KYR) and [1nlRfAkYfqxM-z_FAzt1WXyES_DyFTlmi](https://drive.google.com/drive/folders/1nlRfAkYfqxM-z_FAzt1WXyES_DyFTlmi).

**NOTE 1** - Although the input sources are both child folders of an existing deployment folder, in practice they will more likely be separated (e.g. main asset folder and minimal override asset folder). I haven't done this for now as would require setting up and sharing new google drive folders (but reviewers should feel free to do so if wished)

**NOTE 2** - In theory any identically named flows or assets in latter folders should override those in former, however there is no example use case of this in the demo (as would throw duplicate error within the parent debug config). An example is however included in the dev testing set of sheets located at `packages\scripts\test\data\input\sheets_additional`. Functional testers should feel free to add duplicate/override files to any of the test folders above as required (but remove after testing to preserve debug config build), or change the folder ids to use folders with more curated content

## Dev Notes
Decided for now easier to implement within existing google_drive config property instead of managing multiple data sources/provider syntax

In order to handle multiple input sources either workflow step task handlers would need to be able to handle arrays of inputs/outputs, or loop logic written into workflow itself (or a combination of both). I've opted to keep the google drive sync handler as a single input sources and add loop to sync workflow instead as this felt more natural to do. Conversely the appData processor has been refactored to support multiple inputs, as it needs to handle the additional json merging post convert.

Tests were a little fiddly due to required changes to whether duplicate flows are considered as error, and also because we use 2 different types of error loggers in the test suites (file-based logs and console-based logs). These could possibly be tidied up a bit more in the future, but for now tests are at least passing

## Git Issues

Closes #2081

## Screenshots/Videos
**Example assets contents.json** - contains merged subset of files from folders [1xy5z6PQmsrgwLpCCs6HCCUY1-u2H5KYR](https://drive.google.com/drive/folders/1xy5z6PQmsrgwLpCCs6HCCUY1-u2H5KYR) and [1nlRfAkYfqxM-z_FAzt1WXyES_DyFTlmi](https://drive.google.com/drive/folders/1nlRfAkYfqxM-z_FAzt1WXyES_DyFTlmi)
image

### DIFF
--- a/packages/scripts/src/commands/app-data/convert/index.ts
+++ b/packages/scripts/src/commands/app-data/convert/index.ts
@@ -70,11 +70,11 @@ export class AppDataConverter {
   cache: JsonFileCache;
 
   constructor(private options: IConverterOptions, testOverrides: Partial<AppDataConverter> = {}) {
+    console.log(chalk.yellow("App Data Convert"));
     // optional overrides, used for tests
     if (testOverrides.version) this.version = testOverrides.version;
     if (testOverrides.activeDeployment) this.activeDeployment = testOverrides.activeDeployment;
 
-    console.log(chalk.yellow("App Data Convert"));
     // Setup Folders
     const { outputFolder, cacheFolder } = options;
     [outputFolder, cacheFolder].forEach((p) => fs.ensureDir(p));
@@ -85,10 +85,6 @@ export class AppDataConverter {
     this.cache = new JsonFileCache(cacheFolder, this.version);
   }
 
-  /**
-   * Reads xlsx files from gdrive-download output and converts to json
-   * objects representing sheet names and data values
-   */
   public async run() {
     const { inputFolders, outputFolder, cacheFolder } = this.options;
     const filterFn = (relativePath: string) => relativePath.endsWith(".xlsx");
@@ -98,17 +94,18 @@ export class AppDataConverter {
       SHEETS_INPUT_FOLDER: "",
       SHEETS_OUTPUT_FOLDER: outputFolder,
     };
-    // Process each input folder, converting xlsx to json
+    // Processing Steps
     for (const inputFolder of inputFolders) {
+      // 1.1 Generate a list of xlsx files in data source and convert to json
       const folderOutputsHashmap: Record<string, FlowTypes.FlowTypeWithData> = {};
       converterPaths.SHEETS_INPUT_FOLDER = inputFolder;
       const list = generateFolderFlatMap(inputFolder, { filterFn });
       const xlsxConverter = new XLSXWorkbookProcessor(converterPaths);
       xlsxConverter.logger = this.logger;
       const data = await xlsxConverter.process(Object.values(list));
+      // 1.2 Sort and filter output jsons
       const outputs = this.cleanFlowOutputs(data);
-      // merge outputs across folders, with latter processed overriding any flows with same type and name
-      // keep local folder list to still track case where duplicate flows appear in same folder
+      // 1.3 Merge jsons both within input sources (duplicate are errors) and across input sources (duplicates are overrrides)
       for (const output of outputs) {
         const hashName = `${output.flow_type}||${output.flow_name}`;
         if (folderOutputsHashmap[hashName]) {
@@ -125,7 +122,7 @@ export class AppDataConverter {
       }
       this.logger.debug({ step: inputFolder, outputs });
     }
-    // Process jsons as flows
+    // 2.1 - Convert all merged jsons to flow data using flow parsers
     const processor = new FlowParserProcessor(converterPaths);
     processor.logger = this.logger;
     const jsonFlows = Object.values(combinedOutputsHashmap);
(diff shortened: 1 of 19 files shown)


packages/scripts/src/commands/app-data/convert/index.ts
```diff
@@ -84,71 +90,75 @@ export class AppDataConverter {
...
+            details: { flow_type: output.flow_type, flow_name: output.flow_name, inputFolder },
+          });
+        }
+        combinedOutputsHashmap[hashName] = output;
+      }
+      this.logger.debug({ step: inputFolder, outputs });
     }
-    const result = output as IParsedWorkbookData;
-
+    // Process jsons as flows
```
jfmcquade, line 128: It could be worth making these comments even more prominent – I quite like the way that the previous `processPipeline` syntax broke down the pipeline into clear steps. Although it may be that good documentation would be better than comments: I have added #1520 to my list of high priorities

esmeetewinkel (approved):
Great, thanks. Functional test passed (a repo with fully synced data shows no file changes when syncing data on this branch, not before nor after updating the config syntax).

I was wondering if we should indicate in the "_xlsxPath" entry in the content jsons which of the multiple folders this path refers to. 

> While working on code it appears that `google_drive.sheets_filter_function` wasn't being properly passed to google drive sync scripts. I'm not sure if any deployments were using this, but would be good to ensure everything working as intended if they are.

Thanks for flagging and resolving. I can't think of a deployment that is using this.

jfmcquade (approved):
Looks good. I can confirm that the base functionality seems to be working for me with the `debug_gdrive_multiple` deployment.

I've added one minor comment inline.

I can see that it's simpler to use the `google_drive` config property instead of introducing a syntax for managing multiple data sources/providers.
I've been working on #2048 – it may be that after those changes we decide that the syntax for multiple data sources/providers would be sensible. But I probably won't include that in that feature PR as there will likely be enough changes to consider without it.

packages/scripts/src/commands/app-data/convert/index.ts
chrismclarke: Yes I had to remove the step-by-step mode as we needed to more deeply nest the processing (inner loop to extract sheets within a given source and outer loop to merge across sources). I'll add a few more comments though for clarity

chrismclarke:
> I was wondering if we should indicate in the "_xlsxPath" entry in the content jsons which of the multiple folders this path refers to.

Makes a lot of sense to do (for cases where there are multiple), although would take a bit of extra refactoring as right now we don't store the parent google drive folder name but the folder id (easier to use in google drive queries). So we would need some extra code to extract the parent folder data, and populate the xlsxPath parent name if multiple defined.

Would it be possible to add as a low-priority follow-up issue?

PR #2091, opened 2023-10-03
Refactor/gdrive spawn

chrismclarke:
PR Checklist

## Description
As part of trying to unblock/retain some of the in-progress work being done in #2061 

Currently all gdrive workspace scripts are executed using spawned commands from the terminal, e.g. 
```sh
yarn workspace @idemsInternational/gdrive-tools start watch --folder-id 1234abc
```
This PR refactors to separate the command line interface (CLI) from main methods (lib), and exposes the lib methods so that they can be imported and called from code. 

This should in theory make them slightly quick to execute (spawning child processes takes some time, although there are some memory management trade-offs depending on local system), but more importantly can be bundled into a build (future desktop app requirement). 

## Review Notes
Should simply need to test the `yarn workflow sync` command still downloads from google drive as expected. Would be useful to confirm the `--content-watch` configuration also works as well as multiple deployment sources

## Dev Notes
The majority of code changes are simply splitting cli+lib code from one file to two. The more considered changes are those to the providers, and how spawnSync commands have been replaced by import

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade (approved):
Looks good, nice to have the methods themselves exposed.

I've tested running `yarn workflow sync` (including with the `--content-watch` flag) for various deployments. All seems to work as expected, except in the case of running `yarn workflow sync --content-watch` with a deployment that has multiple sheets folders, i.e. `debug_gdrive_multiple`. Doing so returns the following error:
Screenshot 2023-10-05 at 13 46 05

This is due to the `sync_watch` workflow using the deprecated `sheets_folder_id` value (see [line 177 of packages/workflows/src/sync.workflows.ts](https://github.com/IDEMSInternational/parenting-app-ui/blob/e51716ef5b1419fedf4bf99c7adac8cc44d627a6/packages/workflows/src/sync.workflows.ts#L177)). This is a knock-on from #2088 rather than an issue caused by this PR, so it could be dealt with in a follow-up or included here.

chrismclarke:
Thanks for checking @jfmcquade , the content-watch issue makes sense and should hopefully be an easy fix. I'll make a follow-up PR so that we can get this one in first.

PR #2097, opened 2023-10-06
fix:(ci) node out of memory issue

chrismclarke:
PR Checklist

## Description

Some recent build [actions have been failing](https://github.com/IDEMSInternational/parenting-app-ui/actions/runs/6434087621/job/17472576030), due to node running out of heap memory. This can be somewhat non-deterministic, depending on a number of factors including operating system, version of 3rd party dependencies used, and overall codebase.

This PR simply configures a fixed memory limit that should be significantly higher than the default (6GB out of 7GB total available to [github actions runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), as OS shouldn't need more than about 500MB or so)

## Review Notes
It will only be clear if this fixes issues if we stop seeing failed builds like the one above.

## Author Notes
This PR updates the github actions on the main repo and the template generator for child repos, however any child repos will need to be manually migrated to include the change.

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade (approved):
Makes sense.

@ChrisMarsh82, would you be able to copy over the changes to the relevant workflow (`app-build`) as it exists in each content repo after merge? Or maybe that's something I should do to acquaint myself with the repos

PR #2107, opened 2023-10-19
fix: template override issues

### LINKED ISSUE #2106: [BUG] Issue with template overrides that reference original template
**Describe the bug**
In the case of a template that overrides a default template whilst also instantiating the default template, an infinite loop is triggered. This seems to be equivalent to the `is_override_target` value being omitted (which is explicitly intended to prevent an infinite loop).

**App version**
0.16.21

**To Reproduce**
This was originally noticed on the `language_select` template on the `plh_tz` deployment, see [web preview](https://plh-teens-tz.web.app/template/language_select).
Screenshot 2023-10-18 at 16 02 22
The issue is present for all deployments. There don't seem to be any recent relevant changes to the relevant templates: [language_select](https://docs.google.com/spreadsheets/d/1iSG1kzK4OTHcoiwWoZ25LZxrJqERg2_iP9369HTi_bQ/edit#gid=823624970), [language_select_tz](https://docs.google.com/spreadsheets/d/18LVVuhE0uuNRfsj3AEKvKzZbPW7sxPUHRn82BbvPcj8/edit#gid=447608578). This suggests the issue is due to recent code changes.

The issue is also present in other cases of template overrides, e.g. `example_temp_override_default`, see [debug web preview](https://idems-debug.web.app/template/example_temp_override_default)

**Debug Template**
See `debug_override` templates [here](https://docs.google.com/spreadsheets/d/1svnypViFpmd5uLhPeIW-4g0YAEOoAtcAMNrMLl7n8Ow/edit#gid=1060209527). The override condition can be satisfied by manually changing the value of the field `debug_boolean`.
Screenshot 2023-10-18 at 16 10 00
Screenshot 2023-10-18 at 16 10 08

When override condition is not satisfied:
Screenshot 2023-10-18 at 15 59 13

When override condition is satisfied:
Screenshot 2023-10-18 at 15 58 57

In the latter case, the behaviour is identical to that when `is_override_target` is omitted.

**Additional screenshots**
An example when the override sheet includes an override of a variable in the default sheet, replicating the behaviour as originally observed:
Screenshot 2023-10-18 at 15 55 57
Screenshot 2023-10-18 at 15 56 10
Screenshot 2023-10-18 at 14 53 17

**Additional information**
As well as the `is_override_target` functionality seemingly broken, the behaviour also suggests an issue with overrides being applied in general. In the case of example_temp_override_default, the template should only be overridden by `example_glob_override_es_sp` in the case that its `override_condition` is met, i.e. when `@fields._app_language.startsWith("es")` evaluates to true. But running locally and inspecting the template on the [web preview](https://idems-debug.web.app/template/example_temp_override_default) suggests that the template override is being applied (or attempting to be applied) even though that condition should not be satisfied.

Screenshot 2023-10-18 at 16 21 32

jfmcquade:
PR Checklist

## Description

#2040 introduced some issues with template overrides. 

1. In the case of a template that overrides a default template whilst also referencing(/instantiating) the default template, an infinite loop was triggered (see #2106)
2. Override conditions were not always correctly evaluated, specifically those utilising string methods on a template variable.

## Testing notes
 Ensure that all override functionality is working as expected. For example, new [debug_overrides](https://docs.google.com/spreadsheets/d/1svnypViFpmd5uLhPeIW-4g0YAEOoAtcAMNrMLl7n8Ow/edit#gid=1060209527) templates, pre-existing [example_list_override](https://docs.google.com/spreadsheets/d/1MpoH3BxhECZRmYM10HZ0pTOoe69FJ-fEW9FwzK-Q6yw/edit#gid=2059672136) templates, pre-existing [example_temp_override](https://docs.google.com/spreadsheets/d/1MpoH3BxhECZRmYM10HZ0pTOoe69FJ-fEW9FwzK-Q6yw/edit#gid=1921555013) templates.

## Dev notes

I'm not sure why these issues weren't caught, or weren't present, at the time of testing #2040. Tracking down the issue proved to be quite difficult due to the multiple override systems, but at least now I have a better understanding of the relevant code. I'm still not sure that my solution to 1. preserves the functionality introduced in #2040 (is there a use case of applying multiple overrides that is now not possible?), so would appreciate a code review.

## Git Issues

Closes #2106

## Screenshots/Videos

Video showing various override debug templates working as expected:
1. New [debug_overrides](https://docs.google.com/spreadsheets/d/1svnypViFpmd5uLhPeIW-4g0YAEOoAtcAMNrMLl7n8Ow/edit#gid=1060209527) templates
2. Pre-existing [example_list_override](https://docs.google.com/spreadsheets/d/1MpoH3BxhECZRmYM10HZ0pTOoe69FJ-fEW9FwzK-Q6yw/edit#gid=2059672136) templates
3. Pre-existing [example_temp_override](https://docs.google.com/spreadsheets/d/1MpoH3BxhECZRmYM10HZ0pTOoe69FJ-fEW9FwzK-Q6yw/edit#gid=1921555013) templates

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/f90ef6e6-d519-4371-909b-349268d54ba7

### DIFF
--- a/src/app/shared/services/data/app-data-variable.service.ts
+++ b/src/app/shared/services/data/app-data-variable.service.ts
@@ -104,6 +104,7 @@ export class AppDataVariableService extends AsyncServiceBase {
       const evaluated = jsEvaluator.evaluate(parsed);
       return evaluated;
     } catch (error) {
+      console.warn(`Error evaluating expression, "${expression}"`, error);
       return parsed;
     }
   }
@@ -120,7 +121,10 @@ export class AppDataVariableService extends AsyncServiceBase {
           isRecursive = true;
         }
         for (const variableName of Object.keys(variableNameHashmap)) {
-          const evaluated = await this.get(contextPrefix as IVariableContext, variableName);
+          let evaluated = await this.get(contextPrefix as IVariableContext, variableName);
+          if (typeof evaluated === "string" || evaluated instanceof String) {
+            evaluated = `"${evaluated}"`;
+          }
           evaluatedContext[contextPrefix][variableName] = evaluated;
         }
       }
--- a/src/app/shared/services/data/app-data.service.ts
+++ b/src/app/shared/services/data/app-data.service.ts
@@ -81,7 +81,7 @@ export class AppDataService extends SyncServiceBase {
     flow_type: FlowTypes.FlowType,
     flow_name: string,
     /** Keep log of previous override flow names to avoid infinite loops, e.g. flow_a -> flow_b -> flow_a  */
-    overrideHistory: string[] = []
+    is_override_target: boolean = false
   ) {
     const flowContents = this.sheetContents[flow_type][flow_name];
     if (!flowContents) {
@@ -91,9 +91,8 @@ export class AppDataService extends SyncServiceBase {
     }
     // Check for any runtime overrides to flows
     const overrideFlowName = await this.checkFlowOverrides(flowContents);
-    if (overrideFlowName && !overrideHistory.includes(overrideFlowName)) {
-      overrideHistory.push(flow_name);
-      return this.getSheet(flow_type, overrideFlowName, overrideHistory);
+    if (overrideFlowName && !is_override_target) {
+      return this.getSheet(flow_type, overrideFlowName, true);
     }
 
     // Populate flow from cache if exists, or load json if it does not
--- a/src/app/shared/components/template/services/template.service.ts
+++ b/src/app/shared/components/template/services/template.service.ts
@@ -143,7 +143,8 @@ export class TemplateService extends SyncServiceBase {
     await this.ensurePublicServicesReady();
     const foundTemplate = await this.appDataService.getSheet<FlowTypes.Template>(
       "template",
-      templateName
+      templateName,
+      is_override_target
     );
     // const foundTemplate: FlowTypes.Template = template.find((t) => t.flow_name === templateName);
     if (foundTemplate) {


esmeetewinkel (approved):
Functional test passed on `plh_teens` and `debug` deployments.

N.B. on my end the error looks a bit different than in your screenshot, instead of the screen displaying the overridden title many times I get "Aw, snap!"

image

src/app/shared/services/data/app-data-variable.service.ts
```diff
@@ -120,7 +121,10 @@ export class AppDataVariableService extends AsyncServiceBase {
           isRecursive = true;
         }
         for (const variableName of Object.keys(variableNameHashmap)) {
-          const evaluated = await this.get(contextPrefix as IVariableContext, variableName);
+          let evaluated = await this.get(contextPrefix as IVariableContext, variableName);
+          if (typeof evaluated === "string" || evaluated instanceof String) {
+            evaluated = `"${evaluated}"`;
```
chrismclarke, line 126: nit(blocking) - This is quite a significant change to a core method. Could you try to identify the use case where this is required and add to the existing spec tests for the service?

You'll also see that this change breaks the existing tests by adding extra quotation marks where not required.

I think this highlights an importance to try get the ng test suite fully operational (some work started in #2080 which was meant to be handed over, but possibly wasn't or fell down priority list?)

src/app/shared/services/data/app-data.service.ts
```diff
@@ -81,7 +81,7 @@ export class AppDataService extends SyncServiceBase {
     flow_type: FlowTypes.FlowType,
     flow_name: string,
     /** Keep log of previous override flow names to avoid infinite loops, e.g. flow_a -> flow_b -> flow_a  */
-    overrideHistory: string[] = []
+    is_override_target: boolean = false
```
chrismclarke, line 84: nit(blocking)
I'm a bit unsure about this, as I think making it binary would prevent an otherwise fine case of nested overrides (like in the comment). Would it be possible to add tests to the service spec file and confirm if the behaviour is as-expected or not?

chrismclarke (changes requested):
Thanks for looking into this @jfmcquade , can appreciate the challenge of understanding all the layers of abstraction.

I really think as a result of this we need to be cautious about making any changes which could cause regressions elsewhere. I think ideally any work on this bug should include spec tests that both capture the before behaviour and can validate the after behaviour still works for standard test cases

src/app/shared/services/data/app-data-variable.service.ts
chrismclarke: I've added comments to the app-data and app-data-variable spec files to include code to run test standalone (e.g. `yarn ng test --include src\app\shared\services\data\app-data-variable.service.spec.ts`

jfmcquade: Sure enough, I had misunderstood the issue. 

I was getting the following log when trying to run the example_temp_override_default template:
Screenshot 2023-11-07 at 16 50 01
The override condition, `@fields._app_language.startsWith("es")` (found in the [contents list](https://docs.google.com/spreadsheets/d/1MpoH3BxhECZRmYM10HZ0pTOoe69FJ-fEW9FwzK-Q6yw/edit#gid=1745157248)) was being parsed as `za_en.startsWith("es")`, which I took to be a general issue with fields storing string values. In fact, I think it is general behaviour affecting all cases of _applying a string method_ to the value of a field. The issue can be resolved by rewriting the condition as `"@fields._app_language".startsWith("es")`. Do you think this is expected behaviour? That debug template appears to be the only template that uses `startsWith()`.

I have reverted the change to this method in the latest commit

chrismclarke: Interesting, does it work if using delimited syntax `{@fields._app_language}.startsWith("es")`?
Not super sure what the expected behaviour here is, might be good to write a couple tests and update template as required

jfmcquade: I've added some tests to [src/app/shared/services/data/app-data-variable.service.spec.ts](https://github.com/IDEMSInternational/parenting-app-ui/pull/2107/files#diff-8873dfedef4ee3fd95f8af1a3d11a9856aff28b2c9cdc6683952d0daa53493cc), which capture the current behaviour.

Screenshot 2023-11-09 at 15 15 57

I think this is not expected behaviour, and we should amend the app data variable service, see new issue: #2141

src/app/shared/services/data/app-data.service.spec.ts
```diff
@@ -66,5 +132,22 @@ describe("AppDataService", () => {
...
+    expect(data.rows?.[0].id).toEqual("a_id1");
+  });
+
+  it("handles flow overrides", async () => {
+    const data = await service.getSheet("data_list", "flow_d");
+    expect(data.rows?.[0].id).toEqual("a_id1");
+  });
+
+  // In the case that flow_b is overridden by flow_c, which in turn is overridden by flow_b,
+  // the loop should be broken after the initial override
```
jfmcquade, line 147: This was what I took to be the expected behaviour, but if we wanted it to allow `flow_a` -> `flow_b` -> `flow_a`, we could change [line 100 of app-data.service.ts](https://github.com/IDEMSInternational/parenting-app-ui/pull/2107/files#diff-d5415ab90f2bbaee6aba223424f7e14faef364f1f635f0db46909bf5bdba7e70R100) from `overrideHistory.push(flow_name)` to `overrideHistory.push(overrideFlowName)`

src/app/shared/services/data/app-data.service.ts
jfmcquade: I think I wasn't clear on what behaviour tracking the `overrideHistory` was supposed to handle, and had it confused with the core use case of the PR's associated issue. But I think now I see the difference: `overrideHistory` is intended to handle the case where `flow_a` is overridden by `flow_b` which in turn is overridden by `flow_a`.

I've updated the PR with [changes to the app data service](https://github.com/IDEMSInternational/parenting-app-ui/pull/2107/files#diff-d5415ab90f2bbaee6aba223424f7e14faef364f1f635f0db46909bf5bdba7e70) which I think handle both cases, and [added tests](https://github.com/IDEMSInternational/parenting-app-ui/pull/2107/files#diff-3e326d5bce4e0471362869d3f6fed1af63ce3de890634b1ce0e9c91c618118f2) to capture the behaviour.

jfmcquade:
Thanks, @chrismclarke, see my replies to your inline comments, I think I've dealt with the issues.

src/app/shared/services/data/app-data-variable.service.ts
jfmcquade: Me and Esmee are wondering if it might make sense for condition fields to require an `@calc()` block in order to pass javascript. As I understand it, our `condition` columns (`override_condition` and `condition_list` on campaigns) evaluate any javascript that is passed. But it could be confusing for authors to understand when they can and can't write javascript directly in a cell

chrismclarke: Ah, yeah I see the challenge here - comparing the two cases:

1. `hello @field.name`
2. `@field.name.startsWith("A")`

In case (1) we want the string to just replace and return `hello my_name`
In case (2) we want the string to replace wrapped as a string `"my_name".startsWith("A")`

Currently we focus on supporting case (1). The parser attempts to evaluate, but reads `hello` and `my_name` as variables which are not defined, and has a fallback in case of such errors to just return the full expression as-is (as text)

I think the best way to also support case (2) would be to internally store a mapping of replaced strings when evaluating, which can then be used in follow-ups. E.g. for both cases above the mapping would simply be variables
```
{my_name: "my_name"}
```
Then when it comes to evaluation, case 1 will still just return as text (failing to recognise `hello` as a variable), but case 2 should be able to parse `my_name.startsWith("A")`, substituting the `my_name` variable for `"my_name"` string

I'll try make these changes in a follow-up PR as not so much targetted on the issue at hand

chrismclarke: Moving over to https://github.com/IDEMSInternational/parenting-app-ui/pull/2146

src/app/shared/services/data/app-data.service.spec.ts
chrismclarke: Just to clarify, are we essentially asking whether in the end we should show `flow_a` vs `flow_b`?
I'd expect stopping at `flow_b` - although not sure I fully understand. In either case as long as immediate issue fixed happy for that to be a future update

chrismclarke (approved):
Thanks for all the updates, the changes look a lot more focused now and the expanded tests a really nice addition to have. At this point I think should be good to go assuming functional tests passed

I've also built on this with #2146 to address the issue raised in the [comment above](https://github.com/IDEMSInternational/parenting-app-ui/pull/2107#discussion_r1395572216)

src/app/shared/services/data/app-data.service.spec.ts
jfmcquade: Yes, that's what I was asking. I agree that stopping at `flow_b` seems right, and that seems clearer to me after not thinking about it for a while!

esmeetewinkel (approved):
Functional test passed on `debug` deployment. I'm happy for this PR to be merged.

On `plh_teens_tz` deployment this also resolves the error, however I'm not seeing the template override correctly because of an unrelated spreadsheet issue that this revealed: The `plh_tz` deployment was renamed `plh_teens_tz` but the changes were not made to the spreadsheets, where this is used as an override condition. I'll make the necessary replacements there. 
image

esmeetewinkel (changes requested):
Actually, after making the renaming changes on the `plh_teens_tz` deployment as suggested in my previous review, the issues are **not** resolved. 

In this case we have a template language_select which should be overridden by either language_select_tz or language_select_za depending on the deployment. Looking at the error in the console the app seems to be mixing up the two, somehow. 
image

jfmcquade:
> Actually, after making the renaming changes on the `plh_teens_tz` deployment as suggested in my previous review, the issues are **not** resolved: I'm still seeing the default template rather than the override.
> 
> In this case we have a template language_select which should be overridden by either language_select_tz or language_select_za depending on the deployment. Looking at the error in the console the app seems to be mixing up the two, somehow. image

Thanks, @esmeetewinkel. In this case, the issue is to do with JS string evaluation, which @chrismclarke fixed in the supplementary PR, #2146. I've merged those changes into this PR now, so you if you could functionally test this one again,  both issues should be resolved: 
1. Recursive overrides 
2. Override conditions which contain string evaluation (e.g. `@fields.deployment_name == "plh_za"`

esmeetewinkel (approved):
Functional test passed 

image

PR #2111, opened 2023-10-24
feat: add content version field, display value in sidebar

jfmcquade:
PR Checklist

## Description

Adds a field, `_content_version`, to store the value of the content version. This value is read from the `config.git.content_tag_latest` value that is set in the deployment config, and which gets automatically incremented as part of the `repo publish` workflow.

(With `APP_SIDEMENU_DEFAULTS.should_show_version` set to `true`, which is default) the app sidebar will now show the content version number if it exists. The app version/builder number is hidden in a tooltip, displayed on hover on desktop and on click on mobile (see screen recordings below). If no content version number exists, the app version/builder number will show, as before. A possible follow-up would be to consider the case of no content version: perhaps a warning should be displayed, as every app with content should have a version number for the content.

## Git Issues

Closes #

#### Desktop

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/ef66cfb7-c459-433b-be11-39f74a278c3b

#### Mobile

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/8633c101-568b-4d95-bdad-8536c275f49e

### No content version

Screenshot 2023-11-06 at 11 25 32


src/app/app.component.html
```diff
@@ -11,7 +11,9 @@
       <ion-toolbar color="primary">
         <ion-title>{{ sideMenuDefaults.title }}</ion-title>
         <div class="app-version" slot="end">
-          <span *ngIf="sideMenuDefaults.should_show_version">{{ APP_VERSION }} </span>
+          <span *ngIf="sideMenuDefaults.should_show_version"
+            >{{ APP_VERSION }}{{ CONTENT_VERSION ? "(" + CONTENT_VERSION + ")" : "" }}
```
chrismclarke, line 15: nit(blocking)
I'm guessing content_version is more interesting to the user anyway? How about we add a span here with an ngIf statement so that if there is content version we just show the content version and use a [title attribute](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_global_title) to display the full title with app_version?

A second span could just show regular app version if no content version exists, or possibly some other text to make it clear that there is not versioned content (would there be a way to ensure we do have versioned content?)

chrismclarke (changes requested):
One minor comment about what best to show to user, but otherwise makes sense to me

src/app/app.component.html
jfmcquade: Thanks, I've implemented this. As the title attribute content doesn't display on touch screens (no hover functionality), I've custom CSS to display the content on click when on mobile, see [latest commit](https://github.com/IDEMSInternational/parenting-app-ui/pull/2111/commits/bf2b55a923c4547f0dd599b383cc1f4a362a8be9).

esmeetewinkel (approved):
Functional test passed.

PR #2113, opened 2023-10-25
Fix/api dbs

chrismclarke:
PR Checklist

- [ ] PR title descriptive (can be used in release notes)

## Description

- Update api core packages (minor patches only to avoid breaking changes)
- Add testing infrastructure and recreate issue seen in production (entries not writing to correct db)
- Refactor way in requests get mapped to models that have specific db connection query interfaces
- Add global connection manager to simplify code in deployment service (and because deployment service is recreated on a per-request basis)

## Review Notes
I doubt the code changes will make much sense (a suuuper deep dive into nestJS/sequelize issues), but it might be useful if you are able to confirm that you can at least setup the e2e tests (see notes in updated documentation) and run the test suite to confirm similar outcomes.

## Dev Notes
This issue was really awful to debug, so many moving pieces on top of a lot of layers of obscurity with how nest-js -> sequelize-typescript -> sequelize -> postgres, with various levels at which services are instantiated (sometimes globally, sometimes per request, not always as expected (e.g. global services still created per-request when interacted with via middleware).

The new testing infrastructure should hopefully help a lot moving forwards, but it still feels like what we're trying to achieve with the database goes far beyond what the system really should be used for.

Ultimately there were a few core issues which should now be resolved, namely:
- Lack of testing infrastructure making it hard to debug. A small set of e2e tests have been created to help debug issues.
- Use of middleware to set default db client just wasn't working as expected. Instead middleware now only bootstraps db and specific connection determined at a later step
- The way in which nestJS models default to using default db connection - this is now manually when accessed through deployment service (but still may cause issue 
- The fact that services declared globally were actually being instantiated on a per-request basis due to... reasons (eurgh, could go into details if interested but honestly a headache).
- Memory management/allocation when lots of requests in parallel writing to objects that have lots of deep/circular references (again, could go into the details but don't expect them to wow anyone)

**Follow-up Tasks (Post Merge)**

- [ ] Refactor remaining code that calls model constructor (e.g. `new User(...)`), in favour of deployment service methods (e.g. `this.model.create(...)`
- [ ] Testing additional methods (beyond user create)
- [ ] Documentation (?)
- [ ] Deploy to server
- [ ] Consider [NestCLS](https://docs.nestjs.com/recipes/async-local-storage) in case might be able to help simplify some of the logic

## Git Issues

Closes #2098

## Screenshots/Videos

**Set of tests created (passing)**
image

**Example output** - 5 users written concurrently to 5 different databases
```json
 [                                                                               
      {
[...]

PR #2115, opened 2023-10-27
Fix/api dbs 2

chrismclarke:
PR Checklist

## Description
Follow-up PR for issues linked to #2113, namely:

- [x] Refactor remaining code that calls model constructor (e.g. `new User(...)`), in favour of deployment service methods (e.g. `this.model.create(...)`
- [x] Testing additional methods (beyond user create)

**Not addressed (future follow-up)**
- [ ] Documentation (?)
- [ ] Deploy to server
- [ ] Add tests for hooks (e.g. db trigger that flattens contact field json)
- [ ] Address potential test system memory leak (tests appear to work in isolation but bootstrap possibly causing memory leak preventing full spec suite to pass)
- [ ] Consider [NestCLS](https://docs.nestjs.com/recipes/async-local-storage) in case might be able to help simplify some of the logic
- [ ] Review (legacy?) contact_fields endpoint and consider deprecating (think include in app_users endpoint but need to confirm)

## Review Notes
The created tests can be run locally via `yarn workspace api test:e2e`

If possible it would also be good to do some functional testing, running the api locally and an app deployment to check that expected functionality works to:
- Sync user, contact fields, feedback and notification interaction to database
- Sync data from deployment with custom db name, ideally within quick succession of previous deployment (ensure switching happening on demand as expect).

There is already some [documentation](https://idemsinternational.github.io/parenting-app-ui/developers/server-development/) that provides 2 ways of testing api locally, although I can't remember if there might have been some challenges in the past that require update. I also believe there to be an existing deployment config that specifies a custom db name, but equally debug config could be quickly changed

When testing db data sync this can be done either/both from within the app (e.g. using template that syncs user data), or from the api interactive interface (usually runs on port 3000)

## Dev Notes
This PR focuses on small refactors and adding test for all endpoints to replicate the fixes required for writing to multiple dbs concurrently. It also tidies up the previous PR tests by splitting functionality and moving to associated code module folders.

All endpoints should now in the very least contain some basic tests to create, update and list data, as well as check functionality when passing custom deployment db name header.

While running tests locally I've noticed a bit of inconsistency related to the db bootstrap/teardown process where there might be a potential memory leak (sometimes suite fails suggesting to use `--detectOpenHandles` flag, but then passes with flag). I don't think this is vital for now as tests still pass in isolation and aren't integrated into other CI processes, however something likely will want to address in the future.

If requiring debug when running tests, it might be useful to comment out the `afterAll` hook that wipes data written to the database, e.g. `afterAll(async () => await testDBTeardown());`, so that the data can be inspected after run. There is an additional `beforeAll` hook that also wipes database before test runs so not really an issue if any data left hanging. It might also be handy to inspect http responses with some extra console logs before expect statements, e.g. `console.log({ status, body });`

Going through all the code there's still quite a number of inconsistencies with things like endpoint names (e.g. `contact_field` vs `contact_fields`), handling of app_user_id (should it be passed as request field or appended to endpoint), and type definitions (various ways data-transfer-object (DTO) and db schema (model) interoperate, with some services not reading from either and just working from `any`). It's beyond the scope of this PR to make too many more changes and I'm not aware of many immediate use cases to expand the current apis, so I assume best to just leave for now and revisit at some point in the future with the context of potential method migration to supabase (if desirable).

## Git Issues

Closes #

## Screenshots/Videos
**Test Output** - Note skipped tests due to deprecated(?) `contact_fields` endpoint which should be removed in the future
image


esmeetewinkel (commented):
It might be most helpful for @jfmcquade to do this testing, I'm a couple of steps behind so will need some handholding...

- When I run `yarn workspace api start:watch` I get the error below. I guess I need to do/install something else firstimage

- Looking at the docs it seems I need to install postgres13, which I don't seem to have (although I did install pgAdmin 4 v7 recently which is running fine)image

- When installing postgres16 I was prompted to input a database superuser pwimage
I used the one I found in the `.env` but not sure this was correct, and I encountered some problems later on in the installation process.

chrismclarke:
> * When I run `yarn workspace api start:watch` I get the error below. I guess I need to do/install something else first image

might just need `yarn install` - the auto installer only works when running the main start command

> * When installing postgres16 I was prompted to input a database superuser pw image

This will be a super user for your own local system. So you can just make it anything you will remember. After installing locally there will be additional steps to connect to that database using pgadmin, and creating the necessary databases/users for the app to use. Definitely a bit of an involved process so might be good if @jfmcquade can try first and then help setup once we know things are actually working as expected.

jfmcquade (approved):
OK, I've tested this using the included tests and done some functional testing, and it looks good to me. A few details worth noting.
### e2e tests
I observed the behaviour you mention in the notes: when running `yarn workspace api test:e2e`, I reliably get some failures and a warning regarding memory leaks. 

all

Then running with the `--detectOpenHandles`, all (non-skipped) tests pass reliably. 

all --detectOpenHandles

Likewise the tests pass reliably when I run each "suite" in isolation, e.g. `yarn workspace api test:e2e -t app_feedback`

app_user
AppController
deployment
app_feedback
app_notification_interaction

### Functional testing
I am still getting errors relating to the umzug package (see [this comment](https://github.com/IDEMSInternational/parenting-app-ui/pull/2113#pullrequestreview-1699885839), as well as Esmee's above). I've pushed [a commit](https://github.com/IDEMSInternational/parenting-app-ui/pull/2115/commits/60188ad8273c93f55dcc25a57e7afec1afb85a5a) to revert the umzug version number, which fixes the problem for me, but perhaps you want to reassess.

Using the `debug_sync_id` template, I can sync user data, including contact fields (as in the `contact_fields` column of the `app_users` table – is the `contact_fields` table deprecated?). This works to the default database, `plh`, if none is provided in the deployment config. If a custom `db_name` is provided, then the user data is synced to the named database, with the db being created if it doesn't already exist.

Testing requests in quick succession is a bit tricky, but I can confirm that if I make a request via the interactive API interface to one database, and sync to a different database through the app template in quick succession (a matter of seconds), both requests seem to successfully update their respective databases.

I haven't repeated all these tests for feedback, but I've tested the base functionality (syncing to default or custom db) for each, with positive results.

In the case of notifications interactions, I've tried to test by running in my local emulator (`ionic cap run android -l --external`). However when I try to sync data from the emulator, the connection to the API is refused. I haven't ever knowingly synced from my emulator to the locally running databases via the API, so I don't imagine this is a new problem (possibly something to do with CORS??).
Screenshot 2023-11-01 at 14 21 50

chrismclarke:
Thanks for the thorough checks @jfmcquade 
Happy to pin umzug one version down (although weird I can't replicate, I'm guessing its node version dependent (?) but happy to just see if resolved next time working on the api and updating packages. Likewise the issue with test memory leaks

 Yes I'm reasonably sure the `contact_fields` db is deprecated (at least couldn't see any references to it in frontend code), but can't remember if there's any external tools that use (n8n?). Should review and remove at some point in the future, for now I've added some code comments more as a reminder
 
 As for communicating from emulator, I expect you'd need to use the local ip of your machine instead of localhost (e.g. `192.168.x.x`), but probably not a hole worth getting stuck too far down - I think probably good to go after this with making the server updates and listing whatever else as follow-up issues

PR #2117, opened 2023-10-31
fix: pop-ups dismiss on completed or uncompleted

### LINKED ISSUE #2116: [FEATURE] Pop-up should close when it emits completed or uncompleted
**What?**
It would be helpful to expose a button action that has the same effect as clicking the "cross" in the top left corner of a pop-up. This can be `click | emit: completed ` and `click | emit: uncompleted`. Currently nothing happens when a pop-up emits completed or uncompleted.

**Why?**
Needed for FacilitatorApp Malaysia deployment.

[screen-capture (63).webm](https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/7ee0ccdc-2f7a-4cf3-a43a-6b71a0d42cdb)

jfmcquade:
PR Checklist

## Description

Pop-ups will now dismiss when the child template emits "completed" or "uncompleted". This was already the case for full-screen pop-ups, but now extends to non-full-screen pop-ups.

This seems to be the expected behaviour, however it is possible that this may interfere with some existing functionality, so if necessary we could expose `dismissOnEmit` as an action parameter (and possibly default to `false`).

## Testing

See new [debug_pop_up_dismiss](https://docs.google.com/spreadsheets/d/1XDugKSTqqEScibrWH0OkTTFf-lqsZREob37EeieQUWI/edit#gid=963655109) template for testing functionality.

Previous [example_pop_ups](https://docs.google.com/spreadsheets/d/1K9JZ9Glnj73CFJmpfGIRa5dRA2TBxSsJmaguV6GL8Pc/edit#gid=402992224) seems to be out-of-date – the functionality described in the comments does not work as stated (see deprecated logic [here](https://github.com/IDEMSInternational/parenting-app-ui/pull/2117/files#diff-4ee811ea134405d45640bbcf1e83f030cb27921c99115381bd4889f7544d74d0L154)).

## Dev notes

With the current implementation, if we wanted to expose `dismissOnEmit` as an action parameter then it would need to be passed between methods via query params, which doesn't seem ideal and would only further complicate the pop-up and nav logic.

## Git Issues

Closes #2116

## Screenshots/Videos

[debug_pop_up_dismiss](https://docs.google.com/spreadsheets/d/1XDugKSTqqEScibrWH0OkTTFf-lqsZREob37EeieQUWI/edit#gid=963655109) template
Screenshot 2023-10-31 at 15 40 09

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/f559ca03-5fe8-47c6-b64f-1dbd5f01b139

### DIFF
--- a/src/app/shared/components/template/services/template-nav.service.ts
+++ b/src/app/shared/components/template/services/template-nav.service.ts
@@ -304,6 +304,7 @@ export class TemplateNavService extends SyncServiceBase {
       templatename: popup_child,
       parent: container,
       showCloseButton: true,
+      dismissOnEmit: true,
     };
     // If trying to recreate a popup that already exists simply mark as visible
     const existingPopup = this.openPopupsByName[popup_child];


esmeetewinkel (approved):
Functional test passed. I'm happy for this to be the default behaviour. I'm not sure it's worth putting in the time now to expose this as a setting. I agree it looks like this was always the intended behaviour (from [example_pop_ups](https://docs.google.com/spreadsheets/d/1K9JZ9Glnj73CFJmpfGIRa5dRA2TBxSsJmaguV6GL8Pc/edit#gid=402992224)) and I think the knock-ons this will have on other apps will be minimal as we barely use pop-ups in this way up to now. 

There's one side case that the PR does not seem to address: When a pop-up has a button with a go-to action to another template, and that template emits completed, then the user gets redirected to the pop-up, independent of whether the initial button contained an action dismissing the pop-up or not (debug example added to [debug_pop_up_dismiss](https://docs.google.com/spreadsheets/d/1XDugKSTqqEScibrWH0OkTTFf-lqsZREob37EeieQUWI/edit#gid=963655109))

[screen-capture (64).webm](https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/05001721-e983-4c33-ae17-59159fe8a946)

It would make sense to me that, if the initial button both has an action dismissing the pop-up and a go-to action, and the second template emits completed, then the user is redirected to the template underlying the pop-up.

PR #2122, opened 2023-11-02
feat: add release labeler action

chrismclarke:
PR Checklist

## Description

Add various actions to help streamline PR labelling and release notes, specifically:

1. Create labeler action that will 
a) Check if pr title assigned in conventional commit standard (not strictly enforced)
b) Assign labels based on PR title (if written in conventional commit way)
c) Assign additional labels based on code path changes (currently just for any code updated in `documentation` folder

2. Export list of repo labels and populate to `.github/labels.yml`. Add action that can be used to sync labels from this list programatically (potentially useful in future if sharing labels across repos)

3. Rename all actions related to github prs consistently (e.g. `gh-pr-labeler`, `gh-sync-labels`)

## Review Notes
Can see this pr automatically labeled with `feature` label due to PR title.
We will likely need to wait until after merge to see if the `gh-release-drafter` action picks up the changes and drafts notes correctly

## Dev Notes
I couldn't actually find any actions that assign labels based on commits, only PR name (proposed in `https://github.com/bcoe/conventional-release-labels/issues/27` but not implemented), so this PR focuses on role of PR title more than anything else. 

It also lays some groundwork for future systems that may or may not want to be implemented to share labels and actions across projects, but recommend trying to get a good workflow going here first before trying to broaden scope too much.

Also included a note in one of the actions that we may want to consider an alternate [https://github.com/marketplace/actions/release-please-action](https://github.com/marketplace/actions/release-please-action) than release-drafter as a way to keep a rolling PR that can create a release on merge (instead of manually creating release)

## Git Issues

Closes #

## Screenshots/Videos


.github/workflows/gh-pr-labeler.yml
```diff
@@ -0,0 +1,39 @@
+name: Github Release Labeler
```
jfmcquade, line 1: Should this be "Github PR Labeler"?

jfmcquade (approved):
Nice, looks good. Minor comment inline

chrismclarke:
> Nice, looks good. Minor comment inline

Thanks @jfmcquade 
I've fixed the typo with 6e11cd4d2edf96b78fa89cecbc74d491ca785550  and also realised I was missing a few labels that we might want to use on other PRs and included with b15f012471e78fc8040bf9a9aa1ba7f1af234f31

.github/workflows/gh-pr-labeler.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: bcoe/conventional-release-labels@v1.3.1
+        with:
+          # Labels assigned based on pr name prefix
+          type_labels: |
+            {
+            "feat": "feature",
+            "fix": "fix",
+            "breaking": "breaking",
+            "refactor": "maintenance",
+            "chore":"maintenance"
```
jfmcquade, line 33: What about "docs":"documentation"?

jfmcquade, line 33: What about "docs":"documentation"?

jfmcquade (changes requested):

.github/workflows/gh-pr-labeler.yml
chrismclarke: Nice idea - the step below auto-assigns documentation tag based on file path but nice to include here and in the release notes template also. Will update now

chrismclarke: Updated type labels and template, should hopefully be good to go now

jfmcquade (approved):
All looks good to me.

@esmee, want to have a quick look before merging?

esmeetewinkel (approved):
Looks good to me. Two questions: 

- I wasn't aware of story board points, but read up on them a bit. Is the suggestion to use these going forward? 
- If a PR has multiple tags (e.g. it could include a new feature and its documentation) would that appear twice in list of PRs that are included in the release, once under the Feature header and once under Documentation header?

chrismclarke:
> * I wasn't aware of story board points, but read up on them a bit. Is the suggestion to use these going forward?

I don't think I'm aware of these either, a quick search shows the following repo which says is deprecated so not sure exactly what is being proposed
https://github.com/pkosiec/github-projects-story-points

> * If a PR has multiple tags (e.g. it could include a new feature and its documentation) would that appear twice in list of PRs that are included in the release, once under the Feature header and once under Documentation header?

I'm not actually sure, I don't see any way of specifying hierarchy so would assume yes. but I think we'll only know once the action has run

chrismclarke:
On the plus side it looks like the first tagged pr updates correctly for the generated release notes
https://github.com/IDEMSInternational/parenting-app-ui/releases

image

chrismclarke:
Just to answer the question on multiple tags, yes it does look like 2 entries are added. I don't think this is necessarily a bad thing (I expect to change behaviour we'd either need to raise a request on the release-drafter git repo or fork it and add our own implementation, which probably wouldn't be worth it at this point)

image

PR #2123, opened 2023-11-03
Server/updates 1.4.x

chrismclarke:
PR Checklist

## Description

Various fixes required when deploying app to server
- Bump node 16 -> 18
- Fix issue not picking package version
- Fix issue not pinning yarn to expected version
- Update methods to install workspace npm production modules
- Code tidying

Additionally on server bumped disk from 30GB -> 50GB as was running low on space, and took snapshots of disk both before `idems-apps-20231102a` and after `idems-apps-20231102b` update 

## Review Notes
Changes already deployed to server so should do functional testing on server and can merge here if working as expected

## Git Issues

Closes #

## Screenshots/Videos

**Latest version of api (1.4.3) running on server**
image

### DIFF
--- a/packages/api/Dockerfile
+++ b/packages/api/Dockerfile
@@ -7,7 +7,7 @@
 # $docker run -p 80:80 --name apps-api --rm idems/apps-api
 
 ### STAGE 1: Build ###
-FROM node:16.14-alpine AS build-env
+FROM node:18-alpine AS build-env
 # Set working directory
 WORKDIR /usr/src/app
 # Add node_modules to path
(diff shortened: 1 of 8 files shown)


esmeetewinkel (approved):
I've tested whether my IDs in different app deployments sync correctly to the tables - all syncing correctly. I confirmed that a user ID that used to sync to the PLH DB table but has now been taken out to sync to a separate table (debug) still exists in the old table but their data is no longer updating there - see screenshots below.

#### My debug app ID (see https://idems-debug.web.app/template/debug_sync_id)
- appearing in PLH DB table: last sync was Oct 6 
   image

- appearing in debug table: last sync was Nov 3 image

#### My PLH TZ app ID (see https://plh-tz.web.app/template/user_code) 
- appearing in PLH DB table: last sync was Nov 3
   image

- does not appear in debug table

jfmcquade (approved):

chrismclarke:
Great, thanks for checking @esmeetewinkel - yes no data will be deleted from the other databases but updated data should go into the right location moving forwards (could decide to manually delete at some point in the future)

Issue #2121, opened 2023-11-02
[FEATURE] Sheet type for JavaScript functions


esmeetewinkel:
**What?**
Ability to author custom JavaScript functions from spreadsheets. It would be good to encourage/enable users to use *any* JavaScript, however, we would then need a parser for these so that we can replace any “new” JavaScript that won’t work on e.g. old Google Chrome. 

**Why?**
Deployments (in this case: `plh_facilitator`) at times require specific functions that are a bit involved to author in a regular template sheet, or are use multiple times.  

**How?**
- [ ] Test whether at  the moment it works to author a JavaScript function in a regular data list (containing key-value pairs) and call it from within `@calc(@data.my_functions.function_id.column_name(input))`
- [ ] In future we could have a data list with a flow subtype “function” which would get parsed in a specific way.

PR #2124, opened 2023-11-03
feat: add calc function: merge_sub_arrays

jfmcquade:
PR Checklist

## Description

Adds a new function available in all `@calc()` blocks: `merge_sub_arrays()`. This is a hack to expose some particular array manipulation logic to authors, for a specific purpose in the upcoming `plh_facilitator_my` deployment release.

`merge_sub_arrays()` takes 3 arguments: an array of arrays (expected to be a partition of a set) and 2 values that appear in sub-arrays. The function looks for the sub-array containing one value and merges it with the sub-array containing another, returning a mutation of the original partition. See screenshots and debug sheets below for usage.

## Notes

I initially tried writing the function in a data list and referring to it in a `@calc()` block. See [debug_functions_data_list](https://docs.google.com/spreadsheets/d/19bOZpBK-spjWaU3gE-EbJoqghmuvYqi7uGaSEZvh0YM/edit#gid=1422144003) and [debug_calc_functions](debug_calc_functions)
Screenshot 2023-11-03 at 10 37 53
Screenshot 2023-11-03 at 10 39 17

However, this approach does not work as the evaluated string from `@calc()` gets _returned_ in the resulting javascript, so additional functions cannot be defined. Further thought is required to enable authors to write functions in data lists and call them in `@calc()` blocks, see #2121.
Screenshot 2023-11-03 at 10 12 06

## Git Issues

Closes #

## Screenshots/Videos

See [debug_calc_functions](https://docs.google.com/spreadsheets/d/19bOZpBK-spjWaU3gE-EbJoqghmuvYqi7uGaSEZvh0YM/edit#gid=1276574703) template for usage
Screenshot 2023-11-03 at 10 37 03

Screenshot 2023-11-03 at 10 37 17

### DIFF
--- a/src/app/shared/components/template/services/template-calc.service.ts
+++ b/src/app/shared/components/template/services/template-calc.service.ts
@@ -120,6 +120,31 @@ const CALC_FUNCTIONS: IFunctionHashmap = {
   },
 
   /**
+   * Mutate an array of arrays (expects a partition of a set): look for the sub-array
+   * containing one value and merge it with the sub-array containing another
+   * @param partition e.g. [[1, 2], [3], [4, 5, 6]]
+   * @param value1 e.g. 3
+   * @param value2 e.g. 5
+   * @returns A new partition, e.g. [[1, 2], [3, 4, 5, 6]]
+   */
+  merge_sub_arrays: (partition: any[][] | string, value1: string, value2: string) => {
+    try {
+      if (typeof partition === "string" || partition instanceof String)
+        partition = JSON.parse(partition as string) as any[];
+      const mergedArray = partition
+        .filter((arr) => arr.includes(value1) || arr.includes(value2))
+        .flat();
+      const partitionUpdated = partition.filter(
+        (arr) => !arr.includes(value1) && !arr.includes(value2)
+      );
+      partitionUpdated.push(mergedArray);
+      return partitionUpdated;
+    } catch (error) {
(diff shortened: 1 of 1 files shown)


src/app/shared/components/template/services/template-calc.service.ts
```diff
@@ -119,6 +119,31 @@ const CALC_FUNCTIONS: IFunctionHashmap = {
...
 
+  /**
+   * Mutate an array of arrays (expects a partition of a set): look for the sub-array
+   * containing one value and merge it with the sub-array containing another
+   * @param partition e.g. [[1, 2], [3], [4, 5, 6]]
+   * @param value1 e.g. 3
+   * @param value2 e.g. 5
+   * @returns A new partition, e.g. [[1, 2], [3, 4, 5, 6]]
+   */
+  merge_sub_arrays: (partition: any[][] | string, value1: string, value2: string) => {
```
chrismclarke, line 130: I don't think this name is very clear (I would expect merging the arrays to concatenate all the child entries into a single entry. If specific function just for one temp context I think probably call it something like `plh_merge_families` or something similarly explicit.

I would also recommend putting the function in its own file and adding a lightweight spec test

chrismclarke: I've pushed potential changes to #2127

PR #2128, opened 2023-11-04
Fix/web browser startup

### LINKED ISSUE #2126: [BUG] Slow initial load on web preview on mobile
**Describe the bug**
When accessing the web preview of the app on mobile, the page hangs for about a minute before the content is rendered. This was first observed for `plh-facilitator-my` deployment, but seems to be an issue for all web previews. See screenshots section for more details

**App version**
0.16.21

**To Reproduce**
On a mobile browser, navigate to plh-facilitator-my.web.app (or another web preview). On my device (Pixel 8), the page is blank for around a minute before the app loads.

**Screenshots and videos**
The following screenshots are of Chrome devtools connected to my phone in USB debugging mode. They show the console logs of the startup services intialising. In each case, there is a pause of around 60 seconds after the `TemplateAsset` service finishes initialising before the `LocalNotifications` service logs that its initialisation has failed (along with the `LocalNotificationInteraction` service and `Campaigns` service). The screen just shows the background colour of the app with no content until this point, at which the expected app content renders.

plh-facilitator-my.web.app
Screenshot 2023-11-03 at 16 40 12

The Performance tab doesn't seem helpful – it only lets me record for 45 seconds, which isn't long enough to capture the eventual page load
Screenshot 2023-11-03 at 16 44 12

plh-tz
Screenshot 2023-11-03 at 16 50 31

idems-debug.web.app
Screenshot 2023-11-03 at 16 51 54

**Additional context**
It appears the issue is with the local notifications service itself, and specifically its initialisation. 

One possible culprit is the `requestPermission()` method. The fact that this bug is only present in the specific case of visiting the web preview on mobile suggests that it's likely related to some native mobile features, for example, the issue could be with the check we run within `requestPermission()` to see if the platform is native, or else with Capacitor's own `LocalNotifications.requestPermissions()`, also within this method. I note in the [docs for Capacitor's LocalNotifications plugin](https://capacitorjs.com/docs/apis/local-notifications), there is mention of changes between Android 12 and 13 with regards to `LocalNotifications.requestPermissions()`, so possibly this issue was introduced by the migration to Capacitor 5 and support for Android 13, #2046.

Screenshot 2023-11-03 at 17 18 46

As the local notifications service `init()` method involves some database activity, and as this issue affects all web previews on mobile and has only just been observed, it is possible that the issue is related to the recent changes to the database(s) and APIs introduced with #2115. However I'm not sure at this stage (does the db service even connect to the remote databases?).

chrismclarke:
PR Checklist

## Description

Running web apps from android mobile fail as an error is thrown from the local notifications service which blocks app initialisation. This is due to a lack of support for the capacitor notifications api within android web (`https://github.com/ionic-team/capacitor/issues/3472`)

This PR provides a small fix by using a try-catch statement to avoid throwing the error (failing silently, like in cases where notifications are disabled by the user)

## Review Notes
Can try the preview link from a mobile browser and make sure the app loads as expected.

I've additionally added a new serve configuration so that the local live-reload app can be run on any device on the same wifi network via `yarn ng serve --configuration=external`. If serving like this then any device on local network should be able to connect to the app by typing in the local ip of the machine serving (on windows get via `ipconfig` command, various other methods exist on linux/mac - should be something like `192.168.x.x`.

I've also added some extra instructions within the `capacitor.config.ts` file for how to do a similar process but instead of live-reload from web app use live-reload on native app (e.g. parenting app running in an emulator). This is similar to above, with an additional step of defining the ip in the config file and syncing to the native build.

Feel free to document either of these methods in the app docs if you think useful to have (info likely also in capacitor docs)

## Git Issues

Closes #2126

## Screenshots/Videos

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -147,6 +147,10 @@
             },
             "ci": {
               "progress": false
+            },
+            "external": {
+              "browserTarget": "app:build:development",
+              "host": "0.0.0.0"
             }
           }
         },
--- a/capacitor.config.ts
+++ b/capacitor.config.ts
@@ -20,6 +20,15 @@ const config: CapacitorConfig = {
   },
   server: {
     androidScheme: "http",
+    /**
+     * NOTE - to support live-reload on external device (e.g. emulator)
+     * 1) Uncomment url and replace with local ip to serve live-reload on local device
+     * 2) Sync to capacitor `npx cap sync`
+     * 3) Serve via `yarn ng serve --configuration=external`
+     * 4) Run app from android studio `npx cap open android` and run
+     * Local browser (localhost:4000), device app and device browser ([ip]:4200) should all be able to access served app
+     **/
+    // url: "http://192.168.50.67:4200",
   },
 };
 
--- a/src/app/shared/services/notification/local-notification.service.ts
+++ b/src/app/shared/services/notification/local-notification.service.ts
@@ -136,22 +136,33 @@ export class LocalNotificationService extends AsyncServiceBase {
         return false;
       }
     }
-    // Return state of already granted
-    const { display } = await LocalNotifications.checkPermissions();
-    if (display === "granted") {
-      return true;
+
+    try {
+      // Return state of already granted
+      const { display } = await LocalNotifications.checkPermissions();
+      if (display === "granted") {
+        return true;
+      }
+      // Use notifications api to check permissions. Run in parallel with a 5-second
+      // timeout to resolve in cases where prompt does not appear or user fails to interact with it
+      const granted = await Promise.race([
+        new Promise<boolean>((resolve) => setTimeout(resolve, 5000, false)),
+        new Promise<boolean>(async (resolve) => {
+          const { display } = await LocalNotifications.requestPermissions();
+          resolve(display === "granted");
+        }),
+      ]);
+      // console.log("[Notifications Enabled]", granted);
+      return granted;
+    } catch (error) {
+      // Catch case where capacitor throws error on android web
+      // (notifications supported in browser but method of constructing fails)
+      // https://github.com/ionic-team/capacitor/issues/3472
+      // https://github.com/ionic-team/capacitor-plugins/blob/main/local-notifications/src/web.ts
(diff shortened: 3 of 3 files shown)


jfmcquade (approved):
That's great, I can confirm that when running the PR web preview on my phone, I do not experience the long load time issue.

And that's really nice to have a way of live reloading from other devices on the network. I think it would be good to document, I've not got time now but have added an issue: #2129.
In the case of live-reloading on a native app, I've been using the following command to launch an emulator instance running a live-reloading build: `ionic cap run android -l --external`, taken from [these docs](https://ionicframework.com/docs/cli/commands/capacitor-run). It has the advantage of not needing to launch Android Studio (I think the emulator options are those that have already been configured using Android Studio). Does the new `yarn ng serve --configuration=external` have advantages over this method do you think?

chrismclarke:
> That's great, I can confirm that when running the PR web preview on my phone, I do not experience the long load time issue.
> 
> In the case of live-reloading on a native app, I've been using the following command to launch an emulator instance running a live-reloading build: `ionic cap run android -l --external`, taken from [these docs](https://ionicframework.com/docs/cli/commands/capacitor-run). It has the advantage of not needing to launch Android Studio (I think the emulator options are those that have already been configured using Android Studio). Does the new `yarn ng serve --configuration=external` have advantages over this method do you think?

Looking at the docs the `external` flag basically does the same job of setting the bind address, but would cover cases where external bind is not `0.0.0.0` (think maybe running inside a container environment or some specific os's?)
https://github.com/ionic-team/ionic-cli/blob/develop/packages/%40ionic/cli/src/lib/serve.ts#L40
https://github.com/ionic-team/ionic-cli/blob/develop/packages/%40ionic/utils-network/src/index.ts#L30

So I think this method should probably be preferable (only downside I can think of is if we want to use in an environment where we are using capacitor but not ionic, which is the case for the picsa app I adapted the code from)

Assuming it can also be used to access web version in same way by specifying local ip. When documenting we should test this and can probably remove the angular.json and capacitor.config.ts code changes from this pr in favour of the ionic cli methods.

Issue #2131, opened 2023-11-05
[Discussion] Template calc system improvements


chrismclarke:
**Details**
As part of work on plh a number of custom calc functions were required to simplify what would otherwise be very complex templating logic. These calculations are currently hardcoded in to the app, however we will want to come up with a better system for extending calc functions more generally.

A couple potential solutions
**Config-based calculations**
Could support a field in app config to provide custom calculations. The benefit of this would be ease of integration with the existing calc statements (just needs a line to import from app config) and using existing app compiler to support calcs written in typescript/esnext). Drawback would be harder to share functions across deployments

**Flow-based calculations**
Provide a custom flow type from which users can define calculations to be made globally available. The benefits of this would be more accessible to authors (?) and ability to share across deployments, drawbacks would be the calcs being one step further removed from runtime environment so would likely need custom parser/interpreter to ensure syntax is valid

**Additional Considerations**
- How to support spec tests for deployment-based calcs (e.g. allow author to include test input/output in sheets, pass a spec.ts sheet etc.)
- Confirm if possible for defined functions to have access to other functions (either defined, or potentially imported from a 3rd party lib? In the latter case would likely require bundling calc statements with all dependencies inlined)
- How best to ensure functions written have valid syntax that works in target browsers (e.g. whether require additional linting and compilation, or if outside scope of the system)

PR #2132, opened 2023-11-06
fix: plh-function shared references

chrismclarke:
PR Checklist

## Description
When a calc function tries to refer to another calc function everything works fine in development, however in production there is a mismatch in the names of functions made available to evaluator context, and the names of the functions as minified by the angular compiler. 

This PR adds a small workaround so that all calc functions can also be accessed via a `window.calc` method, which preserves the names. 

## Review Notes
Create a production build
```sh
yarn build
```
Use npm `serve` module to serve the contents of the production build locally (reads from local `seve.json` for config)
```sh
npx serve
```

## Dev Notes
I've added to #2131 as a reminder to try and address this issue in a better way in the future if possible (currently still possible for developers to write functions that work locally but not production)

## Git Issues

Closes #

## Screenshots/Videos
Debug sheets working in production serve (port 5000). Previously was white-screen

image


esmeetewinkel (approved):
Functional test passed

Issue #2120, opened 2023-11-02
Pop-up follow up issue


esmeetewinkel:
Functional test passed. I'm happy for this to be the default behaviour. I'm not sure it's worth putting in the time now to expose this as a setting. I agree it looks like this was always the intended behaviour (from [example_pop_ups](https://docs.google.com/spreadsheets/d/1K9JZ9Glnj73CFJmpfGIRa5dRA2TBxSsJmaguV6GL8Pc/edit#gid=402992224)) and I think the knock-ons this will have on other apps will be minimal as we barely use pop-ups in this way up to now. 

There's one edge case that the PR does not seem to address: When a pop-up has a button with a go-to action to another template, and that template emits completed, then the user gets redirected to the pop-up, independent of whether the initial button contained an action dismissing the pop-up or not (debug example added to [debug_pop_up_dismiss](https://docs.google.com/spreadsheets/d/1XDugKSTqqEScibrWH0OkTTFf-lqsZREob37EeieQUWI/edit#gid=963655109))

[screen-capture (64).webm](https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/05001721-e983-4c33-ae17-59159fe8a946)

It would make sense to me that, if the initial button both has an action dismissing the pop-up and a go-to action, and the second template emits completed, then the user is redirected to the template underlying the pop-up. 

I'm happy to merge this PR as is and address this in a follow-up PR.

_Originally posted by @esmeetewinkel in https://github.com/IDEMSInternational/parenting-app-ui/pull/2117#pullrequestreview-1710097784_

esmeetewinkel:
Updated [debug_pup_up_return](https://docs.google.com/spreadsheets/d/1XDugKSTqqEScibrWH0OkTTFf-lqsZREob37EeieQUWI/edit#gid=10879997) template to include examples of both `click | emit: completed` and `completed | emit: completed` on the pop-up button action. 

<img width="520" alt="image" src="https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/9e9cd1af-a26d-4773-9457-9cb59cc6c7d9">

jfmcquade:
This issue describes desired behaviour where a pop-up that triggers a `go_to` action should dismiss when the child template (the target of the `go_to`) emits `completed`, returning back to the parent template that triggered the initial pop-up. I will refer to these three templates as "parent", "intermediary" and "child" from now on.

[...]

PR #2134, opened 2023-11-06
feat: new go_to action argument: dismiss_on_return

jfmcquade:
PR Checklist

## Description

Offers a workaround for #2120. Please see [this comment](https://github.com/IDEMSInternational/parenting-app-ui/issues/2120#issuecomment-1795487136) on that issue for context.

Adds a parameter to the `go_to` action, `dismiss_on_return`, to be used in the specific case of using a `go_to` from an active pop-up. If this parameter is passed `true`, then on returning to the parent template from which the pop-up was launched, the pop-up will be closed.

Whilst the `dismiss_on_return` action arg does what it says on the tin and so is potentially a legitimate feature, it is a workaround/hack for our current use case and doesn't fully demonstrate the desired functionality. The intermediary template (the pop-up) will dismiss on return to the parent template regardless of how the user navigates there. This is to say, it will not only dismiss when a `completed`/`uncompleted` action has been emitted by the child template, it will also dismiss, for example, when the user navigates back from the child template using the back button. A true fix for #2120 would pass information back from the child template and only dismiss the intermediary template in the case of a `completed`/`uncompleted` emission from the child template (see the comment on the issue).

## Git Issues

On merge, #2120 should probably be closed and a new issue should be made to capture the desired nav/pop-up functionality

## Screenshots/Videos

Updated [debug_pop_up_return](https://docs.google.com/spreadsheets/d/1XDugKSTqqEScibrWH0OkTTFf-lqsZREob37EeieQUWI/edit#gid=10879997) template:
Screenshot 2023-11-06 at 16 58 18

Demo

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/f5987ee5-d419-4fae-9e32-6ad04ce3c86a

### DIFF
--- a/src/app/shared/components/template/services/template-nav.service.ts
+++ b/src/app/shared/components/template/services/template-nav.service.ts
@@ -5,7 +5,7 @@ import { ModalController } from "@ionic/angular";
 import { first } from "rxjs/operators";
 import { FlowTypes } from "src/app/shared/model";
 import { SyncServiceBase } from "src/app/shared/services/syncService.base";
-import { arrayToHashmapArray } from "src/app/shared/utils";
+import { arrayToHashmapArray, parseBoolean } from "src/app/shared/utils";
 import { ITemplatePopupComponentProps, TemplatePopupComponent } from "../components/layout/popup";
 import { ITemplateContainerProps } from "../models";
 import { TemplateContainerComponent } from "../template-container.component";
@@ -79,11 +79,16 @@ export class TemplateNavService extends SyncServiceBase {
   public handleNavAction(action: FlowTypes.TemplateRowAction) {
     // TODO: Find more elegant way to get current root level template name
     const parentName = location.pathname.replace("/template/", "");
-    const [templatename] = action.args;
+    const [templatename, key, value] = action.args;
     const nav_parent_triggered_by = action._triggeredBy?.name;
     const queryParams: INavQueryParams = { nav_parent: parentName, nav_parent_triggered_by };
     // handle direct page or template navigation
     const navTarget = templatename.startsWith("/") ? [templatename] : ["template", templatename];
+
+    // If "dismiss_on_return" is set to true for the go_to action, dismiss the current popup before navigating away
+    if (key === "dismiss_on_return" && parseBoolean(value)) {
+      history.replaceState({}, "", location.pathname);
+    }
     return this.router.navigate(navTarget, {
       queryParams,
       queryParamsHandling: "merge",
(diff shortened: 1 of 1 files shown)


src/app/shared/components/template/services/template-nav.service.ts
```diff
@@ -79,11 +79,16 @@ export class TemplateNavService extends SyncServiceBase {
...
-    const [templatename] = action.args;
+    const [templatename, key, value] = action.args;
     const nav_parent_triggered_by = action._triggeredBy?.name;
     const queryParams: INavQueryParams = { nav_parent: parentName, nav_parent_triggered_by };
     // handle direct page or template navigation
     const navTarget = templatename.startsWith("/") ? [templatename] : ["template", templatename];
+
+    // If "dismiss_on_return" is set to true for the go_to action, dismiss the current popup before navigating away
+    if (key === "dismiss_on_return" && parseBoolean(value)) {
+      history.replaceState({}, "", location.pathname);
```
jfmcquade, line 90: Originally I tried using `this.location.go(location.pathname)` and `router.navigate()`, but in that case the page with the query params defining the popup remained in the history, meaning that upon return to the parent template, navigating back would cause the popup to reappear.

esmeetewinkel (approved):
Functional test passed. Thanks for creating a work-around on such a short notice! 

Syntax (not so important since this is a temporary solution anyway): I had imagined instead of 
`click | go_to: example_emit: dismiss_on_return: true;`
something like 
`click | go_to: example_emit; click | dismiss_on_return: true;`
as I hadn't really seen the dismiss on return as a property of the go-to action. In my mind `dismiss_on_return` could equally apply to a pop-up navigating to another (maybe full-screen) pop-up. Or maybe it is a property of the pop-up (and so a parameter to the action that creates the pop-up)?

src/app/shared/components/template/services/template-nav.service.ts
chrismclarke: Like the idea of allowing authors to explicitly pass a param to manage this instead of having to rely tracking weird nav history. Although I'm not sure if this line would inadvertently remove additional url info that we might want to keep. 

An alternative snippet that I've just tested locally (and appears to work) might be:
```ts
 if (key === "dismiss_on_return" && parseBoolean(value)) {
      const { popup_child } = this.route.snapshot.queryParams;
      if (popup_child) {
        const popupDismissParams: INavQueryParams = {
          popup_child: null,
          popup_parent: null,
          popup_parent_triggered_by: null,
        };
        // alter route history so that on back-nav popup will not be present
        await this.router.navigate([], {
          relativeTo: this.route,
          queryParams: popupDismissParams,
          queryParamsHandling: "merge",
          replaceUrl: true,
        });
        // Dismiss open popup (without await to allow rest of nav to proceed await)
        this.dismissPopup(popup_child);
      }
    }
```

However as the PR is already merged I'll leave for now and you can revisit if issue comes up

PR #2135, opened 2023-11-06
refactor!: popup-nav-dismiss

chrismclarke:
PR Checklist

## Description

Possible alternate implementation for behaviour in #2134

## Breaking changes
Added by @jfmcquade: Also renames the parameter passed to the `go_to` action: `dismiss_on_return` -> `dismiss_pop_up`. This more accurately describes the behaviour of the feature. This is a breaking change that will require changes to the templates that implement this feature, which appear to be [in_person_unreported](https://docs.google.com/spreadsheets/d/1S7q_hrEpSG-owuRC9oEwMHnTa-ywTJwT2_SEc1rNF1g/edit#gid=1606639158) and [virtual_unreported](https://docs.google.com/spreadsheets/d/1S7q_hrEpSG-owuRC9oEwMHnTa-ywTJwT2_SEc1rNF1g/edit#gid=1696774054). (Alternatively we could support both syntaxes to avoid deprecation, but it should be easy to make the changes at this stage and avoid including the deprecated syntax in the code).

## Git Issues

Closes #

## Screenshots/Videos

Updated [debug_pop_up_return](https://docs.google.com/spreadsheets/d/1XDugKSTqqEScibrWH0OkTTFf-lqsZREob37EeieQUWI/edit#gid=10879997) template:
Screenshot 2023-11-07 at 15 38 28

Updated behaviour showing preservation of query params (previously all query params would have been removed on navigation away from to the pop-up):

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/11d1a1d1-3946-43a5-b4b3-e4a5c45044c5

### DIFF
--- a/packages/data-models/flowTypes.ts
+++ b/packages/data-models/flowTypes.ts
@@ -373,7 +373,6 @@ export namespace FlowTypes {
     "audio_end",
     "audio_play",
     "changed",
-    "close_pop_up",
     "emit",
     "feedback",
     "go_to",
--- a/src/app/shared/components/template/services/template-nav.service.ts
+++ b/src/app/shared/components/template/services/template-nav.service.ts
@@ -76,7 +76,7 @@ export class TemplateNavService extends SyncServiceBase {
   /*****************************************************************************************************
    *  Nav Actions
    ****************************************************************************************************/
-  public handleNavAction(action: FlowTypes.TemplateRowAction) {
+  public async handleNavAction(action: FlowTypes.TemplateRowAction) {
     // TODO: Find more elegant way to get current root level template name
     const parentName = location.pathname.replace("/template/", "");
     const [templatename, key, value] = action.args;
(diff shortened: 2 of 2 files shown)


chrismclarke:
> Thanks, @chrismclarke, I like your refactor. I've updated the PR description and also changed the name of the passed parameter `dismiss_on_return` -> `dismiss_pop_up`, which was agreed to be a more accurate description of the functionality.
> 
> I think this is ready for a functional test. And I think worth merging in anticipation of any query params issues rather than waiting to see if any emerge.

Thanks for the improvements and identifying the breaking change - all makes sense to me. In hindsight, do we already have a click action to `dismiss_popup`? If so that could have been the solution from the start (i.e. add a popup dismiss action before the nav one), but no harm giving authors multiple options I suppose...

esmeetewinkel (commented):
> In hindsight, do we already have a click action for popup_dismiss? If so that could have been the solution from the start (i.e. add a popup dismiss action before the nav one), but no harm giving authors multiple options I suppose...

I'm not aware we have such an action, but this would make lot of sense to me.

jfmcquade:
> > In hindsight, do we already have a click action for popup_dismiss? If so that could have been the solution from the start (i.e. add a popup dismiss action before the nav one), but no harm giving authors multiple options I suppose...
> 
> I'm not aware we have such an action, but this would make lot of sense to me.

It appears there was a `close_pop_up` action ID listed in the code, but no logic to actually handle the functionality, so I've [removed its reference](https://github.com/IDEMSInternational/parenting-app-ui/pull/2135/commits/b6aba7f5edc27d444ede1418b42217d5fcd73822).

I agree that a standalone `pop_up_dismiss` action might have been a simpler approach... Although I don't think making use of the `TemplateNavService.dismissPopup` method without modifications would work, I think I did try that and the popup wasn't dismissed when navigating back to the page.

In any case, I think this PR should be ready to merge now.

esmeetewinkel (approved):
Functional test passed and renaming addressed (will sync content in `plh_facilitator_my` repo post merge).

PR #2144, opened 2023-11-16
Fix: test visual ci action

chrismclarke:
PR Checklist

## Description

Whilst bumping `axios` as part of #2142 it was identified that `test-visual` workflow is currently broken as it relies on the `plh_global` deployment that is no longer included in the repo. Whilst the action does use the common build action that does have support for pulling remote repos, there is an extra `deployment set` as part of the follow-up process required to pull the `contents.json` list of app sheets from deployment src code.

**Main changes**
- Refactor action to retrieve contents.json from `www` build folder instead of src code (meaning we no longer need to set deployment as part of the action and can just use the output from the build action)
- Update default behaviour to always serve the built `www` folder (previously did have option the run locally dev server via `yarn start` command, however makes sense if it always uses production build for comparison)

**Additional changes**
-  the target deployment has been swapped from `plh_global` to `debug` as this contains the component demos and examples that are most likely to identify visual regressions (includes commits from #2142)
- `test-visual-generate` action fixed to trigger on master push (accidentally had if condition copied from compare action, meaning it hasn't been run in past 3 months) - recommend in future we might want to only generate following release
- Manually run the test-visual-generate action to allow compare to run

## Review Notes
Should be able to run locally following instructions updated in readme
```
yarn build
yarn workspace test-visual start compare --clean
```
## Git Issues

Closes #2142

## Screenshots/Videos


jfmcquade (approved):
Great, all looks good. 

I can run the `test-visual start compare` workflow locally (although [by default macOS uses port 5000 for some reason](https://nono.ma/port-5000-used-by-control-center-in-macos-controlce) 🙃)

chrismclarke:
> Great, all looks good.
> 
> I can run the `test-visual start compare` workflow locally (although [by default macOS uses port 5000 for some reason](https://nono.ma/port-5000-used-by-control-center-in-macos-controlce) 🙃)

I hardcoded port 5000 just to not confuse with dev app potentially running on 4200, so not a mac thing this time

jfmcquade:
> I hardcoded port 5000 just to not confuse with dev app potentially running on 4200, so not a mac thing this time

I meant that by default on Mac, port 5000 is already in use (for receiving AirPlay) so the workflow doesn't succeed unless I turn off that functionality. I feel like 5000 is a port that's pretty commonly used by developers, so I'm not sure why macOS doesn't use another more obscure one

chrismclarke:
> I meant that by default on Mac, port 5000 is already in use (for receiving AirPlay) so the workflow doesn't succeed unless I turn off that functionality. I feel like 5000 is a port that's pretty commonly used by developers, so I'm not sure why macOS doesn't use another more obscure one

Ah, that makes sense (the issue, not why the port is in use). Yeah I know firebase serve uses 5000 and I think vercel serve might also (or perhaps they previously did and then swapped to 3000...). In any case, if it becomes bothersome feel free to either change the hardcoded value or just expose an extra variable in commander to specify port (but guessing the action will be run very rarely locally)

PR #2146, opened 2023-11-17
Fix/string evaluation js

### LINKED ISSUE #2141: [BUG] AppDataVariableService evaluator does not handle string values correctly
As highlighted in #2107, the evaluator `AppDataVariableService` introduced in #2040 does not handle string values correctly. See tests added to [src/app/shared/services/data/app-data-variable.service.spec.ts](https://github.com/IDEMSInternational/parenting-app-ui/pull/2107/files#diff-8873dfedef4ee3fd95f8af1a3d11a9856aff28b2c9cdc6683952d0daa53493cc)

Compare to `appStringEvaluator.evaluate()` method found in `packages/shared/src/models/appStringEvaluator/appStringEvaluator.ts`, which explicitly adds string delimiters.

chrismclarke:
PR Checklist

**TODO**
- [ ] Builds on #2107, should update post-merge for cleaner diff. Main changes in cd11a4b

## Description
Fix issue where dynamic string replacements do not support further JS evaluation (e.g. `@fields.name.startsWith("A")`)

## Review Notes
See updated test cases now passing
```js
 it("supports JS evaluation on replaced string values", async () => {
    await service.set("field", "test_value", "hello");
    const evaluated = await service.evaluateExpression("@field.test_value.startsWith('h')");
    expect(evaluated).toEqual(true);
  });
  it("evaluates expressions with string value", async () => {
    await service.set("field", "test_value", "hello");
    const evaluated = await service.evaluateExpression("@field.test_value === 'hello'");
    expect(evaluated).toEqual(true);
  });
```

## Author Notes
It was previously suggested that manual string delimeters could be used as a workaround, e.g. `"@field.test_value".startsWith('h')`. 

This should still work (non-breaking change), but no longer be necessary. I'm not sure if we have any example templates that include this behaviour that can also be checked as part of this PR

## Git Issues

Closes #2141

## Screenshots/Videos

### DIFF
--- a/src/app/shared/services/data/app-data-variable.service.spec.ts
+++ b/src/app/shared/services/data/app-data-variable.service.spec.ts
@@ -70,8 +70,8 @@ describe("AppDataVariableService", () => {
 
   it("parses expressions", async () => {
     await service.set("field", "expression_field_1", "value_expression_field_1");
-    const evaluated = await service.parseExpression("hello @field.expression_field_1");
-    expect(evaluated).toEqual("hello value_expression_field_1");
+    const { parsed } = await service.parseExpression("hello @field.expression_field_1");
+    expect(parsed).toEqual("hello value_expression_field_1");
   });
 
   it("evaluates expressions", async () => {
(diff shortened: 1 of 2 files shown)


jfmcquade (approved):
Looks good, I'm going to merge into the feature branch for #2107 as I think these fixes are required for the functional tests on that PR to pass

PR #2157, opened 2023-11-29
Chore(core)!: Ng 17 updates

chrismclarke:
PR Checklist

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade (changes requested):
I've done some testing based on the packages you identify in the "Description", see notes below.

- Node version. This is particularly worth being aware of for troubleshooting issues that may arise for those working on authoring the various apps

### Tested - working
- angular/fire v7->v17(alpha)
	- Tested auth in emulator using [example_google_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit#gid=579616705) template, see video below

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/b3ae6689-4933-4fce-bfb8-db94ed129410

- firebase v9->v10
	- See above
- ngx-extended-pdf viewer v16->v18
	- comp_pdf template appears to function as expected. I note the difference in the visual test screenshots: this looks like a difference in how the pages are displayed on a very tall viewport: in both cases the pdf viewer correctly focuses on page 3, but whereas the "before" allows the pdf to scroll past the last page, the "after" does not. See below
    Before:
    comp_pdf before
    After:
    comp_pdf after

- katex v0.15->v0.16
	- comp_latex appears to be working as expected

## Tested - issues
- ng2-nouislider v1->v2
	- comp_slider is broken
	slider
	- when exiting a template with a slider component, I get the following error
	slider error

- rxdb v13->v14
	- comp_data_items throws the following error
	- comp_data_items

### Not explicitly tested
- Angular v15 -> v17
- ngx-matomo v1 -> v6
- ngx-lottie v7->v10
- ionic-angular v6->v7

### Other issues
- The video below shows 2 distinct issues relating to template loading and navigation
	- The list of templates available at `/template` no longer auto updates based on search, you need to press enter for the search query to have an effect
	- Sometimes, selecting a particular template from the list does not trigger the template to load

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/3bc0d654-1cad-4955-a87b-bbe3d6bf85bf

chrismclarke:
> Other issues
The video below shows 2 distinct issues relating to template loading and navigation
The list of templates available at /template no longer auto updates based on search, you need to press enter for the search query to have an effect
Appears to be regression introduced by ionic (related to `https://github.com/ionic-team/ionic-framework/discussions/25532`), where change event only fires after _comitted_ change (i.e. user de-focuses input box). I've replaced the `ionChange` event binding to use `ionInput` which fires on input instead b549c66. I also checked the codebase and only instance of ion-searchbar I could find, and only other calls of ionChange are for elements without input so I expect should work as expected (e.g. radio buttons, checkboxes, where click event should commit the change).

> Sometimes, selecting a particular template from the list does not trigger the template to load
I can't recreate, so let me know if the previous fix has resolved the issue

chrismclarke:
**comp_audio**
Separate issue detected when checking - appears the seek bar is mis-aligned from player. Is this something we've knowingly changed @jfmcquade ? Looking at web preview compared to this branch I no longer see the `ion-range` component in use (not sure if change is from ionic or us?)

image

chrismclarke:
> rxdb v13->v14
comp_data_items throws the following error

On closer review this warning is currently thrown on master, so not a regression introduced. The issue appears to be related to multiple function calls all to initialise a specific db collection in parallel (3 requests all to populate db docs, the first passes and next 2 fail). So not a breaking issue, but I've gone ahead and applied a fix to this branch (as I had done that before realising the issue was in master) 993b06f

jfmcquade:
> * ng2-nouislider v1->v2
>   
>   * comp_slider is broken
>   
>     slider
>   
>   * when exiting a template with a slider component, I get the following error
> 
> slider error

After banging my head against a wall looking into the issue with `comp_slider`, it seems to have started reliably working for me. 

Screenshot 2023-12-08 at 15 41 09

What _appeared_ to fix the issue was manually reverting the version number of `nouislider` in `package.json` to `^14.0.0` and running `yarn install`, before manually changing it back to `^15.7.1` and running `yarn install` again. Before I did this, I was testing by alternately checking out `master` and `core/ng-17` and running `yarn install each time`, to test the functionality on the respective branches. Every time I checked out `core/ng-17` ran `yarn start` (after `yarn install`), the `nouislider` component wouldn't render any content inside it. Now, after manipulating `package.json`, it seems that it reliably does render correctly.
If this was the source of the fix, then I can only assume it was some kind of caching error where the correct package version was not actually being used? Although yarn should handle such issues. I may have mis-diagnosed the solution, but there have been no changes to local files tracked by git.

jfmcquade:
> Appears to be regression introduced by ionic (flagged in https://github.com/ionic-team/ionic-framework/blob/main/BREAKING.md#searchbar), where change event only fires after _comitted_ change (i.e. user de-focuses input box). I've replaced the `ionChange` event binding to use `ionInput` which fires on input instead [b549c66](https://github.com/IDEMSInternational/parenting-app-ui/commit/b549c66c59436f7a361f2268d5c160ad96c2dc1b). I also checked the codebase and only instance of ion-searchbar I could find, and only other calls of ionChange are for elements without input so I expect should work as expected (e.g. radio buttons, checkboxes, where click event should commit the change).

I have scanned through the ionic changelog looking for other potential issues. The only thing that stands out is, what you've identified here: various other components now also do not emit `ionChange` when they used to. For example, for `ion-checkbox`:
> `ionChange` is no longer emitted when the `checked` property of `ion-checkbox` is modified externally.

So if we change the value of the checkbox component from elsewhere in the template, `ionChange` will not be triggered and any actions set to trigger on the the `changed` trigger on the checkbox will not be triggered. As far as I can tell, this isn't something we currently do, but is worth being aware of.

An equivalent change has been made to `datetime`, `ion-range`, `ion-input`, `ion-select`, `ion-textarea`, `ion-radio-group` – I've looked into how we use each (if at all) and I don't believe the change affects us.

One other component which has this change and for which we do use `ionChange` is `ion-toggle`. We don't use `ionChange` on `ion-toggle` in the `toggle-bar` component, but we do use it in the `debug-toggle` component. As far as I can tell, the debug toggle is the only way to set debug mode, so I don't foresee any issues.

chrismclarke:
> What _appeared_ to fix the issue was manually reverting the version number of `nouislider` in `package.json` to `^14.0.0` and running `yarn install`, before manually changing it back to `^15.7.1` and running `yarn install` again. Before I did this, I was testing by alternately checking out `master` and `core/ng-17` and running `yarn install each time`, to test the functionality on the respective branches. Every time I checked out `core/ng-17` ran `yarn start` (after `yarn install`), the `nouislider` component wouldn't render any content inside it. Now, after manipulating `package.json`, it seems that it reliably does render correctly. If this was the source of the fix, then I can only assume it was some kind of caching error where the correct package version was not actually being used? Although yarn should handle such issues. I may have mis-diagnosed the solution, but there have been no changes to local files tracked by git.

Super weird and imagine a total headache, but good to know things working now at least. So in terms of outstanding issues is it just `comp_audio` now as far as we're aware?

esmeetewinkel (changes requested):
While doing functional testing, I found issues with a couple of dynamic components. 

1. A line appears below a text box when you click it to type text
   image
2. Checkbox "unticked" background colour changed to light grey
   image
3. Toggle bar visually broken (actions are working fine): in FALSE state the ball doesn't appear entirely on the LHS of the control, and the TRUE state doesn't work at all. The same problem (but reversed) happens when the starting value of the toggle was TRUE. Also it seems that our custom toggle styles (with tick and cross icons) are no longer interpreted. 

   [screen-capture (76).webm](https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/2875ad35-6492-4ff6-8304-f6e60a2c748a)

chrismclarke:
Thanks for checking all these @esmeetewinkel 

@jfmcquade 
> 1. A line appears below a text box when you click it to type text

Looking through changelogs I can't actually find when/why this has started being applied, but it comes from an `.input-highlight` class which we may want to consider overriding (e.g. adding `display:none`) if we do not want this behaviour on any components that use an ion-input. 

> 2. Checkbox "unticked" background colour changed to light grey

Looks like this should have been caught from the [Breaking Changes](https://github.com/ionic-team/ionic-framework/blob/main/BREAKING.md#checkbox) doc. Would you be able to fix and take another pass for any similar potential issues?

> 3. Toggle bar visually broken (actions are working fine): in FALSE state the ball doesn't appear entirely on the LHS of the control, and the TRUE state doesn't work at all. The same problem (but reversed) happens when the starting value of the toggle was TRUE. Also it seems that our custom toggle styles (with tick and cross icons) are no longer interpreted.

I'm guessing this is related to the ionChange updates described in breaking changes, but very weird given that we don't actually use that method (guessing there's some knock-on for how it handles click events). From what I can tell the click event is firing twice, so guessing there's some deeper changes to underlying elements used. I've fixed by adding an additional event block in 2c18594

> 4. On the debug template `comp_odk_form` the back button in the top left corner appears tiny
This is a much more awkward one, it appears a lot of the ionic components have updated styles to move away from fixed units (e.g. 48px) to relative ones (e.g. 3rem). Throughout most of the app this has no noticeable effect (as default em/rem tends to be 16px), but the odk component interferes with this.
The changes are somewhat out of control, as the odk component has to import external stylesheets (where I'm assuming the changes are coming from), and have no view encapsulation to display correctly (hence why changes are able to bleed out). 

So I'll need to take a deeper dive to try figure out a plan for this. I'm not sure if we're actively using this component anywhere, it might be worth temporarily disabling whilst we get the rest of changes merged in as the impact persists across all templates viewed after the odk component has been loaded

esmeetewinkel:
> > 7\. On the debug template `comp_odk_form` the back button in the top left corner appears tiny
> >     This is a much more awkward one, it appears a lot of the ionic components have updated styles to move away from fixed units (e.g. 48px) to relative ones (e.g. 3rem). Throughout most of the app this has no noticeable effect (as default em/rem tends to be 16px), but the odk component interferes with this.
> >     The changes are somewhat out of control, as the odk component has to import external stylesheets (where I'm assuming the changes are coming from), and have no view encapsulation to display correctly (hence why changes are able to bleed out).
> 
> So I'll need to take a deeper dive to try figure out a plan for this. I'm not sure if we're actively using this component anywhere, it might be worth temporarily disabling whilst we get the rest of changes merged in as the impact persists across all templates viewed after the odk component has been loaded

I can confirm we're not actively using the ODK component in any deployment at the moment.

chrismclarke:
ODK form styling should fixed with 582452b. 

I'm not entirely sure why they were originally setup the way they were, but now should be working as expected

esmeetewinkel (changes requested):
I get a number of warnings when running `yarn start` (after first running `yarn install`): 

image

image

Further to my issues flagged [above](https://github.com/IDEMSInternational/parenting-app-ui/pull/2157#pullrequestreview-1783600440): 
1. Still outstanding (blocking) - @jfmcquade Chris asked you to address [above](https://github.com/IDEMSInternational/parenting-app-ui/pull/2157#issuecomment-1857967567)
2. Still outstanding, but I don't actually think the new style looks worse than the original style so I'm happy to leave as is. 
3. Toggle bar "firing twice" is fixed, but custom toggle styles (with tick and cross icons) are still broken
   
image   vs    image

   
4. Addressed, functional test passed

jfmcquade:
Thanks @esmeetewinkel for bumping this. I'll have a look this week and try to resolve what I can (including merge conflicts)

chrismclarke:
To summarise from call:
- Would be good to further bump minor versions on ionic (7.6.x) and angular (17.2.x) as could include bug fixes for issues encountered
- Will need to do quick re-review of ionic changelog for v7 breaking changes, review output from screenshot tests and also manual test of more interactive components

jfmcquade:
I've performed the planned updates, and added a few minor fixes. I believe we've now addressed all of the sisues that came up in the PR comments. @chrismclarke, you can see my changes since your last commit [here](https://github.com/IDEMSInternational/parenting-app-ui/pull/2157/files/d070edd3fd2bdc5212736f93d21f5d3b17975ec8..e3018dd2670663c3d8101ae2d153f9f590acc87c#diff-5d3156d82ae832a88c85fdebd5afadc2b8226f1ebf4f06a2127267d1b3546a95).

### Ionic
Updated `@ionic/angular` 7.5.6 -> 7.7.3.
I read through the release notes [here](https://github.com/ionic-team/ionic-framework/releases), I couldn't find any changes that I'd expect to cause any issues, except for some updates to the `ion-toggle` component.
Our toggle component (which uses [ion-toggle](https://ionicframework.com/docs/api/toggle)) uses a now legacy syntax. I've made a PR to update this: #2212.

### Angular
Updated Angular 17.0.x -> 17.2.2.
Running `ng update` suggested the following package updates:

      @angular-eslint/schematics                 17.1.0 -> 17.2.1        
      @angular/cli                               17.0.10 -> 17.2.1        
      @angular/core                              17.0.9 -> 17.2.2         
      @angular/fire                              17.0.0-next.0 -> 17.0.1  
      @ngx-matomo/tracker                        1.3.6 -> 4.1.2           

I've consulted the [Angular changelog](https://github.com/angular/angular/blob/main/CHANGELOG.md) and didn't find any expected problems.

#### Matamo
When running `ng update @angular/core`, Angular CLI flagged that `@ngx-matomo/router` and `@ngx-matomo/tracker` required updating (^1.3.3 -> ^4.1.2) to maintain compatibility with `@angular/core`. After applying these updates, I discovered that we don't actually use these packages as far as I can tell, and they have been replaced with [ngx-matomo-client](https://www.npmjs.com/package/ngx-matomo-client), which is updated to the latest version in this PR. So I have removed `@ngx-matomo/router` and `@ngx-matomo/tracker`.

#### Sentry
Likewise `@sentry/angular` also required an update, changing the package from [@sentry/angular](https://www.npmjs.com/package/@sentry/angular) to [@sentry/angular-ivy](https://github.com/getsentry/sentry-javascript/tree/master/packages/angular-ivy) docs [here](https://docs.sentry.io/platforms/javascript/guides/angular/). The API appears to be the same, none of the method calls (solely in `src/app/shared/services/error-handler/error-handler.service.ts`) have any issues flagged within the code.

chrismclarke:
Thanks @jfmcquade - I appreciate that I'm the one who opened the PR so can't give a review, but gone through all your changes and looks good to me. I've also added a note #2211 to update sentry import there accordingly.

Agreed we can check on sentry and matomo post merge, and thanks for adding the follow-up issue for ion-toggle

jfmcquade (approved):
I believe this should be good to merge. As mentioned, sentry and matamo functionality should be tested soon after merge. 

@esmeetewinkel I'm not sure how necessary it is to functionally test all the relevant components again, but probably worth looking over the PR comments and re-testing anything that stands out, and also looking in particular at the toggle bar and checkbox components.

esmeetewinkel (changes requested):
Functional testing of text box and check box passed. 

The "slider bar" on the slider component has disappeared
image

The text input within a combo box appears with a green (?) line below it and green cursor
image

Should we worry about the errors and warnings appearing in the console? 
image
image

jfmcquade:
Thanks, @esmeetewinkel.

I've pushed changes to fix the first 3 issues you mention:
- slider component
- combo box text input
- console error: `Acquiring an exclusive Navigator LockManager failed`

Regarding the yellow warnings logged in your final screenshot:
I'm only getting the `RxDB dev-mode warning` one:
Screenshot 2024-02-27 at 16 03 53
Are there any steps you take that cause those warnings to appear, e.g. navigating to a particular template?
I've definitely seen the warnings related to ODK before, I think we just have to accept them for now as they are related to the ODK package we're using. 

As far as the 3 warnings regarding "CommonJS or AMD dependencies" are concerned, @chrismclarke could you advise? Should I be adding these packages to `allowedCommonJsDependencies` in `angular.json`? Or should we be looking to move away from these packages altogether?

chrismclarke:
> Regarding the yellow warnings logged in your final screenshot: I'm only getting the `RxDB dev-mode warning` one:

During compilation I expect you probably see a lot of warnings coming from `./src/app/shared/components/template/components/odk-form`, but these won't always appear in the console depending whether it is first load or a reload action.

There's not a huge amount we can do about all the odk warnings at this stage. It's 3rd party code that we've essentially copy-pasted into the repo, so not going to conform to our syntax guidelines. The best solution would probably be to get it all running as part of a standalone repo and just importing the compiled version from there, that's very far down a list of priorities I already have on a separate project I'm working on (PICSA), so I think possibly not worth too much time investment from here.

> As far as the 3 warnings regarding "CommonJS or AMD dependencies" are concerned, @chrismclarke could you advise? Should I be adding these packages to `allowedCommonJsDependencies` in `angular.json`? Or should we be looking to move away from these packages altogether?
Yes you simply need to add the names of the `allowedCommonJsDependencies` in the `angular.json` file. CommonJS is never going away, some packages will never migrate to ESM and that's not a fundamental issue (yes, harder for compilers to optimise, but a well-crafted commonJS package will still be better than a poorly assembled ESM one), so I think a bit of overreach on the angular warnings that can be ignored (given the common actions are to either ignore or add to the list of warnings to suppress).

chrismclarke:
@jfmcquade 
Judging by `https://github.com/angular/angular-cli/pull/26047` it might now just be possible to set `'*'` in the list of `allowedCommonJsDependencies` to surpress all warnings

jfmcquade:
Thanks @chrismclarke. The new wildcard option does seem to work for suppressing the CommonJS warning messages.
Screenshot 2024-02-28 at 15 10 16

@esmeetewinkel I think this should be ready to merge once you're satisfied I've addressed the points in your last comment

esmeetewinkel (approved):
Functional test passed

PR #2159, opened 2023-12-06
chore: revert packages/api/package.json to use LF line endings

jfmcquade:
PR Checklist

## Description

In a repeat of the case outlined in #1973, a dependabot PR (in this case #2158) changed the line-endings of `packages/api/package.json`, causing issues with changes to that file that refused to be committed locally. This reverts the changing of the line endings in that file (favouring LF).

In the future, unless we are able to configure dependabot in some way, we should look out for deps PRs where `packages/api/package.json` has been changed in this way, and manually revert the changes before merging.


chrismclarke (approved):
I'm guessing there's probably some ci tool that can test line-endings but probably not worth it given that's its a pretty easy fix if/when it does crop up. But otherwise all looks good to me for now

PR #2160, opened 2023-12-07
Feat: share assets; save assets to device

### LINKED ISSUE #1951: [FEATURE] Share files e.g. PDF
**What?**
Following on from #1941, we would like to be able to share files from within the app. 

**Why?**
An initial use-case is sharing PDFs via WhatsApp

**How?**
The functionality of the share feature introduced by #1941 should be expanded to expose the file-sharing feature supported by the Capacitor Share API. There will need to be a method for converting the relative asset path provided by authors into a local filepath URL (e.g. a path beginning `file://` on Android)  – this will have some overlap with the file-handling methods introduced by the [remote assets](https://github.com/IDEMSInternational/parenting-app-ui/milestone/4), and should be straightforward to include after the latest incremental remote assets PR is merged: #1923.

### LINKED ISSUE #2064: [FEATURE] Download PDF to device
**What?**
There's been a request by the FacilitatorApp to have download of PDF files to the device. 

**Why?**
The reasoning is that 
1. People didn't find the in-app PDF viewer easy to use because it doesn't take up the full screen, and they were using devices with relatively small screens. They prefer using another PDF viewer on their device. 
2. People want to save the PDF to their device to be able to access it easily afterwards.

I do feel that these particular issues could also be resolved by making our PDF viewer and use thereof better...

**How?**
I'm not sure if this should be a feature of the PDF viewer (PR https://github.com/IDEMSInternational/parenting-app-ui/pull/1913) or that this is more related to sharing files [#1951](https://github.com/IDEMSInternational/parenting-app-ui/issues/1951).

jfmcquade:
PR Checklist

## Description

Adds functionality to share files and download files.

### Share
- New template actions:
	- `share: file`
		- Pass path of asset to share, e.g. `share: file: images/example/jasper_1.jpg`
	- `share: text`
		- Pass string to share, e.g. `share: text: My text`
	- `share: url`
		- Pass a  URL to share, e.g. `share: url: https://www.idems.international/`
- The latter two replace the legacy `share` action, e.g. `share: My text`. This syntax is still supported for backwards compatibility, but should be considered deprecated.

See videos below demonstrating the share functionality on web and Android platforms.

### Save to device
- New template action: `save_to_device`
- Expects 1 argument: the path to an asset file (in usual format), e.g. `save_to_device: images/example/jasper_1.jpg`
- Behaviour:
  - Web
    - Download file
  - Native (Android)
    - Save file to "Downloads" folder in external device storage and open
    - Repeatedly triggering `save_to_device` with the same asset path passed in will overwrite the original file each time
  - See videos below for demonstration

### Open external
- New template action: `open_external`
- Expects 1 argument: the path to an asset file (in usual format), e.g. `open_external: images/example/jasper_1.jpg`
- Behaviour:
  - Web
    - Open file in new tab, or download if not viewable in browser
  - Native (Android)
    - Save file to app's internal cache on device and open it externally to the app.
      - Differs from `save_to_device` in that the file is not permanently saved and is not accessible through external storage
  - See videos below for demonstration

## Testing

Using both the PR web preview and the appetize link, test all of the new functionality (ideally try multiple Android versions with the latter).

The relevant templates are [feature_share](https://docs.google.com/spreadsheets/d/1VJFxbuUtKLiw9HucShQalalKcVpOKjryOzfKQlBHE_o/edit#gid=1060209527), [feature_share_file](https://docs.google.com/spreadsheets/d/1VJFxbuUtKLiw9HucShQalalKcVpOKjryOzfKQlBHE_o/edit#gid=1105407426) and [feature_save_open_file](https://docs.google.com/spreadsheets/d/1_5XijOqleQJSZzWthlH4Zkyw4TmKCiup8b69yglihcQ/edit#gid=960980632).

## Dev notes

New permissions are required, namely `READ_EXTERNAL_STORAGE` and `WRITE_EXTERNAL_STORAGE`.

The [Capacitor docs](https://capacitorjs.com/docs/apis/filesystem#directory) says the following regarding writing to external storage on Android:
> It's not accesible on Android 10 unless the app enables legacy External Storage by adding android:requestLegacyExternalStorage="true" in the application tag in the AndroidManifest.xml. It's not accesible on Android 11 or newer.

And they offer no alternative method of writing to external storage. However, testing on Android 12, the functionality does work as expected (see videos below).

If any of this causes an issue, we could potentially remove the "save to device" functionality for now, and rely on the "open external" functionality (which does not write to the external storage or need extra permissions).

## Git Issues

Closes #1951 
Closes #2064 

### Share

[feature_share](https://docs.google.com/spreadsheets/d/1VJFxbuUtKLiw9HucShQalalKcVpOKjryOzfKQlBHE_o/edit#gid=960980632) template
Screenshot 2023-12-08 at 11 33 24

[feature_share_file](https://docs.google.com/spreadsheets/d/1VJFxbuUtKLiw9HucShQalalKcVpOKjryOzfKQlBHE_o/edit#gid=1105407426) template
Screenshot 2023-12-08 at 11 33 49

#### Share in browser (Safari, macOS)

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/7e5bf63e-1a7a-4eac-b544-3de4ad3de64f

#### Share on Android (emulator)

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/24d22e61-5e5a-420f-9480-a74bb6d7713a

### Save to device

[feature_save_open_file](https://docs.google.com/spreadsheets/d/1_5XijOqleQJSZzWthlH4Zkyw4TmKCiup8b69yglihcQ/edit#gid=960980632) template
Screenshot 2023-12-13 at 13 26 53

#### Web (Chrome, macOS)

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/67cbe4b9-07e5-4cf1-a416-aef6eeb9585a

#### Android (emulator)

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/42b2bb1f-0ab1-476c-b73b-6bf6c4ff7699

### Open external

[feature_save_open_file](https://docs.google.com/spreadsheets/d/1_5XijOqleQJSZzWthlH4Zkyw4TmKCiup8b69yglihcQ/edit#gid=960980632) template
Screenshot 2023-12-13 at 13 26 53

#### Web (Chrome, macOS)

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/89542bb0-6845-490f-be3e-4e1b396b002d

#### Android (emulator)

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/253c8061-787e-45b4-8652-88f356acdba3


src/app/shared/services/file-manager/file-manager.service.ts
```diff
@@ -1,34 +1,72 @@
...
+    const blob = (await this.templateAssetService.fetchAsset(relativePath, "blob")) as Blob;
+    try {
+      // On native devices, download file to native storage and trigger prompt to open with native apps
+      if (Capacitor.isNativePlatform()) {
+        const { localFilepath } = await this.saveFile(blob, relativePath);
+        FileOpener.open({ filePath: localFilepath, openWithDefault: false });
+      }
+      // On web, open the file in the browser, ready for download
+      else {
+        const fileUrl = window.URL.createObjectURL(blob);
```
chrismclarke, line 52: nit(non-blocking)
As noted in https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static#memory_management, when creating object urls usually you want some means of disposing of these to free up memory (particularly if creating large video/pdf objects)

Would it maybe be a simpler solution to just download in standard web way (allow browser to download from asset directory and user open from their downloads). There's various tools that can help with this, e.g.
https://www.npmjs.com/package/downloadjs
https://github.com/eligrey/FileSaver.js

There's also various manual ways to achieve similar functionality, think this article mentions a couple
https://blog.logrocket.com/programmatically-downloading-files-browser/

packages/data-models/flowTypes.ts
```diff
@@ -380,6 +380,7 @@ export namespace FlowTypes {
     "audio_end",
     "audio_play",
     "changed",
+    "download",
```
chrismclarke, line 383: nit(blocking) given that the _download_ functionality doesn't necessarily download (e.g. convert to temp object url and open), then would it be worth using a combination of `download` and `open` or `file_open` or `open_external` to make clearer what subset of behaviour expected?

src/app/shared/services/share/share.service.ts
```diff
@@ -33,11 +54,59 @@ export class ShareService extends SyncServiceBase {
...
+            const blob = (await this.templateAssetService.fetchAsset(relativePath, "blob")) as Blob;
+            // @capacitor/share can only share files saved to "Cache" directory
+            const { localFilepath } = await this.fileManagerService.saveFile(
+              blob,
+              relativePath,
+              "Cache"
+            );
+            if (localFilepath) {
+              const { activityType } = await Share.share({ url: localFilepath });
+              console.log("[SHARE] Content shared to", activityType);
```
chrismclarke, line 80: nit(non-blocking) should we be removing the temp file after share?

chrismclarke (changes requested):
Really nice feature to have, I think most of the implementation code is looking good (see a few minor comments inline), however I think we might want to tidy up a bit the naming related to downloads, and consider what functionality to extend in the future.

I would probably expect

### Download
**Web** - trigger the browser download, so that user gets the file populated to their local downloads folder and use their browser to handle opening in any external application (e.g. chrome opens pdfs in a new tab but xls files in excel)

**Native** - copy file to external download folder and notify user saved to downloads and give the option to open

In both cases I think the main use-case is so that people can still access the assets when app not running. 

### File Open
**Web** - probably just open asset link in new tab which should automatically handle downloading non-viewable files and showing image/video/pdf in tab

**Native** - use fileopener plugin to allow user select app to open file using

---
It might be a case that given the overlap between these two actions, combining into a single action, e.g. `open_external` might make sense given that technically all files are already downloaded on native

Open to suggestion though, as I appreciate the above is still a bit confusing

packages/data-models/flowTypes.ts
chrismclarke: I'd probably lean more towards keeping `share` `open` and `download` as standalone actions as I'm not sure grouping them together adds much value (particularly given you can currently share non-files, e.g. text, and in the future we may want to support opening non-files, such as programmatically opening a web url (using browser, instead of in-app navigation). 

Maybe we have top level `share`, `download` and `open` actions which then support some extra param syntax, e.g. `open : asset_path | external:true`

android/app/src/main/AndroidManifest.xml
```diff
@@ -8,7 +8,8 @@
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
-        android:theme="@style/AppTheme">
+        android:theme="@style/AppTheme"
+        android:usesCleartextTraffic="true">
```
jfmcquade, line 12: This is [apparently required](https://github.com/diachedelic/capacitor-blob-writer?tab=readme-ov-file#android) by `capacitor-blob-writer`, although I hadn't come across any issues while testing.

esmeetewinkel (approved):
This is great, thanks! Functional test passed. 

I confirmed that when using `save_to_device` on Android to save one file multiple times, the file only appears once - nice one.

jfmcquade:
I've done some more research into the new permissions needed to save files to the `Download` folder (triggered by the `save_to_device` action), and I think it may be worth excluding that functionality for the imminent release.

The [capacitor filesystem docs](https://capacitorjs.com/docs/apis/filesystem#android) state that in order to `Directory.ExternalStorage`, in Android 10 and older, the permissions `READ_EXTERNAL_STORAGE` and `WRITE_EXTERNAL_STORAGE` are required. These are considered "dangerous" ([docs here](https://developer.android.com/reference/android/Manifest.permission#READ_EXTERNAL_STORAGE)) and so require additional checks. Namely, according to [these docs](https://support.google.com/googleplay/android-developer/answer/9888170#:~:text=All%20files%20access%20permission), they must:
> successfully pass an appropriate access review prior to publishing. Apps allowed to use this permission must clearly prompt users to enable 'All files access' for their app under 'Special app access' settings.
They must also [complete a permissions declaration form](https://support.google.com/googleplay/android-developer/answer/9214102?visit_id=638382392344002420-2235604704&rd=1#zippy=%2Cstep-evaluate-requested-permissions).

[Devs report](https://stackoverflow.com/questions/76999561/how-to-save-file-to-downloads-folder-on-android-33) that adding these permissions can mean that the update is rejected by Google because this is not core functionality of the app (which seems to apply in our case).

So whilst the functionality of saving to the Downloads folder should work fine on devices running Android 11 and higher (see the first comment on [this question](https://stackoverflow.com/questions/76803442/android-what-permission-s-is-need-to-download-a-file-to-downloads-directory)), for earlier versions of Android, we would need those permissions which comes with significatn complications. 

I propose removing this functionality for now and removing those requested permissions, and at least temporarily setting `save_to_device` to save the file to the app's ring-fenced storage rather than the general "Downloads" folder.

@chrismclarke, let me know if you have any experience of similar permissions issues or general advice.

jfmcquade:
UPDATE: The above should be fixed by [this latest commit](https://github.com/IDEMSInternational/parenting-app-ui/pull/2160/commits/bb663e4c2fa3dff3d420f04300c61c12e8fb6c38). The permissions are no longer required if we save the file to the device's `Documents` folder, as opposed to `EXTERNAL_STORAGE/Download` folder. I think this is because saving in the former location does in fact save to a location ring-fenced for the app, but files saved there do show up in the user's general Documents folder, where no equivalent exists for the Downloads folder.

See new demonstration video:

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/b98e2597-466e-40c0-8ab8-ec10e2ad135a

android/app/src/main/AndroidManifest.xml
chrismclarke: I think it can also be defined by the `capacitor.config.ts` `server.cleartext` property, I have come across this issue myself before so makes sense to pre-emptively enable (capacitor says also required for live-reload from android api 28)
https://capacitorjs.com/docs/config

chrismclarke (approved):
Code all looks really good to me, I like the added functionality that save_to_device also opens (removing the need to chain both methods back-to-back)

Happy to merge once functional testing passed

src/app/shared/services/file-manager/file-manager.service.ts
```diff
@@ -1,43 +1,146 @@
...
+    subdirectory?: string;
+  }) {
+    const { data, targetPath, directory = "Data", subdirectory = "" } = options;
+    const path = (subdirectory ? subdirectory + "/" : "") + `${this.cacheName}/${targetPath}`;
     // Docs for write_blob are here: https://github.com/diachedelic/capacitor-blob-writer#readme
-    const src = await write_blob({
-      directory: Directory.Data,
-      path: `${this.cacheName}/${relativePath}`,
-      blob,
+    const localFilepath = await write_blob({
```
chrismclarke, line 129: nit(non-blocking)   
I can see in the blob-writer docs that duplicate named files will be overwritten. Is this a method we should consider exposing to authors in the future along with some UI popup/confirm/prompt if desirable? e.g. `handle_overwrite : 'overwrite' (default) / 'prompt'`

jfmcquade:
> @jfmcquade when I'm testing "save" now on Appetize I can't find the file in the file explorer:

I *think* this is just a quirk of the appetize viewer. You can find the file if you through the storage location named `sdk_gphone***`, where it appears inside the enclosed `Documents` folder, but not the top-level one in the sidebar. See screen capture below. Although I've just been testing a few times in a row, and I'm sure it did appear in the `Recent` folder during one test (after first finding it through `sdk_gphone*** > Documents > etc.`), so it appears to be temperamental.

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/8c1b843d-226f-4c61-9cee-37c3c9c1b51c

Anyway, I think this issue is only present when running in an emulator, due to the storage itself being emulated in some sense. @chrismclarke (sorry to keep tagging you...), do you have any insight?

PR #2161, opened 2023-12-11
Added actions to be called by content repositories

ChrisMarsh82:
## Description

Added actions to be called from content repositories. All actions must inherit secrets from repos that run the actions.

### DIFF
--- a/.github/workflows/contentflows/reusable-android-build.yml
+++ b/.github/workflows/contentflows/reusable-android-build.yml
@@ -1,0 +1,190 @@
+##################################################################################
+#         About
+##################################################################################
+# Reuseable workflow to be called from content repos. 
+# Build and deploy app bundle for android
+#
+#         Version : 1.0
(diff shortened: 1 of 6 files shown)


ChrisMarsh82:
Merging without review in order to test actions

chrismclarke:
I think you can probably just add a [branch push trigger](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore) to test actions on push to the branch under development instead of merging into master. 

Generally preferred as master deployment has a couple of side-effects associated with it, such as redeploying the hosting sites, and allows for code review

chrismclarke:
Also, just to check - is this so that deployment repos don't have to store their own github actions? Makes a lot of sense, but I'm wondering if this means we should also be removing the current templating system we have to add actions to a repo?

Issue #2136, opened 2023-11-07
[FEATURE] Deployment repo android build improvements


chrismclarke:
**What?**
When creating android releases on deployment branch main app repo is checked out during action, and therefore has the default `capacitor.config.ts`. It would be useful to allow authors to specify things like app_id and other config overrides in the deployment config.ts, and include overwrite as part of android build workflows

Additionally the android versioning script works from the main package.json, but would probably make more sense to assign based on content version (or combined content/builder)

Ideally these updates can then be incorporated into a new template in `packages/actions/templates`

It would be worth discussing any other requirements for content repos or how best to prioritise/achieve these items

**Other details**
Workarounds for issue above in deployment repo:
https://github.com/IDEMSInternational/plh-facilitator-app-my-content/blob/main/.github/workflows/android_build.yml

chrismclarke:
Just to comment that ideally whatever improvements made should ideally replace the custom string edits made to handle version codes in `.github/workflows/reusable-android-build.yml`, and deprecate the changes introduced #2171

PR #2170, opened 2023-12-15
chore: increase version number to 0.16.24

esmeetewinkel:
PR Checklist

## Description

PR increase version number to 0.16.24

### DIFF
--- a/.github/workflows/reusable-android-build.yml
+++ b/.github/workflows/reusable-android-build.yml
@@ -129,7 +129,7 @@ jobs:
             echo "Version Code: $VERSION_CODE"
             echo "Version: $VERSION"
             # This will need to change currently looking for specific version and not place holder
-            sed -i -e "s/16022/$VERSION_CODE/g" -e "s/0.16.22/$VERSION/g" ./android/app/build.gradle 
+            sed -i -e "s/16024/$VERSION_CODE/g" -e "s/0.16.24/$VERSION/g" ./android/app/build.gradle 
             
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable-android-build.yml
```diff
@@ -129,7 +129,7 @@ jobs:
             echo "Version Code: $VERSION_CODE"
             echo "Version: $VERSION"
             # This will need to change currently looking for specific version and not place holder
-            sed -i -e "s/16022/$VERSION_CODE/g" -e "s/0.16.22/$VERSION/g" ./android/app/build.gradle
```
esmeetewinkel, line 132: @jfmcquade Should this be added to the action `yarn scripts version`? This is what I normally use to bump the version but it only updates the other two places...

jfmcquade (approved):
I've also bumped the version number in the github workflow `reusable-android-build.yml`, which is used by content repos. Currently it's still looking for a specific version number (now increased to match the othee files) to replace with the target version number specified by the repo that is invoking the action.

.github/workflows/reusable-android-build.yml
jfmcquade: Good point. Really we want to action #2136, but this is probably a good idea in the short term, I'll have a look

chrismclarke: If we're updating the android build number for content repos shouldn't we be using a variable coming from the content repo itself? E.g. the content repo version number...

PR #2171, opened 2023-12-15
hotfix: yarn scripts version also updates strings targeted by android build action

jfmcquade:
PR Checklist

## Description

As suggested in [this comment](https://github.com/IDEMSInternational/parenting-app-ui/pull/2170#discussion_r1428199826), adds an ugly workaround whereby running `yarn scripts version` updates the version values targeted for replacement by `.github/workflows/reusable-android-build.yml` to match the new values populated to `./android/app/build.gradle`.

As mentioned, this is a temporary workaround pending directly addressing #2136.

## Testing

With this PR branch checked out, run `yarn scripts version` with a sensible value, and ensure that the changes to all 3 files look correct

## Git Issues

Closes #

### DIFF
--- a/packages/scripts/src/commands/version.ts
+++ b/packages/scripts/src/commands/version.ts
@@ -1,7 +1,7 @@
 import * as fs from "fs-extra";
 import { Command } from "commander";
 import inquirer from "inquirer";
-import { APP_BUILD_GRADLE_PATH, MAIN_PACKAGE_PATH } from "../paths";
+import { APP_BUILD_GRADLE_PATH, MAIN_PACKAGE_PATH, ANDROID_BUILD_ACTION_PATH } from "../paths";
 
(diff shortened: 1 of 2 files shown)


chrismclarke (commented):
Think I'm somewhat out on the loop on this, didn't even realise we had this set of reusable actions/workflows for content repos (can see added over course of various prs without review and also a few commits direct to master which I'd strongly discourage against as a practice)

Given now we're updating scripts to fix what is already a temporary workaround I'd probably suggest better to just make a more-permanent fix to allow authors control over the capacitor config and android build process

esmeetewinkel (approved):
Functional test passed. Merging this as this makes my life strictly easier while awaiting improvements suggested in https://github.com/IDEMSInternational/parenting-app-ui/issues/2136

PR #2175, opened 2024-01-03
feat: text component 'caption' variant

jfmcquade:
PR Checklist

## Description

Adds a new variant for the text component: `caption`. Applying this variant to a text component makes it suitable to sit directly beneath an image component as the image's caption.

## Git Issues

This was requested as a feature for the AE App.

Closes #

## Screenshots/Videos

See example in [comp_text](https://docs.google.com/spreadsheets/d/1iZHeAAZAQ8q-XNKrW_ww3yvPUVGwc407uhAF7vu1TtM/edit#gid=569531329) template.

Screenshot 2024-01-04 at 16 05 06

Screenshot 2024-01-04 at 16 06 14


src/app/shared/components/template/components/text/text.component.scss
```diff
@@ -0,0 +1,7 @@
+[data-variant~="caption"] {
+  font-size: var(--font-size-text-small);
+  text-align: center;
+  position: relative;
```
jfmcquade, line 4: This seemed more straightforward than trying to override the multiple levels of margin applied to the nest of elements.

chrismclarke (changes requested):
I'm not sure if the implementation fully makes sense to me as it appears just to be formatted text with an offset margin (and not necessarily related to an image). This might be a bit fragile to changes in surrounding rows (e.g. if image is removed or different default margins applied)

Would it make more sense to implement as part of the image component instead, so that the user can provide a `caption` parameter?

In doing so you could use the standard web caption element, which as a bonus would have improved accessibility and screen-reader support. It would also be tied directly to an image and not be impacted by more general row layout changes https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption
https://web.dev/learn/accessibility/images

PR #2176, opened 2024-01-05
Feat: standalone task progress bar component

### LINKED ISSUE #2150: [FEATURE]: Task group progress bar
**What?**
New "progress bar" navigation for a task group ("category") that contains of multiple tasks ("articles", corresponding to a single screen). N.B. In the PLH Kids use case we have three layers of task nesting: module - category - article

image

1. progress bar appears if a task group contains more than one task. Progress bar is 1/n filled on the first task, 2/n on the second task, and filled up entirely on the last task. 
2. "cross" appears in the top right corner and exits the current task group (taking the user back to the module landing page)
3. "next" button at the bottom of the screen takes user to next task within the task group. If at the last task of the task group, this will take the user to the first task of the following task group. (unclear whether this should be the highest priority task group or the highest priority uncompleted task group). If at the last task of the last task group of the module, should take back to module landing page.
4. "back" button appears only when a task group has multiple tasks, on all but the first task.

**Why?**
Required for PLH Kids deployment but would apply to FacilitatorApp and PLH Teens deployments in future too. 

**How?**
3. Moving from "completing one category" to the next will be using a function authored in the code to evaluate which task is next (as discussed with @jfmcquade)

jfmcquade:
PR Checklist

## Description

Fleshes out the standalone version (as opposed to being instantiated by the `task-card` component) of the `task-progress-bar` component. The component is passed the following params:

`task_group_data`: The name of the task group for which to track the completion of its subtasks
`completed_field`: The name of the field that stores the completion status of the task group
`progress_units_name`: The displayed name of the units of progress associated with the constituent tasks of the task group. Default "sections".
`show_text`: Show the text associated with progress bar, e.g. "8 sections". Default false
`variant`: A list of named style variants of the component, separated by spaces or commas. Supported values: "green" | "secondary".
`dynamic`: Specify whether to use the dynamic data feature. Should be true if the task_group_data has a "completed" column, else should be false if the task_group_data has a "completed_field" column.

Task group data lists can store their completion status in one of two ways:
1. A "completed_field" column, referencing a field that stores the completion status
2. A "completed" column, with values updated dynamically

Support for 2. is a work in progress. See below.

### Case 1: static data list, completion status stored in fields

[comp_task_progress_bar](https://docs.google.com/spreadsheets/d/1KenOF4C7BjVmg07zONWQgRygVWMDaXCOGLOSfC5vx7g/edit#gid=603128459)

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/0911f489-fc07-4e9b-9988-0913a60cf434

### Case 2: dynamic data list, completion status stored in `completes` column

Dynamic 1

Dynamic 2

## Git Issues

Closes #2150

## Screenshots/Videos

### DIFF
--- a/packages/data-models/appConfig.ts
+++ b/packages/data-models/appConfig.ts
@@ -200,6 +200,14 @@ const ASSET_PACKS = {
   folderName: "asset_packs",
 };
 
+const TASKS = {
+  enabled: true,
+  // The name of the field that tracks the currently highlighted task group
+  highlightedTaskField: "_task_highlighted_group_id",
+  // The top-level list of task groups (e.g. modules)
+  taskGroupsListName: "workshop_tasks",
+};
+
 const APP_CONFIG = {
   APP_FIELDS,
   APP_HEADER_DEFAULTS,
@@ -220,6 +228,7 @@ const APP_CONFIG = {
   NOTIFICATIONS_SYNC_FREQUENCY_MS,
   NOTIFICATION_DEFAULTS,
   SERVER_SYNC_FREQUENCY_MS,
+  TASKS,
 };
 // Export as a clone to avoid risk one import could alter another
 export const getDefaultAppConfig = () => clone(APP_CONFIG);
--- a/src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.scss
+++ b/src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.scss
@@ -18,12 +18,14 @@
     border-radius: var(--ion-border-radius-small);
     background-color: var(--ion-color-gray-400);
   }
-  &.completed {
+  &.completed,
+  &[data-variant~="green"] {
     .progress-bar {
       background-color: var(--ion-color-green);
     }
   }
-  &.highlighted {
+  &.highlighted,
+  &[data-variant~="secondary"] {
     .progress-bar-background {
       background-color: var(--ion-color-secondary-100);
     }
--- a/src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
+++ b/src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
@@ -1,6 +1,9 @@
 import { Component, EventEmitter, Input, OnInit, Output } from "@angular/core";
 import { TaskService } from "src/app/shared/services/task/task.service";
-import { getStringParamFromTemplateRow } from "src/app/shared/utils";
+import {
+  getBooleanParamFromTemplateRow,
+  getStringParamFromTemplateRow,
+} from "src/app/shared/utils";
 import { TemplateBaseComponent } from "../base";
 import { TemplateFieldService } from "../../services/template-field.service";
 import { AppDataService } from "src/app/shared/services/data/app-data.service";
@@ -39,6 +42,8 @@ export class TmplTaskProgressBarComponent extends TemplateBaseComponent implemen
   }
 
   getParams() {
+    // If component is being instantiated by a parent component (e.g. task-card), use Input() values for properties.
+    // If component is being explicitly instantiated from a template, get the params from the template row
     if (this._row) {
       this.taskGroupDataList = getStringParamFromTemplateRow(this._row, "task_group_data", null);
       this.taskGroupCompletedField = getStringParamFromTemplateRow(
@@ -51,6 +56,7 @@ export class TmplTaskProgressBarComponent extends TemplateBaseComponent implemen
         "progress_units_name",
         "sections"
       );
+      this.showText = getBooleanParamFromTemplateRow(this._row, "show_text", false);
     }
   }
 
--- a/src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
+++ b/src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
@@ -10,6 +10,26 @@ import { AppDataService } from "src/app/shared/services/data/app-data.service";
 import { IProgressStatus } from "src/app/shared/services/task/task.service";
 import { CampaignService } from "src/app/feature/campaign/campaign.service";
 
+interface ITaskProgressBarParams {
+  /** The name of the task group for which to track the completion of its subtasks */
+  taskGroupDataList: string;
+  /** The name of the field that stores the completion status of the task group */
+  taskGroupCompletedField: string;
+  /**
+   * The displayed name of the units of progress associated with
+   * the constituent tasks of the task group.
+   * Default "sections"
+   */
+  progressUnitsName: string;
+  /**
+   * Show the text associated with progress bar, e.g. "8 sections"
+   * Default true
+   */
+  showText: boolean;
+  /** A list of named style variants of the component, separated by spaces or commas. */
+  variant: "green" | "secondary";
+}
+
 @Component({
   selector: "plh-task-progress-bar",
   templateUrl: "./task-progress-bar.component.html",
@@ -21,11 +41,12 @@ export class TmplTaskProgressBarComponent extends TemplateBaseComponent implemen
   @Input() highlighted: boolean | null;
   @Input() progressStatus: IProgressStatus;
   @Input() progressUnitsName: string;
+  @Input() showText: boolean;
   @Output() progressStatusChange = new EventEmitter<IProgressStatus>();
   @Output() newlyCompleted = new EventEmitter<boolean>();
+  params: Partial<ITaskProgressBarParams> = {};
   subtasksTotal: number;
   subtasksCompleted: number;
-  showText = true;
 
   constructor(
     private taskService: TaskService,
@@ -42,21 +63,34 @@ export class TmplTaskProgressBarComponent extends TemplateBaseComponent implemen
   }
 
   getParams() {
-    // If component is being instantiated by a parent component (e.g. task-card), use Input() values for properties.
     // If component is being explicitly instantiated from a template, get the params from the template row
     if (this._row) {
-      this.taskGroupDataList = getStringParamFromTemplateRow(this._row, "task_group_data", null);
-      this.taskGroupCompletedField = getStringParamFromTemplateRow(
+      this.params.taskGroupDataList = getStringParamFromTemplateRow(
+        this._row,
+        "task_group_data",
+        null
+      );
+      this.params.taskGroupCompletedField = getStringParamFromTemplateRow(
         this._row,
         "completed_field",
         null
       );
-      this.progressUnitsName = getStringParamFromTemplateRow(
+      this.params.progressUnitsName = getStringParamFromTemplateRow(
         this._row,
         "progress_units_name",
         "sections"
       );
-      this.showText = getBooleanParamFromTemplateRow(this._row, "show_text", false);
+      this.params.showText = getBooleanParamFromTemplateRow(this._row, "show_text", false);
+      this.params.variant = getStringParamFromTemplateRow(this._row, "variant", "")
+        .split(",")
+        .join(" ") as ITaskProgressBarParams["variant"];
+    }
+    // If component is being instantiated by a parent component (e.g. task-card), use Input() values for params.
+    else {
+      this.params.taskGroupDataList = this.taskGroupDataList;
+      this.params.taskGroupCompletedField = this.taskGroupCompletedField;
+      this.params.progressUnitsName = this.progressUnitsName;
+      this.params.showText = this.showText;
     }
   }
 
@@ -65,7 +99,7 @@ export class TmplTaskProgressBarComponent extends TemplateBaseComponent implemen
   }
 
   async evaluateTaskGroupData() {
-    const dataList = await this.appDataService.getSheet("data_list", this.taskGroupDataList);
+    const dataList = await this.appDataService.getSheet("data_list", this.params.taskGroupDataList);
     const subtasks = dataList?.rows || [];
     this.subtasksTotal = subtasks.length;
     this.subtasksCompleted = subtasks.filter((task) =>
@@ -75,13 +109,13 @@ export class TmplTaskProgressBarComponent extends TemplateBaseComponent implemen
       this.progressStatus = "completed";
       this.progressStatusChange.emit(this.progressStatus);
       // Check whether task group has already been completed
-      if (this.templateFieldService.getField(this.taskGroupCompletedField) !== true) {
+      if (this.templateFieldService.getField(this.params.taskGroupCompletedField) !== true) {
         // If not, set completed field to "true"
-        await this.setTaskGroupCompletedStatus(this.taskGroupCompletedField, true);
+        await this.setTaskGroupCompletedStatus(this.params.taskGroupCompletedField, true);
         this.newlyCompleted.emit(true);
       }
     } else {
-      await this.setTaskGroupCompletedStatus(this.taskGroupCompletedField, false);
+      await this.setTaskGroupCompletedStatus(this.params.taskGroupCompletedField, false);
       if (this.subtasksCompleted) {
         this.progressStatus = "inProgress";
         this.progressStatusChange.emit(this.progressStatus);
--- a/src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
+++ b/src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
@@ -1,4 +1,12 @@
-import { Component, EventEmitter, Input, OnInit, Output } from "@angular/core";
+import {
+  ChangeDetectionStrategy,
+  ChangeDetectorRef,
+  Component,
+  EventEmitter,
+  Input,
+  OnInit,
+  Output,
+} from "@angular/core";
 import { TaskService } from "src/app/shared/services/task/task.service";
 import {
   getBooleanParamFromTemplateRow,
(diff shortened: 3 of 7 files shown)


src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
```diff
@@ -88,10 +182,28 @@ export class TmplTaskProgressBarComponent extends TemplateBaseComponent implemen
...
+  async setTaskGroupCompletedStatus(isCompleted: boolean) {
+    console.log(`Setting ${this.params.taskGroupCompletedField} to ${isCompleted}`);
+    await this.templateFieldService.setField(this.params.taskGroupCompletedField, `${isCompleted}`);
   }
 
-  async setTaskGroupCompletedStatus(taskGroupCompletedField: string, isCompleted: boolean) {
-    console.log(`Setting ${taskGroupCompletedField} to ${isCompleted}`);
-    await this.templateFieldService.setField(taskGroupCompletedField, `${isCompleted}`);
+  // Adapted from data-items component
+  private async subscribeToData() {
```
jfmcquade, line 194: @chrismclarke Should we also be unsubscribing inside an `ngOnDestroy()` for this component (and for the `data-items` component)? If not, why not?

packages/data-models/appConfig.ts
```diff
@@ -200,6 +200,14 @@ const ASSET_PACKS = {
   folderName: "asset_packs",
 };
 
+const TASKS = {
+  enabled: true,
+  // The name of the field that tracks the currently highlighted task group
+  highlightedTaskField: "_task_highlighted_group_id",
```
chrismclarke, line 206: nit(non-blocking)
Given the properties already sit under `Tasks` can probably get rid of the work `task` from `highlightedTaskField` and `taskGroupsListName` property. Perhaps something like `activeGroupField`

packages/data-models/appConfig.ts
```diff
@@ -200,6 +200,14 @@ const ASSET_PACKS = {
...
+  // The top-level list of task groups (e.g. modules)
+  taskGroupsListName: "workshop_tasks",
```
chrismclarke, line 208: nit(question)
I'm also not 100% sure what role the task groupList name plays - is this something that needs to be hardcoded into the app or something that should be passed to a `task_group` component (can it only ever be one fixed value so one task list per app??). If it is a default perhaps just `defaultGroupName` ?

src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.scss
```diff
@@ -18,12 +18,14 @@
     border-radius: var(--ion-border-radius-small);
     background-color: var(--ion-color-gray-400);
   }
-  &.completed {
+  &.completed,
+  &[data-variant~="green"] {
```
chrismclarke, line 22: nit(non-blocking)
Green as a success color is quite culturally-specific. Instead of using a hardcoded color would it make more sense to expose a theme variable that can override? E.g. `--task-progress-bar-color` which defaults to using the app primary color. Alternatively if really wanting to override as a one-off might be better to allow user to specify via parameter

src/app/shared/services/task/task.service.ts
```diff
@@ -3,34 +3,50 @@ import { TemplateFieldService } from "../../components/template/services/templat
...
+    this.subscribeToAppConfigChanges();
+    if (this.tasksFeatureEnabled) {
+      await this.getListOfTaskGroups();
+      if (this.taskGroups.length > 0) {
+        this.evaluateHighlightedTaskGroup();
+      }
     }
   }
 
+  subscribeToAppConfigChanges() {
```
chrismclarke, line 42: nit(non-blocking) does this need to be a public method? if not better to denote as private and move below public methods

src/app/shared/services/task/task.service.ts
```diff
@@ -3,34 +3,50 @@ import { TemplateFieldService } from "../../components/template/services/templat
 import { AppDataService } from "../data/app-data.service";
 import { arrayToHashmap } from "../../utils";
 import { AsyncServiceBase } from "../asyncService.base";
+import { AppConfigService } from "../app-config/app-config.service";
+import { IAppConfig } from "../../model";
 
 @Injectable({
   providedIn: "root",
 })
 export class TaskService extends AsyncServiceBase {
```
chrismclarke, line 12: nit(blocking)
It would be useful to have a few basic tests for public methods. I appreciate there's quite a lot exposed via config, but at least having some basic mock data should help understand the main ideas (as it stands very hard to evaluate or review the logic for setting/getting tasks)

src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
```diff
@@ -1,81 +1,175 @@
...
 import { CampaignService } from "src/app/feature/campaign/campaign.service";
+import { Subscription, debounceTime } from "rxjs";
+import { DynamicDataService } from "src/app/shared/services/dynamic-data/dynamic-data.service";
+
+interface ITaskProgressBarParams {
+  /**
+   * TEMPLATE PARAMETER: task_group_data.
+   * The name of the task group to track the progress of (through the completion of its subtasks)
+   */
+  taskGroupDataList: string;
```
chrismclarke, line 28: nit(non-blocking) - as mentioned previously ,can probably omit the word `task` and would also be good to specify it's the data list `name` (easily confused with entire list). So would suggest `dataListName` or `groupDataListName` (not sure how much group is implied or not)

src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
```diff
@@ -1,81 +1,175 @@
...
+  /**
+   * TEMPLATE PARAMETER: completed_field.
+   * The name of the field that stores the completion status of the task group
+   */
+  taskGroupCompletedField: string;
```
chrismclarke, line 33: nit(non-blocking) - same as above, maybe just `completedField`

src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
```diff
@@ -1,81 +1,175 @@
...
+   * TEMPLATE PARAMETER: show_text.
+   * Show the text associated with progress bar, e.g. "8 sections"
+   * Default false
+   */
+  showText: boolean;
+  /**
+   * TEMPLATE PARAMETER: variant.
+   * A list of named style variants of the component, separated by spaces or commas.
+   * */
+  variant: "green" | "secondary";
```
chrismclarke, line 51: nit(blocking) - mentioned in the scss but now looking at the options laid out really doesn't quite feel right as a list of options - so would say more blocking.

src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
```diff
@@ -1,81 +1,175 @@
...
+  /**
+   * TEMPLATE PARAMETER: dynamic.
+   * Specify whether to use the dynamic data feature:
+   * Should be true if the task_group_data has a "completed" column,
+   * else should be false if the task_group_data has a "completed_field" column
+   * */
+  useDynamicData: boolean;
```
chrismclarke, line 58: nit(non-blocking) - would it be confusing for authors to interpret what `dynamic` should be? I'm guessing this should actually be the default behaviour if available. Would it be better here for authors to just pass a `completed_column_name` variable (which defaults to `'completed'`) and look to deprecate the `completed_field` syntax?

Alternatively we could add a means to subscribe to field value changes to keep dynamic behaviour

src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
```diff
@@ -1,81 +1,175 @@
...
 
   get progressPercentage() {
     return (this.subtasksCompleted / this.subtasksTotal) * 100;
   }
 
-  async evaluateTaskGroupData() {
-    const dataList = await this.appDataService.getSheet("data_list", this.taskGroupDataList);
-    const subtasks = dataList?.rows || [];
+  async getTaskGroupDataRows() {
+    const dataList = await this.appDataService.getSheet("data_list", this.params.taskGroupDataList);
```
chrismclarke, line 140: nit(non-blocking) - interop with other services should probably be done from the `taskService` itself and not component level, particularly so we can write tests for the behavior (although less signficant for smaller functions)

src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
```diff
@@ -1,81 +1,175 @@
...
+
+  /**
+   * Fetch the completion status for each subtask of the task group
+   * and calculate the task group's completion percentage to display.
+   * Task group data lists can store their completion status in one of two ways:
+   * 1. A "completed_field" column, referencing a field that stores the completion status
+   * 2. A "completed" column, with values updated dynamically
+   * @param dataRows The data rows of the task group
+   */
+  async evaluateTaskGroupData(dataRows: any[]) {
```
chrismclarke, line 152: nit(blocking) - this method in particular would probably be better if possible to split heavy lifting to task service with tests added

src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
chrismclarke: Subscriptions should always be unsubscribed. I struggle to think of any case where this is not preferable (maybe some query likely to be repeatedly reused across the app, but even then probably better to just recreate). So if not present in data-items component would be good to add

chrismclarke (changes requested):
Nice feature to have! I can't really go too much in depth on the actual task evaluation logic without some base tests to work from, so would suggest these be added either here or in an immediate follow-up.

I also haven't functionally tested, but have done my best to look through all the code and make a number of mostly code quality nit-picks/suggestions.

chrismclarke (changes requested):
Nice feature to have! I can't really go too much in depth on the actual task evaluation logic without some base tests to work from, so would suggest these be added either here or in an immediate follow-up.

I also haven't functionally tested, but have done my best to look through all the code and make a number of mostly code quality nit-picks/suggestions.

packages/data-models/appConfig.ts
jfmcquade: Good point, see reply in [this mattermost thread](https://chat.idems.international/all/pl/jps5afo4c7r69qsp8o74o37d5c)

src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
chrismclarke: @jfmcquade - I can't see any new commits, could you clarify how this has been resolved?

jfmcquade: Sorry, I was working through the comments and marking them as resolved as I made changes locally, but I haven't pushed my changes yet as I thought it would be better to wait until I'm ready to request another review. I thought better to hold off rather than give you lots of notifications before I've finished addressing all the point – but I'm guessing that you get notifications for me marking comments as resolved anyway? In future, I'll keep track of the comments locally and only mark as resolved on github when I've pushed the corresponding changes

src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
jfmcquade: Good point, I don't think `dynamic` would make any sense to authors. I've opted to replace with a param called `completed_column_name`, with default value "completed", and a check to see whether the named column is present in the source data list. If the column is not present, the component will look for a column named "completed_field" and will update non-dynamically

packages/data-models/appConfig.ts
jfmcquade: Discussed in [this WIP RFC](https://docs.google.com/document/d/1qVcUri2DAhyiFx6ihtFLeVeJjoS1ba0rdxntRg3AGec/edit).

jfmcquade:
Thanks @chrismclarke. I'm still yet to write unit tests, but I think I've addressed your other comments

chrismclarke:
Thanks @jfmcquade 
I'll wait until there are some tests in to do a second review as given the number of changes it will have to be a full re-review anyways, so be nice to be able to follow what's going on at the same time. Thanks for adding the RFC doc and example templates also, useful to have for sure and makes sense that we migrate more of the refactoring tasks there

packages/data-models/appConfig.ts
chrismclarke: Linked explicitly with comment:
https://docs.google.com/document/d/1qVcUri2DAhyiFx6ihtFLeVeJjoS1ba0rdxntRg3AGec/edit?disco=AAABC5ZxhSw

Will resolve here for now

packages/data-models/appConfig.ts
chrismclarke: Will also resolve here and link to:
https://docs.google.com/document/d/1qVcUri2DAhyiFx6ihtFLeVeJjoS1ba0rdxntRg3AGec/edit?disco=AAABC5ZxhSw

src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.scss
chrismclarke: Re-opening this comment - pr changes still shows the `data-variant~="green"` property but I can't see when checked out locally (??). I also see a hardcoded `--ion-color-green` locally - just want to check what you decided on this?

chrismclarke: Oh wait - it only shows the diff if viewing the outdated code (as comment line no longer exists on latest). So I think it has been addressed (although a little confusing) - could you just add a quick clarification comment?

esmeetewinkel (commented):
> show_text: Show the text associated with progress bar, e.g. "8 sections". Default false

Assuming that in practice this means it doesn't show i.e. default = blank

> config.app_config.TASKS.enabled = "false" (the default value is true)

I think it would make more sense for the default to be false, and a deployment to opt-in to using features rather than opt-out. In that way someone trying to make a simple app, with the simplest possible config, doesn't need to know what tasks are and make sure to disable them.

image
Getting the warning `[HIGHLIGHTED TASK GROUP] - No highlighted task group is set` when toggling the task completion on `comp_task_prog_bar_dynamic` template

_to be continued_

esmeetewinkel:
When testing feature_task_group, my _task_highlighted_group_id is set to blank whenever I toggle any task completion. 

image

jfmcquade:
@chrismclarke I am still having issues with running tests, we can discuss tomorrow if I haven't been able to resolve.

Thanks for the review, @esmeetewinkel, my response is below:

>> show_text: Show the text associated with progress bar, e.g. "8 sections". Default false
> Assuming that in practice this means it doesn't show i.e. default = blank

That is correct

>> config.app_config.TASKS.enabled = "false" (the default value is true)
>
> I think it would make more sense for the default to be false, and a deployment to opt-in to using features rather than opt-out. In that way someone trying to make a simple app, with the simplest possible config, doesn't need to know what tasks are and make sure to disable them.

I agree, I have update the PR with that change. NB, that now means that the PR contains breaking changes, I have updated the description accordingly.

> When testing feature_task_group, my _task_highlighted_group_id is set to blank whenever I toggle any task completion (potentially the same as issue highlighted above). I tried resetting the app / empty cache and hard reload / clear storage, but the issue persists

Ah, I should have predicted this. Because the debug templates are using a non-default task group for `config.app_config.TASKS.taskGroupsListName` (namely [feat_task_groups](https://docs.google.com/spreadsheets/d/1j4glnh-PLx5ZkQXBsdjjMOAo5gO_AM4pQbprTLsuSdo/edit#gid=443313797)), the feature requires an update to the deployment config. I've included those updates in [this PR](https://github.com/IDEMSInternational/app-debug-content/pull/47) on the debug repo. After merging, syncing, re-setting your deployment, and restarting your local app, you should find that the feature behaves as expected.

esmeetewinkel:
> I agree, I have update the PR with that change. NB, that now means that the PR contains breaking changes, I have updated the description accordingly.

1. I tested that indeed on a deployment without tasks (plh_facilitator_mx) the error "workshop_tasks not found" has disappeared.  
2. I also updated the configs of the two deployments that use tasks:
   i. [plh_teens](https://github.com/IDEMSInternational/plh-teens-app-content/pull/13): When running this deployments on this PR code branch (with or without deployment config update) I'm getting the error 
   image

   ii. [wash](https://github.com/IDEMSInternational/wash-app-content/pull/3)
   I'm getting the error below, but I'm guessing that
image

jfmcquade:
>       [comp_task_prog_bar_dynamic](https://docs.google.com/spreadsheets/d/1KenOF4C7BjVmg07zONWQgRygVWMDaXCOGLOSfC5vx7g/edit#gid=2015154474) seems to have stopped working at all (apart from the button click animation nothing happens when I press the buttons?)
It seems this was due to cell D9 of [comp_task_prog_bar_dynamic](https://docs.google.com/spreadsheets/d/1KenOF4C7BjVmg07zONWQgRygVWMDaXCOGLOSfC5vx7g/edit#gid=2015154474). You pointed out a discrepancy in the way I'd authored the `set_item` action, and I made a change to bring them into alignment. However, I opted for the wrong syntax: it turns the `set_item` action _must_ be authored with a pipe `|` after the action name, e.g. `click | set_item | completed:!@item.completed`. The template should be working now if you sync your debug sheets.

>    2. [plh_teens](https://github.com/IDEMSInternational/plh-teens-app-content/pull/13): When running this deployment on this PR code branch (with or without deployment config update) I'm getting the error
 
Good thing you tested that deployment. The issue was arising because the task progress bars in the task cards on the homepage were erroneously trying to use the dynamic data feature. I've fixed this in the latest commit.

esmeetewinkel (approved):

esmeetewinkel:
Merging this since feature is needed for content deployments now, @jfmcquade will work on tests as a follow-up PR

PR #2177, opened 2024-01-08
fix: dynamic data service multiple queries

chrismclarke:
PR Checklist

## Description
Fix issue where calling multiple dynamic-data service queries in parallel could lead to multiple db collection creation requests throwing rxdb error. Specifically:

- Implement a queue-type system to prevent duplicate db create requests when creating multiple queries in parallel
- Add test case

## Dev Notes
Additional work is included to refactor the app-data mocks 56a749a, to fix a regression issue brought in by changes to the app-data mocks in #2107. Data had been copy-pasted from the app-data mock to the dynamic-data mock (which is super fragile), breaking tests when source app-data mock changed. Now mock services can specify their own data when implementing the app-data mock. 

This is also another good reminder why we should try to  fix up the rest of the ng test suite and integrate into ci (raised in #2080 and various messages on mattermost). Not sure if this is something still on your radar @jfmcquade ?

One other regression discovered (coming from #1989) was the breaking of a test previously designed to throw error when updating row that does not exist. I can see the code changes removed this behaviour (I'm assuming you decided this is no longer a desirable QA check @jfmcquade (?)), so I've removed the test (see updated spec). If we don't get the ng test suite integrated then it would be good to at least make a mental note to check existing tests when making changes to corresponding services for both dev and review.

## Review Notes
Tests can be run via `yarn ng test --include src/app/shared/services/dynamic-data/dynamic-data.service.spec.ts`, also hoping will pass the case you raised in #2176 @jfmcquade (?)

## Git Issues

unblocks #2176

## Screenshots/Videos
**Before** - spec test for parallel requests added (failing)
image

**After** - all tests passing (with one qa test removed)
image

### DIFF
--- a/packages/shared/src/utils/async-utils.ts
+++ b/packages/shared/src/utils/async-utils.ts
@@ -1,0 +1,8 @@
+/** helper function used for dev to wait a fixed amount of time */
+export function _wait(ms: number) {
+  return new Promise<void>((resolve) => {
+    setTimeout(() => {
+      resolve();
+    }, ms);
+  });
+}
--- a/packages/shared/src/utils/cli-utils.ts
+++ b/packages/shared/src/utils/cli-utils.ts
@@ -39,11 +39,3 @@ export function pad(str: string | number, chars: number) {
   const padChars = Math.max(chars - str.length + 1, 0);
   return str + new Array(padChars).join(" ");
 }
-/** helper function used for dev to wait a fixed amount of time */
-export function _wait(ms: number) {
-  return new Promise<void>((resolve) => {
-    setTimeout(() => {
-      resolve();
-    }, ms);
-  });
-}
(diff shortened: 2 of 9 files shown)


jfmcquade (approved):
This is great, thanks. I can confirm that the associated tests are passing for me locally, and it does seem to fix the problem I was having on #2176.

I think the reason I removed the error thrown for trying to update a row that does not exist is that without it, the `DynamicDataService.update()` method functions as an upsert, which was desirable. Point taken about about checking existing tests when making changes to corresponding services.

Yes, fixing up ng-tests (#2080) is definitely still on my radar, but it hasn't been prioritised. Expanding the test coverage has been included in a document of "technical debt" items (with the term being used quite loosely) that we'll hopefully be explicitly contracted to complete. But I think it's still unlikely to be prioritised in the next few months.

PR #2187, opened 2024-01-30
fix: js-evaluator escape characters

### LINKED ISSUE #2184: [BUG] Data pipe operation doesn't work on data list with \n
**Describe the bug**
The data pipe filter operation does not work on a data list that contains `\n` (even if this is in a column that has nothing to do with the filter condition). 

**App version**
v0.16.23 

**To Reproduce**
Release the debug sheet, sync debug content and observe errors in terminal as in screenshot below. 

**Sheet Name**
[article_filter](https://docs.google.com/spreadsheets/d/1ElqsKnRT3nYAJZ0C_in8swQNgZvVTnzedCiAMycw9Og/edit#gid=154244041) data_pipe (draft) in plh_facilitator_mx deployment

**Debug Template**
[debug_data_list and debug_data_pipe](https://docs.google.com/spreadsheets/d/1vtr341TIg0MDq99syARtSK6VmvI3rFjEdYfwgwdb1ek/edit#gid=1595229908) (draft)

**Screenshots and videos**
image

chrismclarke:
PR Checklist

## Description

Fix JSEvaluator to handle javascript escape characters (`\n`, `\r`). These would throw an error if present in text when evaluating, now they should be correctly escaped and still be present in final output.

Add tests to JSEvaluator and affected `filter` data_pipe processes

## Git Issues

Closes #2184

## Screenshots/Videos

### DIFF
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/data_pipe.parser.spec.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/data_pipe.parser.spec.ts
@@ -96,4 +96,33 @@ describe("data_pipe Parser", () => {
     });
     expect(deferred).toEqual(["data_pipe.test_pipe_defer"]);
   });
+
+  // QA - https://github.com/IDEMSInternational/parenting-app-ui/issues/2184
+  // NOTE - test case more explicitly handled by jsEvaluator.spec
+  it("Supports text with line break characters", async () => {
+    const parser = new DataPipeParser({
+      processedFlowHashmap: {
+        data_list: {
+          test_data_list: [{ id: 1, text: "normal" }, { id: 2, text: "line\nbreak" }, { id: 3 }],
+        },
+      },
+    } as any);
+    const ops: IDataPipeOperation[] = [
+      {
+        input_source: "test_data_list",
+        operation: "filter",
+        args_list: "id < 3" as any, // will be parsed during process
+        output_target: "test_output",
+      },
+    ];
+    const output = parser.run({
+      flow_name: "test_line_breaks",
+      flow_type: "data_pipe",
+      rows: ops,
+    });
+    expect(output._generated.data_list.test_output.rows).toEqual([
+      { id: 1, text: "normal" },
+      { id: 2, text: "line\nbreak" },
+    ]);
+  });
 });
--- a/packages/shared/src/models/jsEvaluator/jsEvaluator.spec.ts
+++ b/packages/shared/src/models/jsEvaluator/jsEvaluator.spec.ts
@@ -1,16 +1,21 @@
 import { JSEvaluator } from "./jsEvaluator";
 
+const constants = {
+  a: 1,
+  b: 2,
+  nestedConstant: { array: [1] },
+};
+const functions = {
+  isEven: (n) => n % 2 === 0,
+};
+
 describe("JS Evaluator", () => {
-  const constants = {
-    a: 1,
-    b: 2,
-    nestedConstant: { array: [1] },
-  };
-  const functions = {
-    isEven: (n) => n % 2 === 0,
-  };
-  const evaluator = new JSEvaluator();
-  evaluator.setGlobalContext({ constants, functions });
+  let evaluator: JSEvaluator;
+  beforeEach(() => {
+    evaluator = new JSEvaluator();
+    evaluator.setGlobalContext({ constants, functions });
+  });
+
   it("expression: Math.min(5,7)", () => {
     expect(evaluator.evaluate("Math.min(5,7)")).toEqual(5);
   });
@@ -34,5 +39,14 @@ describe("JS Evaluator", () => {
     const invalidConstants = { default: "hello", new: "test" };
     evaluator.setGlobalContext({ constants: { ...invalidConstants, ...constants } });
     expect(() => evaluator.evaluate("Math.min(a,b)")).toThrowError("Unexpected token 'default'");
+  });
+  it("handles escape characters", () => {
+    // Case 1 - evaluation string with linebreak
+    expect(evaluator.evaluate("'Hello\n'+this.name", { name: "Ada" })).toEqual("Hello\nAda");
+    // Case 2 - this context with linebreak
+    expect(evaluator.evaluate("'Hello'+this.name", { name: "\nAda" })).toEqual("Hello\nAda");
+    // Case 3 - global constant with linebreak
+    evaluator.setGlobalContext({ constants: { name: "\nAda" } });
+    expect(evaluator.evaluate("'Hello'+name", { name: "\nAda" })).toEqual("Hello\nAda");
   });
 });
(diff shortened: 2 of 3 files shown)


chrismclarke:
@jfmcquade - could you look into the failing tests? looks to be related to debug repo config changes as mentioned on the call

jfmcquade:
> @jfmcquade - could you look into the failing tests? looks to be related to debug repo config changes as mentioned on the call

That was indeed the issue, fixed as of [#49](https://github.com/IDEMSInternational/app-debug-content/pull/49) on the code repo

jfmcquade (approved):
The code looks good and I can confirm the tests are passing for me, both for the data pipe (`yarn workspace scripts test`) and the JS evaluator (`yarn workspace shared test`). Good to have the 3 different cases covered in the JS evaluator tests.

esmeetewinkel:
Thanks both. This indeed resolves the error with the line breaks, functional test passed on the `debug` deployment.

The [sheet ](https://docs.google.com/spreadsheets/d/1ElqsKnRT3nYAJZ0C_in8swQNgZvVTnzedCiAMycw9Og/edit#gid=1155509561) on the facilitator app deployment linked initially is still giving me head aches, it's now spitting out other errors `error: SyntaxError: Unexpected identifier 's'`, `error: SyntaxError: Unexpected identifier 'd'`, `error: SyntaxError: Unexpected identifier 're'`

image

List of row IDs, with problematic IDs **bold**:
what_is_safeguarding
during_disclosed_abuse
after_disclosed_abuse
how_to_report_abuse
contact_for_reporting_abuse
crisis_hotlines
**day_1_introduction
day_3_check_in
day_3_mh_stress_self_talk
day_5_goodbye**
day_4_check_in
**day_6_final_chat_sessions_5ux**
faq_h_after_group_ended
**faq_text_support
faq_ground_rules
faq_misbehaviour
faq_small_group_active**
faq_h_still_struggling
faq_h_sad_to_end
**faq_no_participation_general**

I can't determine what the broken / functioning rows have in common, but I'm guessing this is a similar JS thing? Can open a follow-up issue if it makes more sense to merge this PR first.

esmeetewinkel (approved):
Functional test passed

jfmcquade:
> Filtering the [sheet ](https://docs.google.com/spreadsheets/d/1ElqsKnRT3nYAJZ0C_in8swQNgZvVTnzedCiAMycw9Og/edit#gid=1155509561) on the facilitator app deployment linked initially is still giving me head aches, it's now spitting out other errors `error: SyntaxError: Unexpected identifier 's'`, `error: SyntaxError: Unexpected identifier 'd'`, `error: SyntaxError: Unexpected identifier 're'`

I just had a quick look and wasn't able to replicate either on this feature branch or the master branch. I was setting `aritcle_filter` to `released` [here](https://docs.google.com/spreadsheets/d/1ElqsKnRT3nYAJZ0C_in8swQNgZvVTnzedCiAMycw9Og/edit#gid=520802044) and running `yarn workflow sync`, is that right?

esmeetewinkel:
> I just had a quick look and wasn't able to replicate either on this feature branch or the master branch. I was setting `aritcle_filter` to `released` [here](https://docs.google.com/spreadsheets/d/1ElqsKnRT3nYAJZ0C_in8swQNgZvVTnzedCiAMycw9Og/edit#gid=520802044) and running `yarn workflow sync`, is that right?

Try again now - `article_filter` was set to filter my testing copy of `article_data` (`article_test_data) which I'd removed again. I've now changed it to refer to `article_data`.

jfmcquade:
Thanks. I think the issue is still with newlines, but I'm not quite sure what is causing it. If I log the value of `cleanedFuncString` from `jsEvaluator.ts` to the console then it appears that some of the content within a single cell gets correctly logged with visible "\n" characters, but sometimes the newline gets interpreted and actually displays as a newline. I can't work out yet what is causing the discrepancy. Noting here so I can continue looking tomorrow. I've tried tweaking the regexes, but the newlines that do get interpreted seem to be being picked up correctly.

Screenshot 2024-01-30 at 17 22 54

chrismclarke:
@esmeetewinkel @jfmcquade 
I think there's 2 potential further issues here. 

The first is additional characters that need escaping - I'm not exactly sure the full limits of what can/can't be used (all values essentially get combined to form one very long string of javascript code and executed as a function, and I'm not 100% sure the full limitations of working in this way). 

One issue I could identify however is single quotation marks `'` breaking because of the way the long string of javascript code uses single quotation marks around any string values detected from a row. So I've pushed 0ee2ce4 which should hopefully fix that and many of the issues seen (tests updated too).

E.g. search for single quotation marks picked up 30 instances which should now be fixed

image

The second issue is more the question of whether all this data needs to go into the JS Evaluator given that most of it is unused in the evaluation process. E.g. for the filter condition throwing errors it's only the draft column used. 

So I've drafted a follow-up pr #2189 to provide a potential workaround/optimisation in case any rows do still throw error, although if they do would still be good to identify any specific text characters causing issue to try and add a core fix for

jfmcquade:
I've just been doing some functional testing. As you mentioned might be the case (in the description of #2189), the [problematic sheet](https://docs.google.com/spreadsheets/d/1ElqsKnRT3nYAJZ0C_in8swQNgZvVTnzedCiAMycw9Og/edit#gid=1155509561) still caused an error on this PR's branch, but not on the branch for #2189. In this case, it was the row `faq_h_self_harm` which caused the following error to be thrown:
Screenshot 2024-01-31 at 15 23 04

After logging out the value of `cleanedFuncString` and analysing the output, I found that the `section_text_1` key for this row is missing a value:
Screenshot 2024-01-31 at 15 29 24

Upon inspection, it looks like that cell of the sheet contains _only_ newline characters (the highlight colour on my computer is pink, so this shows 2 newline characters selected in the cell):
Screenshot 2024-01-31 at 15 28 57

Deleting the contents of this cell caused the same error to be thrown on a new row, which, after following the same process, turned out to have a cell with a single newline character in also. Having deleted the contents of _this_ cell, the filter runs successfully.

So I'm not sure why it should behave differently, but I think we need to explicitly handle the case of a cell value consisting entirely of newline characters.

esmeetewinkel (approved):
Functional test passed. Merging this now as it resolves the issue at hand, but happy for the issue highlighted in @jfmcquade's last comment to be addressed in a follow-up PR

chrismclarke:
> So I'm not sure why it should behave differently, but I think we need to explicitly handle the case of a cell value consisting entirely of newline characters.

Nice catch! Yeah definitely a case I hadn't thought of. I'm thinking maybe this should be handled further up the chain (e.g. when syncing sheets), as I can't really think of any time a cell really would want to contain just a new line

Maybe we can turn into a low priority issue for someone to fix in the future? The default parser already has a [cleanFieldValues](https://github.com/IDEMSInternational/parenting-app-ui/blob/refactor/data-pipe-spec-filter-constants/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/default.parser.ts#L276) method currently used to remove additional whitespace, so might be the place to add a check/replacement.

chrismclarke:
Issue added https://github.com/IDEMSInternational/parenting-app-ui/issues/2193
Let me know if you want me to provide a quick fix or leave for future

esmeetewinkel:
>Let me know if you want me to provide a quick fix or leave for future

I think we can leave for future. Thanks!

PR #2189, opened 2024-01-31
Refactor/data pipe spec filter constants

chrismclarke:
PR Checklist

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/shared/src/models/dataPipe/operators/filter.ts
+++ b/packages/shared/src/models/dataPipe/operators/filter.ts
@@ -23,8 +23,8 @@ class FilterOperator extends BaseOperator {
 
   public satisfiesFilters(row: any) {
     const evaluator = new JSEvaluator();
-    evaluator.setGlobalContext({ constants: row });
     return this.filterConditions.every((condition) => {
+      evaluator.setGlobalContext({ constants: this.generateEvaluatorConstants(condition, row) });
       try {
         /**
          * row fields can be accessed both from global and local context, e.g.
@@ -38,6 +38,20 @@ class FilterOperator extends BaseOperator {
         throw error;
       }
     });
+  }
+
+  /**
+   * Create a list of all key-value pairs of row data whose keys appear within condition string
+   * This reduces the complexity of later JS evaluation
+   **/
+  private generateEvaluatorConstants(condition: string, row: Record<string, any>) {
+    const constants: Record<string, any> = {};
+    for (const [key, value] of Object.entries(row)) {
+      if (condition.includes(key)) {
+        constants[key] = value;
+      }
+    }
+    return constants;
   }
 }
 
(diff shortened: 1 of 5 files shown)


packages/shared/src/models/dataPipe/operators/filter.ts
```diff
@@ -39,6 +39,20 @@ class FilterOperator extends BaseOperator {
...
   }
+
+  /**
+   * Create a list of all key-value pairs of row data whose keys appear within condition string
+   * This reduces the complexity of later JS evaluation
+   **/
+  private generateEvaluatorConstants(condition: string, row: Record<string, any>) {
+    const constants: Record<string, any> = {};
+    for (const [key, value] of Object.entries(row)) {
+      if (condition.includes(key)) {
```
jfmcquade, line 50: This could lead to some key-value pairs being erroneously included, for example if the condition refers to a "name" column and the row data includes a "first_name" column. Seeing as the purpose of this step is efficiency/reducing complexity of the later JS evaluation, I don't think that's a problem, as it would still be an improvement in such cases.

jfmcquade (approved):
Looks good, and the tests are all passing (`yarn workspace shared test` and `yarn workspace scripts test`).

See minor clarification comment inline.

Happy to merge after #2187 has been merged (see [latest comment](https://github.com/IDEMSInternational/parenting-app-ui/pull/2187#issuecomment-1919366662) on that PR)

packages/shared/src/models/dataPipe/operators/filter.ts
chrismclarke: I think in that case if the condition were `name==='Ada'` then `condition.includes('first_name')` would still return false and not be included.

Although yes the opposite case would be true, `first_name==='Ada'` and `condition.includes('name')`.
Given that the variable can be used within any valid javascript I think the regex gets quite tricky to pin down perfectly (would need regex that allows any valid javascript syntax/operators either side of the condition text but not anything that could be included as part of a variable name - I'm assuming these are actually distinct sets given variable naming restrictions, however I'm less confident writing the regex to do so)

chrismclarke: A couple sources point in various directions to do so
https://mathiasbynens.be/demo/javascript-identifier-regex
https://regex101.com/r/7fjduD/1

However given the likely complexity I think I'd be much more in favor of accepting the limitation, and if desired exposing a config option to operations like `filter` that allow the user to explicitly pass names of columns to include for advanced use

esmeetewinkel (approved):
I tested this by updating [debug_data_list](https://docs.google.com/spreadsheets/d/1vtr341TIg0MDq99syARtSK6VmvI3rFjEdYfwgwdb1ek/edit#gid=1745157248) to contain the issue flagged in https://github.com/IDEMSInternational/parenting-app-ui/issues/2193. I can confirm that this sheet is broken on `master` but works on the PR branch.

Error when syncing `debug` content on `master`:
image

No error when syncing `debug` content on PR branch:
image

PR #2191, opened 2024-01-31
chore: Remove legacy images

esmeetewinkel:
PR Checklist

## Description
As I was playing around with `favicon.png`, I discovered some legacy assets in the code repo under src > assets. This PR is removing the ones that I'm convinced are not used. 

## Review notes
1. Could you confirm I'm not breaking anything by removing these? 
2. Potentially some more images could be removed (remaining ones in screenshot below)

   image


jfmcquade:
Well spotted, I think they can safely be removed. I've pushed a commit that deletes a few more: the only ones that are in use are the `tick.svg` and `cross.svg`, which are being used in the toggle bar component. This should be changed with a refactor of that component, which could be done as part of fixing the toggle bar for #2157

jfmcquade (approved):

PR #2192, opened 2024-01-31
feat: expose favicon asset path to deployment config

### LINKED ISSUE #1773: [FEATURE] Deployment-specific icon in browser bar
**What?**
We want to include an authoring option to change the icon that appears in the browser bar.

Mentioned in PR https://github.com/IDEMSInternational/parenting-app-ui/pull/1769. We will likely need to update the deployments a little so better after PR https://github.com/IDEMSInternational/parenting-app-ui/pull/1763 is sorted.

**Other details**
We may also want to provide best-practice info for authors, e.g.
https://www.emergeinteractive.com/insights/detail/the-essentials-of-favicons/
(tl;dr - a square svg should work fine in most browsers, or otherwise a square png up to 192x192)

jfmcquade:
PR Checklist

## Description

Exposes a property for setting the path to a custom favicon icon from a deployment config, `config.favicon_asset`. The path should be relative to the gdrive assets folder, e.g. "images/icons/favicon.svg". (NB, the property is not named "favicon_path", as the values of properties whose names end with "_path" are converted to absolute paths when the deployment config is compiled, which is not desirable here).

If no custom favicon asset is set, then the default is used. This default icon has been changed from the ionic logomark to the IDEMS trefoil knot logomark.

## Testing

With this feature branch checked out, add the line `config.favicon_asset = "images/icons/favicon.svg"` to `.idems_app/deployments/debug/config.ts` and run `yarn workflow deployment set debug` to compile the config. Now run `yarn workflow sync assets` to ensure the debug favicon asset is downloaded. Upon running `yarn start`, the browser favicon should be the custom icon added [here](https://drive.google.com/file/d/1JL1ViLwJ5yNqsVGZb8MVaiGXBMm5RsAo/view?usp=drive_link).

## Git Issues

Closes #1773

## Screenshots/Videos

Browser tab showing the [custom favicon](https://drive.google.com/file/d/1JL1ViLwJ5yNqsVGZb8MVaiGXBMm5RsAo/view?usp=drive_link) icon:

Screenshot 2024-01-31 at 16 46 29

Browser tab showing the new default favicon:
Screenshot 2024-02-13 at 10 35 43

### DIFF
--- a/packages/data-models/deployment.model.ts
+++ b/packages/data-models/deployment.model.ts
@@ -77,6 +77,7 @@ export interface IDeploymentConfig {
     translated_strings_path?: string;
   };
   web: {
+    /** Relative path of custom favicon asset to load from app_data assets */
     favicon_asset?: string;
   };
   workflows: {
--- a/packages/data-models/deployment.model.ts
+++ b/packages/data-models/deployment.model.ts
@@ -56,6 +56,7 @@ export interface IDeploymentConfig {
     /** filter function that receives basic file info such as relativePath and size. Default `(fileEntry)=>true`*/
     assets_filter_function: (fileEntry: IContentsEntry) => boolean;
   };
+  favicon_asset?: string;
   git: {
     /** Url of external git repo to store content */
     content_repo?: string;
@@ -140,6 +141,7 @@ export const DEPLOYMENT_CONFIG_EXAMPLE_DEFAULTS: IDeploymentConfig = {
     custom_ts_files: [],
     task_cache_path: "./tasks",
   },
+  favicon_asset: "",
   git: {},
   _validated: true,
   _parent_config: null,
--- a/packages/data-models/deployment.model.ts
+++ b/packages/data-models/deployment.model.ts
@@ -56,7 +56,6 @@ export interface IDeploymentConfig {
     /** filter function that receives basic file info such as relativePath and size. Default `(fileEntry)=>true`*/
     assets_filter_function: (fileEntry: IContentsEntry) => boolean;
   };
-  favicon_asset?: string;
   git: {
     /** Url of external git repo to store content */
     content_repo?: string;
@@ -76,6 +75,9 @@ export interface IDeploymentConfig {
     source_strings_path?: string;
     /** translated string for import. Default `./app_data/translations_source/translated_strings */
     translated_strings_path?: string;
+  };
+  web: {
+    favicon_asset?: string;
   };
   workflows: {
     /** path to custom workflow files to include */
@@ -137,11 +139,11 @@ export const DEPLOYMENT_CONFIG_EXAMPLE_DEFAULTS: IDeploymentConfig = {
     source_strings_path: "./app_data/translations_source/source_strings",
     translated_strings_path: "./app_data/translations_source/translated_strings",
   },
+  web: {},
   workflows: {
     custom_ts_files: [],
     task_cache_path: "./tasks",
   },
-  favicon_asset: "",
   git: {},
   _validated: true,
   _parent_config: null,
--- a/src/app/shared/services/seo/seo.service.ts
+++ b/src/app/shared/services/seo/seo.service.ts
@@ -64,10 +64,13 @@ export class SeoService extends SyncServiceBase {
    */
   private getDefaultSEOTags(): ISEOMeta {
     const PUBLIC_URL = location.origin;
+    const faviconPath = environment.deploymentConfig.favicon_asset
+      ? "assets/app_data/assets/" + environment.deploymentConfig.favicon_asset
+      : "assets/icon/favicon.png";
     return {
       title: environment.deploymentConfig.app_config.APP_HEADER_DEFAULTS.title,
       description: "",
-      faviconUrl: `${PUBLIC_URL}/assets/icon/favicon.png`,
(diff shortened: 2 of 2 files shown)


packages/data-models/deployment.model.ts
```diff
@@ -56,6 +56,7 @@ export interface IDeploymentConfig {
     /** filter function that receives basic file info such as relativePath and size. Default `(fileEntry)=>true`*/
     assets_filter_function: (fileEntry: IContentsEntry) => boolean;
   };
+  favicon_asset?: string;
```
chrismclarke, line 59: nit(non-blocking)
Should we consider nesting this within some sort of `web` config/property given that is the only context where this would apply?

packages/data-models/deployment.model.ts
```diff
@@ -140,6 +141,7 @@ export const DEPLOYMENT_CONFIG_EXAMPLE_DEFAULTS: IDeploymentConfig = {
     custom_ts_files: [],
     task_cache_path: "./tasks",
   },
+  favicon_asset: "",
```
chrismclarke, line 144: nit(non-blocking) might be worth putting the default path here just to save having to use conditionals in the seo service

chrismclarke (changes requested):
See a couple minor non-blocking comments inline (can still go ahead with merge if happy to do so)

The other minor issue is that initial favicon is defined in the main `src\index.html` file, which the seo service replaces. So depending how long the seo service takes to load there might be a brief period of wrong favicon shown. So we may want to consider either some workflow that either updates the index.html or overrides the default asset (could probably be done when copying assets to deployment), or putting the default as something blank/clear to hide the changeover. 

If going the asset replacement route then would remove requirement from the seo service, which might be nice in future cases if there is any resason to let deployments opt out of seo

packages/data-models/deployment.model.ts
jfmcquade: I think there would still need to be a conditional/check somewhere because the asset path gets a different prefix depending on whether the default asset is being used, as the custom asset sits inside the `app_data/assets/` dir where the default asset doesn't: `"assets/app_data/assets/<custom_favicon_path>"` vs `"assets/icon/favicon.png"`.

chrismclarke:
> > The other minor issue is that initial favicon is defined in the main `src\index.html` file, which the seo service replaces. ...
> 
> As it's defined in `src/index.html`, I've removed the logic for setting the default path elsewhere.

I'm not entirely sure what you mean by this - I can see the seo service no longer includes the `${PUBLIC_URL}/assets/icon/favicon.png` favicon however by removing that means any deployment that doesn't specify the asset will have a broken favicon which isn't ideal (probably bad for SEO). 

I've reverted to include default favicon fallback in 7fe67e6, could you just double check the custom favicon still works?

I still think unless there is a reason to replace the favicon at runtime (which we currently have no exposed methods for) it would still make more sense to make the file replacement as part of build process, so would be good to create a small issue that someone could take on post-merge (would be a good first issue). 

At the same time it would be good to provide a better default favicon to replace the ionic logo, perhaps someone can come up with a small logo to represent the platform? (possibly side-issue/discussion)

jfmcquade:
> > > The other minor issue is that initial favicon is defined in the main `src\index.html` file, which the seo service replaces. ...
> > 
> > 
> > As it's defined in `src/index.html`, I've removed the logic for setting the default path elsewhere.
> 
> I'm not entirely sure what you mean by this - I can see the seo service no longer includes the `${PUBLIC_URL}/assets/icon/favicon.png` favicon, however by removing it means that any deployment that doesn't specify the asset will have a broken favicon which isn't ideal (probably bad for SEO, so a bit counter-service namesake).

I removed that line from the SEO service because `index.html` includes the tag `<link rel="icon" id="favicon" type="image/png" href="assets/icon/favicon.png" />`. So the default favicon path is already set to "assets/icon/favicon.png", and in the case that no custom path is provided, the SEO service doesn't override it.

But I guess that was a bit of a half-way house between overriding at runtime and not: I can see how, as long as we are doing the runtime override, your changes make the default more explicit. I agree that it seems sensible to do the file replacement as part of the build process instead, I'll make an issue after this PR is merged.

I can confirm that the custom favicon still works with the latest code changes.

And agreed that we should consider a custom default favicon, I've added it to our meeting agenda.

chrismclarke (approved):

chrismclarke:
> I removed that line from the SEO service because `index.html` includes the tag `<link rel="icon" id="favicon" type="image/png" href="assets/icon/favicon.png" />`. So the default favicon path is already explicitly set to "assets/icon/favicon.png", and in the case that no custom path is provided, the SEO service doesn't override it.

Oh interesting, I didn't realise that the service is smart about not overriding what doesn't exist (which I can see now, and yes realise I'm the one who wrote the code 😅 )

> But I guess that was a bit of a half-way house between overriding at runtime and not: I can see how, as long as we are doing the runtime override, your changes make the default more explicit. I agree that it seems sensible to do the file replacement as part of the build process instead, I'll make an issue after this PR is merged.
> 
> I can confirm that the custom favicon still works with the latest code changes.
> 
> And agreed that we should consider a custom default favicon, I've added it to our meeting agenda.

All sounds good, I'd say this should be good to go then

PR #2198, opened 2024-02-09
Updated actions for building Android apps from content repos

ChrisMarsh82:
## Description

Updated gradle and action to now use place holders for version numbers. This will allow content repos to be on different versions.
Deleted old actions for building Android apps from the code repo. This will mean only separate content repos can build Android apps.

Warning: Because this updates the actions to use placeholders this does not want to be merged until a new release is ready. Once merged a new release will need to be made then content repos will need to be updated to use the new release.


chrismclarke:
Not to derail things too much - but I've opened a PR targetting this branch which would allow using the existing versioning script as part of reusable actions.
https://github.com/IDEMSInternational/parenting-app-ui/pull/2206

The benefit I see do doing things this way is we get to avoid the confusing bash scripting (more extensible if we also want to do things like replace IOS build numbers in the future), and it includes validation to ensure version number is incremented

But otherwise I think nice to have the legacy actions tidied up and good to be moving away from fixed value replacements

jfmcquade (approved):
Looks good to me. Incorporates #2206, which I have reviewed. I think I will go ahead and merge so that we can make a new release with these changes included.

@ChrisMarsh82 After merging, could you test the android release action from a deployment repo targeting the master branch? Then, if it all looks good, I'll make a release. After that, could you update all(?) content repos to point to the latest code version.

@esmeetewinkel Something to be aware of: when we want to increase the code version number ahead of a release, we should use the command `yarn run version --package` (or `yarn scripts version --package`). Running the old command, `yarn scripts version` will prompt you to add a flag:
Screenshot 2024-02-21 at 11 38 19
We want to use `--package`, as the Android files should only be changed as part of the publishing to play store action (or by developers when testing builds, for example when testing the in-app updates feature)

### Documentation
There is no existing documentation for publishing to the Play Store, I have added a section to be completed in the [WIP main document of new documentation](https://docs.google.com/document/d/1P7xPqTJcHFibEmv5qbWuXeGKvfCsotEzHnUPaSkHX6M/edit#heading=h.o2er2dytnhyj).
The only reference to the `yarn scripts version` command in the docs currently is in the instructions for testing the [In-App Updates feature](https://idemsinternational.github.io/parenting-app-ui/developers/in-app-updates/#testing), which I have updated as part of this PR.

PR #2174, opened 2023-12-18
feat(scripts): add skip refresh flag to deployment set

chrismclarke:
PR Checklist

- [x ] PR title descriptive (can be used in release notes)

## Description
Add flag to allow skipping auto remote content refresh when setting a deployment

## Review Notes
Set a deployment that has remote repo (e.g. debug), use the `--skip-refresh` flag to ensure updates not pulled from remote
`yarn workflow deployment set [name] --skip-refresh`

## Screenshots
**Refresh step shows as skipped**
image

PR #2199, opened 2024-02-09
Updated actions for content repos

ChrisMarsh82:
## Description
Added skip refresh when calling set deployment from actions. This will stop data from being updated from the spreadsheets and use whats in the repository instead.

@esmeetewinkel This should fix the PR preview action.


jfmcquade (approved):
Looks good to me. Should be tested after merge by building from a content repo.

NB this builds on #2174

PR #2200, opened 2024-02-12
Added appertize workflow for content repos

ChrisMarsh82:
## Description

Created Test-Appertize workflow to push android app to appertize from content repos

### DIFF
--- a/.github/workflows/reusable-android-build.yml
+++ b/.github/workflows/reusable-android-build.yml
@@ -4,7 +4,7 @@
 # Reuseable workflow to be called from content repos. 
 # Build and deploy app bundle for android
 #
-#         Version : 1.0
+#         Version : 1.1
 #
 ##################################################################################
 #         Configuration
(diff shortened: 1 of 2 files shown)


jfmcquade (changes requested):
@ChrisMarsh82 I've found the PR that originally introduced deployment-specific appetize instances, and added the github environments to manage the deployment-specific secrets: #1864.

There's an explanation in that PR's description of why the environments were set up, and I think that now that we have deployment repos we can get rid of the environments (although the environments implementation is quite interesting). It also explains that the `APPETIZE_TOKEN` is indeed the same universally, so that secret could be managed as an organisational secret. I'll send you an email with an API token that can be used here. 

I think for this current PR, the workflow should be changed so that the `APPETIZE_APP_KEY` comes directly from a secret on the deployment content repo, whereas the `APPETIZE_TOKEN` should come from an organisational secret. Maybe this is already the case for the existing action, I'm not certain about where those `secrets.*` values pull from?

One other minor point: there is a typo in the name of the workflow file, it should be `reusable-appetize.yml` rather than `reusable-appeRtize.yml`.

ChrisMarsh82:
@jfmcquade updated typo. 
The secrets work by always coming from the repo. The can have organisational secrets that it can use or override. Therefore no changes are needed for global secrets

jfmcquade (approved):
Looks good. As discussed, I can manually create an Appetize instance for any deployment that needs one, and then save the associated "app key" to the deployment repo's secrets. After that point, the automated Appetize workflow should work to update the instance for the deployment

PR #2206, opened 2024-02-20
refactor: use version script in reusable actions

chrismclarke:
PR Checklist

## Description

Targets #2198 

- Updates version script to be callable by `yarn run version` (alongside original `yarn scripts version`)
- Separates versioning for `package.json` and android `build.gradle`, using flags to specify (e.g. `yarn run version --android`)
- Add android build version check using existing value from `build.gradle`
- Adds support to existing version script to directly provide target version, e.g. `yarn run version 1.0.5 --package --android`, which will automatically populate android build and package.json version codes
- Improves regex to ensure android versionCode and versionName are replaced even if string value appears instead of number (e.g. current `*versionCode*` placeholder text)
- Improves version code generator to allow integer values if desired (e.g. `yarn run version 2` -> code `2000000`)

## Review Notes
Can test variations of setting package.json and android build.gradle versions via
```sh
yarn run version [number] --android --package
```
all args after `version` are optional, prompting input if no version code set and can be applied to android, package or both

Untested in github actions (not sure the best way to trigger a deployment to call)

## Git Issues

Closes #

## Screenshots/Videos
**Example** - if version code not specified input prompted with validation (allows re-editing version)
image

**Example** - output if specifying number directly (with invalid code)
image

### DIFF
--- a/.github/workflows/reusable-android-build.yml
+++ b/.github/workflows/reusable-android-build.yml
@@ -118,26 +118,9 @@ jobs:
             VERSION=$(grep 'content_tag_latest:' $FILE_PATH | sed 's/content_tag_latest: *"\(.*\)",/\1/')
             echo "Extracted Version: $VERSION"
       
-            # Split the version into major, minor, and patch components
-            IFS='.' read -ra VERSION_PARTS <<< "$VERSION"
-            
-            if [[ ${#VERSION_PARTS[@]} -ne 3 ]]; then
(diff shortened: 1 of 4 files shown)


jfmcquade (approved):
Looks good, and I've confirmed with @ChrisMarsh82 that we're happy to include this in #2198 so I will merge.

I've added a minor update to the docs, updating the one place where we reference `yarn scripts version`.

PR #2215, opened 2024-03-01
feat: set_item action: optionally target an index or id

jfmcquade:
PR Checklist

## Description

Adds two new optional parameters for the `set_item` action: `_id` and `_index`. These are used to target a specific item in the data items loop to set the a value on, for example, `click | set_item | _index: 2, completed: true` or `click | set_item | _id: my_some_id, completed: true`.

### _index

NB, the index supplied is evaluated in terms of the standard 0-based indexing of arrays, where the first element of an array has index 0, the second element has index 1, and so on.

JS expressions can be used to specify the target index. Trivially `1 + 2`, for example, to target the item at index `3`, but also these expressions can use the `@item._index` to refer to the index of the current item in the loop. For example, 
`set_item: @item._index + 1` will target the item that appears after the current item in the data items iterator.

## Testing

Experiment with Example 4 in the the [comp_data_items](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=845703892) template – this isn't the prettiest example, but should demonstrate the functionality. And with the [example_task_group_stepper](https://docs.google.com/spreadsheets/d/1w1OpqizCEDqf0FT7tgja3CUgdnXxJ2CGbh8v8N4gpQs/edit#gid=1921555013) template. See screenshots below.

## Git Issues

Closes #

### Action

Updated [comp_data_items](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=845703892)

Screenshot 2024-04-08 at 11 44 16

### Use case demo

[example_task_group_stepper](https://docs.google.com/spreadsheets/d/1w1OpqizCEDqf0FT7tgja3CUgdnXxJ2CGbh8v8N4gpQs/edit#gid=1921555013) template and demo. This is our use case for this new feature.

Screenshot 2024-03-06 at 16 31 41

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/9ba57955-e9bc-4f80-92c5-b4058ce6aa3a

### New tests passing

Screenshot 2024-04-24 at 14 03 20

Screenshot 2024-04-24 at 10 22 53


src/app/shared/components/template/components/data-items/data-items.component.ts
```diff
@@ -80,6 +80,14 @@ export class TmplDataItemsComponent extends TemplateBaseComponent implements OnD
     // TODO - deep diff and only update changed
     this.itemRows = replacedActionRows;
     this.cdr.markForCheck();
+    // Without this second check, and without the delay, templates nested inside data-items loops do not render.
```
jfmcquade, line 83: @chrismclarke – any ideas what could be causing this?

Demo of nested template only rendering after second check:

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/1042772a-46e1-4b33-bba0-31da7fcaf802


Another solution, which causes the nested template to render immediately, is to reset the change detection strategy to the default (currently `OnPush`). Am I right in thinking that we use `ChangeDetectionStrategy.OnPush` to avoid unnecessary checking for changes and DOM updates?

src/app/shared/services/dynamic-data/dynamic-data.service.ts
```diff
@@ -223,4 +237,18 @@ export class DynamicDataService extends AsyncServiceBase {
     }
     return schema;
   }
+
+  evaluateItemIndexString(targetIndexString: string, currentIndex: number) {
+    if (isNaN(Number(targetIndexString))) {
+      try {
+        // HACK use keyword "item_index" to refer to index of current item in loop
+        const jsExpression = targetIndexString.replace("item_index", currentIndex.toString());
```
chrismclarke, line 245: nit(blocking)
Not a huge fan of this arbitrary lookup string. If using some sort of keyword/variable replacement it should ideally be prefixed in a clear way (sometimes we use '@', sometimes '_')

In this case all items already have an _index property, so I think you should be able to refer to @item._index + 1 and it be calculated without any need for this function at all

src/app/shared/services/dynamic-data/dynamic-data.service.ts
```diff
@@ -70,13 +70,27 @@ export class DynamicDataService extends AsyncServiceBase {
...
+      },
       set_items: async ({ args, params }) => {
         const [flow_name, row_ids] = args;
         // Hack, no current method for bulk update so make successive (changes debounced in component)
         for (const row_id of row_ids) {
           await this.update("data_list", flow_name, row_id, params);
         }
       },
+      // "Reset" the data of a given data list by removing its dynamic user overwrites
+      reset_data: async ({ args }) => {
```
chrismclarke, line 91: nit(non-blocking)
It might be better fitting with other item action names to just call it reset_item to remove the data_list overrides for a specific items. Could also add reset_items for bulk

src/app/shared/components/template/components/data-items/data-items.component.ts
chrismclarke: Could you make a more minimal debug sheet? I tried loading the task group stepper but needs extra data from `feat_task_group_1_data` so hard to debug efficiently

chrismclarke:
> In order to reference the index of the current item, I originally imagined using the syntax @item.index at the template level. However there are various complications with this. As the @item. syntax is generally used to refer to columns/keys of the item in the data list, @item.index would be needed to be treated as a special case. Additionally, as @item.index can only be evaluated in the context of of the data items loop in which it finds itself, it would need to pass through, unparsed, until being handled by the data-items component. So using a keyword, item_index, was much more straightforward to implement. If this is acceptable, we could merge this implementation and plan to support another syntax in the future.

As mentioned in the comment, all items already have a readonly `@item._index` property which can be used instead of any new variables.

>A possible alternative syntax (not implemented) came up during development: click | set_item_at_index | index: item_index + 1, active:true or even click | set_item | index: item_index + 1, active:true (as index would be provided explicitly, it may not be confusing to include as an optional parameter on the existing set_item action. This could be substituted for the currently implemented syntax outlined above if it was deemed preferable.

I like the look of this, although I would again refer to `_index` readonly property. We should clarify when to use comma-separated or semi-colon separated (possibly following #2211, as may want to parse action params in a similar way as required). Ideally we want to use a syntax that automatically gets parsed as a json object

So assuming syntax something like:
```
 click | set_item | _index: @item._index + 1; active:true
```
I would also be tempted to let the user pick an item by id, e.g.
```
 click | set_item | id: my_some_id; active:true
```
And updating the logic to first extract `{_index, id, ...writeableProps } = arg`, using the _index or id if available to select item, with default current item in list if not provided.

The benefit of doing things this way is we could then use pretty much identical syntax when adding support for updating items from outside the item list itself (currently not supported but expect a nice-to-have).

src/app/shared/components/template/components/data-items/data-items.component.ts
jfmcquade: Sure, I've made a debug sheet here: [debug_data_items_rendering](https://docs.google.com/spreadsheets/d/14XXwZ7R_7qIBo5xKYuxnPVQbQidI--PPJ7TVvZxwy1k/edit#gid=1422144003).

Here's what I get without the second, delayed mark for check:

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/b0e821df-0437-49a1-ba02-08acaa0d5e83

With the second, delayed mark for check:

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/36875f04-8837-45bf-9d09-d97f23f777b6

And without `changeDetection: ChangeDetectionStrategy.OnPush` (and no second mark for check):

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/608dcca6-7559-4948-933e-7c0cc2f18054

chrismclarke: Potentially resolved with https://github.com/IDEMSInternational/parenting-app-ui/pull/2221

src/app/shared/components/template/services/template-variables.service.ts
```diff
@@ -183,6 +183,11 @@ export class TemplateVariablesService extends AsyncServiceBase {
         return evaluator.fullExpression.replace(/`/gi, "");
       }
 
+      // Do not evaluate if the appropriate context is not available
```
jfmcquade, line 186: This took me ages to figure out! There may be a nicer solution, but this is why something like this is needed:

Even though we have the check lower down (line 393), without this additional check then the string gets evaluated as JS, without the dynamic evaluator value being replaced. This meant that my string `@item._index + 1` was being evaluated as JS before `itemContext` was available, evaluating to `@item._index1` because it was treated as the addition of a string and a number, `"@item._index" + 1`. Then, when the expression was re-evaluated after the data-items component had added the `itemContext`, `@item._index1` was evaluated with the value of `@item_index` replaced. For the item at index `0`, this actually meant the expression evaluated as `01`, i.e. `1`, and so for that one case the evaluation actually gave the expected value, which appeared as a false positive in my testing...

chrismclarke (changes requested):
I think we should put a block on this until after the issues in #2220 resolved...

chrismclarke:
@jfmcquade - might be worth revisiting this one now #2220 is merged. When you get a chance to resolve conflicts let me know and I can give a quick additional check

jfmcquade:
Thanks for the reminder, @chrismclarke, I believe this is ready for review

esmeetewinkel (commented):
Functionally tested `example_task_group_stepper` template, works well. 

For some reason none of the loops in the template `comp_data_items` show anything... (and they do display content on `master`)
image

src/theme/themes/pfr.scss
```diff
@@ -48,7 +48,7 @@
       // radio-button-font-color: var(--ion-color-primary),
       ion-item-background: var(--ion-color-gray-light),
       // task-progress-bar-color: var(--ion-color-primary),
-      // checkbox-background-color: white,,,
+      // checkbox-background-color: white,,,,
```
chrismclarke, line 51: nit(non-blocking):
not sure why this file keeps getting more commas, guessing something to do with code auto-formatters? In any case would suggest discarding the change from this PR and trying to fix root issue separately (maybe just discard all these commented properties?)

chrismclarke (changes requested):

src/app/shared/components/template/services/template-variables.service.ts
chrismclarke: Really useful to catch, it would be good to include such things in spec test if possible

chrismclarke (changes requested):
Thanks for the updates, I've pushed a small commit (ef3047a) to handle more of the nit-pick type things I came across during review, although I think there's enough core changes here that I'd say we really need to add/update spec tests to include the proposed functionality changes before merging. Otherwise it's pretty awkward to test and may be confusing if returning to in the future.

Let me know if you can get a chance to do this, or if you want me to take a look

src/app/shared/services/dynamic-data/dynamic-data.service.ts
```diff
@@ -66,9 +66,41 @@ export class DynamicDataService extends AsyncServiceBase {
...
+            `[SET ITEM] - Setting properties on ${targetRowId}: ${JSON.stringify(writeableProps)}`
+          );
+          await this.update("data_list", flow_name, targetRowId, writeableProps);
+        } else {
+          if (_id) {
+            console.warn(`[SET ITEM] - No item with ID ${_id}`);
+          }
+          if (_index !== undefined) {
+            console.warn(`[SET ITEM] - No item at index ${_index}`);
+          }
```
chrismclarke, line 102: nit(blocking): I feel like this is a large amount of item-specific logic being handled in the dynamic-data service. Would it be possible to move the item selection logic to the items component instead, so that the action is always triggered with the correct item index instead?
There's also quite some debug logs that I think probably should make it into production (would suggest remove console log and merge warn statements

src/theme/themes/pfr.scss
jfmcquade: Yes, I think this is prettier adding a trailing comma, see this open issue: https://github.com/prettier/prettier/issues/7752
I'm reluctant to remove these commented lines, as they represent the variables that are possible to be overridden by a theme (not a great system, but useful for now until it is improved). 
I'll try updating out prettier version, as it [looks like a possible fix was included in v3.0.2](https://github.com/prettier/prettier/blob/main/CHANGELOG.md#do-not-add-trailing-comma-for-grouped-scss-comments-15217-by-auvred).

src/theme/themes/pfr.scss
```diff
@@ -48,7 +48,7 @@
       // radio-button-font-color: var(--ion-color-primary),
       ion-item-background: var(--ion-color-gray-light),
       // task-progress-bar-color: var(--ion-color-primary),
-      // checkbox-background-color: white,,,,
+      // checkbox-background-color: white,
```
jfmcquade, line 51: I'm not sure why this change is included here, as it was already committed as part of #2299. I've tried changing the base of this PR away from master and back again, but this supposed difference remains. (That method did get rid of some erroneous differences that were part of #2294, which I merged into this feature branch before it was merged to master).

jfmcquade:
Thanks @chrismclarke, see my comments and replies inline. 

In terms of tests, I've added some for the `set_item` logic, and some minimal tests for the template-variables service. 
The logic of the template-variables service is pretty complex, and it's hard to generate minimal examples of the the `context` object for testing (I logged out some actual context for a [minimal sheet](https://docs.google.com/spreadsheets/d/1tL6CPHEIW-GPMYjdhVKQToy_hZ1H5qNIBkkh9XnA5QM/edit#gid=114708400) and copied the result for the tests I have written). I think more extensive tests are probably outside the scope of this PR, but the spec file is there and the tests can be run now at least. There is a comment in the `app-data-variable` service that it should ideally be extended to replace the `template-variables` service, so potentially there's a more substantial TODO to handle this refactor, including adding extensive tests. Let me know if you think some more tests here are minimally required for this PR.

chrismclarke:
> Thanks @chrismclarke, see my reply to your latest inline comment.
> 
> In terms of tests, I've added some for the `set_item` logic, and some minimal tests for the template-variables service. The logic of the template-variables service is pretty complex, and it's hard to generate minimal examples of the the `context` object for testing (I logged out some actual context for a [minimal sheet](https://docs.google.com/spreadsheets/d/1tL6CPHEIW-GPMYjdhVKQToy_hZ1H5qNIBkkh9XnA5QM/edit#gid=114708400) and copied the result for the tests I have written).
I think more extensive tests are probably outside the scope of this PR, but the spec file is there and the tests can be run now at least. There is a comment in the `app-data-variable` service that it should ideally be extended to replace the `template-variables` service, so potentially there's a more substantial TODO to handle this refactor, including adding extensive tests. Let me know if you think some more tests here are minimally required for this PR.

Thanks for adding the necessary scaffolding and starting to build out the item tests - agreed the service really isn't set up very well for long term maintainability (lots of complex code and weird caveats - which probably makes testing more important but beyond scope of this PR for sure). I've added 149c868 to try and tidy up a little and add a few more basic public method tests, but will stop there as fair chance we'll want to do a larger refactor at some point in the future. But at least for now it covers the new introduced behaviour so I think should be good to go

chrismclarke (approved):

Issue #1617, opened 2022-11-01
Drawer component visual improvements


EsmeeIDEMS:
(Non-blocking) three minor comments about the look: 
- I **sometimes** seem to have no padding between the drawer and the footer - so the letter "g" in "Things" touches the footer. I can't figure out exactly when this does / does not display correctly.
<img width="242" alt="image" src="https://user-images.githubusercontent.com/74557272/198118314-20a85477-791c-491a-a7b0-bd45e4db4a6b.png">

- Currently the scroll bar on the RHS of the screen appears in front of the drawer, but I think it would make more sense if it was behind the drawer. 
<img width="280" alt="image" src="https://user-images.githubusercontent.com/74557272/198115538-399bd031-a64a-4e50-994f-02046a48dd3a.png">

- When there's no footer the drawer should drop down to the bottom of the screen, taking up the footer's space. Not a use case we have at the moment, but it would make sense to me. 
<img width="279" alt="image" src="https://user-images.githubusercontent.com/74557272/198115813-246f90fc-b13f-45ef-9a8b-362518f5a2c1.png">

_Originally posted by @EsmeeIDEMS in https://github.com/IDEMSInternational/parenting-app-ui/pull/1584#pullrequestreview-1157153240_

PR #2216, opened 2024-03-04
feat: header config: 'collapsing', 'compact, and 'colour' options

### LINKED ISSUE #2154: [FEATURE] Option to remove app header
**What?**
Ability to create a header-less app, see [Figma mockups](https://www.figma.com/file/hUwQvxkWDYKRJltebSayYN/ParentApp-for-Kids?type=design&node-id=215-2993&mode=design&t=xCvLLM5W1VJLmz5P-0). 

image

N.B. @jfmcquade I'm not sure whether to interpret the proposed navigation as a header-less app, or a "minimal header" together with a load of full-screen pop-ups.

**Why?**
Less cluttered design, more screen space to display content. 

**How?**
Find below some notes I took when we discussed this in the dev catchup on 19 Oct 2023 (@jfmcquade please replace with more understandable notes if you have any!)
1. Menu variable in config to hide header
2. If we want a non-sticky (floating) header, proposal to have named layouts, e.g.
   - Header + sidebar
   - Header only
   - Non-sticky header
   - No header
   - Minimal header (only back button, no background colour)
3. Knock-ons for navigation:
   - Header back button
   - iOS navigation - has swipe back action so could confuse with side menu or interacting with a slider component
   - Navigation not custom applied to each template

jfmcquade:
PR Checklist

## Description

Adds 3 new options for styling the main app header: "collapsing", "compact", and "colour". These are exposed at the app config level (as part of the `APP_HEADER_DEFAULTS` object), so can be set in a deployment config or a skin.

|                 | collapsing                                         | variant                                                          | colour                                                                                     |
|-----------------|----------------------------------------------------|------------------------------------------------------------------|--------------------------------------------------------------------------------------------|
| Possible values | boolean                                            | "default", "compact"                                                          | "primary", "secondary", "none"                                                             |
| Default value   | false                                              | "default"                                                            | "primary"                                                                                  |
| Effect                | If true, the header will appear/reappear on scroll | If `compact`, a narrower version of the header is used, with no title | Sets the background colour of the header. As well as setting the header colour to match the page background, "none" also removes the shadow, for a flat look. |

See demos below.

Also refactors the drawer component, which would otherwise be broken by this functionality. This refactor should fix at least some of the issues in #1617

## Testing

In order to test each setting, set its value in the debug deployment config, e.g.:
``` js
config.app_config.APP_HEADER_DEFAULTS.collapsing = true;
config.app_config.APP_HEADER_DEFAULTS.variant ="compact";
config.app_config.APP_HEADER_DEFAULTS.colour = "primary";
```
Then run `yarn workflow deployment set debug` to recompile the config file, before running `yarn start`. I think you'll need to stop the app running and start it again after each time you apply any changes to the config. 

The drawer component does now work as expected even when there is no footer, but you may want to test the interaction of the drawer with the footer, in which cases, use:
``` js
config.app_config.APP_FOOTER_DEFAULTS.templateName = "feat_footer"
```
to target the new [feat_footer](https://docs.google.com/spreadsheets/d/10iy1Vq-pfD2ERBX6xY66-GY87B8_srAnBKz9hOpRYIo/edit#gid=960980632) template.

## Git Issues

Closes #2154

### 1
| collapsing | variant            | colour                 |
|------------|--------------------|------------------------|
| true       | "default" (or not set) | "primary" (or not set) |

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/424f0411-192f-4cb5-81f9-a41ed08feb70

### 2
| collapsing | variant            | colour                 |
|------------|--------------------|------------------------|
| true       | "compact" | "secondary" |

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/947d8581-e8d8-4a7a-8155-79115c77dbda

### 3
| collapsing | variant            | colour                 | Theme                 |
|------------|--------------------|------------------------|------------------------|
| true       | "compact" | "none" | Professional |

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/9e92a85f-f6d3-40d2-9fe5-6155ea8f3fbb

### Interaction with drawer and footer:
[comp_drawer](https://docs.google.com/spreadsheets/d/1NY-FwPH7AZNlBoaS3Oi0E5j20ppdZwt0zbnqh72D1Tk/edit#gid=569531329)

| collapsing | variant            | colour                 |
|------------|--------------------|------------------------|
| true       | "compact" | "primary" |

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/c0bf8cec-e49a-44aa-9ee9-20606de6898a


packages/data-models/appConfig.ts
```diff
@@ -106,8 +106,16 @@ const APP_ROUTE_DEFAULTS = {
   ],
 };
 
+type IHeaderColourOptions = "primary" | "secondary" | "none";
+
 const APP_HEADER_DEFAULTS = {
   title: "App",
+  collapsing: false,
```
chrismclarke, line 113: nit(non-blocking) probably more consistent to use `collapse` (imperative) instead of `collapsing` (present participle)

packages/data-models/appConfig.ts
```diff
@@ -106,8 +106,16 @@ const APP_ROUTE_DEFAULTS = {
...
+  colour: "primary" as IHeaderColourOptions,
+  compact: false,
```
chrismclarke, line 115: nit(non-blocking) could you add a comment to explain what `compact` means. Also is it required or implied by providing heightCompact. If this is more of a style variant would it make more sense to call `variant: 'compact'` (with options for `'default' | 'compact'`)

packages/data-models/appConfig.ts
```diff
@@ -106,8 +106,16 @@ const APP_ROUTE_DEFAULTS = {
...
+  // expose heights (in px) so that "compact" can be applied by config/skin independent of theme
+  heightDefault: 56,
+  heightCompact: 36,
```
chrismclarke, line 118: nit(non-blocking) do we want authors to be able to manipulate these heights? I'm assuming there is reason for the numbers (e.g. minimum padding), so might be better to just expose the `compact` property and use that to manage heights from the scss

src/app/app.component.ts
```diff
@@ -164,9 +167,18 @@ export class AppComponent {
...
       this.appAuthenticationDefaults = this.appConfig.APP_AUTHENTICATION_DEFAULTS;
       this.appFields = this.appConfig.APP_FIELDS;
+      this.shouldCollapseHeader = this.appConfig.APP_HEADER_DEFAULTS.collapsing;
     });
   }
 
+  /** Listen to scroll events to trigger header collapse */
+  @HostListener("ionScroll", ["$event"])
+  scrolling($event: any) {
+    if (this.shouldCollapseHeader) {
```
chrismclarke, line 177: nit(non-blocking) - this will be adding event listeners even when not using collapseHeader. Is there a way to conditionally add the listener instead of conditionally forwarding each event?

src/app/shared/components/header/header.component.ts
```diff
@@ -64,6 +53,26 @@ export class PLHMainHeaderComponent implements OnInit, OnDestroy {
       });
     }
   }
+
+  subscribeToAppConfigChanges() {
+    this.appConfigChanges$ = this.appConfigService.appConfig$.subscribe((appConfig: IAppConfig) => {
+      this.headerConfig = appConfig.APP_HEADER_DEFAULTS;
```
chrismclarke, line 59: nit(non-blocking) might be cleaner to put all the header config customisation into its own function.

src/app/shared/components/header/header.component.ts
```diff
@@ -64,6 +53,26 @@ export class PLHMainHeaderComponent implements OnInit, OnDestroy {
...
+      if (this.headerConfig.colour === "none") {
+        this.colour = undefined;
+      } else if (COLOUR_OPTIONS.includes(this.headerConfig.colour)) {
+        this.colour = this.headerConfig.colour;
+      } else {
+        this.colour = "primary";
+      }
```
chrismclarke, line 66: nit(non-blocking): might be a bit cleaner defining the options as a map/enum so that we can just pick `COLOUR_OPTIONS[this.headerConfig.colour] || 'primary'` instead of all the if statements

src/app/shared/shared.module.ts
```diff
@@ -1,7 +1,7 @@
 import { NgModule } from "@angular/core";
 import { CommonModule } from "@angular/common";
 import { IonicModule } from "@ionic/angular";
-import { PLHMainHeaderComponent } from "./components/plh-main-header";
+import { PLHMainHeaderComponent } from "./components/header/header.component";
```
chrismclarke, line 4: nit(non-blocking): seeing as we've moved from `plh-main-header` maybe worth dropping the `plh` altogher and just using something like `AppHeaderComponent`
Also when referring to `MainHeader` is there some sort of alternative header? Not sure if we can drop the `Main`

src/app/shared/components/common/directives/collapse-on-scroll.directive.ts
```diff
@@ -0,0 +1,92 @@
...
+  constructor(
+    private element: ElementRef,
+    private renderer: Renderer2,
+    private domCtrl: DomController,
+    private appConfigService: AppConfigService
+  ) {}
+
+  ngOnInit() {
+    this.subscribeToAppConfigChanges();
+    if (this.collapsing) {
```
chrismclarke, line 32: nit(blocking) - could you move all this code into a separate function with a description of what it does?

chrismclarke (changes requested):
Thanks @jfmcquade 
I've added a number of minor tidying requests. There's quite a lot to unpack here (appreciate a bunch is non-changes just down to modifying file names/templates), so I'll hold off on a more technical review until after the initial pass comments are updated just to make the process a little easier.

packages/data-models/appConfig.ts
jfmcquade: I've updated `APP_HEADER_DEFAULTS` to include a `heightsMap` instead (see latest code changes):
```js
  heightsMap: {
    compact: 36,
    default: 56,
  }
```

The motivation for including these values in the appConfig was less about exposing them to authors but more to make available universally to the code. We could put this heightsMap object somewhere else and import it statically into the relevant places (header component and collapse-on-scroll directive). 

Or, we could set these values in the theme, i.e. as global CSS variables like `--header-height-default` and `--header-height-compact`. In that case we'd need a way to access them in the TS. Perhaps it would be sensible for the theme service to expose an observable for subscription that contains all of the custom CSS properties (starting with "--"). We already have a method in the theme service, `listAllCssCustomProperties()`. But would these additional subscriptions be worth it to just access these properties?

chrismclarke (approved):
Thanks again @jfmcquade , nice to have the template container subscription (I had been testing a workaround to pick out ion-content elements programtically but seems a lot simpler to do this way for now). 

Other changes to collapse detection look good to me, also thanks for catching the debug setting. I'd say all-in-all should be good to go once functional tests pass. Just to note for @esmeetewinkel most recent screenshots/demos are the ones @jfmcquade added in #2226 comments

esmeetewinkel (approved):
Functional test passed. 

I believe every reference to the `collapsing` property needs to be `collapse` - maybe worth updating since the PR notes will be the basis for future documentation

image

esmeetewinkel:
@jfmcquade minor comment that can be done as a follow up: 

When the colour is none, could the icons that appear on the header appear in `primary`? E.g. currently black but I'd prefer this to be dark-blue like the text

image

PR #2220, opened 2024-03-05
feat: add data-items eval context fields

chrismclarke:
PR Checklist

## Description

Update `items` and `data_items` template row types to add support for metadata `@item._first` and `@item._last` to more easily determine if an item is the first or last rendered row (after potential filter operations).

## Author Notes
- Adds new fields to refer to `@item._first` and `@item._last` to determine if first/last in rendered list. This is calculated after any other operators (e.g. filter) so should be accurate (manually tested by adding a `filter: completed` parameter to the list, a few more items and then toggling actions - I didn't keep those changes in the debug template to keep things more minimal but feel free to add if useful).
- Adds `@item._id` property, which is the same as `@item.id`. This is duplicated more for consistency, I don't think it will make much sense to deprecate the `id` parameter as that matches the excel sheet column name. 
- This should be applied to both `data_items` and `items` templates (only tested with data_items)
- Updated [component_data_items](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=845703892) sheet

## Dev Notes
- Fixes `@item._index` now parsed as integer (previously string)
- Fixes issue where data_item _index would not be dynamically updated (e.g. when using filter operations), and where rows would have their action_list updated before evaluation. Both of these may have been part of the issue identified in #2215 (TBC, if possible to confirm @jfmcquade whether it does or not post merge)

## Git Issues

Closes #

## Screenshots/Videos
**Updated [component_data_items](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=845703892) sheet**

Screenshot 2024-03-12 at 11 55 26

Screenshot 2024-03-12 at 11 55 59


jfmcquade:
Thanks, very helpful.

See my update/fix in [these commits](https://github.com/IDEMSInternational/parenting-app-ui/pull/2220/files/0d71384d2dffc8ca7613671fe12977b565c33fa3..5589c3c3d2134e527ba88251701ae3c50e80ea93) (the _index should be relative to the rows of the data list, not the rows of the rendered template).

I haven't tested extensively but I have merged into my branch for #2215 and it has been helpful. I'm still working on that PR but will update soon.

chrismclarke:
> See my update/fix in [these commits](https://github.com/IDEMSInternational/parenting-app-ui/pull/2220/files/0d71384d2dffc8ca7613671fe12977b565c33fa3..5589c3c3d2134e527ba88251701ae3c50e80ea93) (the _index should be relative to the rows of the data list, not the rows of the rendered template).

I don't quite follow what your changes do.

Previously the lastRowIndex was calculated by
```
const lastRowIndex = rows.length - 1;
```
Now the total rows is calculated by 
```
 const rowIds = Object.values(dataList).map((v) => v.id);
 const lastRowIndex = rowIds.length - 1;
```
But given that the dataList has just been mapped to ids the length will remain unchanged, so I don't think that change achieves anything.

So then the next (and I'm assuming more essential part) of the code which I'm struggling to follow/understand also...
Before _first and _last properties are calculated based on the available data (as passed to the items following any additional sort/filter ops). But you've changed to instead lookup the item in the original data list (?)... surely in doing so we lose all positioning following the filter/sort ops

E.g. I've updated the demo to include a 3rd item and filtering data based on completed column. With your updates then we lose first/last as completed items are toggled we get some weird first/last states not matching up with the data

image

So my only guess is we're trying to address different problems. Could you please update the example sheet to make it clearer what this should achieve (and why it fails with the original implementation)?

jfmcquade:
> So my only guess is we're trying to address different problems. Could you please update the example sheet to make it clearer what this should achieve (and why it fails with the original implementation)?

I think the issue is that we shouldn't be looping over `rows` to assign these metadata fields, but instead should be looping over the filtered data list (which is not the same). Although my changes don't properly fix the issue...

The `rows` array contains all rows to be rendered in the data items block, which don't necessarily map 1-to-1 onto our list of items.

I've pulled the example into its own temporary debug sheet [here](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=968442221) to get clearer console logs, adding in some static content rows within the data items loop.

Screenshot 2024-03-07 at 09 41 50

Here's the output using the code before I made my changes:

Screenshot 2024-03-07 at 09 41 40

Screenshot 2024-03-07 at 09 41 24

In this case we're comparing each individual template row to the array of all rows. This array has a length of 6, whereas we're only rendering content associated with 2 items. The last item gets correctly assigned `true` for `_last` because the last template row in the items loop is associated with that item, but these values will only happen to be correct if the template is set up a certain way.

chrismclarke:
Ah - I finally get it! Thanks, quite a few layers of abstraction to work through (highlights that could also be quite nice to include spec files for components that have complex behaviour, or moving the code to spec file that can be more easily tested)

To address the issue I've done the following in b042a1f:

- Update the ItemProcessor to return both the rendered rows and the item list (updated from any pipe operations such as list filter)
- Update the DataItems component to calculate the updated item index by looking up the included evalContext item id from the newly returned item list (I can see this is what you were doing for with your update)
- Update the temp_data_items demo to include more explicit example (thanks for adding)
- Improve type definitions and comments

I think now things should finally be working. See updated demo below, where _first and _last update and are both enabled as true for items lists that are filtered to single entries

[Screenity video - Mar 7, 2024.webm](https://github.com/IDEMSInternational/parenting-app-ui/assets/10515065/bc5fd553-7381-4fbc-8998-c0aa30eedee9)

jfmcquade (approved):
Looks good, I think the code is clear and the comments are helpful.

I've moved the example back to the [comp_data_items](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=845703892) sheet as Example 5 (and deleted `temp_data_items`). I really like the example, a satisfying way of demonstrating the first/last functionality. I'll update the PR description to include the up-to-date demo.

esmeetewinkel (approved):
Functional test passed.

PR #2221, opened 2024-03-05
fix: data-items cdr

chrismclarke:
PR Checklist

## Description
Fix issue where templates rendered from data-items would not detect changes. All template container elements now include a call to mark changes when child content rendered.

I assume the issue is due to some way by which angular cdr limits change detection for child components from a parent with onPush strategy (haven't tracked down source reference for this, just a hunch).

We could also consider moving all components (notably template and container) to onPush strategy as well, although given the potential for highly breaking changes might make more sense to wait until we implement an alternative rendering system.

Issue discussion in #2215:
https://github.com/IDEMSInternational/parenting-app-ui/pull/2215#discussion_r1508967183

Debug sheet:
https://docs.google.com/spreadsheets/d/14XXwZ7R_7qIBo5xKYuxnPVQbQidI--PPJ7TVvZxwy1k/edit#gid=1422144003

## Git Issues

Closes #

## Screenshots/Videos

[Screenity video - Mar 5, 2024.webm](https://github.com/IDEMSInternational/parenting-app-ui/assets/10515065/33a22a10-785b-46d9-b1eb-484a6ccf640c)

### DIFF
--- a/src/app/shared/components/template/template-container.component.ts
+++ b/src/app/shared/components/template/template-container.component.ts
@@ -1,4 +1,5 @@
 import {
+  ChangeDetectorRef,
   Component,
   ElementRef,
   EventEmitter,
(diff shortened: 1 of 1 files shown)


jfmcquade (approved):
Great, thanks, this does appear to fix the issue

PR #2222, opened 2024-03-06
fix: changeDetectorRef error

jfmcquade:
PR Checklist

- [x] PR title descriptive (can be used in release notes)

## Description

Following #2221, possible fix for issue identified [on mattermost](https://chat.idems.international/all/pl/persax3scjgajnbnoy5wp768fc).

## Git Issues

Closes #

## Screenshots/Videos

_If useful, provide screenshot or capture to highlight main changes_

PR #2223, opened 2024-03-06
fix: cdr injection

chrismclarke:
PR Checklist

## Description

Fix regression from #2221 where calls to change detection mechanisms fail for popup templates.

The likely reason for this is the changeDetector is being called during service constructor methods, which can happen before the app has had a chance to inject the changeDetector (likely happens after all other constructors). 

This fix defers retrieval until after explicitly required for a template, and changes the `emitDecoratorMetadata` tsconfig setting which has been linked to [similar issues](https://github.com/search?q=emitDecoratorMetadata+repo%3Aangular%2Fangular&type=issues&s=created&o=desc) and removed in default angular starter templates (although I think it now defaults to `false` anyway when using esbuild, but possibly not webpack compiler)

## Review Notes
Should check before/after running any deployment that processes startup templates

## Git Issues

Closes #2222 (previous fix attempt)

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/template-container.component.ts
+++ b/src/app/shared/components/template/template-container.component.ts
@@ -60,19 +60,21 @@ export class TemplateContainerComponent implements OnInit, OnDestroy, ITemplateC
   templateBreadcrumbs: string[] = [];
   private componentDestroyed$ = new Subject<boolean>();
   debugMode: boolean;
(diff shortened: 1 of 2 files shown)


esmeetewinkel (approved):
Functional test passed

PR #2230, opened 2024-03-11
chore: update Firebase BoM; refactor: auth actions

### LINKED ISSUE #2133: Warning: SafetyNet Attestation API discontinued
**Details**
When creating a Play Store Release I'm getting the error below, which @ChrisMarsh82 said he wanted to speak to @chrismclarke about.

**Screenshots and videos**
image

jfmcquade:
PR Checklist

## Description

Updates the Android Firebase library versions, in an attempt to address the warning in #2133. 

Also refactors the auth actions to use a new `auth` namespace:
- `google_auth` action is replaced by `auth: sign_in_google`
  - `google_auth` syntax is still supported for compatibility, but should be considered deprecated
- Adds a new action: `auth: sign_out`, which signs the user out through Firebase. Currently the only use case for this is debugging

See [example_google_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit#gid=579616705) template below for the actions in use.

## Git Issues

Closes #2133

### Demo of [example_google_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit#gid=579616705) template:

Screenshot 2024-03-13 at 13 38 36

The initial behaviour on clicking the "Sign in with Google" button was confusing to me (see demo below), especially as it didn't match what I observed when testing the feature as part of reviewing #2517 (see screen recording in [this comment](https://github.com/IDEMSInternational/parenting-app-ui/pull/2157#pullrequestreview-1759676894)). This meant that I spent much longer than I'd like to admit trying to debug the seemingly broken `signInWithGoogle` functionality, finding it to be also "broken" on master. In fact, the feature was working as expected, but I was already logged in on my emulator and there is no confirmation that the user is logged in. Whilst this lack of feedback is still an issue (although the signed in user is stored in a field), I was able to debug by adding a new `auth_sign_out` action, which allows me to test the Google auth functionality. The console logs in the demo below are temporary debug logs, now removed.

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/1dc097e9-34dc-4d87-9bbf-91c71126e666

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -43,8 +43,6 @@ dependencies {
     // When using the BoM, you don't specify versions in Firebase library dependencies
     implementation 'com.google.firebase:firebase-crashlytics'
     implementation 'com.google.firebase:firebase-perf'
-    // https://github.com/mixpanel/mixpanel-android/issues/744
-    implementation("com.google.firebase:firebase-iid")
 }
 
 apply from: 'capacitor.build.gradle'
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -37,7 +37,7 @@ dependencies {
     implementation project(':capacitor-cordova-android-plugins')
 
     // Import the BoM for the Firebase platform
-    implementation platform('com.google.firebase:firebase-bom:29.3.0')
+    implementation platform('com.google.firebase:firebase-bom:32.7.2')
 
     // Declare the dependencies for the Crashlytics and Analytics libraries
     // When using the BoM, you don't specify versions in Firebase library dependencies
--- a/src/app/shared/components/template/services/instance/template-action.service.ts
+++ b/src/app/shared/components/template/services/instance/template-action.service.ts
@@ -226,10 +226,6 @@ export class TemplateActionService extends SyncServiceBase {
         );
         const processor = new TemplateProcessService(this.injector);
         return processor.processTemplateWithoutRender(templateToProcess);
-      case "google_auth":
-        return await this.authService.signInWithGoogle();
-      case "auth_sign_out":
-        return await this.authService.signOut();
       case "emit":
         const [emit_value, emit_data] = args;
         const container: TemplateContainerComponent = this.container;
--- a/src/app/shared/components/template/services/instance/template-action.service.ts
+++ b/src/app/shared/components/template/services/instance/template-action.service.ts
@@ -228,6 +228,8 @@ export class TemplateActionService extends SyncServiceBase {
         return processor.processTemplateWithoutRender(templateToProcess);
       case "google_auth":
         return await this.authService.signInWithGoogle();
+      case "auth_sign_out":
+        return await this.authService.signOut();
       case "emit":
         const [emit_value, emit_data] = args;
         const container: TemplateContainerComponent = this.container;
--- a/src/app/shared/services/auth/auth.service.ts
+++ b/src/app/shared/services/auth/auth.service.ts
@@ -30,11 +30,11 @@ export class AuthService extends SyncServiceBase {
   }
 
   public async signInWithGoogle() {
-    return FirebaseAuthentication.signInWithGoogle();
+    return await FirebaseAuthentication.signInWithGoogle();
   }
 
   public async signOut() {
-    return FirebaseAuthentication.signOut();
+    return await FirebaseAuthentication.signOut();
   }
 
   public async getCurrentUser() {
--- a/src/app/shared/services/auth/auth.service.ts
+++ b/src/app/shared/services/auth/auth.service.ts
@@ -30,11 +30,11 @@ export class AuthService extends SyncServiceBase {
   }
 
   public async signInWithGoogle() {
-    return await FirebaseAuthentication.signInWithGoogle();
+    return FirebaseAuthentication.signInWithGoogle();
   }
 
   public async signOut() {
-    return await FirebaseAuthentication.signOut();
+    return FirebaseAuthentication.signOut();
   }
 
   public async getCurrentUser() {
--- a/src/app/shared/services/auth/auth.service.ts
+++ b/src/app/shared/services/auth/auth.service.ts
@@ -6,6 +6,7 @@ import { filter } from "rxjs/operators";
 import { IAppConfig } from "../../model";
 import { AppConfigService } from "../app-config/app-config.service";
 import { SyncServiceBase } from "../syncService.base";
+import { TemplateActionRegistry } from "../../components/template/services/instance/template-action.registry";
 
 @Injectable({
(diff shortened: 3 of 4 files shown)


chrismclarke (commented):
Nothing really to review in the code, have you tried running on a device/deployment that accesses native features?

android/app/build.gradle
```diff
@@ -37,14 +37,12 @@ dependencies {
...
     // Import the BoM for the Firebase platform
-    implementation platform('com.google.firebase:firebase-bom:29.3.0')
+    implementation platform('com.google.firebase:firebase-bom:32.7.2')
 
     // Declare the dependencies for the Crashlytics and Analytics libraries
     // When using the BoM, you don't specify versions in Firebase library dependencies
     implementation 'com.google.firebase:firebase-crashlytics'
     implementation 'com.google.firebase:firebase-perf'
-    // https://github.com/mixpanel/mixpanel-android/issues/744
-    implementation("com.google.firebase:firebase-iid")
```
jfmcquade, line 47: `firebase-iid` is deprecated as of `firebase-bom` v31.0.0, see [release notes](https://firebase.google.com/support/release-notes/android#bom_v31-0-0). As far as I can tell, we don't currently use mixpanel, so it seems that we can just remove this declaration?

jfmcquade:
@chrismclarke I've updated the PR description with the results of my testing the Firebase Auth functionality. I needed to remove a deprecated dependency in order to get gradle to build the project, see inline comment.

Do you think we need to extensively test all other native features, or should we be reasonably safe to merge?

android/app/build.gradle
chrismclarke: Judging by the previous comment we have never used mixpanel but had to include the the package because of an error being triggered. Hopefully should work now without (don't see anything in the issue updates but will try a quick test locally)

src/app/shared/components/template/services/instance/template-action.service.ts
```diff
@@ -228,6 +228,8 @@ export class TemplateActionService extends SyncServiceBase {
         return processor.processTemplateWithoutRender(templateToProcess);
       case "google_auth":
         return await this.authService.signInWithGoogle();
+      case "auth_sign_out":
+        return await this.authService.signOut();
```
chrismclarke, line 232: nit(non-blocking) appreciate this is likely just to use with some testing templates, but I think might be good to create a follow-up issue to tidy this up a bit. I'd say ideally `google_auth` action should just be replaced by an `auth` action namespace that includes both `sign_in` and `sign_out` child actions.

chrismclarke:
> @chrismclarke I've updated the PR description with the results of my testing the Firebase Auth functionality. I needed to remove a deprecated dependency in order to get gradle to build the project, see inline comment.
> 
> Do you think we need to extensively test all other native features, or should we be reasonably safe to merge?

Thanks for testing auth and catching the issue, looking through the changelog that's the only breaking change I can identify as likely to impact, however still worth very quickly testing all native firebase sdks. As it stands that is just:

**Authentication** 
Tested working with new BOM

**Crashlytics**
Not exposed in authoring so I just hardcoded a quick exception to log at startup and can confirm working as expected
[Crashlytics Link](https://console.firebase.google.com/u/0/project/plh-teens-app1/crashlytics/app/android:international.idems.plh_teens/issues/b2727abf3739a2171944230bb0a723df?time=last-hour)
image

**Performance**
Not exposed, a placeholder `PerformanceMonitorService` was created when we wanted to analyze some app startup issues people were having but seems like never actually used, so I'd say can ignore and decide in future whether to implement or remove)

So I'd say should be good to go

src/app/shared/services/auth/auth.service.ts
```diff
@@ -30,11 +30,11 @@ export class AuthService extends SyncServiceBase {
   }
 
   public async signInWithGoogle() {
-    return FirebaseAuthentication.signInWithGoogle();
+    return await FirebaseAuthentication.signInWithGoogle();
```
chrismclarke, line 33: nit(non-blocking) returning the function call already passes the promise, so adding the `await` here doesn't actually achieve anything (only useful in a try-catch statement)
https://blog.battlefy.com/when-to-use-return-vs-await-vs-return-await-vs-nothing

src/app/shared/services/auth/auth.service.ts
```diff
@@ -30,11 +30,11 @@ export class AuthService extends SyncServiceBase {
...
   }
 
   public async signOut() {
-    return FirebaseAuthentication.signOut();
+    return await FirebaseAuthentication.signOut();
```
chrismclarke, line 37: nit(non-blocking) same as above

chrismclarke (approved):
See minor non-blocking comments inline, otherwise I'd say good to go

jfmcquade:
Thanks @chrismclarke, I've updated to include a new `auth` action namespace in this PR, as I didn't like the idea of putting in a new action with a poor syntax.

esmeetewinkel (approved):
Thanks Johnny. I read through the PR notes and noted the change in action naming.

I was wondering whether `click | auth: sign_in_google` and `click | auth: sign_out` is inconsistent (i.e. whether it should be `sign_out_google`). Is it true that, if in future we would have multiple sign in platforms (Google, email, some Apple platform) then the sign ins for the different platforms would be separate actions whereas `sign_out` signs the user out of any platform they may be signed in to?

jfmcquade:
> I was wondering whether `click | auth: sign_in_google` and `click | auth: sign_out` is inconsistent (i.e. whether it should be `sign_out_google`). Is it true that, if in future we would have multiple sign in platforms (Google, email, some Apple platform) then the sign ins for the different platforms would be separate actions whereas `sign_out` signs the user out of any platform they may be signed in to?

This inconsistency does reflect a genuine difference: it is not possible to be signed in by more than one method at a time, the user is either signed in (via some method) or signed out. So whilst we may end up with various sign in actions corresponding to various auth "providers", e.g. `sign_in_email`, `sign_in_apple`, there will only ever by one `sign_out` method which will sign out the user regardless of which method they used to sign in.

chrismclarke (approved):
Thanks for making the updates to the auth actions too, nice to keep the legacy for now.

Moving forwards I'm not sure what will be the best way to handle deprecations, but perhaps whenever calling something deprecated in the code we could also add a report to glitchtip or similar as a way of identifying content using legacy syntax that needs to be updated?

I would also recommend marking `@deprecated` above the `google_auth` action name partly to give visual clue in vscode, and also to make it easier to track down deprecated code in the future

PR #2233, opened 2024-03-12
style: add drop-shadow to radio button grid component items

### LINKED ISSUE #2231: [FEATURE] Radio button grid to have drop shadow
**What?**
The radio button grid should have a drop shadow. This could be an option, in which case I think it should be the default. 

**Why?**
The graphic designer is trying to make the designs consistent in the sense that 
> a component has a drop shadow if and only if it is clickable

Following that rule, the radio button grid should have a drop shadow.

**Screenshots**
The complaint is that some of the radio buttons (see below) look like images, it's not clear to users that they are clickable.
https://plh-kids-tz.web.app/template/article_connect_learn_6
image

jfmcquade:
PR Checklist

## Description

Adds a drop-shadow effect to the clickable items that appear within the radio button grid component, to match those of the radio group component.

This effect is hardcoded and editable at the theme level, as opposed to being exposed as an authorable parameter on the component.

## Git Issues

Closes #2231

## Screenshots/Videos

[comp_radio_button_grid](https://docs.google.com/spreadsheets/d/1yHLQHxZQuxs34gBU-ny95ZWs9e8-HybC4IAeFxwpcVw/edit#gid=569531329) template

| Before | After |
|-------|-------|
| Screenshot 2024-03-12 at 15 25 25 | Screenshot 2024-03-12 at 15 22 00 |

### DIFF
--- a/src/app/shared/components/template/components/radio-button-grid/radio-button-grid.component.scss
+++ b/src/app/shared/components/template/components/radio-button-grid/radio-button-grid.component.scss
@@ -19,6 +19,7 @@ $text-color: var(--radio-button-font-color, var(--ion-color-primary));
   padding: 8px;
   background: white;
   cursor: pointer;
+  box-shadow: var(--ion-default-box-shadow);
 }
 .item[data-selected="true"] {
   background: $background-selected;


chrismclarke (approved):

esmeetewinkel (approved):
Functional test passed

image

Issue #2241, opened 2024-03-18
[FEATURE] Handle iOS app icon generation


jfmcquade:
**What?**
Follow-up to #2234

It would probably make sense to address this and #2072 together.

esmeetewinkel:
This also includes splash screen image, currently showing on Appetize as 
<img width="158" alt="image" src="https://github.com/IDEMSInternational/plh-facilitator-app-mx-content/assets/74557272/6e8f4924-2547-4967-9a71-377974ff49a2">

jfmcquade:
See [this comment](https://github.com/IDEMSInternational/open-app-builder/pull/2836#issuecomment-2752318603) on #2836 regarding incorporating this step into an ios build action

jfmcquade:
Work has begun on this, see WIP issue #3087. Has currently been pushed down the priority list.

PR #2251, opened 2024-03-20
feat: env-replace workspace

chrismclarke:
PR Checklist

## Description

- Create new `env-replace` workspace used to manage utility methods for future template file replacement
- Add core `replaceFiles` method used to rewrite `.template` files using global environment variables
- Add basic testing suite and tests

## Dev Notes
Whilst drafting a PR to populate android and ios template files I realised the code was quickly becoming unmanageable as there are various combinations of files to convert at different times with variables read from multiple places (e.g. Android config has about 5 file replacements but some need to leave templated strings in whilst others need to replace).

To reduce complexity I've moved the core replacement utility into its own workspace, and will make a follow-up pr to call specific methods from it as required. 

The utility is designed to support various folder and filename search and exclusion patterns, as well as variable names to include/exclude. See full list of configuration variables included and test cases for more details.

Rough next steps would be

- [ ] Update deployment config to provide core variables for use in local development (e.g. APP_ID is currently is only populated using github actions, but to run scripts locally would ideally want the id as part of the deployment android config)
- [ ] Replace all hardcoded config files with templates for android, ios, capacitor and firebase configs
- [ ] Add workflows to run selective template replacements as required. E.g. `yarn workflow android setup` could be used to update all android config files and replace variables such as the appId and versionCode from templates. It could also trigger related workflows such as processing assets.
- [ ] Update github actions to call workflows as required when preparing CI builds

## Review Notes
Test suite can be run via
```sh
yarn workspace @idemsInternational/env-replace test
```

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/@idemsInternational/env-replace/src/index.ts
+++ b/packages/@idemsInternational/env-replace/src/index.ts
@@ -1,0 +1,182 @@
+import { readFileSync, writeFileSync } from "fs";
+import { GlobOptionsWithFileTypesUnset, glob } from "glob";
+import { resolve } from "path";
+
+export interface IEnvReplaceConfig {
+  cwd: string;
+
+  /** List of input folder glob patterns to include, default ['**'] */
+  includeFolders: string[];
+
+  /** Filename pattern to find */
+  fileNameFind: string;
+
+  /** String to replace filename pattern match with to output to different file, default '' */
+  fileNameReplace: string;
+
+  /** Additional variables to populate to env. Will be merged with global env */
+  envAdditional: Record<string, any>;
+
+  /** List of folder glob patterns to exclude,  */
+  excludeFolders: string[];
+
+  /** Specify list of variable names to exclude, default includes all */
+  excludeVariables: string[];
+
+  /** Specify list of variable names to include, default includes all */
+  includeVariables: string[];
+
+  /** Override generated globs from input folder and filename patters */
+  rawGlobInput: string;
+
+  /** Additional options to provide directly to input glob match */
+  rawGlobOptions: GlobOptionsWithFileTypesUnset;
+
+  /** Throw an error if environment variable is missing. Default true */
+  throwOnMissing: boolean;
+}
+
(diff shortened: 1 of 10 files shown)


packages/@idemsInternational/env-replace/src/index.ts
```diff
@@ -0,0 +1,182 @@
...
+
+      // handle replacement
+      const outputName = inputName.replace(fileNameFind, fileNameReplace);
+      this.summary[outputName] = {};
+
+      const replaceContent = this.generateReplaceContent(outputName, sourceContent, replacementEnv);
+      const outputPath = resolve(cwd, outputName);
+      writeFileSync(outputPath, replaceContent);
+    }
+    return this.summary;
```
jfmcquade, line 116: It could be nice to log the paths of the output files here (the keys of `this.summary`) to see what's been generated? Although these files won't necessarily have changed, so perhaps git diffs are a better place to view changes

jfmcquade (approved):
Really nice, and good to have as a standalone package. 

From the proposed next steps:
>  Add workflows to run selective template replacements as required. E.g. `yarn workflow android setup` could be used to update all android config files and replace variables such as the appId and versionCode from templates. It could also trigger related workflows such as processing assets.

I like this idea, I think it would make a lot of sense to include multiple workflows such as generating the native icon/splash assets in a command like that, I will keep in mind when working on #2241. 
I suppose it might be more user friendly to use a syntax like `yarn workflow setup_native --android --ios`, where the flags are optional (and default is both). But as we already have the `android` workflow namespace it's probably not worth refactoring, we can add an equivalent `ios` one.

packages/@idemsInternational/env-replace/src/index.ts
chrismclarke: I had previously included a table log, although was concerned whether github actions would still be able to correctly mask any variables populated from secrets if logging after the various manipulations. No harm including as an optional config param to test later however, I'll include the update with next pr (screenshot below for reference)

image

PR #2167, opened 2023-12-13
Feat: PfR theme

jfmcquade:
PR Checklist

- [x] PR title descriptive (can be used in release notes)

## Description

Adds a new theme specifically for the PfR deployment. 

The new theme uses the colours black and red as the two source colours from which the theme is generated. It also uses the specific shade of green, taken from the logo, as the "success" green colour used throughout the app (e.g. for the green ticks on completed modules). I did experiment with other combinations of the colour scheme, but these resulted in red and green being too prominent, which looks odd but also conflicts with the standard use of red and green as "success" and "warning" colours in apps (although currently we don't use red anywhere by default). See the screenshots at the very bottom of the PR description for the rejected colour schemes.

It may be that further tweaks are necessary as a follow-up, either to the theme override values for specific components, or to the authoring (for example, many of the icons used are blue).

## Dev notes

In the future, deployment-specific themes (or at least the minimal data required for their generation) should live in content repos, see new issue #2168. However this is a significant task that is not currently high priority. Adding a new theme to the code repo, whilst not ideal, does not cause any issues, but it does defy the separation of code from deployment-specific data to include a named "pfr" theme in the code repo. One way around this would be to rename the theme from "pfr" to something more general (e.g. "black-red"), as in theory another deployment may wish to use this theme. However, as we're specifically using the brand colours of PfR, that feels a bit like sleight of hand, and wouldn't solve the problem.

## Testing

[This PR](https://github.com/IDEMSInternational/pfr-app-content/pull/99) in the PfR repo should be merged before testing, as the config has been updated to target the `pfr` theme.

After checking out the PfR deployment and pulling the latest changes, run `yarn start` as usual. Initially, the theme changes will not be visible because the "default" theme, which is available as a fallback, will be applied if it has previously been set as the app's active theme. In order to update the theme, you will need to reset the app using the "Reset app" button in the main menu. This will clear any cached data in your browser associated with the PfR app, giving you the experience of someone running the app for the first time. You should then see the changes to the visual styles.

## Git Issues

Closes #2165

### Before
Screenshot 2023-12-13 at 16 44 45

Screenshot 2023-12-13 at 16 50 52

### After
Screenshot 2023-12-14 at 11 37 28

Screenshot 2023-12-14 at 11 38 44

### Rejected colour schemes

Screenshot 2023-12-14 at 11 42 09

Screenshot 2023-12-14 at 11 41 12

PR #2259, opened 2024-03-26
Feat: Add PLH Facilitator MX theme

esmeetewinkel:
PR Checklist

## Description

Attempt to create a specific theme for the PLH Facilitator MX deployment, using the partner's branding colours. 

I followed the changes made in the PR creating the PfR theme #2167
image

@jfmcquade I was wondering what decides that the text on the header and buttons shows in black rather than white? 
image

## Git Issues

Closes https://github.com/IDEMSInternational/plh-facilitator-app-mx-content/issues/53


chrismclarke:
In case of interest, a quick look at [webaim contrast checker](https://webaim.org/resources/contrastchecker/) gives the following scores for accessibility

image

If using white as the contrast color we get
image

So in either case I think the black text gives better results (passes AA standard, not AAA). Tweaking the primary color slightly (e.g. `#48A3A8`) gives a contrast ratio a little better than 7:1 (which is the threshold for AAA)

image

jfmcquade:
Thanks @chrismclarke, great to know about those resources.

@esmeetewinkel: in terms of what actually decides, there is a utility function in the file `src/theme/themes/utils/contrast-colour.scss` that runs some maths to decide whether the contrast colour should be black or white. 

I haven't looked into the details, but as it says in the comment at the top of the file, the code is taken from [this post](https://sgom.es/posts/2016-12-21-using-sass-to-automatically-pick-text-colors/). The algorithm tries to maximise the contrast ratio, as defined by the WCAG standards that Chris referenced.

Screenshot 2024-04-03 at 16 21 30

esmeetewinkel:
Opted for partner colours blue and orange (not too different from the default theme) for improved contrast. 
Before / After: 
image

image

esmeetewinkel:
See this content PR https://github.com/IDEMSInternational/plh-facilitator-app-mx-content/pull/57 for test preview link

jfmcquade (approved):
Good job figuring out what to change, it looks good to me.

> NB. I thought this round and rectangular button had the same colour, but in this changed theme it is apparent that they are slightly different. Maybe they should be the same? To me the difference doesn't look intentional. @jfmcquade

Good spot. The issue is that one of the buttons has colour `--ion-color-secondary` (via a theme-customisable property `--button-background-secondary`), whilst the other has colour `--ion-color-secondary-400`. Because of the way our colour generating system works, there will always be one colour "weight" that closely resembles the input `secondary` colour, and therefore may appear oddly similar. This will vary depending on the relative lightness of that original input colour, so it's a bit hard to account for cases like these. 

In this instance, I'm pretty sure this could be fixed by removing the `style: orange` parameter in cell E43 of the [parent_profile](https://docs.google.com/spreadsheets/d/1VQTZglKW0oRPXxncnDyix_s9EfGBSZ4intIaIeXu798/edit#gid=1729633565) template. The `orange` parameter is outdated anyway (as the secondary colour could be anything, there are still a handful of variables so-named), and without it the button's background should default to the value of `button-background-secondary` as specified by the theme.

esmeetewinkel:
I'm finding that without the orange parameter the button becomes blue (primary)...

image

jfmcquade:
> I'm finding that without the orange parameter the button becomes blue (primary)...

Ah sorry, you're right. There is no default secondary option. Setting the `style: get_me_going` should work for now (although not very nice). I could make a quick PR to add a `style: background-secondary` option

esmeetewinkel:
Hahaha I had forgotten about that style naming...

I found that the `style: navigation` round button does have the colour I want, only that it's slightly smaller (not a problem) so I'll use that for now.

image

Issue #2265, opened 2024-03-27
Notifications as opt-in/out feature


esmeetewinkel:
**Details**
Currently notifications are enabled for all apps, even if they have not authored any notifications. 

This leads to the slightly jarring and confusing situation where the user is asked to accept the app sending them notifications for an app that doesn't have any notifications.

**Screenshots and videos**
On iOS:
<img width="175" alt="image" src="https://github.com/IDEMSInternational/plh-facilitator-app-mx-content/assets/74557272/be94bde3-e65f-4642-ac31-b6ade7e3fff9">

On web: 
<img width="311" alt="image" src="https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/2f42cd04-ea8f-4e86-b7cd-63f6523c6bde">

On Android:
![image](https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/d1ccefa7-a82e-4777-b0a1-280538ee03de)

esmeetewinkel:
![image](https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/5cf657b0-cce4-46d7-ae57-95b17368c958)

PR #2275, opened 2024-04-04
chore: update ios build files and template

jfmcquade:
PR Checklist

## Description

After using xcode to build and publish an iOS app to App Store Connect, some of the iOS build files were updated. This PR includes those updates in the templated file.

The specific changes enacted by xcode were:
- Adding a `ios/App/App/App.entitlements` file, and a reference to this file in `ios/App/App.xcodeproj/project.pbxproj`
- Updating the name of `App.app` to `${APP_NAME}.app` as referenced in `ios/App/App.xcodeproj/project.pbxproj`. In our case, this was updated to `Formando Conciencia.app`, showing that spaces are supported.

## Git Issues

Closes #

### DIFF
--- a/ios/App/App.xcodeproj/project.template.pbxproj
+++ b/ios/App/App.xcodeproj/project.template.pbxproj
@@ -21,7 +21,7 @@
 /* Begin PBXFileReference section */
 		2FAD9762203C412B000D30F8 /* config.xml */ = {isa = PBXFileReference; lastKnownFileType = text.xml; path = config.xml; sourceTree = "<group>"; };
 		50379B222058CBB4000EE86E /* capacitor.config.json */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.json; path = capacitor.config.json; sourceTree = "<group>"; };
-		504EC3041FED79650016851F /* App.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = App.app; sourceTree = BUILT_PRODUCTS_DIR; };
+		504EC3041FED79650016851F /* ${APP_NAME}.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = App.app; sourceTree = BUILT_PRODUCTS_DIR; };
 		504EC3071FED79650016851F /* AppDelegate.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppDelegate.swift; sourceTree = "<group>"; };
 		504EC30C1FED79650016851F /* Base */ = {isa = PBXFileReference; lastKnownFileType = file.storyboard; name = Base; path = Base.lproj/Main.storyboard; sourceTree = "<group>"; };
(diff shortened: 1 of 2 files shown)


jfmcquade:
@chrismclarke I've requested your review although I appreciate you're not really able to test, I just thought it could be good to sense-check

jfmcquade:
> I can also see `DEVELOPMENT_TEAM = CHLBG6683N;` in a couple places which I'm assuming is the appstore id for the idems developer account (?) and maybe also should be set by a variable?

That's right, and I think that setting it from a variable probably makes sense. Although it's not clear to me where the value of that variable should be specified – it seems to be semi-secret and so maybe shouldn't be in the deployment config, although I can't find a source that states that (chatGPT suggests Apple don't say so explicitly). It could just be a repo- (or organisation-) level secret/variable, but without a local counterpart it would be an additional step for devs to populate that value.

> I'm guessing there will also be some updates required when adding xcode API permissions, but happy to just merge for now in any case and iterate as we go

Yes indeed. And things could get even more complicated if we try to remove functionality from certain apps related to API permissions. We've already got a case where the `plh_facilitator_mx` deployment, which we're publishing to the app store, doesn't use notifications but still requests permissions (see #2265). I know we've previously discussed removing code that is unused by a given app in order to optimise that bundle size. We'd also eventually ideally have the relevant permissions removed from the relevant places in the iOS configuration where possible, so that we weren't requesting permissions with Apple for features that aren't used for a given deployment.

chrismclarke:
> That's right, and I think that setting it from a variable probably makes sense. Although it's not clear to me where the value of that variable should be specified – it seems to be semi-secret and so maybe shouldn't be in the deployment config, although I can't find a source that states that (chatGPT suggests Apple don't say so explicitly). It could just be a repo- (or organisation-) level secret/variable, but without a local counterpart it would be an additional step for devs to populate that value.

Yeah I think that's potentially something partially overlooked with the current variable handling system. I guess the safest way is to just use an `ios.config.json` file that is encrypted and stores any sensitive data. Alternatively it might be worth supporting a deployment-level `.env` file which can be populated from input prompts when trying to populate template files (or populate from github env when running on CI). 

Although for the specific case of development_team the following StackOverflow post suggests it's not actually all that sensitive:
https://stackoverflow.com/questions/44417171/should-i-keep-the-apple-developer-team-id-secret

> Yes indeed. And things could get even more complicated if we try to remove functionality from certain apps related to API permissions. We've already got a case where the `plh_facilitator_mx` deployment, which we're publishing to the app store, doesn't use notifications but still requests permissions (see #2265). I know we've previously discussed removing code that is unused by a given app in order to optimise that bundle size. We'd also eventually ideally have the relevant permissions removed from the relevant places in the iOS configuration where possible, so that we weren't requesting permissions with Apple for features that aren't used for a given deployment.

Yeah that does start to get awkward. I know expo uses a novel system where it automatically generates android/ios core files each time, fully populating all code from inspection of config file (a bit like how capacitor populates app names, ids and a few other things when adding platform) - however they have a whole community repo of specific plugin configurations to handle things like the relevant firebase bom as required, so maintaining our own system like that would be pretty tricky...

At least if focusing on opt-out only then we can use comments in template files to remove as necessary (so core build starts with all possible permissions and code stripped away depending on specific parameters)

Issue #2236, opened 2024-03-15
[FEATURE] - Populate reusable action config files from templates


chrismclarke:
**What?**
Some github actions use custom scripting to populate files using variables stored in github actions context. These scripts can be a little verbose and hard to adapt, and are somewhat detached from local workflows.

E.g. (https://github.com/IDEMSInternational/parenting-app-ui/blob/feat/deployment-firebase/.github/workflows/reusable-deploy-pr-preview.yml#L57-L87)

It might be more useful to implement a templating system that can be used either as part of github actions or locally to populate files as required.

**How?**
There's no shortage of tools around for templating, however preferably it would be good to find one that can be run both locally and as part of CI. Given that github actions can execute any javascript actions directly it may be preferable to use a javascript/node-based solution as opposed to one that is installed through docker, or at least a templating syntax that is interoperable. 

**Other details**
Example tools to consider:
- https://jsonnet.org/
- https://github.com/marketplace?category=&type=actions&verification=&query=sort%3Apopularity-desc+envsubst
- https://github.com/marketplace?category=&type=actions&verification=&query=sort%3Apopularity-desc+replace

Example of template file that could be used as generator (env_substr format)
```json
{
  "projects": {
    "default": "${FIREBASE_PROJECT_ID}"
  },
  "targets": {
    "${FIREBASE_PROJECT_ID}": {
      "hosting": {
        "${FIREBASE_HOSTING_TARGET}": [
          "${FIREBASE_PROJECT_ID}"
        ]
      }
    }
  },
  "etags": {}
}
```

chrismclarke:
Templating system implemented in #2251, android and ios config migrated, so all that remains should be firebase hosting. Will update issue title accordingly

PR #2276, opened 2024-04-04
fix: android workflow variables

chrismclarke:
PR Checklist

## Description

Fix issue where android workflow would try to populate firebase templates

## Review Notes
Should be able to successfully run the android workflow command with any deployment set that is configured for android
```sh
yarn workflow android
```

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/scripts/src/tasks/providers/android.ts
+++ b/packages/scripts/src/tasks/providers/android.ts
@@ -43,8 +43,9 @@ const configure = async ({ appId, appName, versionName }: IAndroidBuildOptions) 
       VERSION_CODE: versionCode,
       VERSION_NAME: versionName,
     },
-    // do not overwrite ${applicationId} variable in AndroidManifest template
-    excludeVariables: ["applicationId"],
+    // Only replace the following variables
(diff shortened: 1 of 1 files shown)


jfmcquade (approved):
Looks good and `yarn workflow android (configure)` works for me locally. It generates a `firebase.json` file but does not try to populate the `${FIREBASE_HOSTING_TARGET}` variable – should generating the `firebase.json` file ultimately be handled by a different workflow command?

chrismclarke:
> Looks good and `yarn workflow android (configure)` works for me locally. It generates a `firebase.json` file but does not try to populate the `${FIREBASE_HOSTING_TARGET}` variable – should generating the `firebase.json` file from the template ultimately be handled by a different workflow command?

Yes it should, issues previously raised as #2236, but for now shouldn't cause issue as these files are overwritten by shell scripts when running firebase deployment actions

esmeetewinkel (approved):
When running `yarn workflow android configure` I can see the file `firebase.json` gets created.

PR #2278, opened 2024-04-04
fix: test actions

chrismclarke:
PR Checklist

## Description

- Fix test-preview action to handle firebase json as required
- Deprecate unused actions
- Simplify core web-build action (not reusable version) to remove previous branch `deployment/....` logic (now handled by content repos) 
- Update test-visual action

## Review notes
Example `test-preview` [action pass](https://github.com/IDEMSInternational/parenting-app-ui/actions/runs/8562087040)
Example `test-visual` [action pass](https://github.com/IDEMSInternational/parenting-app-ui/actions/runs/8562087073/job/23464736090)

## Author notes
I've removed the functionality for automated actions based on this repo `deployment/[name]` branches - as I'm assuming all content is now managed in content repos (?). If that is the case we may also want to remove the deployment and deployment_dev branches from this repo as a follow-up

## Dev Notes
I had originally planned to try use the current set of reusable actions, essentially triggering them as part of the builder repo (and loading debug content repo), although it made things a little more complicated than I was hoping for so ended up simply refactoring the existing actions - so we still have 2 separate systems in place, one for actions related just to the builder repo and another for reusable actions triggered by content repos.

I did at least manage to remove what was essentially a third system, whereby actions could be triggered against deployment branches. I've also tried to keep it fairly similar to how variables are set on deployment repos, with the following secrets and variables currently set in a `debug` environment for this repo

image

## Git Issues

Closes #

## Screenshots/Videos


esmeetewinkel (approved):
> all content is now managed in content repos (?). We may also want to remove the deployment and deployment_dev branches from this repo as a follow-up

Yes that's correct, I'll remove these branches. Recently one such branch was useful to retrieve a previous deployed app version. But since deleted branches can be restored anyway, I don't see a problem deleting them (and having a cleaner code repo).

jfmcquade (approved):
Looks good to me. Using an environment for the secrets/variables seems like a nice pattern that simplifies the build actions and could be extended in future.

PR #2281, opened 2024-04-05
fix: iOS footer bug

### LINKED ISSUE #2243: [BUG] iOS: footer template
**Describe the bug**
Follow-up to #2234 

On iOS, the icons in the footer template disappear when navigating to some templates.

**Screenshots and videos**

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/693dc219-bab9-47f7-9811-d998befbd334

jfmcquade:
PR Checklist

## Description

Fixes an issue where the icons displayed in the footer on iOS would intermittently disappear.

## Testing

Appetize build [here](https://appetize.io/app/7d672z7p7hzmayf7nq3topocwi?device=iphone14pro&osVersion=16.2).

## Dev notes

The solution ended up being extremely simple, but I'm still not sure what was causing the issue. The width of the `.image-wrapper` element was 0 when the icons were not visible, but it is not clear to me why the width needs to be set explicitly on iOS and not on other platforms. I haven't found discussion of this issue online, but it does appear to be a genuine bug at some level, as the behaviour was inconsistent and sporadic (see video in comments of original issue).

## Git Issues

Closes #2243

## Screenshots/Videos

Demonstration of icons not disappearing in iPhone simulator. As the behaviour was that the icons would occasionally disappear, it is not possible to prove empirically that this has been solved, however they would reasonably consistently disappear when navigating between templates, so we can be confident that the issue is fixed through functional testing (and the fact that the CSS changes directly address the cause of the issue).

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/28eb47b5-2195-48ca-b935-c23d18d9f7f1


esmeetewinkel (commented):
I no longer see the footer icons disappearing 👍 

I did encounter a potentially unrelated issue: 
On the debug emulator build that you link in this PR description, the key board does (most of the time?) not pop up properly on a text box component or on the template list search box

[Screenity video - Apr 5, 2024.webm](https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/5ba5d8db-14e3-404b-ad50-281fa8e187ae)

I'm not sure if this is introduced by this PR. Comparing this against an iOS [build ](https://appetize.io/app/mwyz2tx457dv55dw4eye4z2dpi?device=iphone14pro&osVersion=16.2)(plh_facilitator_mx) you produced yesterday, it seems that the key board does work properly there...
image

esmeetewinkel (approved):
Correction to my previous review: 

I did also encounter the inconsistent keyboard behaviour on yesterday's release, so the issue is not caused by this PR. 

I'll create a separate issue, and approve this.

chrismclarke:
weird indeed, although I do wonder if it might also be linked to the other element positioning bugs... I would probably suggest you add a quick comment in the style code to the issue as it's the type of thing I could imagine another developer viewing as 'unnecessary' and removing in the future

PR #2288, opened 2024-04-10
fix: iOS popup close button behaviour

### LINKED ISSUE #2262: [BUG] iOS: toggle-bar in button, card-portrait variant, appears outside of element
**Describe the bug**
On iOS, the toggle-bar component that should display in the top right corner of a button with `variant: card-portrait` instead appears at the very top of the template content, outside of the button element.

(from [this comment](https://github.com/IDEMSInternational/parenting-app-ui/pull/2250#pullrequestreview-1961207073))
Screenshot 2024-03-27 at 15 56 06

**App version**
0.16.26

**Template Name**
[comp_button](https://docs.google.com/spreadsheets/d/1OmgZICjM5EMT1KgLOU_ovDljRF_SPlpQAgkKWPrNX0s/edit#gid=569531329)

Screenshot 2024-03-27 at 16 01 59

### LINKED ISSUE #2274: [BUG] iOS: popup close button behaviour
**Describe the bug**
On iOS, the popup close button is obscured behind the popup itself when the popup content is large enough to be scrollable. 

This seems related to #2243 in that it can be resolved by toggling seemingly any CSS rule on the element off and then on again.

**To Reproduce**
I have observed this when running in a local simulator, I am yet to observe it when runnin gin appetize or on native device, but haven't tested extensively

**Screenshots and videos**

Using the popup opened by Button 3 on the [example_pop_ups](https://docs.google.com/spreadsheets/d/1K9JZ9Glnj73CFJmpfGIRa5dRA2TBxSsJmaguV6GL8Pc/edit#gid=402992224) template.

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/e575b5d7-cc03-4b9d-a789-7f36048f2cc8

### LINKED ISSUE #2287: [BUG] iOS button shadow
**Describe the bug**
card style button shadow is inconsistent and/or intersects the button, depending on scrolling behaviour.

**App version**
v0.16.29 iOS 

**To Reproduce**
Access any article template on PLH Facilitator MX

**Screenshots and videos**
image

jfmcquade:
PR Checklist

## Description

Resolves a bug where the popup close button would appear behind the popup window in the case that the popup contained scrollable content.

## Testing

Appetize build available [here](https://appetize.io/app/7d672z7p7hzmayf7nq3topocwi?device=iphone14pro&osVersion=16.2).

## Dev notes

A bug on iOS causes some CSS rules to not be applied to some elements under certain conditions, manifesting in the bugs detailed in the linked issues. This PR provides a workaround in the form of a CSS mixin that can be applied to any problematic element, within a new `overrides.ios.scss` file. Originally, this approach was used to provide a workaround for issues #2262 and #2287. These have both now been closed by standalone PRs which address their specific issues more directly: #2289 and #2302 respectively.

## Git Issues

Closes #2274 

## Screenshots/Videos

The close button on a scrollable popup rendering correctly, resolving #2274 ([example_pop_ups](https://docs.google.com/spreadsheets/d/1K9JZ9Glnj73CFJmpfGIRa5dRA2TBxSsJmaguV6GL8Pc/edit#gid=402992224) template)
Pop-up


chrismclarke (changes requested):
Nice you were able to figure out a workaround, although still a bit strange as I thought most of these type of hacks usually effected animations more than static content, but guessing it has some knock-on here for the root issue too.

Given that these hacks are currently being applied to all devices/browsers it might be a bit cleaner to create a separate `overrides.ios.scss` file that keeps all of these in one clear file instead of spread out across various components. 

This would likely require populating a data-attribute to the document body when the main component loads to reference the platform `<body data-platform='ios'>` (reading the platform from capacitor).

That way the override file can scope accordingly, e.g. 
```scss
body[data-platform='ios']{
   ion-button[data-variant~="card"]{
       @include mixins.hack-webkit-reset;
 }
}
```

I'd also probably recommend renaming the mixing to be more illustrative of what it does e.g. `force-gpu-accelearation`

chrismclarke:
Also interestingly, I'm now running the webkit browser using playwright and can replicate a bunch of the ion-toggle issues but I don't see the browser popup close (so that might be mobile-browser specific? - did you ever clarify which issues were on safari desktop as well as mobile app?)...

I'm also noticing a bunch of these issues relate specifically to `ion-toggle` but we seem to be using a legacy version of the component (various deprecated upgrade warnings) - it might be worth trying to update these to see if it fixes anything without the added workarounds

chrismclarke:
Looking more closely at specific `comp_button` - it still feels to me more an issue with how the button is coded. We're using absolute positioning to nest child rows however not explicitly stating what the absolute is relative too (should be parent container), so that leaves a bit of ambiguity to be filled in.

chrismclarke:
@jfmcquade 
See https://github.com/IDEMSInternational/parenting-app-ui/pull/2289
I think it might be a more concrete solution (although will need more work to address other components that have similar problems).

chrismclarke:
> I'll see again if I can address the issues with the other two components directly with that in mind, otherwise I will implement the refactoring of this PR that you suggest above and will suggest that we go ahead with this as a workaround.

Yeah the box-shadow issues seem very weird generally, although maybe it's the background color getting 'lost' somehow (and hardcoding might fix?). I can't replicate either of the outstanding issues on webkit browser so possibly is more a mobile browser issue, which may be fixed through a combination of tidying up ambiguous style rules and enabling hardware acceleration.

I think it would still be preferable if we can fix via regular CSS as it does seem that GPU acceleration may have trade-offs:
https://mercimichel.medium.com/don-t-abuse-enabling-gpu-acceleration-on-ios-95520da2428

But I expect for scroll-specific bugs it may well be the only way to fix

esmeetewinkel (approved):
Functional test passed on Appetize link

src/app/app.component.ts
```diff
@@ -103,6 +109,7 @@ export class AppComponent {
 
   private async initializeApp() {
     this.platform.ready().then(async () => {
+      this.platforms = this.platform.platforms().join(" ");
```
jfmcquade, line 112: Originally I planned to use `Capacitor.getPlatform()`, which just returns "web", "ios" or "android". But there are two potential reasons to prefer `platform.platforms()`:
1. In the future we may want to to extend our platform specific CSS to apply rules based on platform types within this hierachy (e.g. "mobile")
2. When running on a web browser with device mode active, `Capacitor.getPlatform()` returns "web", but `platform.platforms()` returns a list representing the emulated device. E.g. in Chrome with device mode active and set to iPhone SE:
    Screenshot 2024-05-08 at 16 06 57

    - This seems desirable as it allows us to test functionality when running on web, although there may be issues with this that would make the simple platform more suitable

jfmcquade:
@chrismclarke I've requested a re-review as I've refactored according to your suggestions, and updated the PR description to explain that this workaround fix is now applied in just one case, but could be extended to cover other similar cases as they occur.

src/app/app.component.ts
chrismclarke: Makes a lot of sense, also appreciate storing it as string instead of array to use with the css selectors and providing the option for various levels of platform-specific override.

chrismclarke (approved):
Thanks @jfmcquade - looks really good to me, appreciate being able to just apply the styles for a specific platform subset and the separation of core from override scss.

I haven't tested anything on ios, but happy to merge here and just review any further follow-ups for specific issues as they come up

PR #2289, opened 2024-04-10
fix: button child layout positioning

### LINKED ISSUE #2262: [BUG] iOS: toggle-bar in button, card-portrait variant, appears outside of element

chrismclarke:
PR Checklist

## Description

Inconsistencies have been noticed to the way different browsers handle toggle element nested inside button for `comp_button` template. This is likely due to different ways the `position: absolute` property of the rendered children span is interpreted (this should be nearest element that has relative positioning, it appears chrome stops if it doesn't find an element when reaching shadow-dom parent component whilst webkit continues further up dom root).

This PR explicitly sets relative positioning on the button container element to ensure children are rendered with correct positioning

## Review Notes
As the bug has only been noticed on ios/safari browser, testing on windows should be done via the playwright test framework, accessing the underlying webkit browser
```
npx playwright install
npx playwright wk http://localhost:4200/template/comp_button
```

## Git Issues

Closes #2262

## Screenshots/Videos

**Before** - Portrait card toggle positioned in absolute top-right of screen
image

**After** - Portrait card toggle positioned correctly
image

### DIFF
--- a/src/app/shared/components/template/components/button/button.component.html
+++ b/src/app/shared/components/template/components/button/button.component.html
@@ -7,6 +7,7 @@
     (click)="triggerActions('click')"
     [attr.data-param-style]="params.style"
     [attr.data-variant]="params.variant"
+    [attr.data-has-children]="_row.rows ? true : false"
   >
     <ion-icon *ngIf="params.icon" slot="start" src="{{ params.icon | plhAsset }}"></ion-icon>
     <span
@@ -32,6 +33,7 @@
     class="button-container"
     (click)="triggerActions('click')"
     [attr.data-variant]="params.variant"
+    [attr.data-has-children]="_row.rows ? true : false"
   >
     <img *ngIf="params.image" src="{{ params.image | plhAsset }}" />
     <div *ngIf="_row.value" [class]="'button-text ' + params.textAlign">
--- a/src/app/shared/components/template/components/button/button.component.scss
+++ b/src/app/shared/components/template/components/button/button.component.scss
@@ -11,6 +11,10 @@ $background-secondary: var(--button-background-secondary, var(--gradient-seconda
     display: contents;
   }
 }
+// ensure relative position on parent if absolute position children nested inside
+[data-has-children] {
+  position: relative;
+}
 ion-button {
   font-size: var(--buttons-font-size-small);
   font-weight: var(--font-weight-standard);


jfmcquade (approved):
Great, I'm pleased you managed to get to the bottom of this. And good to have a method of testing on a webkit browser, although this was the only one of the three bugs referenced in #2288 that I could recreate on desktop Safari, the other two were only present on iOS safari/native as far as I can tell.

I can confirm that the bug is fixed when running on an iOS simulator.

Screenshot 2024-04-11 at 16 56 14

See the discussion on #2288.

chrismclarke:
> Great, I'm pleased you managed to get to the bottom of this. And good to have a method of testing on a webkit browser, although this was the only one of the three bugs referenced in #2288 that I could recreate on desktop Safari, the other two were only present on iOS safari/native as far as I can tell.

Thanks, yeah I think actually quite good to try and decouple these issues instead of fixing altogether as I have a feeling there's a few issues going on under the hood (as evidenced by the fact some appear in ios and others don't). From what we've unearthed so far I'm guessing a mix of browser CSS spec interpretation/ambiguous component css, browser bugs and other hardware issues (e.g. hardware acceleration).  

So after this I'd probably recommend checking similar components that have absolute positioning layout (ideally all elements that refer to `position:absolute` should also have a parent component with `position:relative` to position in relation to. Unfortuantely this appears to be quite a lot so might just be something we pick up case-by-case as/when things come up
image

Otherwise it might be worth considering updating our visual test system to include other browsers (worth a follow-up issue/discussion?) - This wouldn't be totally trivial as I don't think pupetteer headless browser runner supports webkit, so would mean migrating code to use playwright instead.

PR #2291, opened 2024-04-11
feat: allow style_list to apply directly to text in text component

jfmcquade:
PR Checklist

## Description

There has been an author request to change the colour of text that appears in a text component.

A thorough solution would involve expanding our WIP variant/style system (#1971), e.g. by adding a general style that can be applied via a parameter to any component, such as `style: color-secondary-500`. Currently, this is not a priority and we don't have the capacity to implement such a system properly at this time. Another option would be to add a new property on the component, e.g. `colour`(/`color`), that could take values such as `primary`, `secondary`, `white`, `black` (encouraged) and potentially also `green` and `#32a852` (discouraged/undocumented). But implementing that would be in conflict with our plan for a more general variant/style system.

This PR implements an alternative solution: the rules specified in the `style_list` are now applied such that they should be reflected in the appearance of the actual text itself (as opposed to the enclosing row container). This feature, or something like it, is included in our plans for the variant/style system, with this functionality rolled out to all components in a custom way.

## Git Issues

Closes #

## Screenshots/Videos

Updated comp_text component:

Screenshot 2024-04-11 at 18 25 11

Screenshot 2024-04-11 at 18 24 32

[debug_text_style_list](https://docs.google.com/spreadsheets/d/1b0FK3Eggz76H_C9lRdFVVLUxUBEZPruoyaF0d6YeWSI/edit#gid=1520315646) component demonstrating that applying a value for the `flex` property in the style list functions as expected (dev note: the style is applied both to the row container element as before, and additionally to the `<div>` that contains the text html. The latter has no effect since the div is not inside a flexbox)

Screenshot 2024-04-11 at 18 25 47

Screenshot 2024-04-11 at 18 09 49

### DIFF
--- a/src/app/shared/components/template/components/text.ts
+++ b/src/app/shared/components/template/components/text.ts
@@ -10,6 +10,7 @@ import { getStringParamFromTemplateRow } from "../../../utils";
     [class]="style"
     [ngStyle]="isFalsy ? { display: 'none' } : { display: 'block' }"
     [innerHTML]="type === 'numbered' ? (_row.value | number) : (_row.value?.toString() | markdown)"
+    [style]="_row.style_list | styleList"
     [style.textAlign]="textAlign"
   ></div> `,
   styleUrls: ["./tmpl-components-common.scss"],


chrismclarke:
I think exposing custom css overrides at component level makes sense, much better than adding arbitrary custom styles. For the text component this is pretty intuitive as there's only a single component so I think the solution makes sense (ignoring the slightly messy implementation of the text component more generally). 

When thinking about other row types which have multiple inner components we will need a better way to signify what part to style, e.g. `text_style`, `container_style` etc. I'm assuming all this is captured in one of the working documents somewhere...

PR #2301, opened 2024-04-23
feat: store first app launch value in field

### LINKED ISSUE #2295: Identify users who reinstalled the app in postgres data
**Details**
On PLH Teens TZ, there are a couple of users whose app launch count reduced (which should not be possible). The hypothesis is that they cleared their app cache/data, potentially by uninstalling and reinstalling the app. There is no way to verify this assumption in the postgres data at the moment: 
- We do have a `CreatedAt` timestamp in the postgres data, but this is unchanged for these users so cannot be used to verify the hypothesis. This is expected, since it is the timestamp in which the user ID was first created in the postgres table. 
image
- We do have an `app_events_summary` key visible on the web app, which contains a `first_app_launch` timestamp, but this variable doesn't seem to make its way to postgres. 
   image
   I think this key is legacy (one of its properties is `app_day` which was something specific to one deployment that was discarded a long time ago) and can be removed.

I could already store a `first_app_launch` contact field through the templates, but since this could be useful for debugging any app it might make sense to put it in as a protected variable `_first_app_launch` (or maybe `_app_launch_first` to be consistent with the existing variables `_app_version` and `_server_sync_latest`)

jfmcquade:
PR Checklist

## Description

Whilst we already track the timestamp of a user's first launch of the app, this value is not synced tot he database. This PR stores the value of the app's first launch in a field, enabling it to be synced to the user database for analysis. 

The field name is `_app_first_launch`.

## Dev notes

Initially, I wrote the logic to use the template field service to set the field variable, see https://github.com/IDEMSInternational/parenting-app-ui/pull/2301/commits/ef7220f19a7e56816cebd5b0bf7835f18147fed2.
However that approach hardcodes the field name ("_app_first_launch") into the app-events service, without having it declared in the app config (i.e. as `appConfig.APP_FIELDS.APP_FIRST_LAUNCH`). Therefore it seemed sensible to use local storage service to set the field string directly, matching the pattern used to set most of the other protected variables throughout the app. However I don't think this pattern is ideal, as we're setting a field without using the field service and it requires an additional subscription to the app config service in order to get the field name. 

This system of setting protected fields could probably be rationalised across the code at some point. It may be that it's sensible to store the names of these protected fields a level up, in the deployment config defined in `deployment.model.ts`, but not in the `app_config` object, as it doesn't seem necessary to update these values at runtime. Then the values could be consumed by services without needing a subscription to the app config service. If these values were declared without the `FIELD_PREFIX` prefix, then the fields could be set via the `templateFieldService.setField()` method, which appends the prefix itself. It may be that the current pattern of bypassing the template field service is intentional in the case of protected fields in order to differentiate them from user-set fields, in which case perhaps there should be some protections on `templateFieldService.setField()` to prevent setting a protected field directly with this method.

## Git Issues

Closes #2295, assuming that having access to this value in the postgres database is deemed to be a satisfactory resolution of that issue

## Screenshots/Videos

Screenshot 2024-04-23 at 12 01 22

### DIFF
--- a/src/app/shared/services/app-events/app-events.service.ts
+++ b/src/app/shared/services/app-events/app-events.service.ts
@@ -3,7 +3,8 @@ import { generateTimestamp } from "../../utils";
 import { DbService } from "../db/db.service";
 import { LocalStorageService } from "../local-storage/local-storage.service";
 import { AsyncServiceBase } from "../asyncService.base";
-import { TemplateFieldService } from "../../components/template/services/template-field.service";
+import { AppConfigService } from "../app-config/app-config.service";
+import { IAppConfig } from "../../model";
 
 @Injectable({ providedIn: "root" })
 /**
@@ -19,11 +20,12 @@ export class AppEventService extends AsyncServiceBase {
   summary: IAppEventSummary;
   /** Keep a copy of all app events saved in the database in memoery also, by event_id */
   appEventsById: { [key in IEventId]: IAppEvent[] };
+  appFields: IAppConfig["APP_FIELDS"];
 
   constructor(
     private db: DbService,
     private localStorageService: LocalStorageService,
-    private templateFieldService: TemplateFieldService
+    private appConfigService: AppConfigService
   ) {
     super("App Events");
     this.registerInitFunction(this.intialise);
@@ -31,8 +33,9 @@ export class AppEventService extends AsyncServiceBase {
 
   /** Initialise the app events service by recording an app_launch instance */
   private async intialise() {
-    await this.ensureAsyncServicesReady([this.db, this.templateFieldService]);
-    this.ensureSyncServicesReady([this.localStorageService]);
+    await this.ensureAsyncServicesReady([this.db]);
+    this.ensureSyncServicesReady([this.localStorageService, this.appConfigService]);
+    this.subscribeToAppConfigChanges();
     this.setAppEventsById([]);
     this.loadSummary();
 
@@ -56,6 +59,12 @@ export class AppEventService extends AsyncServiceBase {
     await this.loadAppEvents();
     // recalculate any summary data that might have changed due ot this
     this.calculateEventSummaries();
+  }
+
+  private subscribeToAppConfigChanges() {
+    this.appConfigService.appConfig$.subscribe((appConfig: IAppConfig) => {
+      this.appFields = appConfig.APP_FIELDS;
+    });
   }
 
   private getDBAppEvents(event_id?: IAppEvent["event_id"]) {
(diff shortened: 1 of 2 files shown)


src/app/shared/services/app-events/app-events.service.ts
```diff
@@ -79,6 +93,7 @@ export class AppEventService extends AsyncServiceBase {
     // current app day calculated by finding length of array of subset of all unique app open event dates
     const app_day = [...new Set(appOpenEvents.map((e) => e._created.substring(0, 10)))].length;
     const first_app_launch = this.appEventsById.app_launch?.[0]?._created || generateTimestamp();
+    this.localStorageService.setString(this.appFields.APP_FIRST_LAUNCH, first_app_launch);
```
chrismclarke, line 96: nit(blocking) - something doesn't quite feel right here, we're setting the first_launch field any time event summaries are calculated, which is not only on init but also any time there is any other recorded app event. Really this should only ever happen once, on initial load and only if not already populated. 

I'd probably suggest moving the logic to the main `app.component` for now where we set other hardcoded fields, and just do it somewhat deferred after everything else has been initialised and only if there is not an existing value

chrismclarke:
> Initially, I wrote the logic to use the template field service to set the field variable, see https://github.com/IDEMSInternational/parenting-app-ui/commit/ef7220f19a7e56816cebd5b0bf7835f18147fed2.
> However that approach hardcodes the field name ("_app_first_launch") into the app-events service, without having it declared in the app config (i.e. as appConfig.APP_FIELDS.APP_FIRST_LAUNCH). Therefore it seemed sensible to use local storage service to set the field string directly, matching the pattern used to set most of the other protected variables throughout the app. However I don't think this pattern is ideal, as we're setting a field without using the field service and it requires an additional subscription to the app config service in order to get the field name.
> 
> This system of setting protected fields could probably be rationalised across the code at some point. It may be that it's sensible to store the names of these protected fields a level up, in the deployment config defined in deployment.model.ts, but not in the app_config object, as it doesn't seem necessary to update these values at runtime. Then the values could be consumed by services without needing a subscription to the app config service. If these values were declared without the FIELD_PREFIX prefix, then the fields could be set via the templateFieldService.setField() method, which appends the prefix itself. It may be that the current pattern of bypassing the template field service is intentional in the case of protected fields in order to differentiate them from user-set fields, in which case perhaps there should be some protections on templateFieldService.setField() to prevent setting a protected field directly with this method.

I think given that some protected fields may to be accessed before any other services init (e.g. knowing if it's the first launch to handle data sync/migrate or similar ops) we have pretty good reason to move all of these outside of runtime config. Not to mention how suddenly changing at runtime does nothing to migrate the legacy values. The same can be said for the general localstorage prefix, which again really shouldn't be swapped at runtime but also probably should be more explicitly used in core app config to at least be able to distinguish app variables set by different deployments when running locally.

As a first step in tidying I think I would recommend:
1. move the `APP_FIELDS` out of `appConfig.ts` and instead just hardcode into the localstorage service (as this is where they are used). Likely rename to `PROTECTED_FIELDS` and do not export
2. move the `FIELD_PREFIX` to a deploymentConfig variable that can be read from environment
3. add a method to the localstorageservice to `setProtectedField`, with support just for those listed.
4. update pr to set app first open field from the main `app.component.ts` to run only on first launch and after all other services initialised. Can include a check in case does not exist to just pull from `app-event.service` - in the future we can hopefully set this independently of service
5. check codebase for instances of `localstorage.set(` and try to refactor to use localstorageservice (this is a sync service so no need to worry to much about initialisation at least)

Let me know if you think this approach sounds sensible, and whether you want me to give a pass at a PR or if you're happy to attempt a quick refactor.

chrismclarke (changes requested):

PR #2307, opened 2024-04-30
fix: data pipe append_columns operation can handle boolean strings

jfmcquade:
PR Checklist

- [x] Merge https://github.com/IDEMSInternational/app-debug-content/pull/66 (for diff comparison of output data lists from data pipe)

## Description

Fixes an issue where values of data list columns added via the data pipe `append_columns` operation would not be parsed as boolean values as expected when the value is a boolean string (`"TRUE"`, `"false"` etc.)

Incorporates #2308:
- add generic `parseStringValue` string util and tests to cover conversion of string to boolean, number, null or undefined

## Testing

Run `yarn workspace shared test` and ensure tests are passing as expected:
Screenshot 2024-05-01 at 16 05 01
Screenshot 2024-05-01 at 16 05 22

With this PR's branch checked out, and assuming https://github.com/IDEMSInternational/app-debug-content/pull/66 has been merged, set the active deployment to `debug`. Run `yarn workflow sync_sheets` and `yarn workflow repo open` to inspect the differences. There should just be one changed file, `app_data/sheets/data_list/generated/example_pipe_appended.json` with the diff showing that previous boolean strings have been replaced by boolean values.

Screenshot 2024-04-30 at 14 48 29

## Dev notes

There are multiple places where this `booleanStringToBoolean()` parsing could be applied. Doing so in the `appStringEvaluator` seems appropriate (possibly correcting an oversight) and preserves the type safety of the `appendColumns` operator.

## Git Issues

Closes #

### DIFF
--- a/packages/shared/src/models/appStringEvaluator/appStringEvaluator.spec.ts
+++ b/packages/shared/src/models/appStringEvaluator/appStringEvaluator.spec.ts
@@ -20,7 +20,9 @@ describe("App String Evaluator", () => {
   it("Evaluates boolean strings as booleans", () => {
     expect(evaluator.evaluate("false")).toEqual(false);
     expect(evaluator.evaluate("FALSE")).toEqual(false);
+    expect(evaluator.evaluate("False")).toEqual(false);
     expect(evaluator.evaluate("true")).toEqual(true);
     expect(evaluator.evaluate("TRUE")).toEqual(true);
+    expect(evaluator.evaluate("True")).toEqual(true);
   });
 });
--- a/packages/shared/src/models/appStringEvaluator/appStringEvaluator.spec.ts
+++ b/packages/shared/src/models/appStringEvaluator/appStringEvaluator.spec.ts
@@ -16,4 +16,11 @@ describe("App String Evaluator", () => {
       "Hello Ada Lovelace"
     );
   });
+
+  it("Evaluates boolean strings as booleans", () => {
+    expect(evaluator.evaluate("false")).toEqual(false);
+    expect(evaluator.evaluate("FALSE")).toEqual(false);
+    expect(evaluator.evaluate("true")).toEqual(true);
+    expect(evaluator.evaluate("TRUE")).toEqual(true);
+  });
 });
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/flowParser.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/flowParser.ts
@@ -4,7 +4,7 @@ import { IConverterPaths, IFlowHashmapByType, IParsedWorkbookData } from "../../
 import { arrayToHashmap, groupJsonByKey, IContentsEntry } from "../../utils";
 import BaseProcessor from "../base";
 
-const cacheVersion = 20240412.3;
+const cacheVersion = 20240502.0;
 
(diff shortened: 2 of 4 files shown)


packages/shared/src/models/appStringEvaluator/appStringEvaluator.spec.ts
```diff
@@ -16,4 +16,11 @@ describe("App String Evaluator", () => {
       "Hello Ada Lovelace"
     );
   });
+
+  it("Evaluates boolean strings as booleans", () => {
+    expect(evaluator.evaluate("false")).toEqual(false);
+    expect(evaluator.evaluate("FALSE")).toEqual(false);
+    expect(evaluator.evaluate("true")).toEqual(true);
+    expect(evaluator.evaluate("TRUE")).toEqual(true);
```
esmeetewinkel, line 24: Should this also apply to "True" and "False"? Maybe first convert string to lower case and then compare against "true" / "false" ?

esmeetewinkel (approved):
Functional test passed

packages/shared/src/models/appStringEvaluator/appStringEvaluator.spec.ts
jfmcquade: Good spot. The actual logic that checks the string uses a regex to match strings of any case (see the `booleanStringToBoolean()` function [here](https://github.com/IDEMSInternational/parenting-app-ui/blob/master/packages/shared/src/utils/string-utils.ts)). So this is just an issue with my test cases not being extensive, I can update them.

jfmcquade: https://regexr.com/ is a brilliant tool for testing and understanding regex expressions, if you're not aware
Screenshot 2024-04-30 at 17 26 59

esmeetewinkel: Ah sure, that makes sense - I didn't realise this change was describing test cases only :)

chrismclarke (changes requested):
Nice issue to track down, makes sense as we define `"column : value"` format for append column operators so yes it will always get parsed as a string.

I feel like if addressing boolean strings it might also make sense to cover other data formats at the same time (e.g. number, null/undefined). I also think it might be better in this instance to keep the method targetted just at the data pipe append_columns operator for now (as harder to track knock-ons if using more generally). I've made a PR targetting this branch at https://github.com/IDEMSInternational/parenting-app-ui/pull/2308, let me know what you think @jfmcquade

chrismclarke (approved):
@jfmcquade - I've pushed a minor commit to bump the flowParser cache version and updated the example sheet to include the number/number-string parsing examples. Otherwise I think should be good to go

@esmeetewinkel - See updated [example_data_pipe_append](https://docs.google.com/spreadsheets/d/18CUEEFHYs6FqJmaAiXzTRspWxV1alqSkcr1wuEOHfNw/edit#gid=1738236785), should now correctly parse boolean and number values as part of append operation. This only applies to specific append operation (because that takes inline `key : value` pairs in a single cell which means the sheets to not natively convert format), but if you do find similar issues further down the line (e.g. parameter lists, action lists etc.) then we can try and make more generally applied (a lot of these values are also re-interpreted at runtime so hard to know what is already handled further down the line or not).

See before/after for content sync below

image

PR #2308, opened 2024-04-30
refactor: data-pipe parse string

chrismclarke:
PR Checklist

## Description

Targets branch of pr #2307

- add generic `parseStringValue` string util and tests to cover conversion of string to boolean, number, null or undefined
- update `appStringEvaluator` to no longer parse boolean strings, instead applying new method only to `appendColumns` operator

## Dev notes
I couldn't decide between making the conversion more specific to the `appendColumns` operator or generally to the `appStringEvaluator`, but in the end felt keeping it more specific for now might be better as fewer potential knock-ons and can consider making more general in a follow-up PR if useful.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/shared/src/models/appStringEvaluator/appStringEvaluator.spec.ts
+++ b/packages/shared/src/models/appStringEvaluator/appStringEvaluator.spec.ts
@@ -16,13 +16,4 @@ describe("App String Evaluator", () => {
       "Hello Ada Lovelace"
     );
   });
-
-  it("Evaluates boolean strings as booleans", () => {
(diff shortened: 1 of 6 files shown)


jfmcquade (approved):
Thanks, it's a good idea to handle other data formats alongside booleans. 

I was also unsure about whether to update the `appStringEvaluator` or just the `appendColumns` operator, but I agree with your reasoning that it's better to avoid unwanted knock-ons and therefore restrict the changes to the data pipe operator for now (although we may be faced with a similar issue with the `appStringEvaluator` at some point in the future).

I'll merge this and update my original PR.

PR #2309, opened 2024-05-01
fix: set_item actions nested inside item loop

jfmcquade:
PR Checklist

## Description

Fixes an issue where `set_item` and `set_items` actions would not function as expected when the row was deeply nested within an item loop, e.g. within a `display_group` component inside an item loop.

This was due to the `setItemContext` meta only being applied to actions within the top-level `action_list` inside the items loop.

## Git Issues

Issue identified in [this mattermost thread](https://chat.idems.international/all/pl/spto5y6f53b4xfm1k6p8x91dpc).

## Screenshots/Videos

See new [debug_data_items_actions](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=1007541292) component. The buttons on all actions now function as expected – on `master`, the buttons in the second items loop, nested inside `display_group`s, do not trigger the action correctly and an error is thrown.

Screenshot 2024-05-01 at 15 35 38

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/a86caf98-8ad1-40dc-9d09-35f4dafb7c96

### DIFF
--- a/src/app/shared/components/template/components/data-items/data-items.component.ts
+++ b/src/app/shared/components/template/components/data-items/data-items.component.ts
@@ -89,8 +89,7 @@ export class TmplDataItemsComponent extends TemplateBaseComponent implements OnD
   }
 
   /**
-   * Update item dynamic evaluation context and action lists to include relevant
-   * item data
+   * Update item dynamic evaluation context and action lists to include relevant item data.
    * @param templateRows List of template rows generated from itemData by item processor
    * @param itemData List of original item data used to create item rows (post operations such as filter/sort)
    * @param dataListName The name of the source data list (i.e. this.dataListName, extracted for ease of testing)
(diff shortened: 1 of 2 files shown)


esmeetewinkel (approved):
Functional test passed

src/app/shared/components/template/components/data-items/data-items.component.ts
```diff
@@ -93,44 +93,59 @@ export class TmplDataItemsComponent extends TemplateBaseComponent implements OnD
...
+        }
+        if (a.action_id === "set_items") {
+          // TODO - add a check for @item refs and replace parameter list with correct values
+          // for each individual item (default will be just to pick the first)
+          a.args = [setItemContext];
+        }
+        return a;
+      });
+    }
+    if (row.rows) {
```
chrismclarke, line 145: nit(non-blocking): might be good to add a quick comment to make explicit why checking, e.g
```
// ensure item children with nested rows (e.g. display groups) also inherit item context
```

chrismclarke (approved):
Thanks @jfmcquade, nice to have the use-case covered, also very much appreciate adding spec tests for the component, really helps clarify the behaviour and will make it much easier to work on in the future.

I've added one minor blocking comment inline. I also noticed that it would probably be possible to reduce the diff by simply adapting the original code in the same way as you've done with the recursive function, namely keeping the original `templateRows.map(r=>...` function and simply adding the recursive call
```ts
 if (r.rows) {
      r.rows = r.rows.map((childRow) => this.setItemMeta(childRow, itemData));
    }
```
But I'm not strongly opposed to having two functions to handle the setItemMeta process so I'll leave you to decide if you think worth a quick refactor or not.

jfmcquade:
> I also noticed that it would probably be possible to reduce the diff by simply adapting the original code in the same way as you've done with the recursive function, namely keeping the original `templateRows.map(r=>...` function and simply adding the recursive call
> 
> ```ts
>  if (r.rows) {
>       r.rows = r.rows.map((childRow) => this.setItemMeta(childRow, itemData));
>     }
> ```
> 
> But I'm not strongly opposed to having two functions to handle the setItemMeta process so I'll leave you to decide if you think worth a quick refactor or not.

Ah, good point. Refactoring along these lines actually makes the logic simpler than this, because `setItemMeta()` takes an array of rows. See latest diff for all commits.

Issue #2310, opened 2024-05-02
[BUG] Data items cannot loop over list with translatable column


esmeetewinkel:
**Describe the bug**
Data items cannot loop over list with translatable column (whereas items can). When attempting to do so, the console displays an error 
```
error-handler.service.ts:65 ERROR RxError (SC1): RxError (SC1):
fieldnames do not match the regex
Given parameters: {
regex:"^[a-zA-Z](?:[[a-zA-Z0-9_]*]?[a-zA-Z0-9])?$"
fieldName:"_translations"}
    at newRxError (rx-error.js:98:10)
    at checkFieldNameRegex (check-schema.js:34:21)
    at checkField (check-schema.js:47:102)
    at check-schema.js:138:9
    at Array.forEach (<anonymous>)
    at traverse (check-schema.js:135:29)
    at check-schema.js:142:7
    at Array.forEach (<anonymous>)
    at traverse (check-schema.js:135:29)
    at validateFieldsDeep (check-schema.js:145:3)
```
**App version**
v0.16.31

**Debug Template**
[debug_data_items_translate](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=545128588) which loops over the data list [debug_data_list
](https://docs.google.com/spreadsheets/d/1vtr341TIg0MDq99syARtSK6VmvI3rFjEdYfwgwdb1ek/edit#gid=1595229908)

**Screenshots and videos**
![image](https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/64c7635e-eb21-49f8-bba3-fcd0b1a449fa)
![image](https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/be63fafa-bcfc-4b63-9741-7a0d78a12bd8)

PR #2312, opened 2024-05-02
fix: data items are sorted according to original order by default

### LINKED ISSUE #2311: [BUG] Data items sorts data list by default
**Describe the bug**
By default, a data items loop sorts its input data list alphabetically by the `id` column values. Instead, the default should be for it to leave the data list order untouched. Sorting should be optional. 

**App version**
v0.16.31

**Use case**
In `plh_kids` we're looping over a list with id's learn_1, learn_2, learn_3 up to learn_11 (in that order) and I discovered they were displayed in order learn_1, learn_10, learn_11, learn_2, etc

**Debug Template**
[debug_data_items_order](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=1681985300) which loops over [debug_data_list_order](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=275599747)
image

**Screenshots and videos**

image image
N.B. I have no clue why the spacing between the text components is so different on `items` and on `data_items`!

jfmcquade:
PR Checklist

## Description

As documented in #2311, item rows within a `data-items` loop are sorted according to their ID value by default. Whilst we do have a `sort` operation that can be applied to the items via the parameter list of the data-items component, which can be used to sort the list based on some property of the items, there is currently no way to reference the original order of the items in the data list.

This PR makes it the default that data-items render in the order that matches the original data list.

## Git Issues

Closes #2311

## Screenshots/Videos

Template [debug_data_items_order](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=1681985300), reading from [debug_data_list_order](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=275599747) showing items displayed in the expected order.

Screenshot 2024-05-02 at 16 38 00

Screenshot 2024-05-02 at 16 38 16

Passing tests:

Screenshot 2024-05-09 at 12 29 47

### DIFF
--- a/src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
+++ b/src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
@@ -48,7 +48,7 @@ const SCHEMA: RxJsonSchema<any> = {
     flow_name: { type: "string", maxLength: 64 },
     flow_type: { type: "string", maxLength: 64 },
     row_id: { type: "string", maxLength: 64 },
-    row_index: { type: "integer", minimum: 0, maximum: 10000, multipleOf: 1 },
+    row_index: { type: "integer", minimum: 0, maximum: 10000, multipleOf: 1, final: true },
     data: {
       type: "object",
     },
--- a/src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
+++ b/src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
@@ -37,7 +37,7 @@ const SCHEMA: RxJsonSchema<any> = {
   title: "base schema for id-primary key data",
   // NOTE - important to start at 0 and not timestamp (e.g. 20221220) as will check
   // for migration strategies for each version which is hugely inefficient
-  version: 1,
+  version: 2,
   primaryKey: "id",
   type: "object",
   properties: {
@@ -48,17 +48,22 @@ const SCHEMA: RxJsonSchema<any> = {
     flow_name: { type: "string", maxLength: 64 },
     flow_type: { type: "string", maxLength: 64 },
     row_id: { type: "string", maxLength: 64 },
+    row_index: { type: "integer", minimum: 0, maximum: 10000, multipleOf: 1 },
     data: {
       type: "object",
     },
   },
-  required: ["id", "flow_type", "flow_name", "row_id", "data"],
-  indexes: ["flow_type", "flow_name", "row_id"],
+  required: ["id", "flow_type", "flow_name", "row_id", "data", "row_index"],
+  indexes: ["flow_type", "flow_name", "row_id", "row_index"],
 };
 const MIGRATIONS: MigrationStrategies = {
   // As part of RXDb v14 update all data requires migrating to change metadata fields (no doc data changes)
   // https://rxdb.info/releases/14.0.0.html
   1: (doc) => doc,
+  2: (oldDoc) => {
+    const newDoc = { ...oldDoc, row_index: 0 };
+    return newDoc;
+  },
 };
 
 interface IDataUpdate {
--- a/src/app/shared/services/dynamic-data/dynamic-data.service.ts
+++ b/src/app/shared/services/dynamic-data/dynamic-data.service.ts
@@ -182,7 +182,7 @@ export class DynamicDataService extends AsyncServiceBase {
     const initialDataWithMeta = initialData.map((el) => {
       return {
         ...el,
-        index_original: initialData.indexOf(el),
+        row_index: initialData.indexOf(el),
       };
     });
 
--- a/src/app/shared/services/dynamic-data/dynamic-data.service.ts
+++ b/src/app/shared/services/dynamic-data/dynamic-data.service.ts
@@ -209,10 +209,8 @@ export class DynamicDataService extends AsyncServiceBase {
   /** Retrive json sheet data and merge with any user writes */
   private async getInitialData(flow_type: FlowTypes.FlowType, flow_name: string) {
     const flowData = await this.appDataService.getSheet(flow_type, flow_name);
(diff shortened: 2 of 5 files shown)


src/app/shared/services/dynamic-data/dynamic-data.service.ts
```diff
@@ -178,9 +178,17 @@ export class DynamicDataService extends AsyncServiceBase {
     if (initialData.length === 0) {
       throw new Error(`No data exists for collection [${flow_name}], cannot initialise`);
     }
-    const schema = this.inferSchema(initialData[0]);
+    // add index property to each element before insert, for sorting queried data by original order
+    const initialDataWithMeta = initialData.map((el) => {
+      return {
+        ...el,
+        index_original: initialData.indexOf(el),
```
jfmcquade, line 185: Originally I was going to name this field "_index", however RxJs throws an error when trying to use a field name that starts with an underscore: `RxError: fieldnames do not match the regex:"^[a-zA-Z](?:[[a-zA-Z0-9_]*]?[a-zA-Z0-9])?$"`. Using "_index" might also get it confused with the `_index` of an item in `itemContext`, which is the index within the array of items to render, after any operations have been performed on the original data.

So it's an open question as to what this field could be named. It should be something that we're not going to overwrite as a property of the item. With #2310 in mind, we could prefix protected properties with a custom prefix: e.g., in this case we call the field something like `PROTECTED_index`, and as a solution to #2310, we append any field name that starts with `_` with the prefix "protected", e.g. `PROTECTED_translations`. Thoughts?

chrismclarke (changes requested):
Ah of course, given the `id` column is a primary key we lose all default sorting for any data not in alphabetical order. I like the solution to populate and track the additional metadata fields, a couple minor suggestions:

1. I think it would be good to include the metadata field as part of the schema for both the `persistedMemory` and `reactiveMemory` adapters. That way we can also mark as an [index](https://rxdb.info/rx-schema.html#indexes) column type to make it more efficient to retrieve ordered data. 

2. We may also want to consider marking the column as a [final](https://rxdb.info/rx-schema.html#final) column type to prevent user writes to the column, although should first test if we're still able to populate and repopulate data on initial load with updated index (possibly just consider later).

3. For the naming convention appreciate we can't start with underscore. Given we already use `row_id` as a column for persisted data I'd probably recommend using `row_index` instead of `index_original`. This can be defined in the schemas as an integer type. I think probably no need to specifically mark `PROTECTED` columns as I think even if users included their own `row_index` column it will just get overwritten by the generated one, and if wanting to provide a better way for users to sort data then this can be handled elsewhere.

4. Instead of relying on sorting data within component I think it might be better to make the default behaviour to return rows in the `row_id` order unless otherwise specified. So likely updating the adapter `query$` method to automatically specify the [query index](https://rxdb.info/nosql-performance-tips.html#set-a-specific-index) as `row_id` 

Let me know if this makes sense and if you need me to do anything else for support on this PR

chrismclarke:
> This does seem to work and still allow us to update the index if the data changes. See demo below (with some waiting time cut out): I make a user write over some data, then change the order of that data (changing the index for each item) and make additional writes. You can see that the `row_index` property remains the same until a write is made specifically to that item, at which point it successfully updates to reflect the new order. Does this cover the case you were thinking of? 

Great, yeah that works really nicely, because the `row_id` is a top-level property and only the inner `data` property is persisted/merged then it means we will always preserve the `row_index` from parsing the sheet. I also noticed in the rxdb code that the final field is only enforced in dev mode (guessing performance reasons, seems reasonable to me given we're more likely to explore implementation during dev anyways), so production is a little more forgiving anyways.

> I'm not quite sure how to put this into a spec test, but I've added one that confirms the `row_index` property cannot be overridden via the `DynamicDataService.update()` method. Are there any more cases related to setting the column to `final` that might not be covered?

I think that's sufficient, our spec tests cover the metadata population and inner data merge so would be a bit artificial adding tests right now for things that don't exist/won't happen. I manually tested migration switching between branches, had looked at adding a spec test but decided would be more effort than its worth as we'd need to totally decouple the schema/migration files from the service code (to be able to replace it more easily), which doesn't seem worthwhile when we're unlikely to change these much again in the future.

src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
```diff
@@ -48,17 +48,22 @@ const SCHEMA: RxJsonSchema<any> = {
...
-  indexes: ["flow_type", "flow_name", "row_id"],
+  required: ["id", "flow_type", "flow_name", "row_id", "data", "row_index"],
+  indexes: ["flow_type", "flow_name", "row_id", "row_index"],
 };
 const MIGRATIONS: MigrationStrategies = {
   // As part of RXDb v14 update all data requires migrating to change metadata fields (no doc data changes)
   // https://rxdb.info/releases/14.0.0.html
   1: (doc) => doc,
+  2: (oldDoc) => {
+    const newDoc = { ...oldDoc, row_index: 0 };
```
chrismclarke, line 64: +1 - I took a quick look to see if there was any way to get the actual index but not passed to rxdb migration executor so I think your workaround to just make identical and use secondary sort index based on id is the best option

chrismclarke (approved):
Thanks @jfmcquade - a really like this implementation, nice to be able to handle all the logic just from within the service code and also appreciate the tests and migration implementations.

I've added one minor code tidying commit with 307225 (think the auto wrap left an empty type interface for whatever reason), but otherwise I'd say should be good to go

PR #2304, opened 2024-04-26
feat: make local context available in data-items loop

jfmcquade:
PR Checklist

- [x] PR title descriptive (can be used in release notes)
 
 ## TODO

- [x] Builds on #2305, merge that first for clearer diffs
- [x] Builds on the spec test files added by #2215, although should be able to be merged independently and resolve any conflicts afterwards

## Description

Allows rows within a data-items loop to reference local variables (#2303).

## Dev notes

This feels like a hack, and adding another caveat to the template-variables service logic doesn't feel great, but it does seem to work.

NB: view the code changes for `template-variables.service.ts` with whitespace changes hidden in order to see meaningful diff.

## Git Issues

Closes #2303

## Screenshots/Videos

[debug_data_items](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=1560865754)

Screenshot 2024-04-26 at 12 35 03

Screenshot 2024-04-26 at 12 30 11

[debug_data_items_local](https://docs.google.com/spreadsheets/d/14XXwZ7R_7qIBo5xKYuxnPVQbQidI--PPJ7TVvZxwy1k/edit#gid=1990453574)

Screenshot 2024-04-26 at 12 36 07

Screenshot 2024-04-26 at 12 30 17

Issue #2303, opened 2024-04-25
[BUG] data_items doesn't find local variables


esmeetewinkel:
**Describe the bug**
Local variables set within a template outside a data_items show as `undefined` when evaluated within the data_items context.

This might be related to issue https://github.com/IDEMSInternational/parenting-app-ui/issues/1598

**App version**
v0.16.31

**Template Name**
[module](https://docs.google.com/spreadsheets/d/1MeuVFYp7ovMCABdgC5c1mExgDK8Xop13fzzBNR5EAPo/edit#gid=880798806) on `plh_kids` deployment

**Debug Template**
[debug_data_items_local](https://docs.google.com/spreadsheets/d/14XXwZ7R_7qIBo5xKYuxnPVQbQidI--PPJ7TVvZxwy1k/edit#gid=1990453574)

**Screenshots and videos**
![image](https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/15f4c041-68ad-4cc1-9f79-a3bdf94af745)

![image](https://github.com/IDEMSInternational/parenting-app-ui/assets/74557272/05454a83-05f3-4836-b720-be70a889ffd7)

esmeetewinkel:
Closed by

PR #2314, opened 2024-05-02
feat: data-items local context (alt)

chrismclarke:
PR Checklist

## Description

Alternate implementation to #2304

- 

closes #2304

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/components/data-items/data-items.component.ts
+++ b/src/app/shared/components/template/components/data-items/data-items.component.ts
@@ -147,6 +147,10 @@ export class TmplDataItemsComponent extends TemplateBaseComponent implements OnD
         rows: [],
       },
     } as any);
+    // HACK - still want to be able to use localContext from parent rows so copy to child processor
+    processor.templateRowMap = JSON.parse(
+      JSON.stringify(this.parent.templateRowService.templateRowMap)
+    );
     await processor.processContainerTemplateRows();
     return processor.renderedRows;
   }
@@ -185,9 +189,8 @@ export class TmplDataItemsComponent extends TemplateBaseComponent implements OnD
       parsed[listKey] = listValue;
       for (const [itemKey, itemValue] of Object.entries(listValue)) {
         if (typeof itemValue === "string") {
-          parsed[listKey][itemKey] = await this.templateVariablesService.evaluateConditionString(
-            itemValue
-          );
+          parsed[listKey][itemKey] =
+            await this.templateVariablesService.evaluateConditionString(itemValue);
         }
       }
     }
--- a/src/app/shared/components/template/services/instance/template-row.service.ts
+++ b/src/app/shared/components/template/services/instance/template-row.service.ts
@@ -36,7 +36,10 @@ export class TemplateRowService extends SyncServiceBase {
   public templateRowMap: ITemplateRowMap = {};
   public renderedRows: FlowTypes.TemplateRow[]; // rows processed and filtered by condition
 
-  constructor(private injector: Injector, public container: TemplateContainerComponent) {
+  constructor(
+    private injector: Injector,
+    public container: TemplateContainerComponent
+  ) {
     super("TemplateRow");
     /**
      * Avoid initialisation logic and prefer to ensure services ready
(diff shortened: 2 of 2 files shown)


esmeetewinkel:
> To address the issue seen in the [debug_kids](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=2079679226) template I've modified the sheet slightly (see comments on sheet)

> I couldn't get the stop_loop condition working properly (I think the data doesn't update frequently enough to work as expected), so instead just added a condition to the display_group (so slightly less efficient as will still process all items before stopping).

I realised my debug sheet wasn't entirely clear because I was using the id as an example column to check for, which has unique values. I've updated the debug sheet now to reflect the use case where there is multiple rows that match the condition (e.g. looking for rows of a particular `type`, which there are multiple of). I _do_ actually need the loop to stop after the first time the condition is met - or I need another way to find "the first row where the condition is met".

> I think this can also be handled by using the parameter_list to provide a filter operation (can't find debug sheet but pretty sure we included that functionality previously)

We do have the parameter list filter operation on data_items, however, it doesn't accept comparisons to local variables (e.g. `@item.type == @local.type`) so it's not of use here. If that were possible, however, I think that would avoid the stop_loop issue as well, because I could simply get away with using a parameter list on the `begin_data_items` row being `@item.type == @local.type; limit: 1`.

jfmcquade:
Thanks, @chrismclarke, I definitely think this implementation is preferable to #2304. I've copied over a test from #2304 (https://github.com/IDEMSInternational/parenting-app-ui/pull/2314/commits/79323f0cb99c87e6ed9d81b61d68536414cabe72), I think the rest of that PR can be abandoned. 
In terms of adding tests for the data items logic, I did add some tests to `data-items.component.spec.ts` as part of #2309, I agree it would be good to expand these at some point. As part of that follow-up, do you think it would be sensible to extract the items related logic out of both the data-items component and the dynamic data service and into a new `data-items` service (as floated [here](https://github.com/IDEMSInternational/parenting-app-ui/pull/2215#discussion_r1577981965))? This would likely make testing easier (for example if some of the complex logic of the data-items component was in public methods of a standalone service).

chrismclarke:
> Thanks, @chrismclarke, I definitely think this implementation is preferable to #2304. I've copied over a test from #2304 ([79323f0](https://github.com/IDEMSInternational/parenting-app-ui/commit/79323f0cb99c87e6ed9d81b61d68536414cabe72)), I think the rest of that PR can be abandoned. In terms of adding tests for the data items logic, I did add some tests to `data-items.component.spec.ts` as part of #2309, I agree it would be good to expand these at some point. As part of that follow-up, do you think it would be sensible to extract the items related logic out of both the data-items component and the dynamic data service and into a new `data-items` service (as floated [here](https://github.com/IDEMSInternational/parenting-app-ui/pull/2215#discussion_r1577981965))? This would likely make testing easier (for example if some of the complex logic of the data-items component was in public methods of a standalone service).

I've started to stub out some tests in https://github.com/IDEMSInternational/parenting-app-ui/pull/2315 as I think might get a bit messy if adding here

chrismclarke:
> We do have the parameter list filter operation on data_items, however, I believe it doesn't accept comparisons to local variables e.g. `filter: @item.type == @local.type` (or at least items doesn't #1598) so it's not of use here. If that were possible, however, I think that would avoid the stop_loop issue as well, because I could simply get away with using a parameter list on the `begin_data_items` row being `filter: @item.type == @local.type; limit: 1`.
> 
> Edited: @jfmcquade [pointed out on Mattermost](https://chat.idems.international/all/pl/aqxaifzrztbaup5mre9r1coauo) a temporary (hacky) way to do this using javascript functions, example [here](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit#gid=2079679226)

Thanks for the clarifications. That makes sense, I think probably best to handle in a follow-up PR once merged

jfmcquade (approved):

esmeetewinkel (approved):
Functional test passed

esmeetewinkel:
Closes #2303

Issue #2254, opened 2024-03-22
[FEATURE] Accommodate left-to-right languages


esmeetewinkel:
**What?**
App to accommodate translations to left-to-right languages. This might involve reflecting the entire screen layout horizontally, e.g.
- left-aligned text would need to be right-aligned
- centred text would stay centred
- "next" buttons that appear on the right and "previous" buttons that appear on the left would need to be flipped
- images probably should not be flipped (this would make sense in some cases, but not in all). We could go around this by providing language-specific images in case they do need to be reflected. 

**Why?**
FacilitatorApp Kuwait is planning to [deploy in Arabic](https://github.com/IDEMSInternational/plh-kids-app-kw-content/issues/2), which is left-to-right.

chrismclarke:
Ionic has a blog post on adding support for RTL, I think the post makes it sound simpler than it actually is as a lot of our custom content has a somewhat hardcoded layout (like you identified with nav buttons), so we might need a bunch of extra styling overrides to accommodate as required

https://ionic.io/blog/ionic-and-rtl

jfmcquade:
A few interesting points here: https://twitter.com/kepano/status/1777813094619840535

chrismclarke:
Discussed on call, recommend keeping in config (or app_config) a languages property that stores language metadata, such as right-to-left e.g.
```ts
languages:{
   en_rtl:{
      rtl: true
  }
}
```
Then when setting the language should lookup app config in case there's any extra metadata to apply.

esmeetewinkel:
Issues found post-merge of #2317: 

- [ ] Toggle bar + debug toggle have not swapped 
- [ ] Left-aligned buttons need to become right-aligned
- [ ] Need to confirm markdown works in RTL translation (in particular, bulleted / numbered lists)
- [ ] Need to confirm punctuation (`.`, `;`, `:`) at the end of a line works in RTL translation (it seems that this displays on the left automatically)
- [ ] Cards with icon / toggle don't display sensibly 
- [ ] Controls on video component have not reflected

jfmcquade:
[...]

PR #2317, opened 2024-05-07
feat: add support for RTL languages

jfmcquade:
PR Checklist

## Description

Adds minimal support for right-to-left languages. As identified in #2254, there are likely to be further changes required to address issues with the app when in RTL "mode" (the changes go beyond just the appearance of text). These could be added as follow-ups following user testing with someone who understands the expected behaviour of an RTL app.

The metadata for a language (currently just a boolean value for `rtl`, but could be expanded to include other properties in future) can be set at the level of the deployment config. For example, the debug deployment config has been updated in https://github.com/IDEMSInternational/app-debug-content/pull/69 to support an RTL language with the name "en_rtl" (the language can be set via the template [feature_rtl_language](https://docs.google.com/spreadsheets/d/1DkWnJrpR6pjHpPiNNfGXR4sJpCXiaRnElebeorh8sLY/edit#gid=569531329).

Screenshot 2024-05-07 at 14 56 19

## Git Issues

Addresses #2254, but this issue should remain open for tracking

## Screenshots/Videos

[feature_rtl_language](https://docs.google.com/spreadsheets/d/1DkWnJrpR6pjHpPiNNfGXR4sJpCXiaRnElebeorh8sLY/edit#gid=569531329) template:

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/f1d1db96-d28c-45cd-ad9f-6cd3eb5fa575

### DIFF
--- a/packages/data-models/appConfig.ts
+++ b/packages/data-models/appConfig.ts
@@ -51,6 +51,15 @@ const APP_LANGUAGES = {
   default: "gb_en",
 };
 
+export interface ILanguageMeta {
+  rtl?: boolean;
+}
+const APP_LANGUAGES_META: { [languageCode: string]: ILanguageMeta } = {
+  en_rtl: {
+    rtl: true,
+  },
+};
+
 /** Specifies the skins available to the current app build.
  * "defaultSkinName": the name of the skin used by default for a given build
  * "available": an array of all skins that are available to the current build */
@@ -221,6 +230,7 @@ const APP_CONFIG = {
   APP_INITIALISATION_DEFAULTS,
   APP_AUTHENTICATION_DEFAULTS,
   APP_LANGUAGES,
+  APP_LANGUAGES_META,
   APP_ROUTE_DEFAULTS,
   APP_SIDEMENU_DEFAULTS,
   APP_FOOTER_DEFAULTS,
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -30,7 +30,7 @@ import { AppDataService } from "./shared/services/data/app-data.service";
 import { AuthService } from "./shared/services/auth/auth.service";
 import { LifecycleActionsService } from "./shared/services/lifecycle-actions/lifecycle-actions.service";
 import { AppConfigService } from "./shared/services/app-config/app-config.service";
-import { IAppConfig } from "./shared/model";
+import { IAppConfig, ILanguageMeta } from "./shared/model";
 import { TaskService } from "./shared/services/task/task.service";
 import { AppUpdateService } from "./shared/services/app-update/app-update.service";
 import { RemoteAssetService } from "./shared/services/remote-asset/remote-asset.service";
@@ -57,6 +57,8 @@ export class AppComponent {
   footerDefaults: IAppConfig["APP_FOOTER_DEFAULTS"];
   /** Track when app ready to render sidebar and route templates */
   public renderAppTemplates = false;
+  languagesMeta: { [languageCode: string]: ILanguageMeta };
+  languageRtl: boolean;
 
   constructor(
     // 3rd Party Services
@@ -104,6 +106,8 @@ export class AppComponent {
   private async initializeApp() {
     this.platform.ready().then(async () => {
       this.subscribeToAppConfigChanges();
+      this.subscribeToLanguageChanges();
+
       // ensure deployment field set correctly for use in any startup services or templates
       localStorage.setItem(this.appFields.DEPLOYMENT_NAME, this.DEPLOYMENT_NAME);
       localStorage.setItem(this.appFields.APP_VERSION, this.APP_VERSION);
@@ -171,6 +175,14 @@ export class AppComponent {
       this.footerDefaults = this.appConfig.APP_FOOTER_DEFAULTS;
       this.appAuthenticationDefaults = this.appConfig.APP_AUTHENTICATION_DEFAULTS;
       this.appFields = this.appConfig.APP_FIELDS;
+      this.languagesMeta = this.appConfig.APP_LANGUAGES_META;
+    });
+  }
+
+  /** Update meta for current language */
+  private subscribeToLanguageChanges() {
+    this.templateTranslateService.app_language$.subscribe((lang) => {
+      this.languageRtl = this.languagesMeta?.[lang]?.rtl;
     });
   }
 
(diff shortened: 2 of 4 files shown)


packages/data-models/appConfig.ts
```diff
@@ -51,6 +51,15 @@ const APP_LANGUAGES = {
   default: "gb_en",
 };
 
+export interface ILanguageMeta {
+  rtl?: boolean;
+}
+const APP_LANGUAGES_META: { [languageCode: string]: ILanguageMeta } = {
```
chrismclarke, line 57: nit(non-blocking) - Not sure we really need the `APP_` prefix, in my mind that is more for cases of app-specific UI, not so much underlying settings (although not remotely consistent right now). But given this all sits in `appConfig` I feel like something along the lines of `appConfig.LANGUAGE_SETTINGS` or `appConfig.LANGUAGE_CONFIG` or `appConfig.LANGUAGES` might be a little more human readable than `appConfig.APP_LANGUAGES_META`... but I don't have an overly strong opinion (open to your thoughts too @esmeetewinkel )

src/app/app.component.ts
```diff
@@ -171,6 +175,14 @@ export class AppComponent {
       this.footerDefaults = this.appConfig.APP_FOOTER_DEFAULTS;
       this.appAuthenticationDefaults = this.appConfig.APP_AUTHENTICATION_DEFAULTS;
       this.appFields = this.appConfig.APP_FIELDS;
+      this.languagesMeta = this.appConfig.APP_LANGUAGES_META;
+    });
+  }
+
+  /** Update meta for current language */
+  private subscribeToLanguageChanges() {
```
chrismclarke, line 183: nit(non-blocking) - it might be a little bit tidier to simply keep a `rtl` public property in the language service and bind that in the component html instead of having intermediate variable within the component ts. You could also avoid the need to use subscriptions by defining the variable using signals, so that the translate service sets the signal value on language change and the app component html reads the signal value

chrismclarke (approved):
Nice to start on this for sure - guessing there will be quite a few issues to address in the future but appreciate the scaffolding debug sheet and direction binding.

As a minor nit, may be worth including a number on the debug sheet to clearly show `1. English` , `2. English RTL` (more obvious order is reversed), and possibly including translations that should how the text would appear, e.g. `hsilgnE .1` and `LTR hsilgnE .2` (assuming that's comparable?)

packages/data-models/appConfig.ts
esmeetewinkel: Not sure I feel strongly either, but do also like to keep things human readable if possible. 

We do have an existing property `config.app_config.APP_LANGUAGES.default` which should maybe also be taken into account. I think currently the full list of languages is not specified in the config, is that correct? So you would only need to specify a language in the config if you want to change any of its default settings? 

Maybe something like: `config.app_config.LANGUAGES = { en: { default: true} , en_rtl: { rtl: true } }`

(default values of `default` and `rtl` are both `false`)

jfmcquade: > I think currently the full list of languages is not specified in the config, is that correct?
Yes that's right. I think that there is no master list of available languages currently and any value can be passed as the language code to the `emit: set_language` action. So a deployment's available languages are just those that are authored to be passed as values to that action, from any template, and it's up to the authors to ensure that the specified language codes have associated translations and assets. This seems a bit fragile, but a full overhaul is maybe outside of the scope of this PR. Requiring authors to specify the available languages in the deployment config would make sense to me, then when setting the language via the `emit: set_language` action we could run a check against the specified languages, throwing an error if the target language isn't in the list (this is what we do for the skins and themes, for example).

> Maybe something like: `config.app_config.LANGUAGES = { es: { default: true } , en_rtl: { rtl: true } }`
I like this suggestion, you're right that having separate system additional to `config.app_config.APP_LANGUAGES.default` doesn't make much sense. One negative I can see is that this syntax requires authors (at least those who are editing the deployment config) to explicitly write and understand nested javascript objects. But we already do require this (for example for the Firebase config settings) so I don't think it's a problem.

So a few of questions probably for @chrismclarke:
1. If we want to migrate to a system where languages and their metadata are handled within a single property in the app config, can we do better than the proposed syntax, `config.app_config.LANGUAGES = { es: { default: true } , en_rtl: { rtl: true } }`?
  - In this syntax, adding a language without any non-default metadata would require adding an empty object, e.g. `en_gb: {}` or else arbitrarily specifying a default property, e.g. `en_gb: { default: false }`. For themes and skins, we use an array to list the available entities and then use a separate property to specify which one is default. But here we already have two metadata properties (`default` and `rtl`) and may want to expand to have more in the future.
2. How would be best to handle the migration to this new system to minimise breaking changes to all deployment configs?
  - I'm thinking it would be sensible to mark the `APP_LANGUAGES` property as deprecated but continue to support it. This property is only used to get the name of the default language, so it should be straightforward to look in the new `LANGUAGES` property for a default language instead (with a fallback to the deprecated property)
3. Would it be desirable to require an exhaustive list of available languages in the config, and throw an error(/warning?) when trying to set the language to one that is not explicitly named?

chrismclarke: Good questions - my initial thoughts would probably be:

1. Probably better to store as an array so we can use the order to imply default if not explicitly set (or make it always first entry), and track more metadata from the start, e.g.
```
[
{country_code:'tz', language_code:'sw', label: 'Swahili'},
{country_code:'za', language_code:'af', label: 'Afrikaans'},
]
```
This would then allow authors in the future to more easily filter languages by country if useful to do so. If wanting to pass from a sheet then probably we could include a `fromSheet(...)` operator in the config that takes the name and reads in from there (again, future planning). At least if we have the list in the config then we can run QC checks to make sure language exists, and expose a protected variable to allow authors to see available list and select, e.g. `@field._language_options` or `@filed._language` (for active)

2. Agree using a new property and likely using scripts to migrate during deprecation process. I'd probably recommend swapping out `LANGUAGES` for `LOCALISATION` as we may use the property to contain more settings like how to handle dates within the current configuration.

3. I'd say still worth letting it be optional (if authors have hardcoded lists of languages into other sheets and want to use to select then no reason to stop them), but just use documentation to highlight benefit of using config (i.e. QC checks and protected variable population)

chrismclarke: Looping back to this PR then begs the question whether we want a short-term/intermediate solution or to do a slightly larger refactor now...? I don't have a strong feeling either way, I guess it depends on how urgent we want the RTL support.

esmeetewinkel (approved):
Functional test passed. I can see a few issues but will track these in a follow-up issue that we can address post merge.

packages/data-models/appConfig.ts
chrismclarke: @jfmcquade - might be worth pulling out more actionable parts of this discussion to #2254

PR #2296, opened 2024-04-19
Feat: user dynamic_data sync and import

chrismclarke:
PR Checklist

- [x] PR title descriptive (can be used in release notes)

## Description
**Main Changes**
- Add debug page to view current user contact fields and dynamic data
- Add migration to include `dynamic_data` column with server db
- Update app sync and import methods to include dynamic_data

**Additional Changes**
- Add a debounce to reduce the frequency user dynamic data is written to disk (max once per 500ms). This seemed a sensible optimisation given that import operations happen 1 row and a time, so could easily trigger a significant number of operations which may block the main thread and make the app hang
- Remove API `/tables` endpoint (as suggested in #2277)

**NOTE**
Production changes will only take effect once merged and server code updated (follow-up task)

## Review Notes
In order to populate dynamic data recommend viewing http://localhost:4200/template/comp_data_items

User debug page available at  http://localhost:4200/user

Sync testing requires a running development server environment (API + DB)

## Dev Notes
We don't currently have a set of spec tests to add db persist debounce test to, and feels beyond scope of this PR to create all from scratch. 

The persist operation has been tested locally by adding console logs to `src\app\shared\services\dynamic-data\adapters\persistedMemory.ts` public `persistStateToDB` and private `handleStatePersist` methods, and checking when importing user data that the the `handleStatePersist` method is only invoked once despite multiple calls to `persistStateToDB`.

## Author Notes
The `/user` debug page hasn't been included in the debug content app menu (but could in the future)

The list of dynamic_data is formatted as a single column in the existing `app_users` table, nested first by flow_type (for now all are data_lists) and then by flow_name. E.g. for the synced user the `dynamic_data` column entry is formatted as:
```json
{
  "data_list": {
    "comp_data_items_list": {
      "id_1": {
        "id": "id_1",
        "label": "Task 1",
        "completed": true
      },
      "id_2": {
        "id": "id_2",
        "label": "Task 2",
        "completed": false
      },
      "id_3": {
        "id": "id_3",
        "label": "Task 3",
        "completed": false
      }
    }
  }
}
```

## Git Issues

Closes #2246

## Screenshots/Videos
**User Sync and Import demo**
Create dynamic data entries using the `comp_data_items` template, and view available user data on new user debug page. 
Sync user data to db (running locally), and then clear cache to generate a new user profile.
Import initial user to see contact fields and dynamic data restored.

[...]

PR #2318, opened 2024-05-08
Server/updates 1.5.0

chrismclarke:
PR Checklist

## Description

- Bump API version to `1.5.1`
- Fix issue where local api connection failing (SSL not enabled locally)

## Review Notes
The updates have already been deployed to the server along with updates from #2296, so this PR is more just for reference.

It should now at least be possible to check the updates from #2296 on a live server (if using a deployment that has user backup/restore and been recently updated to include code)

## Dev Notes
I'm assuming the reason now that errors were being seen is either changes to the auto-update on postgres (pinned as `v13` but could have included in minor patch), or to the API packages themselves (been a while since we've pushed server changes and remember a few non-deployed from #2270

The error message seen was `no pg_hba.conf entry for host` (with various ip/db name combinations following), as described in `https://github.com/docker-library/postgres/issues/482`. Most solutions seemed to suggest updating postgres data files to support connection from any source, however this wasn't easy to implement as the postgres data files are mapped to a docker volume and not easily editable (like example that maps to local disk). So instead I managed to find a fix that forces NestJS to connect to the database using SSL instead.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "api",
-  "version": "1.4.4",
+  "version": "1.5.1",
   "scripts": {
     "prebuild": "rimraf dist",
     "build": "nest build",
--- a/packages/api/src/db/index.ts
+++ b/packages/api/src/db/index.ts
@@ -32,6 +32,14 @@ export class DBInstance {
         database: this.dbName,
         // disable verbose migration logs in test
         logging: process.env.NODE_ENV === "test" ? false : true,
+        // Fix SSL issue
+        // https://dev.to/rodjosh/connectionerror-sequelizeconnectionerror-no-pghbaconf-entry-for-host-in-heroku-postgresql-using-sequelize-3icj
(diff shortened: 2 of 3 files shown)


esmeetewinkel (approved):
I can confirm dynamic data is coming through on the server as expected
image
with values of the form `{"data_list": {"comp_data_items_list": {"id_1": {"id": "id_1", "_id": "id_1", "label": "Task 1", "_index": "@item._index + 1", "completed": false}, "id_2": {"id": "id_2", "_id": "id_1", "label": "Task 2", "_index": "0", "completed": true}}, "comp_task_prog_bar_dyn_data": {"id_1": {"id": "id_1", "completed": false, "row_index": 0}, "id_2": {"id": "id_2", "completed": false, "row_index": 1}, "id_3": {"id": "id_3", "completed": true}, "id_4": {"id": "id_4", "completed": true}, "id_5": {"id": "id_5", "completed": true}}}}`

PR #2320, opened 2024-05-14
Refactor/app fields

chrismclarke:
PR Checklist

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/data-models/appConfig.ts
+++ b/packages/data-models/appConfig.ts
@@ -25,28 +25,6 @@ const DYNAMIC_PREFIXES = [
   "raw",
 ] as const;
 
-/**
- * All localstorage fields will be prefixed with this
- * TODO - this has not been consistently applied so refactoring required
- * */
-const FIELD_PREFIX = "rp-contact-field";
-
-/**
- * Fieldnames hardcoded into the app
- * TODO - these have not been consistently applied so refactoring required
- * */
-const APP_FIELDS = {
-  APP_AUTH_USER: `${FIELD_PREFIX}._app_auth_user`,
-  APP_FIRST_LAUNCH: `${FIELD_PREFIX}._app_first_launch`,
-  APP_LANGUAGE: `${FIELD_PREFIX}._app_language`,
-  APP_SKIN: `${FIELD_PREFIX}._app_skin`,
-  APP_THEME: `${FIELD_PREFIX}._app_theme`,
-  APP_VERSION: `${FIELD_PREFIX}._app_version`,
-  CONTENT_VERSION: `${FIELD_PREFIX}._content_version`,
-  DEPLOYMENT_NAME: `${FIELD_PREFIX}._deployment_name`,
-  SERVER_SYNC_LATEST: `${FIELD_PREFIX}._server_sync_latest`,
-};
-
 const APP_LANGUAGES = {
   /** Language used during first load. If translations do not exist will default to source strings (gb_en) */
   default: "gb_en",
@@ -217,7 +195,6 @@ const TASKS = {
 };
 
 const APP_CONFIG = {
-  APP_FIELDS,
   APP_HEADER_DEFAULTS,
   APP_INITIALISATION_DEFAULTS,
   APP_AUTHENTICATION_DEFAULTS,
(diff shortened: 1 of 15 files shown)


jfmcquade (commented):
Thanks Chris, it's nice to see this system be rationalised somewhat, I like the changes to various services for the better.

A few notes below.

### Error when calling `getField()` without a key
An error was being thrown on launch because the task service calls `getField()` with an `undefined` value for the key.

I've pushed a commit which addresses this commit: https://github.com/IDEMSInternational/parenting-app-ui/pull/2320/commits/35d6674e7136e7309517e9790b1302b4c913c7bc
Alternatively, perhaps the check should sit at the level of the `LocalStorageService.get()` method.

### Protected field values
A spot check on the protected field values suggests that they are preserved between `master` and this feature branch, and can be updated as expected (e.g. setting theme and skin).

The `_app_first_launch` field appears to be set correctly

### User sync/import
As stated in the "Author Notes" section of #2270, protected fields are currently excluded from the user import action, so we wouldn't expect to see these values populated from the server when performing the user sync action. This is potentially something to consider as a follow up. Syncing other fields and dynamic data works as expected, with one caveat: when importing user data from the server, the value of the field `_server_sync_latest` is saved to a field called `server_sync_latest`, without the `_` prefix. See demo below:


https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/f0d546c8-4c92-4118-b1f3-3d51e616d696



### Field prefix
I have never understood the prefix "rp-contact-field", I assume this is inherited from RapidPro? Is now the time to rethink this prefix? EDIT: I see this addressed in the first item of the Future TODOs checklist

chrismclarke:
> ### Error when calling `getField()` without a key
> An error was being thrown on launch because the task service calls `getField()` with an `undefined` value for the key.
> 
> I've pushed a commit which addresses this commit: [35d6674](https://github.com/IDEMSInternational/parenting-app-ui/commit/35d6674e7136e7309517e9790b1302b4c913c7bc) Alternatively, perhaps the check should sit at the level of the `LocalStorageService.get()` method.

Interesting, what was the error. When I just try in my local browser `LocalStorage.getItem(undefined)` it returns `null` without any error. No objection to the catch in any case, more just curious.

> ### Protected field values
> A spot check on the protected field values suggests that they are preserved between `master` and this feature branch, and can be updated as expected (e.g. setting theme and skin).
> 
> The `_app_first_launch` field appears to be set correctly

Thanks for checking

> ### User sync/import
> As stated in the "Author Notes" section of #2270, protected fields are currently excluded from the user import action, so we wouldn't expect to see these values populated from the server when performing the user sync action. This is potentially something to consider as a follow up. Syncing other fields and dynamic data works as expected, with one caveat: when importing user data from the server, the value of the field `_server_sync_latest` is saved to a field called `server_sync_latest`, without the `_` prefix. See demo below:

Nice catch, I can see the `server.service` manually populates the contact_field property before sending to the server but doesn't add the prefix like the storage service does. This is the only place in the code I can see `PROTECTED_FIELDS` accessed directly, although noticed I had also forgotten to include the `_` prefix when removing a protected field so have instead added a protected field name accessor to avoid similar issues in the future 847c980b2dd382125d4900fa5863d270e36ef1ea

> ### Field prefix
> I have never understood the prefix "rp-contact-field", I assume this is inherited from RapidPro? Is now the time to rethink this prefix? EDIT: I see this addressed in the first item of the Future TODOs checklist

Yes all inherited from early focus on rapidpro which doesn't make a huge amount of sense anymore (although useful mostly as a prefix to disambiguate multiple apps on localhost). Would be nice to remove but we would first need a migration system to copy old values to new system

chrismclarke:
Thanks for looking over @jfmcquade - I've addressed the 1 issue you picked up on, let me know if you think any other changes are required. I've moved follow-up issues to https://github.com/IDEMSInternational/parenting-app-ui/issues/2322, feel free to update anything else you think of

esmeetewinkel (approved):
I can confirm `rp-contact-field._app_first_launch` is populated correctly. 

I noticed that there's a new contact field `app_events_summary`. Should that start with `_`?
image

Could someone resolve the conflicts between this PR branch and master?

chrismclarke:
> I can confirm `rp-contact-field._app_first_launch` is populated correctly, and also skin / theme survive when changing from master to the PR branch.
> 
> I noticed that there's a new contact field `app_events_summary`. Should that start with `_`? 

Mentioned briefly in the dev notes, this has previously been stored without the `rp-contact-field` prefix and so wasn't getting synced. Now it has the prefix but the legacy version will remain unless browser cache cleared. As for making it a protected `_` variable - that's not possible at this time as the protected variables don't yet support nested json (mentioned as possible follow-up). So for now I think probably just leave as-is and deal with in the future

> Could someone resolve the conflicts between this PR branch and master?
Done.

PR #2332, opened 2024-06-06
docs: replace all references to "parenting-app-ui" in the docs with "open-app-builder", including links to online documentation site

ChrisMarsh82:


jfmcquade (approved):
Good catch. I've gone ahead and replaced all other references within docs to "parenting-app-ui" with "open-app-builder". There are still some references within the code, but I think it makes sense to update all of the docs separately and at once.

PR #2340, opened 2024-06-20
test: remote assets spec test scaffolding

jfmcquade:
PR Checklist

## Description

- Adds scaffolding for spec tests for remote asset and file manager services.
- Minor code tidying of those services

## Git Issues

Closes #


src/app/shared/services/file-manager/file-manager.service.ts
```diff
@@ -48,11 +47,11 @@ export class FileManagerService extends SyncServiceBase {
...
   /**
-   * Save an asset file to device storage.
+   * Save an asset file to device storage, to be accessible to the user.
    * On native devices, save file to native storage and optionally trigger prompt to open with native apps.
    * On web, prompt standard browser download.
    * @param options.relativePath The relative path to an asset, in standard authoring syntax
-   * @param options.open On native devices, open the file after download. Default: true
+   * @param options.open If true, then the file will be opened after download on native devices. Default: true
    */
   private async downloadTemplateAsset(options: {
```
chrismclarke, line 56: nit(non-blocking) Not specifically related to the code but I've seen a few mentions of the `Documents` directory and just wanted to check if this is the folder you want? If using the `Data` directory then on IOS this is a proxy for the Documents directory, but on Android will be local application data directory. I'm not sure what is best (depends on whether you want these files to be accessible to other apps or not) 
https://capacitorjs.com/docs/apis/filesystem#directory

src/app/shared/services/remote-asset/remote-asset.service.spec.ts
```diff
@@ -1,21 +1,56 @@
...
   });
 
   it("should be created", () => {
     expect(service).toBeTruthy();
   });
 
-  it("generates core asset pack from asset contents", () => {
-    // TODO
+  it("generates a core asset pack from asset contents", () => {
+    const coreAssetPack = service["generateCoreAssetPack"](MOCK_ASSETS_CONTENTS_LIST);
```
chrismclarke, line 53: nit(non-blocking): any particular reason to access the method in this way instead of just `service.generateCoreAssetPack`? Doesn't make any functional difference, just curious

chrismclarke (approved):
thanks for separating this out into smaller chunks, will definitely make it easier to review the main code later. 

I've included a couple minor (non-blocking) comments, but otherwise all looks good to me

chrismclarke (approved):

src/app/shared/services/remote-asset/remote-asset.service.spec.ts
jfmcquade: This was a way I found of accessing a private method in the test. I'm not quite sure if it's a common or desirable pattern but it seems to work for cases like this.
Have you got any thoughts about testing private methods? I know that it's often suggested that only public methods should have unit tests, but in a case like this it seems like a good idea to have a test for this method, and the only reason to make it public would be to make it available for testing. I suppose it could live elsewhere as a util and have an associated unit test, but it is specifically a remote assets/asset pack util so it seems right that it should sit in the service file to me.

src/app/shared/services/file-manager/file-manager.service.ts
jfmcquade: Thanks for flagging. This was written before we has iOS integration and so I was only considering the Android case, but the intention was that this method (triggered by the `save_to_device` action) would download the file and make it accessible to other apps.

Those docs are a bit ambiguous to me, each of the possible storage locations is described slightly differently, even when they seem to be referring to the same location, and the only explicit mention that files will be available publicly to other apps is for the combination of `DOCUMENTS` and the Android platform. I think I'll have to handle this with an explicit check for the platform and change the storage directory accordingly – my best guess would be that I want to target `EXTERNAL_STORAGE` on iOS, but I can experiment with which location works for making the file publicly available.

src/app/shared/services/remote-asset/remote-asset.service.spec.ts
chrismclarke: Oh, interesting! I think I remember trying similar before, cool that it just works like that.

As a general rule of thumb I'd say most of the time can assume private methods will be called somewhere from public methods, and so can usually be tested indirectly. In this case the private method is called during a separate private init so might be harder to track except using the methods mocked from the `templateAssetService` which is reads from and updates as part of the init workflow. 

Utils are another way to go, although this method doesn't really do much in the lines of manipulation (converts hashmap to array, which possibly we could add general util for given we have `arrayToHashmap`, and populates as a flow type). So I think for this case I'd lean in to making the `hashmapToArray` a util with spec tests, moving the side-effect code `addRuntimeFlowToContents` to the init flow instead of within the function call, and skipping the need to test that it essentially just populates the generated list within the flow structure (as typescript would give error if that structure changed). 

But more generally I don't have any issue with what you've done - tests are useful to have, I agree best to not change core code just to accomodate tests (unless no reasonable way around), and particularly if code gets longer/more complex adding intermediate tests for private methods still makes sense in theory to me.

PR #2342, opened 2024-06-21
fix: data pipe merge operator handles case of no input source

### LINKED ISSUE #2341: [BUG] Data pipe `merge` operation errors
I can't get the `merge` operation to work (might be just my lack of understanding). See [example_data_pipe_merge](https://docs.google.com/spreadsheets/d/1imxKEBQ31qpdu7NeXuH3UyK88G_G1jitLdw9CRZ9ZIY/edit#gid=751466229) 

**Use case**
_PLH Kids KW deployment_: I want to compose "modules" out of "articles" that appear on different lists. To achieve this, I'm [generating this data_pipe](https://docs.google.com/spreadsheets/d/11GlK8DVUWwHr1AAKbqta7SEI1kHbNyW3nTGHCrOoyiA/edit#gid=1227678473) over the list of modules: 
image

When syncing this sheet, I get the terminal error 
image

jfmcquade:
PR Checklist

## Description

The data pipe `merge` fails and throws an unhandled error in the case that no `input_source` is provided. This PR handles that case by throwing a custom error in that case. 

Also tidies some of the pre-existing data pipe operator tests (thanks, @chrismclarke).

#2341 also suggests the existence of another possible issue to do with applying the merge operator to data lists that contain nested JSON data. I think it is sensible to handle this case as a follow-up if it does persist after applying the changes of this PR. In that case, a new debug sheet that exhibits the broken behaviour would be required.

## Testing

See [example_data_pipe_merge](https://docs.google.com/spreadsheets/d/1imxKEBQ31qpdu7NeXuH3UyK88G_G1jitLdw9CRZ9ZIY/edit#gid=751466229). Run a sync and ensure no errors are thrown, then inspect the outputs of the data pipe to confirm they are as expected.

## Git Issues

Closes #2341 

## Screenshots/Videos

Attempting to run the data pipe as defined by [example_data_pipe_merge](https://docs.google.com/spreadsheets/d/1imxKEBQ31qpdu7NeXuH3UyK88G_G1jitLdw9CRZ9ZIY/edit#gid=751466229) in the following state:
Screenshot 2024-06-21 at 10 33 15
now logs the following error:
Screenshot 2024-06-25 at 12 51 02

### Passing tests
Screenshot 2024-06-25 at 12 51 55

### DIFF
--- a/packages/shared/src/models/dataPipe/operators/filter.spec.ts
+++ b/packages/shared/src/models/dataPipe/operators/filter.spec.ts
@@ -23,10 +23,10 @@ describe("Filter", () => {
     expect(outputIDs).toEqual(["id_1", "id_3", "id_4"]);
   });
   it("Filters with 'this' context", () => {
-    const nestedData = testData.names.map((entry) => {
-      entry["nested"] = { first_name: entry.first_name };
-      return entry;
-    });
+    const nestedData = testData.names.map((entry) => ({
+      ...entry,
+      nested: { first_name: entry.first_name },
+    }));
     const testDf = new DataFrame(nestedData);
     const output = new filter(testDf, ["this.nested.first_name === 'Ada'"]).apply();
     const outputIDs = output.column("first_name").values;
--- a/packages/shared/src/models/dataPipe/operators/merge.spec.ts
+++ b/packages/shared/src/models/dataPipe/operators/merge.spec.ts
@@ -58,17 +58,4 @@ describe("Merge Operator", () => {
     const expectedNested = ['{"value":"test"}', "undefined", "undefined", "undefined"];
     expect(output.column("nested").values).toEqual(expectedNested as any);
   });
-  it("Merges multiple lists from args_list, with no input_source", () => {
-    const emptyDf = new DataFrame([]);
-    // merges data - additional nationality column appended for all entries and populated for available
-    const output = new merge(emptyDf, ["names", "nationality_data"], testPipe).apply();
-
-    expect(output.index).toEqual(["id_1", "id_2", "id_3", "id_4"]);
-    // merges new nationality column
-    const expectedNationalities = ["British", "French", undefined, undefined];
-    expect(output.column("nationality").values).toEqual(expectedNationalities);
-    // merges existing name overrides
-    const expectedNames = ["override", "Blaise", "Charles", "Daniel"];
-    expect(output.column("first_name").values).toEqual(expectedNames);
-  });
 });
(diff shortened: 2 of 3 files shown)


chrismclarke (approved):
Looks fine to me, although still have the issue in the back of my mind as to whether it actually makes sense to use the args list this way (given we have the input_source column), particularly given potential long-form args_list syntax in the future, e.g. `merge_source: list_b; join_type: outer`

https://chat.idems.international/all/pl/bji1h4pxbfbzdeiso9pzcx46re

jfmcquade:
> Looks fine to me, although still have the issue in the back of my mind as to whether it actually makes sense to use the args list this way (given we have the input_source column), particularly given potential long-form args_list syntax in the future, e.g. `merge_source: list_b; join_type: outer`
> 
> https://chat.idems.international/all/pl/bji1h4pxbfbzdeiso9pzcx46re

Thanks, we discussed this on a call and decided that it would be better not to support the authoring style of providing a list of data sources in the `args_list` and no data source in the `input_source`, for the reasons you have given. I have reworked this PR, see the updated description.

We noted that the [Data Pipelines RFC](https://docs.google.com/document/d/1Pk2gYNhtu68XJo-U-h6RY8ZD8gOPaCnLOlCft4IjLY4/edit#heading=h.ooapz64v1x3w) includes the proposal to support an explicit key/value syntax for the arguments passed to the `args_list`, which we approved of.

We also noted that it might make sense for cases like the `merge` operator to support passing a list of input sources to the `input_source` column (although not to the `args_list` column). And also that there may be good reasons to also support a list of outputs to the `output_target` column. I'm not sure where the best place to capture this is, potentially as part of that RFC, or I could write a new issue.

esmeetewinkel:
> https://github.com/IDEMSInternational/open-app-builder/issues/2341 also suggests the existence of another possible issue to do with applying the merge operator to data lists that contain nested JSON data. I think it is sensible to handle this case as a follow-up if it does persist after applying the changes of this PR. In that case, a new debug sheet that exhibits the broken behaviour would be required.

Debug sheet: [example_data_pipe_merge_nested](https://docs.google.com/spreadsheets/d/1imxKEBQ31qpdu7NeXuH3UyK88G_G1jitLdw9CRZ9ZIY/edit?gid=1005960371#gid=1005960371)
which merges a [data list with a translatable column](https://docs.google.com/spreadsheets/d/1imxKEBQ31qpdu7NeXuH3UyK88G_G1jitLdw9CRZ9ZIY/edit?gid=698527884#gid=698527884) (which is nested as .json) 
image

with another data list. Syncing this gives the error: 
image

esmeetewinkel (approved):

chrismclarke:
I think this is good to merge - I'll address the specific issue raised by @esmeetewinkel in a follow-up PR

PR #2344, opened 2024-06-25
Fix/data pipe merge nested

### LINKED ISSUE #2343: [BUG] Data pipe merge operator doesn't handle nested lists
> https://github.com/IDEMSInternational/open-app-builder/issues/2341 also suggests the existence of another possible issue to do with applying the merge operator to data lists that contain nested JSON data. I think it is sensible to handle this case as a follow-up if it does persist after applying the changes of this PR. In that case, a new debug sheet that exhibits the broken behaviour would be required.

Debug sheet: [example_data_pipe_merge_nested](https://docs.google.com/spreadsheets/d/1imxKEBQ31qpdu7NeXuH3UyK88G_G1jitLdw9CRZ9ZIY/edit?gid=1005960371#gid=1005960371)
which merges a [data list with a translatable column](https://docs.google.com/spreadsheets/d/1imxKEBQ31qpdu7NeXuH3UyK88G_G1jitLdw9CRZ9ZIY/edit?gid=698527884#gid=698527884) (which is nested as .json) 
image

with another data list. Syncing this gives the error: 
image

_Originally posted by @esmeetewinkel in https://github.com/IDEMSInternational/open-app-builder/issues/2342#issuecomment-2188756018_

chrismclarke:
PR Checklist

## Description
Builds on top of #2342 to address specific bug where `_translations` data could not be included in in the base data_list of a merge operator.

## Review Notes
Only changes beyond #2342 are to address the bug identified in #2343. So if functional testing passed should be able to assume good to go (I've already reviewed the code written in #2342 and seen changes since)

## Dev Notes
The core issue appeared to be a bug within the `danfo.js` dataframe `apply` method - this should pass an array of all row values to manipulate, however it omitted any values that were formatted as nested json, and hence when using a _translations column the overall return array would be 1 element shorter than expected (dimension error).

I've updated the test spec to cover, runnable via `yarn workspace shared test`
See screenshots below for before/after (should capture same issue seen).

I've also changed the console error on empty data frame to throw, and included in the tests so that it is more explicit to any author (thrown errors are identified in final output logs whereas console errors can be more easily missed)

## Git Issues
Closes #2343

## Screenshots/Videos

Tests - Before
image

Tests - After
image

### DIFF
--- a/packages/shared/src/models/dataPipe/operators/merge.spec.ts
+++ b/packages/shared/src/models/dataPipe/operators/merge.spec.ts
@@ -58,17 +58,4 @@ describe("Merge Operator", () => {
     const expectedNested = ['{"value":"test"}', "undefined", "undefined", "undefined"];
     expect(output.column("nested").values).toEqual(expectedNested as any);
   });
-  it("Merges multiple lists from args_list, with no input_source", () => {
-    const emptyDf = new DataFrame([]);
-    // merges data - additional nationality column appended for all entries and populated for available
-    const output = new merge(emptyDf, ["names", "nationality_data"], testPipe).apply();
-
-    expect(output.index).toEqual(["id_1", "id_2", "id_3", "id_4"]);
-    // merges new nationality column
-    const expectedNationalities = ["British", "French", undefined, undefined];
-    expect(output.column("nationality").values).toEqual(expectedNationalities);
-    // merges existing name overrides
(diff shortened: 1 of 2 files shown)


esmeetewinkel (approved):
Functional test passed

jfmcquade (approved):
Looks good. Good that the solution can be this minimal workaround, and nice to have the test coverage. I can also see why throwing an error in the case of no `inpuit_source` is preferable to just logging it.

PR #2351, opened 2024-07-09
Fix/consistency with button animations

### LINKED ISSUE #2232: [FEATURE] Consistency across button interactions
**What?**
Currently, button-like components can exhibit one of three different behaviours when clicked on:
1. An animated ripple effect, e.g. most variations of the button component

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/a157ce21-9724-45a7-978f-54f28134476a

2. A static colour change, e.g. button with `variant: card-portrait`

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/b0109fa7-afae-46f5-ba3a-2fc384d895f3

3. No interaction, e.g. tile components

https://github.com/IDEMSInternational/parenting-app-ui/assets/64838927/cc4471eb-cc45-434a-be09-d85784b5d731

These interactions should be standardised across all button-like components, likely favouring the ripple effect.

**Why?**
Ensuring that similar components feel and behave similarly improves the UX

**How?**
We should be able to add the [ion-ripple-effect](https://ionicframework.com/docs/api/ripple-effect) to all relevant clickable components

kwAsant:
PR Checklist

## Description

Adds the ripple effect to most buttons specified.
This ensures that consistency is maintained.

## Git Issues

Closes #2232 

## Screenshots/Videos
Evidence is for button types I am currently aware of:
1. Button card (no changes were made)
https://github.com/IDEMSInternational/open-app-builder/assets/88288104/c33ba940-3295-446a-903a-d85c73ae9244

2. Button card portrait
https://github.com/IDEMSInternational/open-app-builder/assets/88288104/19d51b3e-0b5b-48bf-888e-c578d45b24d9

3. Disabled parameter buttons
https://github.com/IDEMSInternational/open-app-builder/assets/88288104/930b66ae-6b91-46fa-a314-1a74804e0b2b

4. Button colour buttons
https://github.com/IDEMSInternational/open-app-builder/assets/88288104/827b7440-9afb-4fbc-9f7a-0b42bf142ce8

5. Button width
https://github.com/IDEMSInternational/open-app-builder/assets/88288104/7e66672d-25d9-434e-80e5-6271f7877d65

6. Button height
https://github.com/IDEMSInternational/open-app-builder/assets/88288104/6f369893-940a-49cc-8710-11e9c9e617b2

7. Button text colour
https://github.com/IDEMSInternational/open-app-builder/assets/88288104/c7799b23-309c-4b29-ba82-d8e5984122dc

The rest were omitted because they were under the subheading: "To be removed"

### DIFF
--- a/src/app/shared/components/template/components/button/button.component.html
+++ b/src/app/shared/components/template/components/button/button.component.html
@@ -49,6 +49,7 @@
         [parent]="parent"
         [attr.data-rowname]="_row.name"
       ></plh-template-component>
+      <ion-ripple-effect></ion-ripple-effect>
     </span>
   </div>
 </ng-container>
--- a/src/app/shared/components/template/components/button/button.component.html
+++ b/src/app/shared/components/template/components/button/button.component.html
@@ -25,6 +25,7 @@
         [attr.data-rowname]="_row.name"
       ></plh-template-component>
     </span>
+    <ion-ripple-effect></ion-ripple-effect>
   </ion-button>
 
   <!-- "card-portrait" variant. This variant is not achievable using an ion-button -->
--- a/src/app/shared/components/template/components/button/button.component.html
+++ b/src/app/shared/components/template/components/button/button.component.html
@@ -31,11 +31,12 @@
   <!-- "card-portrait" variant. This variant is not achievable using an ion-button -->
   <div
     *ngSwitchCase="true"
-    class="button-container"
+    class="button-container ion-activatable"
     (click)="triggerActions('click')"
     [attr.data-variant]="params.variant"
     [attr.data-has-children]="_row.rows ? true : false"
   >
+    <ion-ripple-effect type="unbound"></ion-ripple-effect>
     <img *ngIf="params.image" src="{{ params.image | plhAsset }}" />
     <div *ngIf="_row.value" [class]="'button-text ' + params.textAlign">
       {{ _row.value }}
--- a/src/app/shared/components/template/components/tile-component/tile-component.component.scss
+++ b/src/app/shared/components/template/components/tile-component/tile-component.component.scss
@@ -41,7 +41,7 @@ $background-secondary-light: var(
       --box-shadow: none;
 
       box-shadow: none;
-      --ripple-color: transparent;
+      // --ripple-color: transparent; // why was this there?
       img {
         width: 60px;
         height: 60px;
@@ -61,6 +61,7 @@ $background-secondary-light: var(
   box-shadow: var(--ion-default-box-shadow);
   min-height: var(--tile-default-height);
   border-radius: var(--ion-border-radius-standard);
+  overflow: hidden;
 }
 
 /***********************************************************************************************
--- a/src/app/shared/components/template/components/button/button.component.scss
+++ b/src/app/shared/components/template/components/button/button.component.scss
@@ -92,6 +92,7 @@ ion-button[data-variant~="card"] {
   padding: 1.5rem 0.5rem 0.5rem;
   // TODO: make a variable
   width: 160px;
+  overflow: hidden;
 
   &:active {
     background-color: var(--ion-color-gray-light);
--- a/src/app/shared/components/template/components/tile-component/tile-component.component.html
+++ b/src/app/shared/components/template/components/tile-component/tile-component.component.html
@@ -24,6 +24,7 @@
         alt=""
       />
       <ion-icon *ngIf="!isCustomIcon" [name]="icon_src"></ion-icon>
+      <ion-ripple-effect></ion-ripple-effect>
     </ion-button>
   </div>
 </div>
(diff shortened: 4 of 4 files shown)


src/app/shared/components/template/components/tile-component/tile-component.component.scss
```diff
@@ -41,7 +41,7 @@ $background-secondary-light: var(
       --box-shadow: none;
 
       box-shadow: none;
-      --ripple-color: transparent;
+      // --ripple-color: transparent; // why was this there?
```
jfmcquade, line 44: I think it's there to avoid the unexpected behaviour of the icon having a ripple effect when you click it, but not the whole "tile" element. Since the component uses an `ion-button` within it, this sub-element gets the ripple effect applied automatically. I've updated the [example_tiles](https://docs.google.com/spreadsheets/d/1UWZDMskUnAxkGMmEv9uEpFetXQLdMRdu2ecXEKcbTjo/edit?gid=1595229908#gid=1595229908) template so that the image links are no longer broken. Uncomment this line and experiment with it yourself, it's strange that just the icon has an effect when clicked.

I'm not entirely sure why the component is written this way, using an `ion-button` as a wrapper around where it could just render the icon in a `div`. I suspect this is just some old code that could be rewritten. However, I don't think that's worth worrying about – I think you should revert the changes to the `tile-component` files for now so that this PR only affects the `button` component.

Also, for comments like this that are intended for the reviewer of the PR, they should be added as a comment within github once you've created the PR (I can show you how to do this), rather than in the code itself. If we were to merge these changes, this comment would be permanently committed to the codebase and wouldn't make sense to someone coming across it without any context.

src/app/shared/components/template/components/button/button.component.html
```diff
@@ -25,16 +25,18 @@
...
   <!-- "card-portrait" variant. This variant is not achievable using an ion-button -->
   <div
     *ngSwitchCase="true"
-    class="button-container"
+    class="button-container ion-activatable"
     (click)="triggerActions('click')"
     [attr.data-variant]="params.variant"
     [attr.data-has-children]="_row.rows ? true : false"
   >
+    <ion-ripple-effect type="unbound"></ion-ripple-effect>
```
jfmcquade, line 39: From the [docs](https://ionicframework.com/docs/api/ripple-effect#type), it looks like the two types are  `"bounded"` and `"unbounded"`, so `"unbound"` will have no effect. I don't think we want the effect to be unbounded, so you can remove this `type` attribute.

jfmcquade (changes requested):

src/app/shared/components/template/components/tile-component/tile-component.component.scss
kwAsant: Oh whoops, I forgot to remove that comment. When I checked, I think it made the ripple not easy to see and once I removed it, it became visible and had the default colour it needed.

src/app/shared/components/template/components/button/button.component.html
kwAsant: Will do!

src/app/shared/components/template/components/button/button.component.html
```diff
@@ -25,16 +25,18 @@
         [attr.data-rowname]="_row.name"
       ></plh-template-component>
     </span>
+    <ion-ripple-effect></ion-ripple-effect>
```
jfmcquade, line 28: The `card` variant already had a ripple effect, so this can be removed as it is redundant

src/app/shared/components/template/components/button/button.component.html
```diff
@@ -47,6 +49,7 @@
         [parent]="parent"
         [attr.data-rowname]="_row.name"
       ></plh-template-component>
+      <ion-ripple-effect></ion-ripple-effect>
```
jfmcquade, line 52: This can be removed – the button itself already has a ripple effect applied by line 39

jfmcquade (changes requested):
Apologies, I've just returned to review this PR. See my comments inline.

Additionally, there seems to be a discrepancy between the ripple effect applied to the button `card` variant and the `card-portrait` variant. The former is smooth and slow, but the latter appears fast and glitchy (possibly two ripple effects being applied?). See video below for a demo, see if you can figure out how to resolve this.

https://github.com/user-attachments/assets/145d9e22-ce7f-42dd-89ca-a3ef34b87c59

Also I think it might be best to revert the changes made to the tile component so that this PR just applies to the `card-portrait` button variant. Then you could update the PR title and description accordingly.

kwAsant:
> Apologies, I've just returned to review this PR. See my comments inline.
> 
> Additionally, there seems to be a discrepancy between the ripple effect applied to the button `card` variant and the `card-portrait` variant. The former is smooth and slow, but the latter appears fast and glitchy (possibly two ripple effects being applied?). See video below for a demo, see if you can figure out how to resolve this.
> 
>  ripple.mov 
> 
> Also I think it might be best to revert the changes made to the tile component so that this PR just applies to the `card-portrait` button variant. Then you could update the PR title and description accordingly.

Improved by changing the duration of all ripple effects by using `ion-ripple-duration`. This allows the animations to be consistent across all buttons. 

https://github.com/user-attachments/assets/cd1faa94-bff4-4688-80d9-b224c51de94c

Due to the dimensions of the buttons, I think it may create an idea that it is moving faster than the other. From reading the doc, the ripple effect works by trying to move across the button to its dimensions. It seems to aim to do it by an unknown time, but it depends on how wide and long the button is.

PR #2352, opened 2024-07-17
Chore: Migrate toggle bar to use modern ion toggle syntax

### LINKED ISSUE #2212: [FEATURE] toggle bar component: migrate to use modern ion-toggle sytnax
**What?**
The `toggle_bar` component makes use of the `ion-toggle` component, adding custom styling. This styling makes use of a syntax that ionic now considers to be [legacy](https://ionicframework.com/docs/api/toggle#using-the-legacy-syntax). We should update to use the new "[modern syntax](https://ionicframework.com/docs/api/toggle#using-the-modern-syntax)".

**Why?**
This legacy syntax will be deprecated in an upcoming major release of ionic.

**How?**
Migrating the basic styling should be reasonably straightforward, but accounting for our custom icons, enabled via the `show_tick_and_cross` parameter, is likely to be a bit more involved. We should consider replacing our custom on/off labels with a feature introduced for `ion-toggle`, accessed via a new [enableOnOffLabels](https://ionicframework.com/docs/api/toggle#on--off-labels) property. It appears these icons can't be customised: there is only one option for Android and one option for iOS, so we should decide whether it would be acceptable to use these instead of our custom implementation.
Native ionic on/off labels (from [here](https://ionicframework.com/docs/api/toggle#on--off-labels)):
Screenshot 2024-02-23 at 14 08 49
Screenshot 2024-02-23 at 14 08 54
Screenshot 2024-02-23 at 14 08 59
Screenshot 2024-02-23 at 14 09 04

kwAsant:
PR Checklist

## Description

- Migrated from the legacy syntax for the ion-toggle component (see [docs](https://ionicframework.com/docs/api/toggle)) to the newly updated one
Then for the last bullet point, I would add something like:
- Updated the [sheet](https://docs.google.com/spreadsheets/d/17SbJITgQSdNs3PNusDJrkbmCLDutOH9jN_qiea8r0zo/edit?gid=569531329#gid=569531329) for testing 
- Instead of using `showTickAndCross`, there is a new parameter called `showIcons`. `showTickAndCross` is still supported but should be considered legacy
- Now, we use `variant` to be an identifier for `ios`, `android` along with `icon` and `in_button`.

## Git Issues

Closes #2212

## Screenshots/Videos

Toggles off:

image

image

Toggles on:

image

image


src/app/shared/components/template/components/toggle-bar/toggle-bar.html
```diff
@@ -11,12 +11,11 @@
...
     >
-      <ion-toggle
-        mode="md"
-        [checked]="!!_row.value"
-        (click)="handleClick($event)"
-        legacy="true"
-      ></ion-toggle>
+      <ion-toggle 
+        mode="ios"
+        [enableOnOffLabels]="params.showTickAndCross"
```
jfmcquade, line 16: This should also be affected by the new `showIcons` param (see my other inline comment). In order to handle the default being to show the icons whilst supporting the legacy functionality of `show_tick_and_cross: false`, I think the logical statement you'll want here is `params.showTickAndCross && params.showIcons`. As long as both properties default to the value `true`, that should handle the logic correctly.

jfmcquade (changes requested):
- From the screenshots, it now looks like the "handle" of the toggle bar is too close to the edge of the component, I would expect more of a gap all the way around. Comparing the screenshots to the examples in [the docs](https://ionicframework.com/docs/api/toggle), it looks like we must be applying some custom styling (likely via the `--handle-spacing` CSS property) that should be updated/removed so that there is a uniform border all the way around
    Screenshot 2024-07-17 at 11 46 34

- See my inline comments for some other suggested changes. After making these changes, it would also be good to update the PR description to be a bit more descriptive. It should include a reference to the new `show_icons` parameter and explain how this replaces the legacy `show_tick_and_cross` parameter, which is itself still supported.
- A minor point: the PR title should follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) conventions. In this case I would probably say this is a "chore", so I would add the prefix `chore: ` to the title. This means that the PR will be tagged correctly by our automated system when included in [release notes](https://github.com/IDEMSInternational/open-app-builder/releases).
- You could consider exposing the `mode` property to authors so that they can decide whether the mode should be `ios` or `md` when authoring the component in a template. I think it might make sense to handle this via the `variant` parameter, with options including `ios` and `android` (where `ios` is the default). This is an extension, so I would only consider this once you've got the rest working.

src/app/shared/components/template/components/toggle-bar/toggle-bar.scss
```diff
@@ -9,34 +9,13 @@ ion-toggle {
 
 ion-toggle {
   --track-background: #b4b7b7;
```
jfmcquade, line 11: This colour should be read from our theming system rather than being hardcoded. I haven't shown you this yet, but try replacing this hex value with `var(--ion-color-gray-200)`, and then experiment with different values instead of `200` to find a grey colour that looks right. Using a variable in this way means that when the global theme is changed, this colour will update automatically

src/app/shared/components/template/components/toggle-bar/toggle-bar.ts
```diff
@@ -11,8 +11,9 @@ interface IToggleParams {
   variant: "" | "icon" | "in_button";
   /** TEMPLATE PARAMETER: "style". Legacy, use "variant" instead. */
   style: string;
-  /** TEMPLATE PARAMETER: "show_tick_and_cross" */
+  /** TEMPLATE PARAMETER: "show_tick_and_cross". Legacy, use "show-icons" instead */
   showTickAndCross: boolean;
+  showIcons: boolean;
```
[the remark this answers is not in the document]
kwAsant: Yes, I was supposed to ask about this, I wasn't too sure what to do from it.

src/app/shared/components/template/components/toggle-bar/toggle-bar.html
kwAsant: I will make the default iOS because it looks better generally and it is much clearer to show the tick and cross icons.

src/app/shared/components/template/components/toggle-bar/toggle-bar.scss
kwAsant: Done so, there seemed to be no need to experiment.

kwAsant:
I am struggling to find the specific part of the HTML which is the check mark (when the toggle bar is turned on).

src/app/shared/components/template/components/toggle-bar/toggle-bar.ts
```diff
@@ -66,10 +69,12 @@ export class TmplToggleBarComponent
       "show_tick_and_cross",
       true
     );
+    this.params.showIcons = getBooleanParamFromTemplateRow(this._row, "show_icons", true);
     this.params.style = getStringParamFromTemplateRow(this._row, "style", "");
     this.params.variant = getStringParamFromTemplateRow(this._row, "variant", "")
       .split(",")
       .join(" ") as IToggleParams["variant"];
+    this.platformVariant = this.params.variant.split(" ").includes("android") ? "md" : "ios";
```
jfmcquade, line 77: nit (non-blocking): Maybe worth adding an explanatory comment above this line, something like:
```ts
// The ion-toggle component uses "md" ("material design") and "ios" to refer to visual styles of the component corresponding to the respective platforms.
// see docs here: https://ionicframework.com/docs/api/toggle
```

jfmcquade (approved):
Looks good, well done. I've added one minor comment inline

esmeetewinkel (approved):
Functional test passed while running the app locally.

Could you update the notes to clarify what is the expected behaviour on web, android and ios? I saw Johnny's comment

>  I think it might make sense to handle this via the variant parameter, with options including ios and android (where ios is the default)

So I would expect that 
- if variant is ios the toggle looks like ios on all platforms
- if variant is android the toggle looks like android on all platforms
- if no variant is specified (i.e. the default) then the toggle looks like ios on all platforms

Is that correct? Alternatively I could imagine that the variant specified is only for the web previews of the app, and that the visual style used on android is always android and on ios always ios.

Is it worth updating the naming so that the confusion between the visual style and the platform used to deploy the app disappears?

src/app/shared/components/template/components/toggle-bar/toggle-bar.ts
chrismclarke: (non-blocking): I can see that the _default_ would be `ios` (unless android specified) - which would include behaviour on web. I think given more of our user-base is android makes sense for that to be the default on web also - but not a huge issue either way (the ios style is still nice)

chrismclarke (approved):

jfmcquade:
Thanks everyone. I've added a [minor commit](https://github.com/IDEMSInternational/open-app-builder/pull/2352/commits/d41a3f762cb48709805a21f9ad38b816b6ad312b) which changes and clarifies the default "platform variant" logic:
- If the author specifies the variant as "android" or "ios", then this will always be reflected in how the toggle appears across all platforms
- If no variant is explicitly set by the author, the toggle will default to having the "android" appearance on _Android_ and web, else on iOS it will have the "ios" appearance

src/app/shared/components/template/components/toggle-bar/toggle-bar.ts
```diff
@@ -7,7 +8,10 @@ import {
 } from "src/app/shared/utils";
 
 interface IToggleParams {
-  /** TEMPLATE PARAMETER: "variant" */
+  /**
+   * TEMPLATE PARAMETER: "variant". Setting "ios" or "android" will style the toggle to match the respective
+   * platform, otherwise the default is to match the current device platform, using "ios" on web.
```
esmeetewinkel, line 13: @jfmcquade I think this code comment line disagrees with what you've said in [your comment on the PR](https://github.com/IDEMSInternational/open-app-builder/pull/2352#issuecomment-2305169855)(?): 

>If no variant is explicitly set by the author, the toggle will default to having the "android" appearance on Android and web, else on iOS it will have the "ios" appearance

So - on web is the default to use `ios` or `android`?

jfmcquade: On web the default is `android`. I made this default in response to Chris's [comment](https://github.com/IDEMSInternational/open-app-builder/pull/2352#discussion_r1725936151) that most of our user-base use Android, and so it probably makes sense for the web preview to match the styling that they will see.

PR #2353, opened 2024-07-17
feat: new component, 'text_bubble'

### LINKED ISSUE #2346: [FEATURE] Speech bubbles
**What?**
Option to display content (normally text, images or an audio file) in a "speech bubble". 

image

**Why?**
Requested by ParentApp Kids TZ and KW.

**How?**
We need to be able to specify
- file path for image of the "speaker". (optional - I think even if this file if missing the design makes sense)
- whether the speech bubble "points" to the left- or right-hand-side of the screen.
- background colour of the speech bubble. I think this could be the same variants as the display group has 
image

My initial thought was that this entire component could be interpreted as an extension of the recently implemented display group variants, potentially using the `icon_src` parameter for the image file path (which is already in use by certain styles) and I believe there's also an existing parameter for the image position (top-left, top-right, bottom-left, bottom-right). However, I'm not sure it's in-line with our thinking around variants (since the `parameter_list` would contain a variant + subparameters that potentially make sense to that variant only).

jfmcquade:
PR Checklist

## Description

- Adds a new component: `text_bubble`.
- Minor code tidying of the `display_group` component

### Text bubble component
Content:
- Text passed as the `value` will be rendered as text within the bubble
- Optionally, the component can be instantiated as a `begin_`/`end_` block, in which case any child rows will be rendered inside the bubble (see examples below)

Parameters: 
- `speaker_image_asset`
  - A path to an image to be displayed as the "speaker" of the speech bubble
- `speaker_position`
  - The side on which the image and speech bubble tail will appear, possible values are `"left"` (default) or `"right"`.

Additionally, there are four component variants:
- `gray` (default)
- `primary`
- `secondary`
- `no_border`
  - This can be used in combination with the other variants

Whilst these style changes could be perhaps be better applied via new `style` properties, as proposed in #1971 (e.g. instead of including `variant: primary` in the `parameter_list`, we could include `background-primary; border-primary` in the `style_list`), implementing that system is beyond the scope of this PR, and variants that bundle a set of styles together is an agreed upon pattern.

## Git Issues

Closes #2346

## Screenshots/Videos

[comp_text_bubble](https://docs.google.com/spreadsheets/d/1Yo8X4TrTW2lzEgusDplAQ-y-JJvQOIqXGNLVON1uvRo/edit?gid=569531329#gid=569531329)

Screenshot 2024-07-23 at 15 13 20

Screenshot 2024-07-23 at 15 12 52
Screenshot 2024-07-23 at 15 13 00


chrismclarke (changes requested):
I'm not entirely sure this makes sense as a child of the display-group. It feels like it adds a lot of custom code to the component without benefitting much from it (beyond inheriting a couple of colour variables).

I think it would make a lot more sense as a standalone `text-bubble` (or similar) component. It would also then be easier to extend in the future if wanting things like different shaped callouts (e.g. thought bubble, chat message), and would likely simplify the params passed.

I've tried to migrate the changed and start to scaffold out a new component with https://github.com/IDEMSInternational/open-app-builder/pull/2354

I couldn't quite make sense of the before/after CSS used to generate the tail (not sure if you followed a guide online somewhere, if yes would recommend including a link in the CSS file next time to be more explicit), so instead have implemented the tail in a different way (see notes in file)

@jfmcquade - let me know if you're happy to take over the draft pr or whether you would prefer to merge the display group version first and work on an improved version as a follow-up

jfmcquade:
> I'm not entirely sure this makes sense as a child of the display-group. It feels like it adds a lot of custom code to the component without benefitting much from it (beyond inheriting a couple of colour variables).
> 
> I think it would make a lot more sense as a standalone `text-bubble` (or similar) component. It would also then be easier to extend in the future if wanting things like different shaped callouts (e.g. thought bubble, chat message), and would likely simplify the params passed.

Yes I see what you mean, I agree that a standalone component probably makes more sense. Alongside the motivations you've mentioned, having variant-specific component parameters also felt a bit messy, although will likely sometimes be desirable (variable-_relative_ parameters could be interesting too...). The issue (#2346) does suggest that we'll likely want to put arbitrary content inside the bubble, not just text strings, which was part of the motivation for building on top of the display group, but we do already have other components that render child rows within themselves and I don't think that's a problem.

Thanks for setting up the branch PR for the other implementation, I will indeed work on that rather than suggesting we merge the display group version first.

chrismclarke:
> Yes I see what you mean, I agree that a standalone component probably makes more sense. Alongside the motivations you've mentioned, having variant-specific component parameters also felt a bit messy, although will likely sometimes be desirable (variable-_relative_ parameters could be interesting too...). The issue (#2346) does suggest that we'll likely want to put arbitrary content inside the bubble, not just text strings, which was part of the motivation for building on top of the display group, but we do already have other components that render child rows within themselves and I don't think that's a problem.

I think likely we could aim to support both formats within the same component. I.e if using `text_bubble` then the value is displayed as text, but if using `begin_text_bubble` then we could render child rows. Although I'm guessing this will only become a requirement later as I don't think it's very common to illustrate things like images within text bubbles (maybe if we were using more like a thought bubble or chat interface). But yeah, I think extending all of this onto a more simple base component would probably make the most sense.
 
> Thanks for setting up the PR for the other implementation, I will indeed work on that branch rather than suggesting we merge the display group version first. Then I'll probably merge that branch into this PR's feature branch and update this PR.

Sounds Good

jfmcquade:
> I think likely we could aim to support both formats within the same component. I.e if using `text_bubble` then the value is displayed as text, but if using `begin_text_bubble` then we could render child rows.

Thanks, good suggestion. I've updated this PR accordingly so it should be ready to review

chrismclarke (approved):
Thanks for refactoring @jfmcquade - I really like both the inline and nested block authoring options, feels very intuitive to me. Thanks for updating the preview also and getting it to work with image assets - I've added one minor (non-blocking) comment but overall I'd say looks good to go.

PR #2359, opened 2024-07-31
Fix: Handle cells populated by newline characters

### LINKED ISSUE #2193: bug(parser): handle cells populated only by newline characters
Content synced from google sheets can sometimes contain only newline characters. This is likely a mistake, and should be replaced with empty strings instead.

The default parser already has a [cleanFieldValues](https://github.com/IDEMSInternational/parenting-app-ui/blob/refactor/data-pipe-spec-filter-constants/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/default.parser.ts#L276) method currently used to remove additional whitespace, so might be the place to add a check/replacement. 

_Originally posted by @chrismclarke in https://github.com/IDEMSInternational/parenting-app-ui/issues/2187#issuecomment-1919657779_

kwAsant:
PR Checklist

## Description

It fixes the issue with the cells populated by newline characters(\r,\n, etc).
It also contains a test script (with two new tests) to ensure the code functions as expected.

## Git Issues

Closes #2193 

## Screenshots/Videos

image


packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/default.parser.spec.ts
```diff
@@ -1,3 +1,135 @@
...
+import { FlowTypes } from "data-models";
+import { DefaultParser } from "./default.parser";
+
+const testInputSources = {
+  data_list: { test_data_list: [{ id: 1 }, { id: 2 }, { id: 3 }] },
+};
+
+describe("Default Parser", () => {
+  it("Cleans field values - removes whitespace within empty strings", () => {
+    const parser = new DefaultParser({ processedFlowHashmap: testInputSources } as any);
```
jfmcquade, line 10: Rather than declaring this for each spec test, you can declare it once at the start of the `specDefinitions` in the `describe` function. See line 47 of `packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/data_list.parser.spec.ts` for an example. In your case you want the parser to be instantiated as a `new DefaultParser()`, but you can use the function `getTestFlowParserProcessor`, which is imported into `data_list.parser.spec.ts`, to get an entity of the right type to pass to `new DefaultParser()`. Then you can get rid of `testInputSources`, which I don't think is serving any purpose. See if you can figure out how to declare `parser: DefaultParser` at the top of the spec tests so that it doesn't need to be declared within each one, something like this:

```ts
describe("Default Parser", () => {
  // Instantiate parser here

  it("does something", () => {
    const output = parser.run({

   ... 

```

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/default.parser.spec.ts
```diff
@@ -1,3 +1,135 @@
...
-  it("TODO - add tests", () => expect(true).toEqual(true));
+import { FlowTypes } from "data-models";
+import { DefaultParser } from "./default.parser";
+
+const testInputSources = {
+  data_list: { test_data_list: [{ id: 1 }, { id: 2 }, { id: 3 }] },
+};
+
+describe("Default Parser", () => {
+  it("Cleans field values - removes whitespace within empty strings", () => {
```
jfmcquade, line 9: I think this description could be more accurate, something like: "Cleans field values - handles strings consisting only of whitespace"

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/default.parser.spec.ts
```diff
@@ -1,3 +1,135 @@
...
+      },
+      {
+        type: "text",
+        name: "newline_return_empty",
+        value: "",
+      },
+    ]);
+  });
+
+  it("Cleans field values - Replaces string consisting of only line breaks and new lines", () => {
```
jfmcquade, line 73: This description isn't representative of what the spec test is checking. "Cleans field values - trims whitespace from beginning and end of strings"

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/default.parser.ts
```diff
@@ -69,6 +70,7 @@ export class DefaultParser<
     }
     this.flow.rows = processedRows;
     this.flow = this.postProcessFlow(this.flow);
+
```
jfmcquade, line 73: Worth deleting these inserted newlines, just to keep the diffs clean

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/default.parser.ts
```diff
@@ -285,6 +287,11 @@ class RowProcessor {
       if (typeof this.row[field] === "string") {
         // remove whitespace
         this.row[field] = this.row[field].trim();
+
+        // remove all line breaks in the field using replaceAll() and regex
```
jfmcquade, line 291: This comment doesn't describe what the logic below does (and we wouldn't want to remove all line breaks). It should say something like:
"replace any strings that consist only of whitespace with the empty string"

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/default.parser.ts
```diff
@@ -285,6 +287,11 @@ class RowProcessor {
...
+        if (!this.row[field].match(/\S/)) {
```
jfmcquade, line 292: Nice, I think this is a good way to make the check

jfmcquade (changes requested):
Good job getting the tests to work.

I've made some inline comments. When you've made the requested changes let me know, I have some further suggestions for how the tests could be structured, but good to get them tidied up in their current form first.

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/default.parser.spec.ts
```diff
@@ -1,3 +1,135 @@
...
+  data_list: { test_data_list: [{ id: 1 }, { id: 2 }, { id: 3 }] },
+};
+
+describe("Default Parser", () => {
+  it("Cleans field values - removes whitespace within empty strings", () => {
+    const parser = new DefaultParser({ processedFlowHashmap: testInputSources } as any);
+    const output = parser.run({
+      flow_name: "test_flow_input_empty",
+      flow_subtype: "generator",
+      flow_type: "generated",
```
jfmcquade, line 14: This isn't a valid flow type. I would recommend to use "data_list" here (as it is the most permissive type). You can also get rid of the `flow_subtype` field entirely as it isn't requried

jfmcquade (approved):
Looks good to me. Possibly the test cases could be broken out into multiple, but I think probably good as is

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/default.parser.ts
```diff
@@ -285,6 +286,11 @@ class RowProcessor {
       if (typeof this.row[field] === "string") {
         // remove whitespace
         this.row[field] = this.row[field].trim();
+
+        // replace any strings that consist only of whitespace with the empty string
+        if (!this.row[field].match(/\S/)) {
```
chrismclarke, line 291: (non-blocking): It took me a second to understand this, so you are checking for the case where there is not a single non-whitespace character. It's smart and should work as intended, but the double negative throws me a bit. I might be more tempted to use the regex
```
/^[\s]*$/
```
Where 
`^` indicates the start of the string
`$` indicates the end of the string
`[\s]*` matches case when only whitespace provided

Interactive demo for testing at: https://regexr.com/85b97

But I don't feel too strongly about this, more just for interest

chrismclarke (approved):
Thanks for the fix and adding the tests too, nice to have for sure

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/default.parser.ts
kwAsant: I am happy to work on it, when I am free too. I am not too confident about the regex myself when I wrote it, so if I can improve it, I am happy to.

chrismclarke: The regex you wrote does make sense to me and I think should cover all use-cases successfully (appreciate the tests you added to check) so I don't think any additional work specifically required, more just an observation if interested to replace the double-negative with a positive match statement. Regex is quite fiddly at the best of times, I definitely recommend using the regexr.com website linked above for an interactive testing playground

chrismclarke: I did push one small commit with some comments just as a reminder it would be good to add more tests generally to the parser in the future, so if you ever do find yourself working again on it and are interested that could also be something to look into

PR #2360, opened 2024-08-05
fix: make calc context vars available immediately on init

jfmcquade:
PR Checklist

## Description

A bug was identified with the `plh_facilitator_mx` deployment whereby the user ID (`app_user_id`) was not displayed as expected:

unnamed

A workaround was committed in https://github.com/IDEMSInternational/plh-facilitator-app-mx-content/pull/93, whereby the value of `app_user_id` is fetched from `@fields._app_user_id` as opposed to `@calc(this.app_user_id)`.

This PR addresses the underlying issue of why using `@calc(this.app_user_id)` would sometimes not work. Now, the page will not finish loading until the calc context variables (`app_day`, `app_first_launch`, `app_user_id` and `device_info`) are properly populated (because the calc service now waits for the calc context vars to be available and populates them on init).

## Testing

Whilst I have been unable to fully recreate the bug outlined above (the screenshot shows the template in a state that is only accessible after a successful data sync, at which point the value of `app_user_id` should be fully populated), it is possible to observe the app to in a state where `@calc(this.app_user_id)` is `undefined` whilst `@fields._app_user_id` is correctly populated:

On `master`, sync debug content and navigate to the template [debug_calc_context_variables](https://docs.google.com/spreadsheets/d/125_zehOos_qTDKvhMVqKzUJAIYamS-RVh6cJ4G3VcjY/edit?gid=1422144003#gid=1422144003). After forcing an "Empty cache and hard reload", the template should display as follows:

Screenshot 2024-08-05 at 15 21 09

Now checkout this PR feature branch, `fix/calc-context-vars`. The template [debug_calc_context_variables](https://docs.google.com/spreadsheets/d/125_zehOos_qTDKvhMVqKzUJAIYamS-RVh6cJ4G3VcjY/edit?gid=1422144003#gid=1422144003) should now display both variables populated correctly:

Screenshot 2024-08-05 at 15 24 48

## Git Issues

Closes #

## Screenshots/Videos

New template, [debug_calc_context_variables](https://docs.google.com/spreadsheets/d/125_zehOos_qTDKvhMVqKzUJAIYamS-RVh6cJ4G3VcjY/edit?gid=1422144003#gid=1422144003)

Screenshot 2024-08-05 at 15 15 30

Screenshot 2024-08-05 at 15 26 03


esmeetewinkel (approved):
Functional test passed following the testing steps outlined in the PR notes.

chrismclarke (changes requested):
I think the way we need to wait for the server service to initialise to correctly populate the app user id for calc statements feels a bit off (legacy code issues). 

I think it might be a little cleaner to address that dependency chain more directly, i.e. refactor `src\app\shared\components\template\services\template-calc.service.ts`  to populate the `app_user_id` and `device_info` in the same way that the server service does (via local storage and capacitor api), and entirely remove the dependency on the `ServerService` (which can then be left as-is)

The one caveat would be to make sure that the localstorage value is populated, likely by injecting and waiting init of the `UserMetaService` which populates the app_user_id storage object.

I think it might be easier to just open a new PR instead of reverting this one, let me know if you're happy to do or if you want me to give a first pass.

Issue #2108, opened 2023-10-19
[FEATURE] Disable campaign system for deployments that don't use them


esmeetewinkel:
**What?**
Loading campaigns takes ~300ms, could be taken out for deployments that don't use this to save time/space.

esmeetewinkel:
This also applied to e.g. the PDF reader component, and probably other components/features

esmeetewinkel:
Also holds for notifications https://github.com/IDEMSInternational/parenting-app-ui/issues/2265

PR #2361, opened 2024-08-07
Feat/map component

### LINKED ISSUE #2358: [FEATURE] Map component
**What?**
A component for displaying interactive maps, with various options to display layers of data.

**Why?**
A research group are looking to build a web app to display some data relating to predicting conflicts. The group's aim is to improve conflict prediction, helping forecasters and decision makers. The purpose of app is to provide easier access to their information and data.

INNODEMS are handling the app authoring, see [this content repo](https://github.com/IDEMSInternational/conflict-forecast-content). An integral part of the proposed app is an interactive map, that will display various layers of data. This will require a new map component.

Whilst a map component would be a welcome addition to the app builder, there are currently no other use cases. Along with the fact that the group would like a version of the app very soon (potentially containing a placeholder for the map element), and the fact that dev work on this does not seem to come under any existing contract, it would seem sensible to build a new map component that is designed to meet the immediate needs, without necessarily offering full flexibility. The usual attempts should be made to generalise the component and its API as much as possible, but in this case it need only be minimally generalised.

Potential examples can be found [here](https://conflictforecast.org/) and [here](https://ucdp.uu.se/).

**How?**
In the spirit of "open by default", an open source mapping package should be considered first. Prime candidates would be:
- [OpenLayers](https://openlayers.org/)
- [Leaflet](https://leafletjs.com/)

A comparison of the two libraries can be found [here](https://www.geoapify.com/leaflet-vs-openlayers/). I would lean towards using OpenLayers, which seems to offer greater functionality (potentially with a greater learning curve).

**Other details**
The mapping package is likely to be pretty heavy, and therefore it should ideally be an opt-in feature to avoid adding too much to our app bundle size by default. We do not currently have the ability to opt-in to specific features and have them removed from the build by default, and so it may be that for the time being it makes sense to manage this via a dedicated branch on the code repo that contains the map component code.
This issue does provide motivation for increasing the priority of disabling features in an opt-in way, see #2108 and #2265

jfmcquade:
PR Checklist

## Description

Adds a new component, `map`, for displaying interactive maps with layers visualising data.

This is an initial prototype that will need to be improved upon to meet the needs of a particular deployment (see #2358). 

## Review Notes

As the map component will only be used for a single deployment for the time being, it would be desirable to make it an opt-in feature, and exclude the associated packages from the built app of those deployments that do not make use of it. Currently, we do not have the functionality to do so, see for example #2108 and #2265. In the meantime, it may be desirable to work around this by keeping this feature on a dedicated branch of the code repo in order to not include it in those deployments that do not make use of it.

Looking at the analysis generated by running `yarn analyse`, it looks like the map component and its associated node module files only add 1.4MB to the web app (although I'm not sure I am interpreting the chart correctly).

Screenshot 2024-08-08 at 15 19 46

Screenshot 2024-08-08 at 15 19 28

## Git Issues

Closes #2358 

## Screenshots/Videos

[comp_map](https://docs.google.com/spreadsheets/d/1KgLTrFhqZdGhcSk6Zt0KQoy2mYf7NCqhCJz_zS0KEYw/edit?gid=569531329#gid=569531329)

Screenshot 2024-08-08 at 14 52 13

[comp_map_layer_data](https://docs.google.com/spreadsheets/d/1KgLTrFhqZdGhcSk6Zt0KQoy2mYf7NCqhCJz_zS0KEYw/edit?gid=1695224840#gid=1695224840)

Screenshot 2024-08-08 at 14 54 44

Demo

https://github.com/user-attachments/assets/cb3c8bf5-4aeb-45ec-9cc1-976ce0eb81fa


esmeetewinkel (commented):
Looking great! 

From a data-communication point of view, I think there should be an explanation of what the different colours mean. This could be 
- a legend
- display the toggled-on data for a certain area whenever that area is clicked / tapped / hovered over

package.json
```diff
@@ -98,6 +100,7 @@
     "ngx-lottie": "^10.0.0",
     "ngx-matomo-client": "^6.0.2",
     "nouislider": "^15.7.1",
+    "ol": "^10.0.0",
```
chrismclarke, line 103: nit(non-blocking): nothing to do with you, but really wish openlayers just kept a package named openlayers! incredibly unhelpful when just skimming through deps list

package.json
```diff
@@ -66,9 +66,11 @@
     "@ionic/pwa-elements": "^3.2.2",
     "@sentry/angular-ivy": "^7.102.1",
     "@supabase/supabase-js": "^2.39.0",
+    "@types/chroma-js": "^2.4.4",
     "@types/file-saver": "^2.0.7",
     "bootstrap-datepicker": "^1.10.0",
     "capacitor-blob-writer": "^1.1.14",
+    "chroma-js": "^2.6.0",
```
chrismclarke, line 73: nit(non-blocking): interesting package. I think still probably better to use our mixins where possible to generate CSS at compile time, but definitely should unlock some interesting runtime colour functionality.

package.json
```diff
@@ -125,7 +128,7 @@
     "@angular/language-service": "~17.2.2",
     "@capacitor/cli": "^5.5.1",
     "@compodoc/compodoc": "^1.1.23",
-    "@ionic/angular-toolkit": "^10.0.0",
+    "@ionic/angular-toolkit": "^11.0.0",
```
chrismclarke, line 131: thanks for catching, we've been on angular 17 for a while which should have required this version. Interesting that there weren't more notifications from the cli to update (or maybe I just didn't notice them) - out of interest, how did you realise?

src/app/shared/components/template/components/map/map.component.html
```diff
@@ -0,0 +1,27 @@
...
+                  {{ mapLayer.get("description") }}
+                </p>
+              }
+            </ion-label>
+          </ion-item>
+        }
+      </ion-list>
+    }
+  </div>
+</div>
```
chrismclarke, line 27: +1 All looks really nice and concise (except the amount of boilerplate you need to add a simple toggle in ionic!).

src/app/shared/components/template/components/map/map.component.scss
```diff
@@ -0,0 +1,14 @@
+.map-container {
+  width: 100%;
+}
+
+.map {
+  width: 100%;
+  height: 400px;
```
chrismclarke, line 7: nit(non-blocking): does the map container need a fixed height or can we make it more flexible? Can imagine cases where we might want the map to take up the majority of the screen.

Also just to check, does the map intercept touch events? If so you might either want to limit the width or add some sort of toggle if there is content above/below the map as it might not be possible to scroll the screen if the map is taking up the full view

src/app/shared/components/template/components/map/map.component.ts
```diff
@@ -0,0 +1,203 @@
...
+import BaseLayer from "ol/layer/Base";
+
+interface IMapLayer {
+  id: string;
+  description: string;
+  fill: string;
+  name: string;
+  opacity: number;
+  // a property of the dataset to be plotted
+  property: string;
```
chrismclarke, line 26: nit(non-blocking): better to use `/** */` comment blocks as these will be picked up by vscode intellisense when referring to the interface elsewhere

src/app/shared/components/template/components/map/map.component.ts
```diff
@@ -0,0 +1,203 @@
...
+  type: "vector" | "heatmap";
+  visible_default: boolean;
+}
+
+interface IMapParams {
+  /**
+   * TEMPLATE PARAMETER: extent.
+   * Optional. The extent of the map to be displayed, i.e. the boundaries of the visible map, in EPSG:3857 co-ordinates
+   */
+  extent: number[];
```
chrismclarke, line 42: nit(non-blocking): I'm assuming this terminology is to maintain consistency with OL, but might be worth making more human readable and including an example. E.g. `boundToCoordinates` or similar. I'd also recommend an option to automatically generate bounds based on data in the future (e.g. `boundToData: true`), or possibly the extent of any geojson layers provided

package.json
jfmcquade: yes! And unhelpful when searching files for imports

jfmcquade:
I've created a new branch, `deployment/conflict_forecast`, which will be used to build out the WIP functionality required for the [conflict forecast deployment](https://github.com/IDEMSInternational/conflict-forecast-content), alongside authoring the content.

chrismclarke (commented):
Thanks @jfmcquade 
The implementation looks really nice, I haven't tested the preview myself but screenshots look good.

You raise a good point about bundle sizes and need making template components opt-in/out. I think it would be good to create as a standalone issue for prioritising (as likely the solution for template component opt-in/out will need to be substantially different than the campaign and notification features). 

I'm not personally entirely convinced by the choice of OpenLayers over Leaflet. The article you linked to is about 5 years old and [npm trends](https://npmtrends.com/leaflet-vs-ol) points quite strongly towards leaflet

image

The most recent comparison [blog post](https://thinkgeo.com/blog/leaflet-vs-openlayers-3-which-is-the-better-client-side-javascript-mapping-library) that came up in my search suggests leaflet tends to be much more friendly for simpler use cases - which I expect most of ours to be (complex cases tends to be more along the lines of using alternative map projections and geometries). I've used leaflet a lot myself but never openlayers - still interested to try out but looking at the current boilerplate code it definitely feels much more verbose.

That being said, assuming we get some sort of build workflow in place that can remove components not required by the content then could consider supporting multiple implementations (e.g. adding support for leaflet in the future and possibly migrating the current code to an `ol_map` component or similar), and I'm happy to reserve full judgement until I've spend a bit of time working with the openlayers code

package.json
jfmcquade: I actually can't remember what led me to change this, but I think an error with the open-layers components and a Google search.

I've included this upgrade in #2374

PR #2362, opened 2024-08-12
Chore: Migrating from using the cordova-res package to using the capacitor-assets package for asset generation

### LINKED ISSUE #2072: [FEATURE] Migrate package used for generating Android assets cordova-res->@capacitor/assets
**What?**
The package that we use for dynamically generating Android assets (icon and splash screen) was previously known as "cordova-res", but has changed to "@capacitor/assets". We use the legacy [cordova-res](https://www.npmjs.com/package/cordova-res) package, but should migrate to use [@capacitor/assets](https://github.com/ionic-team/capacitor-assets).

**Why?**
It's possible that the assets generated by cordova-res are in a legacy format not supported since the targeting of api 33 and the migration to Capacitor 5 #2046. Possibly the cause of [this Android build failure](https://github.com/IDEMSInternational/parenting-app-ui/actions/runs/6036316837/job/16378494318).

EDIT - It appears that capacitor-assets does not support being run programmatically, so cannot replace cordova-res like-for-like

kwAsant:
PR Checklist

## Description

- The capacitor-assets library was added to `package.json` using `yarn`.
- Added commands for generating splash images and icons within `android.ts` and `android.workflow.ts` for Android.
- Added commands for generating splash images and icons within `ios.ts` and `ios.workflow.ts` for iOS
- Due to the nature of the `capacitor-assets` package, splash image and icon generation can co-occur using new commands.

## Git Issues

Closes #2072 

## Screenshots/Videos

Unfortunately, I cannot get it for the `yarn` commands (an issue with my machine that I intend to fix later). However, capacitor-assets commands should produce these outputs:

Android:
image
image
image

iOS:image


jfmcquade:
Closing in favour of #2364

PR #2367, opened 2024-08-15
Fix: Action go_to_url does not work when URL is given through variable and includes https://

### LINKED ISSUE #1706: [BUG] Action go_to_url does not work when URL is given through variable and includes https://
**Describe the bug**
When a URL is given to a `go_to_url` action through a variable and the URL includes `https://`, then when launching the website the URL appears as `https//github.com/IDEMSInternational/parenting-app-ui` (so the colon has disappeared). 

**UPDATE**
See #2367 and comments for latest status

**App version**
v0.16.9

**Debug Template**
[debug_go_to_url](https://docs.google.com/spreadsheets/d/1tKO_-9XtYZHMXP8upkylvU6oumEYFZ8_IB_V7qhY8tM/edit#gid=1422144003)
image

**Screenshots and videos**

[screen-capture (16).webm](https://user-images.githubusercontent.com/74557272/211063774-7b108075-f435-42a1-929e-8021a43b79b8.webm)

kwAsant:
PR Checklist

- [X] PR title descriptive (can be used in release notes)

## Description

- It fixes the broken url, by adding a conditional check to see if the `key` in `template-action.service.ts` is already a HTTP link.

## Git Issues

Closes #1706 

## Screenshots/Videos

https://github.com/user-attachments/assets/46f942a5-8a00-4d12-b164-fb7757931305


jfmcquade (changes requested):
Good job figuring out a workaround. However, the pre-existing `go_to_url` logic was already pretty hacky, so I think rather than adding to it, it would be a good idea to replace it completely. 

I've updated [debug_go_to_url](https://docs.google.com/spreadsheets/d/1tKO_-9XtYZHMXP8upkylvU6oumEYFZ8_IB_V7qhY8tM/edit?gid=1422144003#gid=1422144003) with one more use case. You can look at the console logs for each to see the `key`/`value` that is being passed.

Screenshot 2024-08-16 at 09 56 22

Screenshot 2024-08-16 at 09 56 50


Ideally, this hacky logic wouldn't be required at all, because the parser wouldn't split a URL on the ":" character. However it may be pretty difficult to figure out how to handle this. If you want to investigate, one starting point might be the `parseAppDataActionString()` function – could it be changed so that it only splits on the ":" character when it is not part of a URL string? The parser logic is run when you run a `yarn workflow sync`, but sheets that have been already been processed won't automatically be re-processed each time you run this or make changes to the parser logic. So if you do change any parser logic files, you'll want to increment the `cacheVersion` found on line 7 of `packages/scripts/src/commands/app-data/convert/processors/flowParser/flowParser.ts` before running `yarn workflow sync` in order to invalidate the existing cache and trigger the already processed sheets to be processed again with the new logic.


If you can't figure out how to change the parser to not split URLs, I think the existing workaround could be improved by replacing the `case "go_to_url"` logic with something like the following:
```ts
const url = key && value
  ? key + ":" + value
  : key || value
return this.templateNavService.handleNavActionExternal(url)
```
In fact, I think you'll need an additional step to handle the case where the author has omitted "https://" from the original URL string. I would recommend using the [normalize-url](https://www.npmjs.com/package/normalize-url) package, something like:
```ts
const url = key && value
  ? key + ":" + value
  : key || value
return this.templateNavService.handleNavActionExternal(normalizeUrl(url))
```

kwAsant:
> If you can't figure out how to change the parser to not split URLs, I think the existing workaround could be improved by replacing the `case "go_to_url"` logic with something like the following:
> 
> ```ts
> const url = key && value
>   ? key + ":" + value
>   : key || value
> return this.templateNavService.handleNavActionExternal(url)

I have decided to go with the latter, because I was looking for the past couple hours and I don't particularly know what to change.

src/app/shared/components/template/services/instance/template-action.service.ts
```diff
@@ -182,6 +186,8 @@ export class TemplateActionService extends SyncServiceBase {
     // Handle specific actions
     // TODO - Refactor action handlers that call global services to use registry instead
     // NOTE - instance-specific handlers will likely need to remain in service (e.g. set_local)
+    console.log(args);
+    console.log(action_id);
```
jfmcquade, line 190: nit (non-blocking): these debug logs can be removed

src/app/shared/components/template/services/instance/template-action.service.ts
```diff
@@ -198,10 +204,12 @@ export class TemplateActionService extends SyncServiceBase {
         // if there is no http or https then there is no : to separate and we only have a key in the url. This
         // case then adds '//' to the key so it recognises it as external and not local
         // This removes the need to have http or https in the url.
-        if (!value) {
-          value = "//" + key;
-        }
-        return this.templateNavService.handleNavActionExternal(value);
+        // make a prior check for if the key starts with https:// and if it does then it is a url and should be opened
```
jfmcquade, line 207: todo (blocking): This comment is out-dated, as this check isn't being performed. I would suggest replacing all the comments from lines 201 to 207 with a new comment that summarises what's happening on lines 208-212 and why (i.e. because URL strings get separated by the parser at the ":" character in some cases)

src/app/shared/components/template/services/instance/template-action.service.ts
```diff
@@ -198,10 +204,12 @@ export class TemplateActionService extends SyncServiceBase {
...
+        const url = key && value ? key + ":" + value : key || value;
+        console.log(url);
```
jfmcquade, line 209: nit: this log can also be removed

jfmcquade (changes requested):
Thanks, yes I'm not sure that engaging with the parser code is the best use of time at this stage. These changes look good, I've added some suggestions in the inline comments

src/app/shared/components/template/services/instance/template-action.service.ts
```diff
@@ -195,13 +199,15 @@ export class TemplateActionService extends SyncServiceBase {
...
-        // This removes the need to have http or https in the url.
-        if (!value) {
-          value = "//" + key;
-        }
-        return this.templateNavService.handleNavActionExternal(value);
+        // form the url to generate the url from the key and value
+        // it does so by using a ternary operator to check if the key and value are present
+        // if so. it concatenates the key and value with a colon in between
+        // otherwise, it uses the key or value as the url (whichever is present will used, if both, then the key is taken)
+        // because sometimes when generating {key, value}, they may assign the url in just the key or value
```
jfmcquade, line 206: I think all the lines of comments from 199 to 206 could be re-worded to be more concise and clear. It could also be emphasised that this is a bit of a hack-y workaround by preceding it with the word "HACK:" Something like:

```ts
// HACK: URL strings that include the ":" character (as in `https://`) will be split on that character into a key-value pair. 
// If both a key and a value are populated, assume this to be the case and join on the ":" character.
```

jfmcquade (changes requested):
Looking good, I've just added one request to improve the clarity of the comment

jfmcquade (approved):
Looks good. Still a bit of a hacky workaround but an improvement

chrismclarke (changes requested):
@kwAsant  - Thanks for identifying the core issue and coming up with a workaround.

@jfmcquade - if a fix is urgently needed then makes sense to merge, although if not I think there should hopefully be a reasonably quick fix that can be applied at the parser level (as we are likely to need to handle eternal URLs in actions in the future).

esmeetewinkel:
> @jfmcquade - if a fix is urgently needed

A fix is not urgently needed. I'll remove myself from review for now, please request again when ready

jfmcquade:
Closing as we're agreed that, as a fix is not urgently needed, it would be better to address the cause of the problem directly by updating the parser rather than extending the runtime workaround. #1706 will remain open

PR #2370, opened 2024-08-20
Feat: "progress_path" component; "circle" task-card variant

### LINKED ISSUE #2324: [FEATURE] New task "journey" design
**What?**
Display component that displays a list of tasks (which may themselves be task groups) as a route connecting dots. 

image

Screenshot 2024-08-19 at 15 09 10

**Why?**
This design would evoke the feeling of being on a learning journey. Inspired by other learning app designs.

**How?**
- Screen divided into two columns, design needs to scale sensibly also when long text strings are provided as titles
- Each task can have different statuses which will have minor effects on the way it is displayed (such as completed, locked)

jfmcquade:
PR Checklist

## Description

- New component: `progress_path`
- New variant for `task_card` component: `circle`

These are designed to be used in conjunction to display multiple task cards visually on a "journey" or "progress path".

#### Child rows
Each child row will be rendered in a column, alternating between the left and the right side of the screen.

## Git Issues

Closes #2324

## Screenshots/Videos

[comp_progress_path](https://docs.google.com/spreadsheets/d/14BBKG07nqUWDmcscdUMv1AKYSWGBUPCNDT-VcSERHyU/edit?gid=569531329#gid=569531329)

Screenshot 2024-08-27 at 10 54 13

| Theme: default | Theme: professional |
|-----|-----|
| localhost_4200_template_comp_progress_path | localhost_4200_template_comp_progress_path (2) |
| localhost_4200_template_comp_progress_path (1) | localhost_4200_template_comp_progress_path (3) |

Demo showing responsive adaptation to screen sizes:

https://github.com/user-attachments/assets/c2c2bbeb-be3f-4df1-9815-792a09951a4a


src/app/shared/components/template/components/progress-path/progress-path.component.ts
```diff
@@ -0,0 +1,28 @@
+import { Component, OnInit } from "@angular/core";
+import { TemplateBaseComponent } from "../base";
+import { getStringParamFromTemplateRow } from "src/app/shared/utils";
+
+interface IProgressPathParams {
+  /** TEMPLATE PARAMETER: path_segment_asset */
+  pathSegmentAsset: string;
```
chrismclarke, line 7: nit(blocking): given how critical the path segment asset is to rendering the component might make more sense to hardcode the SVG and drop the authoring option. Instead could expose some sort of styling for the path asset (e.g. color). For that you will want to edit the original svg to support currentColor, e.g. `<svg fill="currentColor">`, and then just specify the shade of gray from CSS

chrismclarke: Could also (in the future) consider different styles of path, e.g.
https://zellwk.com/blog/million-devs-paths/

src/app/shared/components/template/components/progress-path/progress-path.component.html
```diff
@@ -0,0 +1,16 @@
+<div class="progress-path-wrapper">
+  @for (childRow of _row.rows | filterDisplayComponent; track trackByRow($index, childRow)) {
+    <div
+      class="progress-path-child-wrapper"
+      [class]="{ odd: $index % 2 === 0, even: $index % 2 !== 0 }"
```
chrismclarke, line 5: nit(non-blocking): interesting, I thought you needed to use `ngClass` to support this type of syntax, nice it works inline! Although I'm guessing you only actually need to track the even state, and actually angular provides `$even` and `$odd`  context variables
https://angular.dev/api/core/@for

chrismclarke (changes requested):
Thanks @jfmcquade - new component looks good. See one blocking comment inline and a few other minor nits. 

I haven't reviewed the `task-card` component changes as don't really have much of a frame of reference for this and the changes look quite big, but as mostly visual I think fine to just leave to @esmeetewinkel to visually inspect

src/app/shared/components/template/components/progress-path/progress-path.component.ts
chrismclarke: Alternatively you could also consider a pure CSS approach like the ones seen in:
https://codepen.io/j471n/pen/vYJaLvm
https://codepen.io/cplepage/pen/EozVXL

These basically just use elements with rounded borders to set up the path, where each row essentially has whatever combination of left-bend, center straight, and right-bend elements. A few more examples at:
https://freefrontend.com/css-timelines/

jfmcquade: Thanks, this was helpful. I did consider both of these routes, but originally chose to pass in an SVG asset as I thought this would give me the quickest results. Returning to this, I've chosen to use an inline SVG, which works quite nicely. And I've exposed the style variables (for the colour of the path and the stroke width of the path) at the _theme_ level, rather than at the template component-instantiation level, see [this commit](https://github.com/IDEMSInternational/open-app-builder/pull/2370/commits/fa704a4ab643f1c8fc455f9e58761bf9c3fdb1ca).

The pure CSS approaches are nice and seem to be the best way to make the path line responsive to different content heights (which would be nice to have). But I agree that they're quite hard to work with – leveraging `::before`/`::after` in that way is pretty opaque to me (although I did end up using something similar for the [text-bubble component](https://github.com/IDEMSInternational/open-app-builder/pull/2353/files#diff-ec59163eb361ae21abc556f9688813c90cdf09c47d0733240c345b4eeec68ac2)).

jfmcquade:
Thanks @chrismclarke, I've acted on your comments and left one inline comment thread open with a reply. See my 3 commits since your last review.

src/app/shared/components/template/components/task-card/task-card.component.scss
```diff
@@ -103,44 +103,111 @@
...
+  $circle-width: 100px;
+
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+
+  // HACK: use custom CSS variable, set in progress-path component, to set flex alignment
+  // based on the index parity of the card component within progress-path.
+  // Default to center when not a child of progress-path
+  align-items: var(--progress-path-flex-align, center);
```
jfmcquade, line 159: It would be nicer if the variable could just pass the index value (or its parity) to be consumed in the CSS, however applying conditional rules based on a CSS custom property is tricky. There is a way to do it via `@container`, but this doesn't have full browser support: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_size_and_style_queries#container_style_queries

chrismclarke:
Thanks for the update @jfmcquade - I revisited the component and tried to tidy up the inline SVG a little (mostly to remove unnecessary nested grouping), although found very quickly that any minor changes had a lot of knock-on effects and was quite tricky to work with.

I've attempted a refactor with https://github.com/IDEMSInternational/open-app-builder/pull/2373, let me know what you think. Also happy to merge here if you want more time to spend on the refactor approach.

chrismclarke (approved):

esmeetewinkel:
> I haven't reviewed the `task-card` component changes as don't really have much of a frame of reference for this and the changes look quite big, but as mostly visual I think fine to just leave to @esmeetewinkel to visually inspect

Looking at the screenshots, the only difference I can pick up is that the buttons in the `radio_group` component have become slightly shorter. I don't think that's a problem.
image
image

PR #2371, opened 2024-08-21
fix: make calc context vars available immediately on init (alt)

jfmcquade:
PR Checklist

## Description

NB: this is an alternative implementation of, and therefore closes #2360, which has been refactored in accordance with @chrismclarke's [comment on that PR](https://github.com/IDEMSInternational/open-app-builder/pull/2360#pullrequestreview-2241488207). Duplicate PR notes below:

A bug was identified with the `plh_facilitator_mx` deployment whereby the user ID (`app_user_id`) was not displayed as expected:

unnamed

A workaround was committed in https://github.com/IDEMSInternational/plh-facilitator-app-mx-content/pull/93, whereby the value of `app_user_id` is fetched from `@fields._app_user_id` as opposed to `@calc(this.app_user_id)`.

This PR addresses the underlying issue of why using `@calc(this.app_user_id)` would sometimes not work. Now, the page will not finish loading until the calc context variables (`app_day`, `app_first_launch`, `app_user_id` and `device_info`) are properly populated (because the calc service now waits for the calc context vars to be available and populates them on init).

## Testing

Whilst I have been unable to fully recreate the bug outlined above (the screenshot shows the template in a state that is only accessible after a successful data sync, at which point the value of `app_user_id` should be fully populated), it is possible to observe the app to in a state where `@calc(this.app_user_id)` is `undefined` whilst `@fields._app_user_id` is correctly populated:

On `master`, sync debug content and navigate to the template [debug_calc_context_variables](https://docs.google.com/spreadsheets/d/125_zehOos_qTDKvhMVqKzUJAIYamS-RVh6cJ4G3VcjY/edit?gid=1422144003#gid=1422144003). After forcing an "Empty cache and hard reload", the template should display as follows:

Screenshot 2024-08-05 at 15 21 09

Now checkout this PR feature branch, `fix/calc-context-vars-alt`. The template [debug_calc_context_variables](https://docs.google.com/spreadsheets/d/125_zehOos_qTDKvhMVqKzUJAIYamS-RVh6cJ4G3VcjY/edit?gid=1422144003#gid=1422144003) should now display both variables populated correctly:

Screenshot 2024-08-05 at 15 24 48

## Git Issues

Closes #

## Screenshots/Videos

New template, [debug_calc_context_variables](https://docs.google.com/spreadsheets/d/125_zehOos_qTDKvhMVqKzUJAIYamS-RVh6cJ4G3VcjY/edit?gid=1422144003#gid=1422144003)

Screenshot 2024-08-05 at 15 15 30

Screenshot 2024-08-05 at 15 26 03


src/app/shared/components/template/services/template-calc.service.ts
```diff
@@ -57,11 +65,20 @@ export class TemplateCalcService extends AsyncServiceBase {
...
     };
   }
 
+  private async setUserMetaData() {
+    if (!this.device_info) {
+      this.device_info = await Device.getInfo();
+    }
+    if (!this.app_user_id) {
+      this.app_user_id = this.localStorageService.getProtected("APP_USER_ID");
+    }
```
chrismclarke, line 79: nit(non-blocking): As this is called on init (and not re-triggered) the `if` blocks aren't necessary. In fact might be good to remove so that the service can update the value if the user changes (e.g. through user import action) - not sure if we've covered this edge case already or not (guessing not).

chrismclarke (approved):
Thanks @jfmcquade - looking a lot tidier now. 
While looking through I thought of one edge case that might be worth checking (handling calc after user import action) - although I expect not too important as the values tend not to be used much in calculations.

chrismclarke:
Tagging @esmeetewinkel for review as would be good to ensure functional test still fixes the issue flagged originally (now that the PR has been refactored)

esmeetewinkel (approved):
Functional test passed on `fix/calc-context-vars-alt` branch

PR #2373, opened 2024-08-22
review: task-journey-component

chrismclarke:
Targets branch for PR #2370

- Recreate svg element to avoid use of nested groups and transformations. 
- Make SVG generation dynamic (to support variable width and height)
- Fix inheritance of stroke color. 
- Remove path-line-weight variable (as image distorts when using different weight)

## Dev Notes
The changes started as minor edits to SVG to inherit color properly and remove unnecessary grouping, however I noticed that the SVG I was working from had been altered since to include more vertical space, which had knock-ons of breaking various layout aspects. It didn't help that the SVG itself had viewbox for one static size and rendering at another (auto-scaling), whitespace within the SVG element and then deeply nested fixed sizes for components that it had to adjust for (e.g task card text content)

This all felt quite fragile, so to address this PR moves more towards programmatic svg generation instead of fixed file, so that at least height and width can be reasonably adjustable (albeit with a few extra offset changes).

I'm still not 100% satisfied with the final result, but interested to see what you think @jfmcquade (see possible variants for _basic_ and _wavy_ ). See example links in the code for a better explanation of generating SVGs programmatically and a preview of the outputs

Longer term I think likely would want to de-couple from the task-card, and/or calculate the task-card sizes programmatically after view init and generate a full svg path once sizes are known. 

PR Checklist

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos
**Generated SVG** - wavy variant
localhost_4200_template_comp_progress_path (3)

**Generated SVG** - basic variant
localhost_4200_template_comp_progress_path (4)

### DIFF
--- a/cspell.config.yml
+++ b/cspell.config.yml
@@ -34,6 +34,7 @@
     "templatename",
     "tmpl",
     "venv",
+    "viewbox",
   ]
 # flagWords - list of words to be always considered incorrect
 # This is useful for offensive words and common spelling errors.
--- a/src/app/shared/components/template/components/progress-path/progress-path.component.html
+++ b/src/app/shared/components/template/components/progress-path/progress-path.component.html
@@ -6,14 +6,16 @@
       [style.margin-bottom]="contentHeight"
       [style.width]="width"
     >
-      <div class="progress-path-child-content">
-        <plh-template-component
-          [row]="childRow"
-          [parent]="parent"
-          [attr.data-rowname]="_row.name"
-          [style.--progress-path-flex-align]="$odd ? 'flex-end' : 'flex-start'"
-        >
-        </plh-template-component>
+      <div class="progress-path-child-content-wrapper" [style.--container-width]="width">
+        <div class="progress-path-child-content">
+          <plh-template-component
+            [row]="childRow"
+            [parent]="parent"
+            [attr.data-rowname]="_row.name"
+            [style.--progress-path-flex-align]="$odd ? 'flex-end' : 'flex-start'"
+          >
+          </plh-template-component>
+        </div>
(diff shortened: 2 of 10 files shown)


jfmcquade (approved):

chrismclarke:
> Ooh, fun! Much nicer to generate the SVG programmatically, I haven't done this before so was previously just drawing an SVG in a vector graphics application and copying it over to the HTML.

Yeah it's a bit of a learning curve (pun intended!), but makes a lot of sense when you get into the details of what makes an SVG. I've tried to link to a few handy resources, I remember watching a video on it a few years back that really opened my mind but can't find unfortunately. 

> One issue with this implementation is that the component wasn't responsive to changes in view width, and overflowed the edge of the screen on smaller screen sizes (assuming we want to support screen sizes down to 320px width)

Nice, yeah I tend to not go down all the way to 320px screens (seen very few in the wild myself), but think 384px isn't as uncommon. I'm sure there would be a way to make fully responsive (possibly setting `preserveAspectRatio=none` and fixing height to allow horizontal scaling), but already dived pretty deep down this rabbit hole so might be better to keep for a v2 if we really need it.


> *An alternative would be to fix the component styling to match how it displays at the minimum screen size, but it seems desirable to make the styling somewhat responsive Demo:

This could be a nice way forward, simply read the screen width on init and use that as a base value - but again I think would likely need a little more work on consistency of child components (probably after brought under the `progress_journey_card` child component or whatever it ends up being named.

> **2. Expose the "wavy"/"basic" setting as a variant template parameter**

Nice to have, think I personally prefer the _wavy_ as a default but not a super strong opinion

PR #2375, opened 2024-08-23
Spike: progress-path component combined svg path

chrismclarke:
PR Checklist

## Description

Attempted implementation to generate a single svg path based on the exact position of rendered content to connect.

## Dev Notes
I don't think it's worth spending anymore time on this (and so will not keep PR open). If revisiting in the future I think might be onto a good idea, although probably instead of generating one dynamic path revert back to individual segments (but programmatically calculating from and to node points from the content). 

Further work also required to make width responsive, and be a bit more systematic in picking where in the child component the line should connect to (e.g. ensuring child content as some sort of element attribute that defines which part is the shape to attach to)

In addition this was started before a few additional commits added to #2373, which would need updating/merging/resolving,.

## Git Issues

Closes #

## Screenshots/Videos
**Example** - Line in black generated as a single line (previous implementation underneath for ref)
localhost_4200_template_comp_progress_path (5)

**Example** - Standalone line
localhost_4200_template_comp_progress_path (6)


chrismclarke:
@jfmcquade - I'm not going to continue working on this (don't think it's worth the effort), but in case we revisit the component in the future it might be of some use

PR #2372, opened 2024-08-21
docs: updates to README and Deployments section

jfmcquade:
PR Checklist

- [x] PR title descriptive (can be used in release notes)

## Description

- Updates README to specify recommended Node version, and for general clarity
- Mirrors changes to README in external docs intro page, `index.md` (this had become out of sync with the README)
- Restructures "Deployments" docs page to put "Import Existing" section at the top
  - This is by far the most commonly relevant section for readers of the docs

## Git Issues

Closes #

### Intro

Docs Intro

### Deployments:

Screenshot 2024-08-21 at 15 21 05

Issue #2185, opened 2024-01-29
[BUG] Spec tests fail in node 20


chrismclarke:
**Describe the bug**
Some scripts tests populate a virtual file system to test various file creation/modification methods, e.g. assets postprocessor spec.
These rely on the package `mock-fs` which is currently broken on Node 20
`https://github.com/tschaub/mock-fs/issues/380`
`https://github.com/tschaub/mock-fs/issues/384`

and may not be fixed any time soon...
`https://github.com/tschaub/mock-fs/issues/383`

**To Reproduce**
Using a system running Node 20:
`yarn workspace scripts test`

**Additional context**
Short term workaround is to avoid Node 20 (stick to 18) for CI and local tests. 

I've tried replacing with alternatives but neither of the options below seem to be a perfect drop-in replacement:

[mock-tmp](https://www.npmjs.com/package/mock-tmp)
Uses slightly different folder structures which would mean refactoring tests

[memfs](https://www.npmjs.com/package/memfs) 
Can't find any way to force memfs methods to be called when running service code. Likely custom utility functions required to replace library fs methods with memfs and restore after tests, in the same way that [mock-fs does](https://github.com/tschaub/mock-fs/blob/main/lib/index.js#L185)

The other workaround is to use all async methods for fs operations both in spec and test code (importing from `fs/promises`), however

chrismclarke:
Just to note, recent angular 17.3 update requires `"^18.13.0 || >=20.9.0"`
which means either using v18 (>18.13.0), or v20 (>20.9.0)

So if wanting to update to angular 17.3 or beyond (when released), will either need developers using node 20.7.0 (last known compatible) to downgrade to node 18 - which isn't ideal.

As of recent comment mock-fs repo owner says open to PRs but not planning to fix, so we may need to discuss alternatives when next planning for core updates
`https://github.com/tschaub/mock-fs/issues/384](https://github.com/tschaub/mock-fs/issues/384#issuecomment-2004249964`

May be of interest @jfmcquade

PR #2376, opened 2024-08-25
Spike/scripts jest tests

chrismclarke:
PR Checklist

## Description
The core repo has quite specific requirement for Node >=18 and <=20.7.0 due to issues with tests outlined in #2185 and #2372.

Attempts had been made with #2186 to fix the implementation of mock-fs, however getting any solution working within the current testing infrastructure was proving difficult.

This PR instead aims to replace the core testing infrastructure (based on Jasmine) with Jest, which in turn has better support for mocking the methods that need replacement. 

**Main Changes**

- Remove jasmine from the scripts test workspace and replace with jasmine
- Refactor test logging methods to better support parallel operations
- Fix broken tests
- Replace broken mock-fs with memfs
- Add lint rules for test best practices (displayed in code editor but not run on CI)

## Review Notes
The changes are pretty hard to follow as there's a mix of find/replace and bits of refactoring. Ultimately the main thing to check is that the test suite still runs and passes locally. 
```sh
yarn workspace scripts test
```

There shouldn't really be any scope for knock-ons to the main scripts themselves or any apps

## Dev Notes
Jest was selected as a successor to Jasmine as ultimately that is the direction the angular team say that they are headed in (https://blog.angular.dev/moving-angular-cli-to-jest-and-web-test-runner-ef85ef69ceca) and also significantly the most popular tools used nowadays (Jasmine is no longer even tracked in stateofjs surveys, and gets mentioned by 16% of respondents in the additional section)

image

Not a huge amount should be different from a test-authoring perspective, focused tests are now defined by `it.only(...)` instead of `fit(...)` but most spies and mocks have similar syntax. It is also possible to only run tests on a single file by name, e.g. `yarn workspace scripts test -t base.spec.ts` (comments have been added to a handful of files to indicate). 

One key difference from a developer perspective is that by default jest runs tests in parallel. This had a number of breaking changes for a number of our tests that read/write and clear local test file folders and log files. I started to refactor tests to work around this (using virtual file systems where possible, in-memory logging options and various other fixes), although eventually realised you could pass a `runInBand` flag to jest to run sequentially rather than in parallel. Interestingly enough this also speeds up the test suite by around 50% (a lot of duplicate cache data gets moved across parallel processes), so I think it's fine to just keep this way.

As for virtual filesystems, mock-fs has been replaced by memfs. The syntax is slightly different so see the updated examples in the code

**Future TODOs**

- [ ] Refactor shared tests to use Jest #2378
- [ ] Consider refactor app tests to use Jest (and web-test-runner instead of karma). Support is still experimental so might be better to wait for angular 19 in case they include more fixes and migration tools. https://angular.dev/roadmap#improve-tooling
- [ ] Update docs (see #2372). Can also consider checking `.nvmrc` and yarn/husky hooks to see if any auto-switching enabled
- [ ] Start testing support for Node 22 (either locally or in CI). Will be marked as active from the end of the year https://nodejs.org/en/about/previous-releases, officially supported from angular 18 https://angular.dev/reference/versions

## Git Issues

Closes #2186

## Screenshots/Videos
**Test suites passing with Jest Runner on Node 20.17.0** 
image


jfmcquade (approved):
Looks good from the PR description, and I can confirm that the tests run as expected for me. 

Screenshot 2024-08-27 at 17 24 17

I can't speak to the details of the code changes (or the extent of work that ended up being required!), but happy to merge.

Not sure of the best way to track the future TODOs. I've broken out one into #2378 and assigned myself as I think it'd be sensible to acquaint myself with the changes.

chrismclarke:
Thanks @jfmcquade 

Thanks for adding the follow-up issue, for the remaining tasks:

**Consider refactor app tests to use Jest** 
As this depends on future angular updates (and will likely be promoted when ready) I think we can just leave to resurface organically when angular starts to push more in that direction.

**Start testing support for Node 22** 
Will likely start happening anyway, I'll plan to swap my go-to dev environment to Node 22 later this year and document any issues I come up against, so again no need to create standalone issue.

**Update docs, check nvmrc and node switching hooks**
I'll do this now and open a follow-up PR. Probably won't dive too much into nvmrc/husky hooks as the use depends a lot on OS used and node tooling (e.g. nvm vs fnm on windows/mac all behave differently), so can probably just update as specific developers request and raise PRs

PR #2385, opened 2024-08-29
Chore: set default node version 20

chrismclarke:
PR Checklist

## Description

Following comments in #2376, update documentation and CI to default to using node v20.

As per https://nodejs.org/en/about/previous-releases, this version should be supported until April 2026 (although we may decide to start testing node 22 support from October)
image

## Review Notes
The only scope for knock-ons would be within github actions, which are hard to test in advance. I'll trigger what actions I can from this branch, assuming that node is mostly just used for app build and screenshot compare (not play/ios signing deploy) so should consider likely stable if passing here.

## Dev Notes
I've pinned explicitly to `20.17.0` as the most recent v20 release. I've been running this version locally without issue so I'd hope everything just works as expected.  

## Git Issues

Closes #

## Screenshots/Videos


chrismclarke:
Yeah I get the punycode deprecation warning on node 22, can see on the node docs it's now considered deprecated:
https://nodejs.org/api/punycode.html

Although a lot of our package use it still, so all the 3rd parties get updated I don't think there's much that can be done

image

PR #2386, opened 2024-08-29
Fix/task card border

chrismclarke:
PR Checklist

## Description
Task cards don't appear to be applying consistent borders due to a combination of factors:
1. Images stretch to try to fill all available space
2. Content issues where images are not perfectly square, centered and applying their own white backgrounds

This is a small fix to apply a separate clipping at the image level to ensure images are contained within a circular region, and that circular region has a minimum distance to the outer container.

## Review Notes
Image whitespace still may look a little off due to the way some images include their own white regions (see red-line version for emphasis). If wanting full consistency the source images will require editing.

## Git Issues

Closes #

## Screenshots/Videos

**Before**
localhost_4200_template_home_screen

If changing white line to red for emphasis

localhost_4200_template_home_screen (1)

**After**
localhost_4200_template_home_screen (2)

If changing white line to red for emphasis

localhost_4200_template_home_screen (3)

### DIFF
--- a/src/app/shared/components/template/components/task-card/task-card.component.scss
+++ b/src/app/shared/components/template/components/task-card/task-card.component.scss
@@ -153,6 +153,7 @@
 
 .circle-card-wrapper {
   $circle-width: 100px;
+  $circle-clip-width: $circle-width / 2 - 4px;
 
   width: 100%;
   display: flex;
@@ -191,8 +192,10 @@
       border: 1px solid rgba(black, 0.07);
       filter: drop-shadow(var(--ion-default-box-shadow));
 
-      .img {
-        width: $circle-width;
+      img {
+        width: 100%;
+        height: 100%;
+        clip-path: circle($circle-clip-width at center);
       }
     }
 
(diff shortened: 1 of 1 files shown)


jfmcquade:
Really nice improvement.

As you note, for full consistency the source images would need to be edited – to achieve the effect of some elements of the illustrations overlapping the edge of the circle that contains them, I think the illustrator would need to decide on a standard border size around the image content and apply this to all images, whether or not the actual illustration overlaps into that border.

src/app/shared/components/template/components/task-card/task-card.component.scss
```diff
@@ -191,8 +191,10 @@
       border: 1px solid rgba(black, 0.07);
       filter: drop-shadow(var(--ion-default-box-shadow));
 
-      .img {
```
jfmcquade, line 194: woops 😅

chrismclarke: yeah, given that angular couples scss tightly with components would be nice if their vscode extension could also pick out unused classes. 

I can see one extension that claims to do that, but not tried myself:
https://marketplace.visualstudio.com/items?itemName=rcore.rcore-unused-css-classes

PR #2388, opened 2024-08-30
refactor: set_item action handling

chrismclarke:
PR Checklist

**TODO**

- [x] Refactor set_item and set_items actions to be independent of dynamic data service
- [x] Add support for `set_item` and `set_items` actions call from outside of loop
- [x] Refactor dynamic data service and action tests
- [ ] Refactor data_items tests
- [x] Improve component test template, add example for testing from outside of loop
- [ ] Add documentation
- [x] Update RFC

## Description
Currently both `set_item` and `set_items` actions only work when called from inside a `begin_data_items` loop, due to the way the item references are populated to the action. This is quite inconvenient, as if an author want to add a _mark all completed_ button they need to render the full list of items with a button for each, and then hide the button for every item except one.

This PR seeks to address that limitation (and the underlying cause), to speficially: 

- Refactor `set_item` and `set_items` actions code and tests to better decouple from item loops
- Add support for calling `set_item` and `set_items` from outside of data_items loops

## Author Notes
The updated syntax to set the value of a data_list row from outside of any loop is now:
```
click | set_item | _list_id: example_list, _id: example_id, completed:true;
```
The underscore properties are used to lookup a given row id within a given data list, and then any other properties provided are updated on the data_list entry itself. 

The syntax for setting items from within a loop is unchanged (allowing _id, _index, or just implied item reference)

It is also possible to bulk set all items from outside an item loop via
```
click | set_items | _list_id: example_list, completed:true;
```

It is not currently possible to bulk set_items for a filtered list, or passing list reference (e.g. @local.some_list), however there is a workaround to only set a list subset from within a data_items loop.

See updated examples in [component_data_items](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892)

**NOTE** - I think now might be a good time to think about improving our documentation on this feature. I tried to do a little as I went both within the code, however found the feature taking longer than anticipated and had to abandon efforts. It would be good if someone could resume, using the example component as a base to write a supporting docs page

## Dev Notes
The main issue with the code was a lot of custom workaround to pass references to rendered data_lists so that the service could support setting by `_index` and have it refer to the same index as the rendered content (not original list). That's now been refactored so the `data_items` component itself manages index replacement, leaving the service implementation much simpler. The service actions have also been separated from the main service for easier development and testing.

The actions contain code to validate args, however the errors thrown aren't very informative. It might be worth considering using a library like [joi yup or zod](https://npmtrends.com/joi-vs-yup-vs-zod) to better handle validation (probably zod)

I've also just noticed some knock-ons to the `data_items` tests that still need updating resolving (started work on the feature over a week ago and lost track). I'll hopefully find time to revisit later this week, but if you happen to review before then do you think you might also be able to take a quick look @jfmcquade?

Basically we need to write new tests to make sure the items loops adds metadata in the new format. It might also be nice to add a test to trigger the action from the items loop for full integration. As we are testing both items setting expected metadata and actions triggering from metadata set in expected format, the test isn't essential - just nice to have in case one side breaks/is updated and the other is not (but hopefully the stronger typing added in this pr would catch)

## Git Issues

Closes #

## Screenshots/Videos


src/app/shared/components/template/components/data-items/data-items.component.ts
```diff
@@ -143,6 +129,50 @@ export class TmplDataItemsComponent extends TemplateBaseComponent implements OnD
...
+      // set_item - auto-populate list id and row id parameters from current item context (can be overridden)
+      if (a.action_id === "set_item") {
+        const params: IActionSetItemParams = {
+          _list_id: this.dataListName,
+          _id: itemId,
+          ...a.params,
+        };
+        // if _index used lookup and replace _id
+        if (params._index) {
+          params._id = itemDataIDs[params._index];
```
jfmcquade, line 152: This does not allow for `params._index` to be an expression, e.g. `this.item._index + 1`

jfmcquade (changes requested):
Thanks for looking into this. I've updated the tests for the component in https://github.com/IDEMSInternational/open-app-builder/pull/2388/commits/a27eb0f68edbf2d76312d66a71272823e891d94e

### Updating arbitrary data

I would suggest that this feature goes beyond the use case highlighted in the PR description of marking all items in a given list as completed: it enables setting any property of any row of any data list (in dynamic data user overwrites), which potentially has broad applications. Whilst this was always possible from within a data-items loop, I think that being able to do so without such a loop is significant.

With this in mind, I think it's worth considering whether the syntax could be changed to reflect this. Currently, the proposed syntax is:
- `click | set_item | _list_id: example_list, _id: example_id, completed:true;`

Does "item" make sense intuitively outside of items loop? It probably does as we're talking about lists, but as we're now setting arbitrary values on arbitrary data list rows, we could consider a syntax in keeping with our "data_list" and "row" terminology, e.g.
- `click | set_data | _data_list_name: example_list, _row_id: example_id, completed:true`. 

This syntax would be in keeping with having a "set" action for each available variable type, e.g. `set_field`/`@field`; `set_local`/`@local`; and now `set_data`/`@data`. The `@item` variable type is only available from within a data-items loop, and so making `set_item` available outside such a loop would break this pattern – indeed up until now, the concept of an "item" has only been employed to refer to each element of the iterable. Although potentially this could get confusing if used in parallel to our existing `set_item` syntax that applies inside data-items loops.
@chrism – I know that setting an arbitrary property on a a given data list row was a feature that you expected to be available to you when you were working on the sheets last month. Do you have any thoughts about this syntax?

### Implementation issue

Additionally I have noticed an issue with this implementation: it does not currently allow for the value of the `set_item` action's `_index` param to be an expression, e.g. `click | set_item | _index: @item._index + 1, completed:false`, as in cell D46 of the [component demo sheet](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892). 

Screenshot 2024-09-16 at 16 29 51
Screenshot 2024-09-16 at 16 30 18

At the point of accessing the relevant property from the list of data IDs, this expression has not been fully evaluated, and is in the form `this.item._index + 1` (see my inline comment on this PR). The previous implementation worked because the `_index` expression was not used to look up the respective value until after the component had been rendered, at which point the expression had been fully parsed with the relevant context.

I'm not quite sure how to handle this, as it's nice that this implementation doesn't have to pass through all the item context to the action logic, but it may be necessary to do so and have the `_index` converted to an `_id` at the level of the `parseSetItemsParams()` method.

chrismclarke:
> After meeting with Esmee, Chris M and David, we agreed that it probably would make sense to use a more general syntax than `set_items`, along the lines of my suggestion above. They also pointed out some further follow-ups for desirable features, some of which are already noted in this PR description:

That makes a lot of sense, although I think it does mean I need to revisit this PR slightly as the proposed changes would quickly supersede existing. I'll work on this in a separate branch and target this one to try and keep the diffs a little cleaner

I think that from within an items or data_items loop, keeping the existing syntax for `set_item` and `set_items` actions makes sense, but ideally they should be thin wrappers around a separate `set_data` action that can also be called generically

> * The ability to pass variable values to the parameters that target the row to update
>   
>   * e.g. `_list_id: @local.some_list, _id: @local.some_id`
> * The ability to filter the data to update, on a property other than `id`
>   
>   * e.g. `_filter: age > 20` or `_query: completed=true`
>   * This could use the same querying syntax as that proposed in the [Data Actions RFC](https://docs.google.com/document/d/1qJi9Nk--CFwGTPMn9ksmOgoVq_dnXrTMlN9a_i7kUyY/edit?pli=1#heading=h.nd960lmg5uuf)
> * The ability to target a data list row through an index when not inside a data list

Yeah agreed would be nice to have and is mentioned in the Data Actions RFC - I think the main thing to decide here is what sort of syntax we really want (assuming that we can somehow map to the rxdb mongoose structure)

PR #2389, opened 2024-08-30
fix: audio player seek

### LINKED ISSUE #2387: [BUG] Refresh/update issue on components within a dynamic template
**Describe the bug**
ComboBox and audio/video players do not update if the template is within dynamically created template like used with PLH Kids KW

**App version**
v0.16.33

**To Reproduce**
See https://plh-kids-kw-uat1.web.app/template/module_selfcare?nav_parent=module_overview_selfcare&nav_parent_triggered_by=task_card_relax
If link doesn't work go to 'Taking Care of Yourself' | 'Relax' then start the audio. The progress on the audio fails to move however, in the template this is derived from it does work. 
See https://plh-kids-kw-uat1.web.app/template/article_relax_1

**Template Name**
outer template
'Module' in https://docs.google.com/spreadsheets/d/1MeuVFYp7ovMCABdgC5c1mExgDK8Xop13fzzBNR5EAPo  
inner template
'article' in https://docs.google.com/spreadsheets/d/1gfiZARKRfqyvaRg8AP20bi2D4K8NrCxmnW-3cMNhX2I

**Debug Template**
Apologies, I am unable to create a separate template that isolates the issue.

chrismclarke:
PR Checklist

## Description
Update the method of change detection used by the audio player to more reliably set the value of the seek progress bar

## Review Notes
Ideally should fix the issue seen in #2387 (was working locally for me). Although feel free to let me know if there are similar issues related to other components which I can look at in follow-up (I couldn't see an example video, but also think the code is handled differently so not sure if it would have the same issue). 

As for the underlying issue (why it works in some place and not in others), hard to be sure and I don't think fully worth a deep dive at this point (angular change detection can be fickle), but happy to look more closely at any other specific issues you notice

## Dev Notes
I was a bit surprised to see the package used to manage audio not have support for playback event tracking (needs management through custom setInterval). Longer term it might be worth considering migrating to a package that has better native audio support, such as
https://github.com/capacitor-community/native-audio

Also, having now worked with signals for the past 6 months or so I really like them and think moving forwards it would be good to slowly migrate the entire codebase towards, particularly where there are highly reactive components. I would perhaps even consider setting up a new `components` workspace where we slowly migrate components one-by-one as a means to track which have been refactored/reviewed, possibly converting to standalone components and including any services where specific to a single component.

I had also thought about marking the component's changeDetection strategy as `onPush`, however decided best not to for now as I'm not sure how easy it would be to test for breaking changes (worth a try for most components in the future though)

## Git Issues

Closes #2387

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/components/audio/audio.component.html
+++ b/src/app/shared/components/template/components/audio/audio.component.html
@@ -19,7 +19,7 @@
       class="audio-range"
       max="100"
       aria-readonly="true"
-      [value]="progress"
+      [value]="progress()"
       (ionChange)="checkChange()"
       (touchstart)="checkFocus()"
       (touchend)="seek()"
@@ -28,10 +28,10 @@
   </div>
   <div class="time">
     <div class="time-value">
-      {{ +currentTimeSong * 1000 | date : "mm:ss" }}
+      {{ +currentTimeSong * 1000 | date: "mm:ss" }}
     </div>
     <div class="time-value">
-      {{ !player ? "00:00" : (player.duration() * 1000 | date : "mm:ss") }}
+      {{ !player ? "00:00" : (player.duration() * 1000 | date: "mm:ss") }}
     </div>
   </div>
   <div class="controls" *ngIf="!errorTxt">
--- a/src/app/shared/components/template/components/audio/audio.component.ts
+++ b/src/app/shared/components/template/components/audio/audio.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, OnDestroy, OnInit, ViewChild } from "@angular/core";
+import { Component, Input, OnDestroy, OnInit, signal, ViewChild } from "@angular/core";
 import { FlowTypes } from "../../../../model";
 import {
   getBooleanParamFromTemplateRow,
@@ -66,7 +66,7 @@ export class TmplAudioComponent
   /** @ignore */
   errorTxt: string | null;
   /** @ignore */
-  progress = 0;
+  progress = signal(0);
   /** @ignore */
   rangeBarTouched: boolean = false;
   /** @ignore */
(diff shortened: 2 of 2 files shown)


ChrisMarsh82:
@chrismclarke This works perfectly for the audio player. Combobox also suffers from the same issue.

chrismclarke:
Thanks @ChrisMarsh82 - I'll make a follow-up PR for the combox, feel free to approve and merge this one first if working as expected

ChrisMarsh82 (approved):

jfmcquade (commented):
Very nice, it's good that using signals fixed the issue, feels like validation of the claims of signals bringing improved changed detection (I can't say I understand the details under the hood, but I also like signals to the extent that I've used them)

I've broken out the issue of refactoring existing components here: #2402. It'd be good to discuss how best to approach this

chrismclarke:
Yeah, they really are a much better mechanism for fine-grained control. The problem with relying on angular's zone.js detection mechanism is that everything happens in a black box and you just gotta kind of hope it works - where the methods listen out for any dom events and then try to re-evaluate everything to see if anything changed (which gets exponentially more expensive as components are nested and arrays/objects in the mix). So zone.js forces you to move to onPush change detection when things stop working so you get some level of manual control, but then that can throw off detection within other components (depending how implemented), and also you have to remember to keep calling the cdr markForCheck methods.

Signals have a few major benefits, by requiring getter/setter functions it means that you essentially calling a function like cdr.markForCheck every time you use either, but not having to remember to code the extra line. Angular knows when you are using the variables so can optimise much better.
https://tech.sparkfabrik.com/en/blog/angular-signals-whats-all-the-fuss-about/

Also the ability to use computed and effect signals massively simplifies complex logic where previously we might have needed to track multiple changes using ngOnChanges or rxjs mergeMap operators. 

If you want a deeper dive into the code behind signals I came across this one (taken a quick skim, I'd be lying if I said it was a thrilling read but definitely the most informative I've seen):
https://angular.love/signals-in-angular-deep-dive-for-busy-developers

PR #2407, opened 2024-09-11
docs: add instructions for avoiding folder path issues

jfmcquade:
PR Checklist

## Description

Adds some instructions regarding the local destination of the cloned repo, in order to avoid possible issues with (example in [this mattermost thread](https://chat.idems.international/all/pl/bkq53syxdbrzmmohinfcoggw9r)).

In general I'm wary of adding too much specific detail to the "Quickstart" set-up docs, as we have to assume that users either have some level of technical proficiency or can acquire it elsewhere. However, it does seem worth including instructions such as this where there might otherwise be issues for some users that do not make themselves immediately apparent.

## Git Issues

Closes #

## Screenshots/Videos

Docs site version:

Screenshot 2024-09-12 at 10 27 05


documentation/requirements.txt
jfmcquade, line 1: This diff was changing `PyYAML==6.0` -> `PyYAML==6.0.2`. PyYAML v6.0 was not installing for me, but updating the dependency to the most recent version fixed the issue.

esmeetewinkel (approved):

documentation/docs/contributors/documentation/running-server.md
```diff
@@ -21,7 +21,7 @@ The scripts below will create a python [virtual environment](https://docs.python
 
     ```sh linenums="1"
     cd documentation
-    python -m venv .venv
+    python3 -m venv .venv
```
chrismclarke, line 24: nit(blocking) how you access python will depend on how you installed python, so I don't think we want to make it specifically python3

README.md
```diff
@@ -22,7 +22,13 @@
 
 ## Installation
 
+### Choosing a location to download the repo
+
+Before running the commands, make sure you’re in the correct folder (the "current working directory") where you want to save the repo locally. You can use the `pwd` command to print the current directory, and `cd` to change the directory. You should ensure that there are no spaces at any point in the current working directory path. For example, if your top level user folder contains a space in the name, consider downloading the repo to a folder located outside of the user folder, e.g. in a new top-level `C:/apps` folder on Windows.
+
```
chrismclarke, line 28: nit(non-blocking): I don't think we need this duplicated here. Can assume people interacting directly with git readme likely more used to cloning repos. I'd say best to just leave the tip in the other docs

documentation/docs/index.md
```diff
@@ -17,6 +17,11 @@
 ## Installation
 
 ### Download the repo with binary assets
+!!! tip "Choosing a location to download the repo"
+
+      Before running the commands, make sure you’re in the correct folder (the "current working directory") where you want to save the repo locally. You can use the `pwd` command to print the current directory, and `cd` to change the directory. You should ensure that there are no spaces at any point in the current working directory path. For example, if your top level user folder contains a space in the name, consider downloading the repo to a folder located outside of the user folder, e.g. in a new top-level `C:/apps` folder on Windows.
+
```
chrismclarke, line 23: nit(blocking)
I think we can trim this down a bit, how about:

> It is best to download the repo to a folder path that does not include any spaces. E.g. if using a user documents directory that includes spaces `/user/my name` you may need to instead re-open the terminal in a different folder instead.

I might also suggest that we add a troubleshooting section where we put this instead along with likely error message/identifier, as I expect this won't really be an issue for the majority of users.

chrismclarke (changes requested):

documentation/docs/index.md
jfmcquade: I've appended it to be your shorter version, much better.

I know what you mean about this not being relevant to most users. And generally I don't like having parts of the main docs that go into detail about specific potential problems. But in this case I think having it in an isolated box upfront might be worth it – an error was only seen at the point of running `yarn workflow sync` after completing all the previous setup steps, at which point a library threw an error along the lines of `No such file or directory: '/user/my'` (splitting the path at the space). I think documenting in this way could catch a few cases without causing much distraction for most users. 

Although possibly I'm motivated by an awareness that there's quite a long list of possible issues that could go in a "troubleshooting" section of the docs...

chrismclarke: Sounds good, you're right that it's a confusing error to lookup in a troubleshooting guide (would still be good to add one in the future). Let's keep the main info box for now then and I'll create a follow-up low-priority issue https://github.com/IDEMSInternational/open-app-builder/issues/2411

chrismclarke (approved):

documentation/requirements.txt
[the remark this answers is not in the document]
jfmcquade: hmm strange. I just tried a few different things to prompt git to treat the file as text, but I'm not able to get it to show up in the diffs. I have now converted the encoding from `UTF-16 LE` to `UTF-8` (using `iconv`), and now when [looking at the diffs](https://github.com/IDEMSInternational/open-app-builder/pull/2407/files#diff-0c80e6a34d230b3977af474a2221d021b00a247218347d3c78237c2163949d1f) in github, the required packages are parsed from the file and displayed in the "rich diff" view. But in the "source diff" view it still says `Binary file not shown.`. Just [looking at the file on the feature branch](https://github.com/IDEMSInternational/open-app-builder/blob/docs/folder-path-issues/documentation/requirements.txt), github is able to display it, so I'm not sure what's going on

chrismclarke (approved):

chrismclarke:
Weird about the python diffs, maybe it's a github thing then. In any case I think this one should be good to go

PR #2408, opened 2024-09-11
Feat (remote assets): handle asset pack containing asset overrides

jfmcquade:
PR Checklist

## Description

Remote assets: Adds functionality to handle case of an asset pack containing asset overrides (previously only top-level/"default" assets were supported).

See below for demo.

## TODO/Next steps
- Generating asset packs and manifest
  - I have been doing this manually, we'll want a workflow(?) to handle bundling targeted assets into an asset pack, generating a manifest in `FlowTypes.AssetPack` format, and uploading everything to supabase
- Debug iOS
  - There are some issues with retrieving files from downloaded asset packs from local storage on iOS. I think a separate follow-up PR makes sense for this issue.
- Further details are included in the [RFC](https://docs.google.com/document/d/10I7lB5q4VzSiGUeenugP9guWUZ8Tnq95/edit#heading=h.gjdgxs), which should be updated after merge

## Git Issues

Closes #

## Screenshots/Videos

The template [debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit?gid=1422144003#gid=1422144003) captures various use cases. Essentially, there are the following asset files included in either the "core" assets (standard assets included in the app bundle) or the "debug_asset_pack_1" asset pack (uploaded to supabase):

|                          | Included in core assets | Included in `debug_asset_pack_1` |
|--------------------------|-------------------------|--------------------------------|
| Image 1                  | image_1 | image_1  |
| Image 2                  | No file | image_2  |
| Image 3 (global variant) | image_3  | No file  |
| Image 3 (es_sp variant)  | No file | image_3 |
| Image 4 (global variant) | image_4 | image_4 |
| Image 4 (es_sp variant)  | image_4 | image_4 |

[Google Drive](https://drive.google.com/drive/u/1/folders/19Kbwidpb_nOKpJAuRCkVLpAzWbKPzNm1) folder structure:
Screenshot 2024-09-25 at 13 28 14

[Supabase storage](https://supabase.com/dashboard/project/lzheiqzxigrkkxloksef/storage/buckets/debug) folder structure: 
Screenshot 2024-09-25 at 13 27 12

[debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit?gid=1422144003#gid=1422144003):

Screenshot 2024-09-25 at 13 29 52

Web demo:

https://github.com/user-attachments/assets/6005e22b-169f-4529-acd3-7de036b710d9

Android demo:

https://github.com/user-attachments/assets/c19aee69-11d0-4811-9a9f-5f9936955094


packages/data-models/deployment.model.ts
```diff
@@ -133,6 +133,14 @@ export interface IDeploymentConfig {
   _parent_config?: Partial<IDeploymentConfig & { _workspace_path: string }>;
 }
 
+/** Duplicate type defintion from data-models (TODO - find better way to share) */
+interface IFlowTypeBase {
```
jfmcquade, line 137: Moved from lower down to group with other relevant definitions

src/app/shared/components/template/components/layout/popup/popup.component.scss
```diff
@@ -18,7 +18,7 @@
     .popup-container {
       height: var(--safe-area-height);
     }
-    .close-button { 
+    .close-button {
```
jfmcquade, line 21: Not sure why this is cropping up now, but reverting it and committing isn't possible as the linter removes the whitespace again when I attempt a commit

esmeetewinkel (commented):
Looks good, videos makes sense. Should we notify the user somehow whether or not the download was successful?

I added the `test - preview` label to the PR to test on web, but got an error saying the `debug` deployment config will need updating first. 
image
@jfmcquade Could you update the `debug` config as an example? I guess I could then test by adding the `test - preview` label to that content repo PR (ensuring that the debug repo targets this code PR branch).

esmeetewinkel:
@jfmcquade It looks to me like the `debug` storage bucket on Supabase sits within a project named `plh`. Should we rename that `open-app-builder`?

image

jfmcquade:
> I added the `test - preview` label to the PR to test on web, but got an error saying the `debug` deployment config will need updating first. @jfmcquade Could you update the `debug` config as an example? I guess I could then test by adding the `test - preview` label to that content repo PR (ensuring that the debug repo targets this code PR branch).

That's strange, the [debug config on the main branch](https://github.com/IDEMSInternational/app-debug-content/blob/main/config.ts) should already be configured correctly. I get the same error on the [live web preview](https://idems-debug.web.app/template/debug_remote_assets), but not locally. I suspect this is something to do with the way that the encrypted supabase config is supposed to be decrypted and incorporated into the deployment config. @chrismclarke can you think of why this might not be working when the deployment is set as part of the github action? Looking at the logs for [the deployment preview action triggered by this PR ](https://github.com/IDEMSInternational/open-app-builder/actions/runs/11124387833/job/30909835949) (see "Set deployment" section of "build"), I can see that initially there are logs complaining that the encrypted configs do not exist, but this is before the deployment has been set. Then when the deployment is set, the encrypted configs do appear to be decrypted successfully, so I may be misdiagnosing the issue.

jfmcquade:
> @jfmcquade It looks to me like the `debug` storage bucket on Supabase sits within a project named `plh`. Should we rename that `open-app-builder`?

I think we probably should. We can specify a target supabase project via the deployment config, so in theory it would be a good idea to keep separate projects for different deployments, or at least groups of deployments like PLH. However, as supabase only allows an organisation to have two free projects at once, it may well make sense to, at least initially, start with a single "open app builder" project that we can use to store everything.

chrismclarke (approved):
Thanks for adding this. I really like the debug template you've setup, does a very good job of illustrating core functionality, although I appreciate it's a bit tricky to manage differences between web and native.

I've added one minor tidying PR, just to move the asset type definitions to their own file
5fdf023610bbebf6c1f9a6a8f59417ef2f3a0681, and have confirmed that the tests and debug sheet are working locally.

I'm not sure quite what's next for this feature, although I would recommend when next working on the code that it might be worth trying to split out the `remote-asset.service` code a bit as the file is getting quite long and hard to keep track of what is and isn't being tested. It might be worth separating the actions to a `remote-asset.actions.ts` with corresponding `remote-asset.actions.spec.ts` and possibly there might be some use having a `remote-asset.utils.ts` to handle methods that are purely data manipulation (no outside service interaction). 

I'd also recommend splitting out the code that initialises supabase into a `supabse.service.ts`

Somewhat unrelated, I can see the `RemoteAssetService` is initialised in the main app as a blocking service. Given that most of the interaction will be from user-generated actions I'd say likely this can be deferred to after main init.

Oh, and I noticed the inline comment about queueing parallel downloads - agreed this would be nice to have although before getting to that point it might be worth considering some sort of UI/popup to display that shows the overall progress of the remote asset download, as I can't really think of a tidy way to include as a variable within templates at this point in time (could be configured later)

chrismclarke:
> That's strange, the [debug config on the main branch](https://github.com/IDEMSInternational/app-debug-content/blob/main/config.ts) should already be configured correctly. I get the same error on the [live web preview](https://idems-debug.web.app/template/debug_remote_assets), but not locally. I suspect this is something to do with the way that the encrypted supabase config is supposed to be decrypted and incorporated into the deployment config. @chrismclarke can you think of why this might not be working when the deployment is set as part of the github action? Looking at the logs for [the deployment preview action triggered by this PR ](https://github.com/IDEMSInternational/open-app-builder/actions/runs/11124387833/job/30909835949) (see "Set deployment" section of "build"), I can see that initially there are logs complaining that the encrypted configs do not exist, but this is before the deployment has been set. Then when the deployment is set, the encrypted configs do appear to be decrypted successfully, so I may be misdiagnosing the issue.

Ah, yeah I see the issue. I get the same problem locally when importing the debug repo for the first time. In order to decrypt the repo the repo first needs to be set (hence on first load it says encrypted not found), but then after being set and decrypted there's no update to the `config.ts` file so it uses the cached `config.json`

A very hacky workaround would be to set the repo, decrypt the repo, and then force the repo to set again (manually deleting `config.json` and `activeDeployment.json`), but really it would be good to address in a better way as this isn't the first time this sort of issue has come up (think there was a similar issue recently when an update to nested skin wouldn't prompt recompile).

Would it be possible to create a standalone issue to address? I'm guessing this might become a high priority sooner if we have any production apps that rely on encrypted config

esmeetewinkel (approved):
Functional test passed on PR preview link

Remaining point for potential follow-up discussion:
> Should we notify the user somehow whether or not the download was successful? I'm thinking of the case where the device doesn't have storage space available.

PR #2421, opened 2024-09-13
chore: pr labeller actions

chrismclarke:
PR Checklist

## Description
The current pr-label action has two sub-steps to update labels based on title and to update labels based on code paths changed. To cover both cases the action needs to be triggered either when the pr title is changed, or whenever any code is changed. This results in both actions being triggered at times when not necessary, and some duplicate logging. This PR aims to fix by:

**Main Changes**

- Separate the `pr-labeller` action into 2 separate actions with different triggers
- Update path labeller action to newer version, refactor label config as required (should result in less duplicate re-labelling)
- Update title action to only trigger on first pr open and when marked ready for review (can't avoid duplicate re-labelling on any pr edit otherwise)

**Adiitional Changes**
- Add auto-labeler config to mark changes to scripts

Note - this will have no impact on the repeated `temporarily deployed to debug...` messages that appear any time an action loads the debug environment. Taking a quick look at  https://github.com/orgs/community/discussions/58018 it seems this is not currently possible.

## Review Notes
- Make changes to pr description, shouldn't trigger any actions
- Make changes to title, shouldn't trigger any actions
- Mark as ready for review, should trigger pr title labelled action
- Push code changes, should trigger paths action

## Dev Notes
We could add more labels for things like `frontend` if touching src components, although not sure if makes sense for now given the potential to misinterpret how changing scripts or data-models can also have an impact on frontend UI components.

I wanted to get the title labeler to retrigger on any changes to the title, however the `edited` trigger option includes any changes to pr description (which I do a lot) resulting in a lot of repeated relabelling (see history below).

So I've now changed to only run on first open and then when requesting review (to make sure updated at that time)

## Git Issues

Closes #

## Screenshots/Videos


chrismclarke:
close in favor of https://github.com/IDEMSInternational/open-app-builder/pull/2421

PR #2422, opened 2024-09-13
chore: update pr labeller actions

chrismclarke:
PR Checklist

## Description
The current pr-label action has two sub-steps to update labels based on title and to update labels based on code paths changed. 

To cover both cases the action needs to be triggered either when the pr title is changed (trigger by the pr `edited` action, which also includes pr description), or whenever any code is changed (triggered by the pr `synchoronized` action). 

This results in both actions being triggered at times when not necessary, and some duplicate logging. This PR aims to fix by:

**Main Changes**

- Separate the `pr-labeller` action into 2 separate actions with different triggers
- Update the title labelled to only trigger on first pr open and when marked ready for review (can't avoid duplicate re-labelling on any pr edit otherwise)
- Update the path labelled to only trigger on first open and when new commits pushed

**Adiitional Changes**
- Update path labeller action to newer version, refactor label config as required (should result in less duplicate re-labelling)
- Add auto-labeler config to mark changes to scripts
- Update title labelled to catch both upper and lower case variants

Note - this will have no impact on the repeated `temporarily deployed to debug...` messages that appear any time an action loads the debug environment. Taking a quick look at  https://github.com/orgs/community/discussions/58018 it seems this is not currently possible.

## Review Notes
- Make changes to pr description, shouldn't trigger any actions
- Make changes to title, shouldn't trigger any actions
- Mark as ready for review, should trigger pr title labelled action
- Push code changes, should trigger paths action

You'll notice that within this PR the only time labels are added is when marked ready for review (as initial trigger on pr load the title of the pr wasn't formatted correctly)

## Dev Notes
We could add more labels for things like `frontend` if touching src components, although not sure if makes sense for now given the potential to misinterpret how changing scripts or data-models can also have an impact on frontend UI components.

I wanted to get the title labeler to retrigger on any changes to the title, however the `edited` trigger option includes any changes to pr description (which I do a lot) resulting in a lot of repeated relabelling (see history in #2421).

So I've now changed to only run on first open and then when requesting review (to make sure updated at that time). I'm also now wondering if this might be the preferred behaviour on the paths labelled action too (in which case could have just keep as a single action??) - this pr doesn't make any changes to docs or scripts so hard to tell the impact...

## Git Issues

Closes #

## Screenshots/Videos


.github/pr-labeler.config.yml
```diff
@@ -1,3 +1,12 @@
-# Additional labels to assign based on file changes
-"documentation":
-  - documentation/**/*
+# Patterns to match to auto-assign labels based on path changes
+# https://github.com/actions/labeler/tree/v5/?tab=readme-ov-file#match-object
```
jfmcquade, line 2: This is nicely extendable. I can see how a `frontend` label could also be useful (or even a `component` label for files in `src/app/shared/components/template/components/**/*`), but as you say in the PR description, such a label would likely not be applied in all cases and so may therefore have limited usefulness

jfmcquade (approved):
Looks good. I can confirm that the triggers work as expected when testing on this PR.

I agree that frequent PR description editing means that using that a s a trigger would cause the action to run with unnecessary frequency. I think triggering the paths action may result in some unnecessary action runs, but I don't think it should be too bad and is likely to catch cases where otherwise we might have to remember to manually convert to draft and then re-mark as ready for review in order to trigger. So I think probably fine as is

PR #2432, opened 2024-09-20
fix: scripts app_config

chrismclarke:
PR Checklist

## Description

Fix regression introduced by #2423 

Default app_config variables are no longer loaded until runtime, however there appears to be an instance where scripts that post-process assets expect to access the app_config theme variables. This would results in any deployments without themes defined being unable to compile

This fix simply adds fallback empty array for cases where deployment doesn't specify themes. An more robust solution should probably include improving type-definitions to asset config is only partial during script processing, and to move the theme config to the top-level deployment config - however likely not worth the time involved right now. 

## Git Issues

Closes #

## Screenshots/Videos


chrismclarke:
Tagging @jfmcquade so you are aware, but planning to merge as a hotfix as potential for knock-ons is minor. As well as I could check I couldn't see any other instances of scripts requiring values from default app config variables.

PR #2436, opened 2024-09-23
chore: Update bug_report.md

esmeetewinkel:
PR Checklist

## Description

Updating the phrasing of the bug report template since we no longer have only templates but also other flow types which a bug might relate to, and also following code/content separation.


jfmcquade (approved):
LGTM

PR #2440, opened 2024-09-25
Fix: generator chained flows

### LINKED ISSUE #2434: [BUG] Can't use output of generated data pipe as input of another
**Describe the bug**
When using a data list that is an `output_target` of a generated `data_pipe` as `input_source` of another `data_pipe` generator, I'm getting the error

```
ERROR! An error was encountered while executing
TypeError: Cannot destructure property '_generated' of '.for' as it is undefined.
```
**App version**
v0.16.34

**To Reproduce**
Inspect debug sheets, mark `gen_filter_tasks_by_colour` as [released](https://docs.google.com/spreadsheets/d/1A_Y1ncmUbaDAEV8W134IJcESq8d-idS_cJNExDmf5r0/edit?gid=341018764#gid=341018764&range=C7)
image

**Deployment Sheets**
[TZ modules data pipes](https://docs.google.com/spreadsheets/d/1LdV8eX-tuo9EkevIx4iILB8t4WIxUJQs_1V6c4-kaoY/edit?gid=594801224#gid=594801224) in PLH Kids TZ deployment (but these sheets are a little involved so better to look at simplified debug example below)

**Debug Sheets**
[debug_gen_data_pipe](https://docs.google.com/spreadsheets/d/1A_Y1ncmUbaDAEV8W134IJcESq8d-idS_cJNExDmf5r0/edit?gid=341018764#gid=341018764)

**Screenshots and videos**
image

chrismclarke:
PR Checklist

## Description

Fix issue where outputs generated from data_pipe or generator operations aren't processed (if output as data_pipe or generator operators)

## Review Notes
As the changes impact the main parser, should ideally test updates against a content repo that has recently been synced to see if there is any content diffs. The only expected change is that `data_pipe` flows should no longer be stored within the generated content (these have already been used to generate the populated data_lists, so aren't useful to the app runtime)

I create a content sync for the latest sheets from master branch on the debug repo at:
https://github.com/IDEMSInternational/app-debug-content/pull/88

Then, using this branch see updated content sync at:
https://github.com/IDEMSInternational/app-debug-content/pull/89

Should also confirm it fixes the issues seen in #2434

## Author Notes
Now that `data_pipe` and `generator` flows are no longer included in final output or git content repo, I'm wondering if we may still want to try and keep some reference to them - maybe just keeping a `_generated_from` property on the list itself as a references to the data_pipe or generator flow that created it? 

We could also consider whether we want to keep generated outputs from data_pipe operations the `generated` data_list subfolder (not sure if the purpose it serves could be better-handled by the proposed `_generated_from` property)

## Dev Notes
The core of the issue is the way that `_generated` flows would be saved to the local flow and then re-populated to the main list at the end of processing. This would fail to trigger any subsequent processing if the generated flow was itself a generator or data_pipe type.

Now, instead of storing lists of `_generated` flows the child data_pipe and generator processor take the outputs and send them back to the main flow processor to process as if they were part of the original list.

The largest part of the code change is actually just with the testing to enable checking the main parent FlowProcessor output from within an individual processor (e.g. data_pipe). This wasn't set up very well for testing as the parent FlowProcessor is tightly coupled with a caching system that writes files to disk - so a new Mock Class was created to substitute in for the jsonFile cache system (to just use an in-memory cache instead). In the future it would be best to try and decouple as much as possible processing logic from disk I/O to make it easier to test and separate concerns a bit more.

## Git Issues

Closes #2434

## Screenshots/Videos

### DIFF
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/data_pipe.parser.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/data_pipe.parser.ts
@@ -3,6 +3,9 @@ import { DataPipe } from "shared/src/models/dataPipe";
 import { DefaultParser } from "./default.parser";
 
 export class DataPipeParser extends DefaultParser<FlowTypes.DataPipeFlow> {
+  /** local hashmap of generated outputs. Used for tests  */
+  private outputHashmap: { [flow_name: string]: { [output_name: string]: any } } = {};
+
   /** If extending the class add additional postprocess pipeline here */
 
   public postProcessFlow(flow: FlowTypes.DataPipeFlow): FlowTypes.DataPipeFlow {
@@ -18,6 +21,9 @@ export class DataPipeParser extends DefaultParser<FlowTypes.DataPipeFlow> {
     }
     try {
       const outputs = pipe.run();
+      // HACK - populate to output hashmap for use in tests. Clone output due to deep nest issues
+      this.outputHashmap[flow.flow_name] = JSON.parse(JSON.stringify(outputs));
+
       this.populateGeneratedFlows(outputs);
       // As the populated flows will be passed directly to the processor queue
       // can just return undefined so that the data pipe will not be stored in outputs
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/data_pipe.parser.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/data_pipe.parser.ts
@@ -29,8 +29,6 @@ export class DataPipeParser extends DefaultParser<FlowTypes.DataPipeFlow> {
   }
 
   private populateGeneratedFlows(outputs: { [output_name: string]: any[] }) {
-    this.flowProcessor.processedFlowHashmap.data_list ??= {};
-
     for (const [flow_name, rows] of Object.entries(outputs)) {
       const flow: FlowTypes.FlowTypeWithData = {
         flow_name,
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/data_pipe.parser.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/data_pipe.parser.ts
@@ -19,28 +19,31 @@ export class DataPipeParser extends DefaultParser<FlowTypes.DataPipeFlow> {
     try {
       const outputs = pipe.run();
       this.populateGeneratedFlows(outputs);
+      // As the populated flows will be passed directly to the processor queue
+      // can just return undefined so that the data pipe will not be stored in outputs
+      return undefined;
     } catch (error) {
       console.trace(error);
       throw error;
     }
-    return this.flow;
   }
 
   private populateGeneratedFlows(outputs: { [output_name: string]: any[] }) {
-    const generated: FlowTypes.DataPipeFlow["_generated"] = { data_list: {} };
     this.flowProcessor.processedFlowHashmap.data_list ??= {};
 
     for (const [flow_name, rows] of Object.entries(outputs)) {
-      generated.data_list[flow_name] = {
+      const flow: FlowTypes.FlowTypeWithData = {
         flow_name,
         flow_subtype: "generated",
         flow_type: "data_list",
         rows,
       };
-      // also populate generated outputs to be available for future input sources
-      this.flowProcessor.processedFlowHashmap.data_list[flow_name] = rows;
+      const deferId = `${flow.flow_type}.${flow.flow_subtype}.${flow.flow_name}`;
+
+      // Pass all generated flows to the back of the current processing queue so that they can be
+      // populated to processed hashmap and referenced from other processes as required
+      this.flowProcessor.deferInputProcess(flow, deferId);
     }
-    this.flow._generated = generated;
   }
 
   private loadInputSources() {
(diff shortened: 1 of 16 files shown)


packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/data_pipe.parser.ts
```diff
@@ -18,29 +21,33 @@ export class DataPipeParser extends DefaultParser<FlowTypes.DataPipeFlow> {
...
         flow_type: "data_list",
         rows,
       };
-      // also populate generated outputs to be available for future input sources
-      this.flowProcessor.processedFlowHashmap.data_list[flow_name] = rows;
+      const deferId = `${flow.flow_type}.${flow.flow_subtype}.${flow.flow_name}`;
+
+      // Pass all generated flows to the back of the current processing queue so that they can be
+      // populated to processed hashmap and referenced from other processes as required
+      this.flowProcessor.deferInputProcess(flow, deferId);
```
chrismclarke, line 49: This is the main change made in both `data_pipe` and generator parser. By pushing the outputs back into the deferred input processor it then processes them in the same way as synced flows.

packages/shared/src/models/templatedData/templatedData.ts
```diff
@@ -77,18 +73,18 @@ export class TemplatedData {
    * Main data conversion method
    * Iterate over data, parse string values and nested objects and arrays recursively
    */
-  public parse<T>(value: T = this.initialValue) {
+  public parse<T>(value: T) {
```
chrismclarke, line 76: additional bug fix - because the `TemplatedData` would use an initial value to parse could hit an infinite loop if the initial value contains an `undefined` child entry (trying to recursively parse would then re-populate the initial value each time). 

So have just made minor code modifications to not include the initialValue

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/generator.parser.ts
```diff
@@ -23,13 +20,23 @@ export class GeneratorParser extends DefaultParser<FlowTypes.GeneratorFlow> {
       return;
     }
     try {
-      this.flow._generated = this.generateFlows(flow, dataListRows);
-      // this.handleOutputs(generated);
+      const generated = this.generateFlows(flow, dataListRows);
```
jfmcquade, line 23: Am I right in thinking that the author could have specified a name for the output flow that is identical to the name of another output flow (or pre-existing flow) and that it wouldn't have been picked up until this point? In that case, the most recently processed flow with that name would overwrite the previous one without throwing any errors, so it could be worth handling here

jfmcquade (approved):
Looks good I think. Good idea to include those content PRs to reference.

I can see how a flow doesn't necessarily need to store the history of how it was created (generated and manually authored flows will be treated the same way at app runtime). But at the same time, I can see how having access to the parsed versions of the generator/data pipe sheets could be useful for debugging, and how having version control on these intermediary sheets (i.e. storing them in the git repo) might be useful. @esmeetewinkel have you found yourself debugging pipes/generators by looking at the parsed JSON version at all?

I can see how `_generated_from` metadata field (or similar) could be sensible for debugging, and may be sufficient vs storing the intermediary templates. As I haven't used the system much I'm not sure what would be necessary/useful.

esmeetewinkel (approved):
Functional test passed - can confirm this fixes the issues seen in https://github.com/IDEMSInternational/open-app-builder/issues/2434

Indeed when performing a content sync on this branch all `data_pipe` and `generator` sheet jsons disappear. These (in particular the data pipe sheets) have occasionally been helpful for debugging, so it would be good if these jsons would still be available to inspect somewhere.

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/generator.parser.ts
chrismclarke: Created https://github.com/IDEMSInternational/open-app-builder/issues/2451 for future follow-up

chrismclarke:
> Functional test passed - can confirm this fixes the issues seen in #2434
> 
> Indeed when performing a content sync on this branch all `data_pipe` and `generator` sheet jsons disappear. These (in particular the data pipe sheets) have occasionally been helpful for debugging, so it would be good if these jsons would still be available to inspect somewhere.
> 
> > maybe just keeping a _generated_from property on the list itself as a references to the data_pipe or generator flow that created it?
> 
> I don't like the duplication that would happen if we stored the parsed data pipe json within every data list it generates. Storing just the data pipe ID as a `_generated_from` property does make sense to me (and making the parsed data pipe json accessible some other way).
> 
> > We could also consider whether we want to keep generated outputs from data_pipe operations the generated data_list subfolder (not sure if the purpose it serves could be better-handled by the proposed _generated_from property)
> 
> I am quite happy that this sits in a generated subfolder, I've found it handy for debugging: Since the git changes on a generated data list are always caused by changes on a different sheet / layer, I don't inspect them in detail when running a content sync. Having these data lists sit in a `generated` subfolder means I can just spot check a few files and then accept the changes on the entire subfolder.

Makes sense, in that case I'll keep the `generator` and `data_pipe` flows within the repo and can just update the code that copies to the app to exclude (as not needed at runtime). I'll also try add a reference to both input/output flows to keep track of relationships (avoiding duplicate content, just by flow type/name)

I think to make things clearer I'll merge here for now and then address that as a follow-up

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/generator.parser.ts
jfmcquade: Great, thanks for making the issue. I hadn't considered the case of intentionally overwriting a flow with a generated one, so good to have that possibility captured

PR #2460, opened 2024-10-08
feat: audio component can be passed transcript text and launch own popup

jfmcquade:
PR Checklist

## Description

Follow up to #2448.

Adds a parameter, `transcript_text` to the audio component. This parameter takes a string value, which should represent the text of the transcript for the associated audio file. If a value for `transcript_text` is provided, then the audio component will show the "info button" icon (i.e. `show_info_button` is set to true), and clicking on this button will automatically launch a pop-up to display the text when the "info button" is clicked, without having to manually author an action, as was the case after #2448 (although custom action handling is still supported).

## Git Issues

Closes #

## Screenshots/Videos

[comp_audio](https://docs.google.com/spreadsheets/d/1utD4Kr0WVjvDwnMbg2wYupaPQ0VY0H2csb86fyi4bPI/edit?gid=603128459#gid=603128459)

Screenshot 2024-10-08 at 17 15 56

Screenshot 2024-10-08 at 17 16 28


src/app/shared/components/template/components/audio/audio.component.html
```diff
@@ -5,9 +5,15 @@ <h3>{{ params.title }}</h3>
...
+            <ion-icon
+              [src]="params.infoIconAsset | plhAsset"
+              (click)="params.transcriptText ? openTranscriptPopup() : clickInfo()"
+            ></ion-icon>
           } @else {
-            <ion-icon name="document-text-outline" (click)="clickInfo()"></ion-icon>
+            <ion-icon
+              name="document-text-outline"
+              (click)="params.transcriptText ? openTranscriptPopup() : clickInfo()"
+            ></ion-icon>
```
chrismclarke, line 16: nit(non-blocking): I'm never a fan of attaching click events directly to icons as they often don't have a very noticeable (or accessible) clickable region. Would instead recommend using a `ion-button` element for the container and attaching the click-handler to it (also less duplication). At the same time it's probably better to just write a simple `handleActionButtonClick` method and deal with the conditional logic there, as you've got 4 inline handlers that will re-evaluate every render even when no button is clicked (not a huge deal, but should just help keep the HTML a little cleaner). E.g.

```html
<ion-button fill="clear" (click)="handleActionButtonClick()">
          @if (params.infoIconAsset) {
            <ion-icon slot="icon-only" [src]="params.infoIconAsset | plhAsset"></ion-icon>
          } @else {
            <ion-icon slot="icon-only" name="document-text-outline"></ion-icon>
          }
        </ion-button>
 ```
 You might also find that with the inherited ion-button styles you don't actually need the custom style class anymore
 
 If you didn't want to duplicate the div you could define it as a named `ng-template` and use `ng-container` to inject into both sections (https://angular.dev/guide/templates/ng-template#using-ngtemplateoutlet), although might be overkill

File: src/app/shared/components/template/components/layout/popup/popup.component.html [hunk 2]
```diff
@@ -8,14 +8,18 @@
...
+      } @else {
+        <plh-template-container
+          class="template-container"
+          [name]="props.name"
+          [templatename]="props.templatename"
+          [parent]="props.parent"
+          [row]="props.row"
+          (emittedValue)="handleEmittedValue($event)"
+        ></plh-template-container>
+      }
```
chrismclarke, line 22: Nice way to be able to toggle between text or template popups

src/app/shared/components/template/components/layout/popup/popup.component.html
```diff
@@ -8,14 +8,18 @@
...
-      <plh-template-container
-        class="template-container"
-        [name]="props.name"
-        [templatename]="props.templatename"
-        [parent]="props.parent"
-        [row]="props.row"
-        (emittedValue)="handleEmittedValue($event)"
-      ></plh-template-container>
+      @if (props.textOnly) {
+        <div class="large standard normal" [innerHTML]="props.text | markdown"></div>
```
chrismclarke, line 12: `large standard normal` - which is it??! 😅

src/app/shared/components/template/components/layout/popup/popup.component.ts
```diff
@@ -53,4 +53,7 @@ export interface ITemplatePopupComponentProps extends ITemplateContainerProps {
   waitForDismiss?: boolean;
   /** Display fullscreen overlayed on top of all other app content */
   fullscreen?: boolean;
+  /** Pass text only to render that text in a popup, bypassing any templates */
+  textOnly?: boolean;
+  text?: string;
```
chrismclarke, line 58: nit(blocking): we're using 2 props to essentially handle one thing, displaying text within the popup (instead of a template).

I think probably better to just call it `popupText` (to distinguish from `template` which should eventually be renamed to `popupTemplate`). That way you also don't have the confusion if `textOnly:true` but no text defined or `text` passed but not `textOnly`

chrismclarke (changes requested):
Really nice you were able to get this working so smoothly, I really like the implementation overall and think it should simplify things for authors nicely. See a few minor comments inline, but otherwise I'd say should be good to go

src/app/shared/components/template/components/audio/audio.component.html
jfmcquade: Thanks, that's much neater now, see https://github.com/IDEMSInternational/open-app-builder/pull/2460/commits/5ecc580258108370b5f327586b41c0531fa134c2

src/app/shared/components/template/components/layout/popup/popup.component.html
jfmcquade: Yeh this is pretty gross... I was using the same pattern to apply those styles as is used by the `text` component, but now I've updated it so that the pop-up component, when called to display text only, uses an instance of the `text` component itself. See https://github.com/IDEMSInternational/open-app-builder/pull/2460/commits/50cc535af47fcd90f96199d1a09b1012b9dff006

The text component styling could still do with some tidying up, but at least this means that it's still only in one place

jfmcquade:
Thanks @chrismclarke, I've addressed your points. I think this should be good to go, pending any opinions you have on my changes to the `text` component to enable it to be instantiated from within a parent component, see https://github.com/IDEMSInternational/open-app-builder/pull/2460/commits/50cc535af47fcd90f96199d1a09b1012b9dff006

chrismclarke:
> Thanks @chrismclarke, I've addressed your points. I think this should be good to go, pending any opinions you have on my changes to the `text` component to enable it to be instantiated from within a parent component, see [50cc535](https://github.com/IDEMSInternational/open-app-builder/commit/50cc535af47fcd90f96199d1a09b1012b9dff006)

I really like leveraging the existing text component to be able to inherit consistent styling, although would much prefer to isolate the text component from these changes. We've already had to update the common popup component to add support for direct text, I think it starts to feel a bit over-engineered to then also be changing the text component to handle case when being rendered directly within a popup. 

So I'd say better to just create an ad-hoc template row in the popup component so that the text can display in the usual way

src/app/shared/components/template/components/layout/popup/popup.component.html
```diff
@@ -8,14 +8,18 @@
...
-        class="template-container"
-        [name]="props.name"
-        [templatename]="props.templatename"
-        [parent]="props.parent"
-        [row]="props.row"
-        (emittedValue)="handleEmittedValue($event)"
-      ></plh-template-container>
+      @if (props.popupText) {
+        <plh-tmpl-text [textInputParams]="{ text: props.popupText }"></plh-tmpl-text>
+      } @else {
```
chrismclarke, line 13:
```suggestion
        <!-- Use a text component row to display popup text -->
        <plh-tmpl-text
          [row]="{ _nested_name: '', name: '', type: 'text', value: props.popupText }"
        ></plh-tmpl-text>
      } @else {
        <!-- Use a container to display popup template -->
```

chrismclarke:
Possibly too many changes to easily merge as "suggested changes" but hopefully you get the idea at least. Let me know what you think

chrismclarke (changes requested):

jfmcquade:
> So I'd say better to just create an ad-hoc template row in the popup component so that the text can display in the usual way

Ah good idea. I'd forgotten how easy it is to pass in a value for `[row]`. I've updated this PR to use that pattern, and reverted the changes to the text component (I think I'll make a follow-up PR just refactoring `text` somewhat)

chrismclarke (approved):
Looks good to me, thanks for making the updates. Oh yeah, nice idea to tidy up the text component, had thought the same while looking at it

esmeetewinkel (approved):
Functional test passed

PR #2462, opened 2024-10-09
fix: encrypted deployment set

chrismclarke:
PR Checklist

## Description

Fix issue where setting config for first time (e.g CI actions) fails to include encrypted deployment values.

## Review Notes
Should be able to confirm if supabase config loaded in test-preview action

## Dev Notes
The reason deployment config would be loaded before decrypt was to handle the case where the name of the deployment isn't known (in which case the deployment set action prompts name input). I've updated the workflow to add a standalone step to handle ensuring a name is provided (either as a command arg or through prompt), and using that can force deployment to be decrypted before compiling the full deployment json.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/scripts/src/commands/deployment/set.ts
+++ b/packages/scripts/src/commands/deployment/set.ts
@@ -68,7 +68,7 @@ export class DeploymentSet {
     }
   }
 
-  private async promptDeploymentSelect() {
+  public async promptDeploymentSelect() {
     const deploymentNames = fs
       .readdirSync(DEPLOYMENTS_PATH, { withFileTypes: true })
       .filter((d) => d.isDirectory())
--- a/packages/scripts/src/commands/deployment/utils.ts
+++ b/packages/scripts/src/commands/deployment/utils.ts
@@ -63,6 +63,7 @@ function parseConfigJson(json: IDeploymentConfigJson) {
 /**
  * Decide if a config needs recompiling by checking if modified timestamps
  * on the .ts and .json files are the same (applied during compilation)
+ * TODO - also consider files from encrypted or other imports (e.g. skins)
  */
 function isConfigUpToDate(tsPath: string, jsonPath: string) {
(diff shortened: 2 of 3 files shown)


jfmcquade (approved):
Nice, thanks for debugging. The changes make sense and inspecting [the logs of the "set deployment" step of the build step within the "test preview" action](https://github.com/IDEMSInternational/open-app-builder/actions/runs/11247767733/job/31271786150), I can confirm that the decryption is now running successfully before the deployment is set.

Screenshot 2024-10-09 at 12 27 10

chrismclarke:
image

Great, yep I can also see from the preview site logs that supabase is now loaded.

jfmcquade:
> image
> 
> Great, yep I can also see from the preview site logs that supabase is now loaded.

Just thought that logging the fully populated deployment here is something of a security risk, as values coming from the encrypted folder could well be sensitive? Hiding only properties identified as coming from the encrypted folder or otherwise identified as "sensitive"/"secret" could be nice, but perhaps it makes sense to just not log the config value on innit for now?

chrismclarke:
Yeah I don't think we really need to log (more of a debug), although anything included in a web build is no longer secure (can always be accessed directly from source code) so shouldn't mistake the use of our encryption system as anything more than limiting what is exposed on github (vs what is exposed to people trying to directly access from one of our sites). It doesn't remove any threats, just limits the accessibility. Removing the logs would be one more step in the same process (would likely favour keeping for the debug site and removing for others).

That's why when using platforms like firebase and supabase that enable direct access from the platform to resources the only way to actually secure is via in-built security rules (e.g. restricting by auth or domain). This is a challenge for us as we also want access from native devices which aren't requesting resources from a known domain under our control. It might be worth reviewing at some point how open our supabase instance is to abuse (also firebase, assuming we don't fully remove it in the near future)

PR #2463, opened 2024-10-09
Update reusable-content-sync.yml

ChrisMarsh82:
Defined secrets explicitly so can be passed from other organisations.  
Removed Deployment Private Key as not used


jfmcquade (approved):
Makes sense. I suppose there isn't a way to test that running from a different org's repo correctly passes in the secrets until after we've merged?

It could be worth keeping the `DEPLOYMENT_PRIVATE_KEY` as an optional secret (and using it to decrypt the config if it is supplied), as possibly in the future there could be a value from the `encrypted` deployment folder that is required for syncing content – e.g. a folder ID or URL that is considered sensitive. I can't really see an immediate use-case for that though, so happy to merge as-is.

PR #2464, opened 2024-10-09
chore: refactor text component

jfmcquade:
PR Checklist

## Description

Minor refactoring to tidy up the `text` component code and file structure. Originally included as part of #2460. 

I had hoped that using the `git mv` command to move/rename the typescript file would mean that the diffs displayed more clearly in github, but apparently not. The changes just consist of:
- Wrapping the template param properties in a typed `params` property
- Extracting the html to its own file
- Using the modern Angular control flow syntax

## Git Issues

Closes #

## Screenshots/Videos

The [comp_text](https://docs.google.com/spreadsheets/d/1iZHeAAZAQ8q-XNKrW_ww3yvPUVGwc407uhAF7vu1TtM/edit?gid=569531329#gid=569531329) component displaying as expected, with no visual changes:
Screenshot 2024-10-09 at 17 52 56


src/app/shared/components/template/components/text/text.component.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+@Component({
+  selector: "plh-tmpl-text",
+  templateUrl: "./text.component.html",
+  styleUrls: ["../tmpl-components-common.scss"],
+})
+export class TmplTextComponent extends TemplateBaseComponent implements OnInit {
+  params: Partial<ITextParams> = {};
+  isFalsy: boolean;
```
chrismclarke, line 18: nit(non-blocking): I can appreciate this accurately describes the value, but not so much what it actually impacts. Maybe something more like `hasTextValue` would be clearer.

chrismclarke (approved):
Thanks for tidying this up. Yeah I have no idea how/when git decides to show as renamed with diff vs deleted + recreated, but in any case the changes are clear enough to follow and look good to me

PR #2468, opened 2024-10-11
feat: add plh_kids_kw theme

jfmcquade:
PR Checklist

## Description

Adds a new theme, `plh_kids_kw`, to be used for making style changes for the `plh_kids_kw` deployment. Currently, this theme is just a deuplicate of the existing `professional` theme.


esmeetewinkel (approved):
Sounds good. There's a chance that Laurie will want a theme very similar to the one to-be-used in PLH Kids Kuwait to be the default one for PLH apps in general, at which point we might want to use a name that doesn't refer to the deployment in specific. However, I think the current name serves the current purpose and we can rename once this has taken shape (and it will be easier to think of a descriptive name once we have something to look at).

PR #2471, opened 2024-10-14
[FEATURE] Nunito Default Font

### LINKED ISSUE #2470: [FEATURE] New Typeface for English version
**What?**
Refactoring the KW App font typeface to use Nunito for the English version

**How?**
image

FaithDaka:
PR Checklist

## Description

Refactors the default font from Roboto to Nunito, with the font files added to 'assets/fonts'. Also includes colour changes to the plh_kids_kw theme

## Git Issues

Closes #2470 

## Screenshots/Videos
image


src/theme/_fonts.scss
```diff
@@ -3,6 +3,17 @@
   --ion-font-family: "Roboto", "Helvetica Neue", sans-serif;
 }
 
+// Override roboto font
+:root {
+  --ion-font-family: "Nunito", sans-serif;
```
jfmcquade, line 8: nit (blocking): It would be good to move this (and the `@font-face` declaration below) to the `plh_kids_kw` theme file. Whilst we may well want to incorporate these changes into the default styles, if we keep them scoped to the `plh_kids_kw` theme for now then we avoid making changes to our existing production apps, and we can decide if/when to roll out the changes independently. 

Ideally, it would be good to be able to expose a new variable for `$font-name` (alongside `$color-primary`, `$color-secondary` and `$page-background` that already exist at the top of the theme files), and infer all the necessary changes from that. However I think that could be tricky as we'd need to account for different font types (we already have `.ttf` and `.woff2` to handle). 

So I think for now it might make sense to do the following:
1. Move the whole `@font-face` declaration block as it is into the theme file (i.e. under line 10 of `src/theme/themes/plh_kids_kw.scss`)
2. Move this variable assignment to the theme file
    - Adding `ion-font-family: "Nunito", sans-serif,` to the list of `$variable-overrides` should work I think – the `--` gets added automatically

Let me know if that doesn't make sense

jfmcquade (changes requested):
See inline comment – I'd like to keep these changes scoped to the `plh_kids_kw` deployment for now

chrismclarke:
Just to quickly add, given all apps run within a (reasonably) modern version of webkit/chrome to run I think we should be able to reliably use `woff2` versions of fonts instead of `ttf` - you'll see that's the format roboto fonts use.

Woff files tend to be a lot smaller (~10x for nunito), so would be good to use so that overall we only add a couple hundred kb to the project instead of the current 2-3MB included in this PR
https://getsnapfont.com/posts/understanding-font-file-formats-ttf-otf-woff-etc

Annoyingly google fonts don't offer woff2 files directly, but you can find most fonts at:
https://gwfh.mranftl.com/fonts/nunito?subsets=latin

Alternatively you can convert using:
https://www.fontsquirrel.com/tools/webfont-generator

I would also recommend matching the current font-weights (400 standard, 700 bold), with italic variant on 400 also. I can't see any other font variants defined in the `_typography.scss` but possibly @jfmcquade can confirm whether any other variants expected.

src/theme/_fonts.scss
FaithDaka: @jfmcquade The scoping of changes does make sense. Given that the font change was required globally, it made sense for me to have that in the variables file but I do understand how that makes deployment somewhat complicated.
I'll work on keeping changes within the theme file.

FaithDaka:
@chrismclarke Yeah I did download the Nunito font files from the Google library which was the `.tff` format. I'll switch to `.woff`, thanks.
And the weight variants were defined in that Nunito `wght` tff file so that they get specified using the `font-weight` CSS attribute but I can define them if that abstracts the font weight.

FaithDaka:
For review @jfmcquade. Style changes kept to theme file

Default Theme: Roboto Font
image

Kuwait Theme: Nunito Font
image

chrismclarke (approved):
Thanks for making the updates @FaithDaka 

Yeah google fonts changed a while back to start using variable fonts, which can give a lot more flexibility to customise fonts, but result in much larger font files. These work well if serving the font from a cdn and only downloading on demand the exact styles needed (and also nice to install on your computer and use locally), but not very well suited for serving offline. 

Given that any fonts needed in the app will have to be bundled into the build, using `woff2` files just for the specific weights we're likely to need is the best option - I can see each of the ones you included are only around 15kb compared to 160kb for the previous ttfs

jfmcquade (approved):
Hi, apologies for the delay in reviewing, but this looks good to me, I will approve and merge

PR #2478, opened 2024-10-22
chore: add plh component folder and example component

jfmcquade:
PR Checklist

## Description

- Adds a folder for demo templates
  - Including a `basic` example component, to demonstrate core component boilerplate and logic
- Adds a folder for PLH-specific components

From `packages/components/demo/README.md`:

##### Intro
This demo package contains example template components that can be instantiated through the template app authoring system

##### Basic component
This package contains a "basic" component for demonstration purposes. The corresponding component demo template spreadsheet can be found [here](https://docs.google.com/spreadsheets/d/1QrlhmASCDfnorV1f4XTkWAUOJlGHopCTruIDutsCY1k/edit?gid=603128459#gid=603128459).

###### Generating component files
New components can be added to a particular component package using the Angular CLI. For example, a new component can be added to the `demo` package by running the following command: 
```sh
ng generate component packages/components/demo/<component-name>
```

###### New template component boilerplate
 To make the component available to the core app code, the component class must be added to `exports` and `declarations` of the `index.ts` file for the component package. Additionally, In order to expose the new component to the spreadsheet template authoring level, a mapping should be added to the relevant `COMPONENT_MAPPING` object. The key that assigned to the value of the component class is the string that will be used in authored templates to instantiate the component. See `packages/components/demo/index.ts` for an example of these patterns.

## Git Issues

Closes #

## Screenshots/Videos

New component demo template, [comp_example](https://docs.google.com/spreadsheets/d/1QrlhmASCDfnorV1f4XTkWAUOJlGHopCTruIDutsCY1k/edit?gid=603128459#gid=603128459):

Screenshot 2024-10-23 at 15 49 11

Screenshot 2024-10-23 at 15 50 12


FaithDaka:
Thank you for the detail @jfmcquade. This is super helpful.

packages/components/demo/example/example.component.scss
```diff
@@ -0,0 +1,10 @@
+.example-container {
+  width: 80px;
+  &[data-variant="red_square"] {
+    background-color: red;
```
chrismclarke, line 4: nit(blocking): could you update to show how to access colour variables instead of hardcoding?

packages/components/demo/example/example.component.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { ComponentFixture, TestBed, waitForAsync } from "@angular/core/testing";
+import { IonicModule } from "@ionic/angular";
+
+import { TmplExampleComponent } from "./example.component";
+
+describe("TmplExampleComponent", () => {
```
chrismclarke, line 6: nit(non-blocking): it would be good to check if this code actually runs, and possibly add a comment to explain how to test this component in isolation (filter tests to just include this). If possible it would also be good to update the test to pass a mock value for `_row` so that it can test against how the component would normally be mounted.

packages/components/demo/example/example.component.ts
```diff
@@ -0,0 +1,46 @@
...
+  /** TEMPLATE_PARAMETER: "variant". Default "red_square" */
+  variant: "red_square" | "yellow_circle";
+  /** TEMPLATE_PARAMETER: "my_boolean_param". A paramater that takes a boolean value. Default true */
+  myBooleanParam: boolean;
+  /** TEMPLATE_PARAMETER: "my_string_param". A string to be displayed */
+  myStringParam: string;
+}
+
+@Component({
+  selector: "plh-example",
```
chrismclarke, line 18: nit(blocking): it would be good to update this component to use the prefix `demo`, and possibly rename the component from `example` -> `basic`, so the full prefix would be `demo-basic`

packages/components/demo/example/example.component.ts
```diff
@@ -0,0 +1,46 @@
...
+  templateUrl: "./example.component.html",
+  styleUrls: ["./example.component.scss"],
+})
+export class TmplExampleComponent extends TemplateBaseComponent implements OnInit {
```
chrismclarke, line 22: nit(blocking): I think we should probably drop the `Tmpl` prefix convention so that new component collections aren't as likely to confilct with each other. I'd say better to prefix with the namespace I've moved the component under, i.e. `DemoBasicComponent`

chrismclarke (changes requested):
@jfmcquade 
Thanks for adding the example. I've gone ahead and moved this into a standalone `components/demo` namespace as an example of how groups of components can be managed together (see 3c2708).

This was surprisingly straight forward, so I'd say looking ahead it would be good to also create similar folders for `plh` components, and we might even want to consider slowly migrating our own `core` components (although possibly better to wait until we do a bit more tidying for how we manage feature-specific components)

I haven't refactored the existing components, and as such the example component won't be able to import any existing, however I did add imports for ionic so we could include example with ionic components in a future demo

Otherwise I think things are looking good. See a couple comments inline, mostly related to the component now living within a dedicated component namespace (so would be good to update component naming conventions to reflect)

packages/components/demo/example/example.component.html
```diff
@@ -0,0 +1,5 @@
+@if (params.myBooleanParam) {
+  <div class="example-container" [attr.data-variant]="params.variant" (click)="handleClick()">
+    {{ params.myStringParam }}
+  </div>
+}
```
chrismclarke, line 5: nit(blocking): really nice to have the examples of accessing params, although could we also include something that accesses the row value as well? (i.e. `_row.value`)

jfmcquade:
Thanks, @chrismclarke. I've acted on your inline suggestions.

> Oh, and if you did want to include a plh-specific example we could either (as part of this PR or a follow-up) create a similar `packages/components/plh` folder and add the `parent-point-counter` component to it as a starter

I've added a `packages/components/plh` folder with some boiler plate, however I couldn't trivially move over the `parent-point-counter` because the component template makes use of `plhAsset` pipe. Importing this into the `PlhComponentsModule` throws an error as it can't declared in more than one module (the `parent_point_box`/"points-item" component also uses the `plhAsset` pipe, so can't be moved over for the same reason). Exposing our pipes, and core components, to our packaged components will require refactoring the existing components somewhat, which I think should be a follow-up. So for now, I've set up the scaffolding for a `plh` components package, but haven't included any components in there

chrismclarke (approved):
Thanks for all the updates @jfmcquade , it's looking really useful.

Yep happy to hold off any larger refactoring for now, although when you do try to handle the `plhAssetPipe` you should be able to just import the into the plh components module in the same way as the template module, i.e.
```
declarations: []
providers: [PLHAssetPipe],
```
So not having to re-declare. Although I haven't tested myself so could be an issue somewhere else.

src/app/shared/components/template/components/index.ts
```diff
@@ -197,3 +195,8 @@ export const TEMPLATE_COMPONENT_MAPPING: Record<
   video: TmplVideoComponent,
   workshops_accordion: WorkshopsComponent,
 };
+
+export const TEMPLATE_COMPONENT_MAPPING = {
+  ...CORE_COMPONENT_MAPPING,
+  ...DEMO_COMPONENT_MAPPING,
```
chrismclarke, line 201: @jfmcquade 
nit(non-blocking): realised after merge, but might be good to also include the plh components mapping here as well for any new components added to the placeholder module

chrismclarke:
@FaithDaka 
Just to loop you back in, this PR took a bit of a refactor but should hopefully now do a better job of illustrating how components can be added to support a specific deployment. See the updated code in `packages/components`

PR #2479, opened 2024-10-22
Kuwait Theme Styling - Accordion, Audio and Button components

FaithDaka:
PR Checklist

## Description

Includes the styling of the accordion, audio and button components under the Kuwait theme.

## Screenshots/Videos

`Button`

image

`Audio`

image

`Accordion`

image


chrismclarke (changes requested):
@FaithDaka 
I think it would be good if you're freed up as much as possible to work on style updates without having to worry to much about knock-ons to other deployments. 

I can see for this PR you've accurately scoped most of the changes to only be applied within the new theme, what I think might be even more useful is if we just create a standalone folder to store all overrides for the theme so that you don't have to worry about individual component scss files. In the future this could also allow us to more easily create custom builds that only includes the style files of themes referenced in the deployment config (smaller bundles, less scope for conflict etc.)

I've opened a refactor PR #2481 which aims to do this, let me know if you get a chance to look at it and whether it makes sense and works for you. Once merged it should make it more easy to distinguish between changes likely or not likely to have any impact elsewhere (and I'd personally be recommending we just keep fast-tracking merge for changes specific to your deployment)

src/app/shared/components/template/components/button/button.component.scss
```diff
@@ -179,9 +179,6 @@ ion-button[data-variant~="card"] {
 .center {
   text-align: center;
 }
-.centre {
-  text-align: center;
-}
```
chrismclarke, line 184: +1 thanks for catching the duplicate code

chrismclarke (approved):
Looks good to me. Now that all the overrides are applied directly from the theme overrides folder I think there's no further requirement to review specific code (more just a case of if it looks how you want it to), although I will still tag @jfmcquade in case there's any specific tips or patterns that he has any advice on from the override styles

FaithDaka:
Alright. Sounds good @chrismclarke

chrismclarke:
> Alright. Sounds good @chrismclarke
> 
> But there's an issue with css specificity for nested components i.e. the overrides file styling has a lower rank than the component's file styling. So i'll have to add those styles to individual files.
> 

I think this generally shouldn't be an issue as I would expect most style overrides to be additive instead of replacement, although I think the button component styles are just quite messy right now. So the option for this case would either be provide a more specific selector e.g. `span.left > p`, or just apply it using the `!important` flag

Personally I think I'd lean more towards using `!important` css more frequently from overrides if they aren't taking precedence (the base components shouldn't ever be using `!important` but overrides should feel free to do so). 

Longer term as we discover styles that we want to more easily expose for override we can be adding variables and removing hardcoded classes from the component (like the `.left` class), e.g. 

_button.component.scss_
```scss
:host{
  --text-align: left;
}

p,span { 
  text:align: var(--text-align) ;
 }
}
```

_overrides.scss_
```scss
plh-button{
  --text-align: center;
}
```

So if you also want to start proposing any changes that are like this I would also say feel free to do so (although these might be a little bit slower to approve depending on how complex they are, vs general style override)

jfmcquade (commented):
This looks great, I really like the visual changes. And I feel good about the way the styling is scoped to the theme. 

I haven't taken a detailed look through the new CSS, but I agree that as long as you've got things looking how you want and the styles won't have knock-ons for other deployments then I'm happy.

It could be worth tagging @esmeetewinkel to review this PR and any like it in future, as she normally carries out functional testing to make sure the components work as expected – and good for us all to be in the loop regarding any changes.

esmeetewinkel:
Thanks @FaithDaka. I added a [PR on the content repo](https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/pull/67) to point the deployment to the new theme

PR #2481, opened 2024-10-23
refactor: plh_kw theme folder

chrismclarke:
PR Checklist

## Description

As part of review of #2479, proposed refactor to move all deployment-specific style overrides to own folder.

This PR targets the #2479 branch.

## Review Notes
To confirm if it still applies the style overrides as expected

## Git Issues

Closes #

## Screenshots/Videos


FaithDaka:
@chrismclarke I like the approach of having one file for all overrides, thank you! 
I wonder if that affects maintenance the larger the file gets, though given that I will be creating a number of custom components for the Kuwait deployment there shouldn't be as much overriding for this theme.

chrismclarke:
> @chrismclarke I like the approach of having one file for all overrides, thank you! I wonder if that affects maintenance the larger the file gets, though given that I will be creating a number of custom components for the Kuwait deployment there shouldn't be as much overriding for this theme.

At that point it's totally up to you however is easiest to organise. E.g. you could have an `overrides` folder with individual per-component scss files, e.g. 

`src/theme/themes/plh_kids_kw/overrides/accordion.scss`, 
`src/theme/themes/plh_kids_kw/overrides/button.scss` 

the name of the file isn't super important as all style changes are scoped to the component selector, so more just for your own organisation. The obvious downside to this approach is that the overrides would be broken if the core component changes name (I expect some will given inconsistency between `plh-some-component`, `tmpl-some-component`, `template-some-component` etc.), however we will do our best internally to remember to do a full codebase search of the selector when changing and can update all references (equally important for cases where one component uses another)

PR #2482, opened 2024-10-23
chore: update husky hook

chrismclarke:
PR Checklist

## Description

I've recently started using [FNM](https://github.com/Schniz/fnm) to manage node version. There seems to be a slight issue when trying to create commits from vscode, where it doesn't detect the environment variables correctly for node/yarn, which are used as part of our git pre-commit hook (calls the `lint-staged` package)

This PR adds explicit support in the hook to re-evaluate the environment for fnm if using.

## Review Notes
The only thing to check is that as a user not using fnm that you can still add commit and trigger the pre-commit hook without issue. So recommend taking this branch and just attempting to add a commit to it. Note that the commits only trigger the `lint-staged` command when applying the actual commit (not just stage), so would need to commit via vscode interactive prompt or via `git commit -m "test commit message"`

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,9 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
+#!/usr/bin/env bash
+
+# HACK - if runnning via fnm on windows then bash subshell may not have access to node/yarn commands
+# fix by re-evaluating fnm env
+if which fnm >/dev/null; then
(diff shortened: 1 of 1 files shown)


jfmcquade (approved):
Looks good. I'm not using `fnm` (I use `nvm` on mac), and can confirm that `lint-staged` runs as expected when adding a commit to this branch

Screenshot 2024-10-23 at 13 05 37

PR #2483, opened 2024-10-24
fix: app-data evaluator

chrismclarke:
PR Checklist

## Description

- Fix regression from #2461 where expressions like `{@row.id}_suffix` would be evaluated as JS (-> `undefined`) instead of templated data (-> `some_id_suffix`) for `data_pipe` append operator
- Fix issue where evaluation would break when using multiple different variable types, e.g. `@row.number > @local.number`
- Update tests to capture new cases

## Review Notes
Should confirm if fixes issue originally outlined in [mattermost thread](https://chat.idems.international/all/pl/krgfwnfdi7yfi8xrd7o53bwhoh)

Should be able to confirm tests passed from github action. 

As for possible knock-ons, it's only the `appendColumns` operator that uses the code, so changes should show up in parser for any given deployment repo that has example content.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/shared/src/models/appDataEvaluator/appDataEvaluator.ts
+++ b/packages/shared/src/models/appDataEvaluator/appDataEvaluator.ts
@@ -1,7 +1,7 @@
 import { JSEvaluator } from "../jsEvaluator/jsEvaluator";
 import { TemplatedData } from "../templatedData/templatedData";
 import { isObjectLiteral } from "../../utils/object-utils";
-import { addJSDelimeters } from "../../utils/delimiters";
+import { addJSDelimiters } from "../../utils/delimiters";
 
 /** Variable context is stored in namespaces, e.g. `{item:{key:'value'},field:{key:'value'}}` */
 type IContext = { [nameSpace: string]: { [field: string]: any } };
@@ -37,7 +37,12 @@ export class AppDataEvaluator {
     this.templatedData.updateContext(context);
   }
   public updateExecutionContext(update: IContext) {
-    this.setExecutionContext({ ...this.context, update });
+    this.setExecutionContext({ ...this.context, ...update });
+  }
+
+  /** Wrapper around templatedData method to list all variables within data **/
+  public listContextVariables(data: any, prefixes: string[]) {
+    return this.templatedData.listContextVariables(data, prefixes);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


packages/shared/src/models/appDataEvaluator/appDataEvaluator.ts
```diff
@@ -97,13 +102,30 @@ export class AppDataEvaluator {
...
+      } catch (error) {
+        // Ignore errors as may be intended as templated data (will attempt next)
+      }
+    }
+    // If error or if not evaluated in JS continue as templated data replacement
+    const parsed = this.templatedData.parse(expression);
+    return parsed;
+  }
+
+  private shouldEvaluateJS(expression: string) {
```
jfmcquade, line 119: Would be nice to have a similar check for all template strings eventually, removing the need to use excessive local vars to concatenate dynamic strings (I know sometimes we'll author a local var just to store a `_` character, so we can provide a string that will be correctly interpreted by the parser, but it'd be nice to be able to write `{@local.prefix}_{@local.suffix}` inline).

Hard to see how to avoid a hack such as this though, when considering the variety of authoring patterns we support, the intention of which is always implicit.

jfmcquade (commented):
Looks good

packages/shared/src/models/appDataEvaluator/appDataEvaluator.ts
chrismclarke: I thought we currently could concatenate no? I thought if explicitly including the curly braces it would work.
On the plus side, now that we have a JS-first approach I think you could also do more things like `"@local.string"+"_suffix"` (haven't tested, but I'd expect either that to work or `@local.string+"_suffix"`... doubt interpolation like `${@local.string}_suffix` would but could also be worth investigating.

But yeah ultimately at some point we will have to either commit to always supporting somewhat hacky/mixed interpretation patterns or be a bit more insistent on how to indicate something is templated vs computed

chrismclarke: oh wait, you mean with the current templating system. Yeah agreed, I'm hoping one day to totally overhaul that and use more of the methods that we've been slowly introducing to be consistently applied across the entire app. But that is quite a task (been doing a lot of stuff linked to it with data-items), likely needing some means to have the new templating system running in parallel with old (via some sort of opt-in deployment config)

PR #2484, opened 2024-10-24
feat: add template base signals and equality check

chrismclarke:
PR Checklist

## Description
Add signals to template base component for more fine-grained updates to row value, parameter_list and action_list updates

This will allow components to refer directly to `value()` property (instead of `_row.value`) and also apply computed effects 

## Review Notes
It isn't expected for this to have much impact as the properties are newly added. The only issues are with any components that provide their own definition for `value`, `parameterList` or `actionList` (instead of accessing via `_row.value`)

Looking at the tests there were a few components that did have conflicts and have been updated accordingly (would be good to confirm timer working as expected as that was slightly less trivial to update).

## Dev Notes
This is a bit of a forward-looking update to provide devs the option of directly linking to row value or parameter list updates using signals. These should be more performant and reliable than responding to `_row` changes more generally.

The changes also include a custom equality checker, as the default for angular signals is the `Object.is` function which only matches equality for strict object equivalence e.g. same array but not two arrays with same entries so that 
```ts
Object.is([],[])  // false
```
https://angular.dev/guide/signals#signal-equality-functions

I have only included the parameterList, actionList and value properties for now as I think these are the main columns likely to be updated dynamically (e.g. containing dynamic refs). Note that the `condition` column will always be interpreted outside of the template component itself (and not render the component if false), and `hidden` column dealt with at the container level.

I had wanted to also include a more general `row()` signal (if wanting to trigger changes generally when full row updates), however this would conflict with the existing row setter functions.

In the future it might be good to start refactoring some components to use the signals (particularly ones that use `@Input() set row` bindings), however this should be done on a case-by-case basis to check for any side-effects.

## Git Issues

Closes #

## Screenshots/Videos


packages/shared/src/utils/object-utils.ts
```diff
@@ -77,3 +77,26 @@ export const sortJsonKeys = <T extends Record<string, any>>(json: T): T => {
...
+/** Minimal deep equality checker, loosely based on lodash _isEqual but for simple primitives only */
+export function isEqual(a: any, b: any) {
+  // handle simple string, boolean, number, undefined, null or same object reference
+  if (a === b) return true;
+  // handle different object types
+  if (typeof a !== typeof b) return false;
+  // handle deep comparison for arrays
+  if (Array.isArray(a)) {
+    const differentEl = a.find((v, i) => !isEqual(v, b[i]));
+    return differentEl ? false : true;
```
jfmcquade, line 90: nit(blocking): This will throw a false positive if a `differentEl` is found that is falsey (e.g. if `a` contains a `0` element at an index that `b` has a different element, e.g. `a = [0]` and `b = [1]`).

How about this instead:
```ts
if (Array.isArray(a)) {
  if (a.length !== b.length) return false;
  return a.every((v, i) => isEqual(v, b[i]));
}
```

jfmcquade (changes requested):
Nice to have for sure. The timer component code is a bit confusing to me but I'm not keen to refactor it now – everything seems to be working as expected from a functional perspective. I also added a minor code tidy in https://github.com/IDEMSInternational/open-app-builder/pull/2484/commits/a63fdb186adbd88e5499988723b496cb41224b71 to clean up a line that was throwing a typescript warning/error.

See my inline suggestion for the equality checker.

packages/shared/src/utils/object-utils.ts
chrismclarke: Really nice catch, you're right `isEqual([true],[false])` would a `true` value. 

I tend to not like `every` operators as there's really no reason to continue processing once you've found an element that differs (can add on a bit of time if dealing with very large arrays). Instead I've added 3a16bc9e3794f6f2baa62a7030feeb3eff0fbe04 which uses the `findIndex` operator to instead focus on the index of the element that differs and ignoring the specific value.

packages/shared/src/utils/object-utils.ts
```diff
@@ -77,3 +77,28 @@ export const sortJsonKeys = <T extends Record<string, any>>(json: T): T => {
...
+  // handle simple string, boolean, number, undefined, null or same object reference
+  if (a === b) return true;
+  // handle different object types
+  if (typeof a !== typeof b) return false;
+  // handle deep comparison for arrays
+  if (Array.isArray(a)) {
+    // find the first element index where there is a mismatch
+    // if all elements are the same `findIndex` returns value -1
+    const differentIndex = a.findIndex((v, i) => !isEqual(v, b[i]));
+    return differentIndex === -1;
```
jfmcquade, line 92: Nice, this seems like a good way to do it. However, we still need a check that the arrays are of equal length first, otherwise it will return a false positive if `b` contains the same elements as `a` plus some additional elements. E.g. `isEqual([1],[1,2])` will return `true` currently

jfmcquade (changes requested):

packages/shared/src/utils/object-utils.ts
chrismclarke: nice point, added length check and updated test with 261a7d96aa565f9b0e36c9db0352a5d1f3ae2d00

chrismclarke: Also realised that will need to explicitly check `b` is an array as relying on previous `typeof` check could match `object` type for both `[]` or `{}` (or even dates or `null` objects)

Updated with b43d534a7e1efc25c92c5508deb7645eea4ce1d4

jfmcquade (approved):
Looking good, just a merge conflict to resolve now...

PR #2485, opened 2024-10-25
chore: move plh-specific components to plh components folder

jfmcquade:
PR Checklist

## Description

Follow-up to #2478.
- Moves two components into the `packages/components/plh` folder created in that PR, updating the relevant module declaration and component mapping
- Refactors the template pipes into a module, which can then be imported into multiple other modules
- Renames and tidies the component files related to `parent-point-box`

The code for these plh components is messy and includes hardcoded paths to assets. I don't think refactoring them is a priority right now, and I'm not sure if we plan to continue actively using them.

@FaithDaka – I think it's worth saying for your benefit that the two components that are now in `packages/components/plh` as of this PR are not good examples of components in themselves! They just happen to be two components that are PLH-specific so I've moved them over. You shouldn't look at the content of those files for an example of how to write component files (instead refer to `packages/components/demo/basic`). However, how they are configured and declared in `packages/components/plh/index.ts` is relevant for adding any new PLH components.

## Git Issues

Closes #

## Screenshots/Videos

Updated [comp_parent_point_counter](https://docs.google.com/spreadsheets/d/13OCh9zMQsI0NuUjy0L-4bJwDF3Do84mhbuQIP7Or61Y/edit?gid=569531329#gid=569531329):

Screenshot 2024-10-25 at 17 18 13

Screenshot 2024-10-25 at 17 11 08

[comp_parent_point_box](https://docs.google.com/spreadsheets/d/10xgdH_zlA7HxyfF9Uq99AnbSC5wQNHoYS9vhvGXwaJ8/edit?gid=569531329#gid=569531329). This component sheet doesn't make sense to me, and some assets are missing, but for the purposes of this PR, what is relevant is that this template displays the same on this PR branch as it does on the debug [web preview](https://idems-debug.web.app/template/comp_parent_point_box), indicating that the component is being made available to the template authoring as expected

Screenshot 2024-10-25 at 17 17 47


chrismclarke (approved):
Thanks for tidying this up and fixing the component demos a bit

I've added one minor tidying commit a841bcc4127358629da9a9417d9a2f06c21bfa19 (given all the pipes were declared in the pipe index felt it made sense to also just export the module from there instead of a separate file), but otherwise I'd say good to go.

I confirmed both the plh component were working and a few others that use pipes, so I'd say all good.

PR #2488, opened 2024-10-28
chore: move plh-specific components to plh components folder ALT

jfmcquade:
PR Checklist

## Description

- Contains the changes originally merged with #2485. See that PR for original notes.
- Adds an additional commit that fixes a regression issue, https://github.com/IDEMSInternational/open-app-builder/pull/2488/commits/ecd948b64bd6a987e1213f97835394413510cb54.

NB #2485 introduced a regression issue and we wanted a quick code release, so I reverted the changes with #2486 (that's why the commits on the current PR are a little wonky). This PR reinstates them, plus an additional fix:
Our spot checks on #2485 didn't cover the case of a component that imports the pipe directly into the `.ts` file and calls the `transform()` method there. There were two such cases of this: the `audio` and `timer` components. In these cases, an error was thrown when the component tried to initialise (screenshot of error included below). A possible workaround is to supply the `PLHAssetPipe` to the `providers` arrays of the components in question, however this pattern of importing a pipe and calling it's method directly doesn't seem ideal, so instead I've refactored to use the method from the associated service instead.

## Git Issues

Closes #

## Screenshots/Videos

Screenshot 2024-10-28 at 15 12 48


chrismclarke (approved):
Oh interesting, yeah that's a bit of a weird use-case. Agreed that a workaround could be to include in the module providers, but a bit of a weird pattern to be calling pipe transform from ts code, so prefer your alternative here.

src/app/shared/components/template/pipes/plh-asset.pipe.ts
```diff
@@ -3,19 +3,14 @@ import { TemplateAssetService } from "../services/template-asset.service";
...
 @Pipe({ name: "plhAsset" })
 export class PLHAssetPipe implements PipeTransform {
   constructor(private templateAssetService: TemplateAssetService) {}
 
   transform(value: string) {
-    if (!value) return "";
-    // keep external links
-    if (value.startsWith("http")) {
-      return value;
-    }
```
chrismclarke, line 18: +1 good idea moving this to the service logic

PR #2490, opened 2024-10-28
feat: youtube component

### LINKED ISSUE #2442: [FEATURE] Streaming of videos
**What?**
Streaming of large assets e.g. videos. 

**Why?**
PLH Kids TZ deployment has many large assets (videos, audios, jpgs). Concern is that downloading to device is not feasible because people won't have space to store these files.

jfmcquade:
PR Checklist

## Description

Adds a new component, `youtube`, for displaying a specified Youtube video in an embedded Youtube player.

Component minimally requires a Youtube URL passed as the `value`, e.g. `https://www.youtube.com/watch?v=0oOAGrStVGI`.
Parameters:
- `allow_fullscreen` (default `true`).

## Git Issues

Closes #2442

## Screenshots/Videos

[comp_youtube](https://docs.google.com/spreadsheets/d/1bQ2I077nSRFpe5aLszbPzhwNysZrAbS7EeEPlwCQ7AA/edit?gid=603128459#gid=603128459) template:

Screenshot 2024-10-29 at 13 05 00

Screenshot 2024-10-29 at 13 05 31


src/app/shared/components/template/components/youtube/youtube.component.ts
```diff
@@ -0,0 +1,75 @@
...
+import { DomSanitizer, SafeResourceUrl } from "@angular/platform-browser";
+import {
+  getBooleanParamFromTemplateRow,
+  getStringParamFromTemplateRow,
+} from "src/app/shared/utils";
+import { TemplateTranslateService } from "../../services/template-translate.service";
+
+interface IYoutubeParams {
+  /** TEMPLATE PARAMETER: video_id */
+  videoId: string;
```
chrismclarke, line 12: nit(blocking): I don't think we need to have both `value` and `parameter_list` used to define the video id, I'd say better to just keep as `value`

chrismclarke (changes requested):

src/app/shared/components/template/components/youtube/youtube.component.ts
```diff
@@ -0,0 +1,75 @@
...
+    const baseYoutubeUrl = `https://www.youtube.com/embed/${this.params.videoId}`;
+
+    // See https://developers.google.com/youtube/player_parameters
+    const youtubeQueryParams = {
+      // Favour white over red for more theme compatibility
+      color: "white",
+      // hide the fullscreen button if allow_fullscreen is false
+      fs: this.params.allowFullScreen ? "1" : "0",
+      // Attempt to set the player's interface language to match the app language
+      hl: this.templateTranslateService.app_language$.value,
```
chrismclarke, line 58: nit(non-blocking): nice idea to try and support the language, although if I think our `language` is more of a localisation code (country+lang, e.g. `tz_sw`), so I'm guessing for this to work would likely need to split the string and take the end characters.

chrismclarke (changes requested):
nice implementation and appreciate the example component. 

My one main (blocking) thought is what would make most sense for authors to pass for video ID. 

I'm guessing makes most sense to just use the value field (and save parameter_list for additional parameters), I'm also wondering whether it might be more intuitive to authors to just paste the full youtube url instead of manually extracting the video id. 

Having a full url would also make it a little bit easier to debug if a, for example, a video wasn't working as expected.

src/app/shared/components/template/components/youtube/youtube.component.html
```diff
@@ -0,0 +1,9 @@
+<div class="youtube-container">
+  <!-- The `iframe` property `allowFullscreen` must be set as a static attribute,
+   see https://angular.dev/errors/NG0910 -->
+  @if (params.allowFullScreen) {
+    <iframe [src]="src" frameborder="0" allowfullscreen></iframe>
+  } @else {
+    <iframe [src]="src" frameborder="0"></iframe>
```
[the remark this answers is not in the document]
jfmcquade: True, I don't see much use for it, especially as we can't disable the rest of the youtube interface for the embedded video, including the external link to youtube as you mention. But I've left in for now as I don't see harm in including it as an option

src/app/shared/components/template/components/youtube/youtube.component.ts
jfmcquade: You're right, from [the docs](https://developers.google.com/youtube/player_parameters) it doesn't look like they support our (non-standard?) format, I've updated to try and extract the country code

chrismclarke:
@jfmcquade - thanks for making the updates. See some proposed changes here: https://github.com/IDEMSInternational/open-app-builder/pull/2493

The main suggestion is to make more explict the parameter list and value parsing steps, so that any additional search params included in the provided value url can also be included (e.g. start time t=1). It also includes more notes about potential url limitations (e.g. different youtube formats), but does not explicitly add any additional functionality at this time.

chrismclarke (approved):
@jfmcquade 
I've added one final tidying commit with 5ea35ef3e411b55ddce47c0db61a56b98d72bb22 to remove the searchParam utils no longer in use and to move the language_code lookup to the translate service itself (given it always depends on it anyways). If happy I'd say good to go

jfmcquade:
Thanks, @chrismclarke, the latest tidying looks good. I'm going to merge now.

@esmeetewinkel, tagging you just to make you aware of the final decisions regarding authoring (reflected in the updated PR description), namely:
- The `youtube` component expects to be passed a full youtube URL, e.g. `https://www.youtube.com/watch?v=0oOAGrStVGI`. In the first iteration, an ID was required, e.g. `0oOAGrStVGI`, but this is no longer supported
- This URL must be passed as the `value` of the component row. In the first iteration, it could also be passed via a `video_id` parameter, but this has been removed.

PR #2491, opened 2024-10-29
chore: move arrayToHashmap function

chrismclarke:
PR Checklist

## Description

Move the `arrayToHashmap` shared util from `file-utils` to `object-utils`. This is required as the `file-utils` contain node-specific logic, and so can't import from the file in web (issues with webpack polyfills for `os`). 
The `object-utils` doesn't contain any node-specific code and makes more sense as the location for this method. 

Also adds basic tests.

## Review notes
I've checked for any files that import the method from `file-utils` however all import from the top-level `shared` so shouldn't have any issues with the moved method.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/shared/src/utils/file-utils.ts
+++ b/packages/shared/src/utils/file-utils.ts
@@ -247,30 +247,6 @@ export function groupJsonByMultipleKeys<T>(json: T[], keys: string[], joinCharac
   return byKey;
 }
 
-/**
- * Convert an object array into a json object, with keys corresponding to array entries
- * @param keyfield any unique field which all array objects contain to use as hash keys (e.g. 'id')
- * @param handleDuplicateKey optional function to trigger when duplicate hash key entry detected.
- * Should return replacement key to populate instead
- */
-export function arrayToHashmap<T extends object>(
-  arr: T[],
-  keyfield: keyof T,
-  handleDuplicateKey = (k: string) => k
-): { [key: string]: T } {
-  const hashmap: { [key: string]: T } = {};
-  for (const el of arr) {
-    if (el.hasOwnProperty(keyfield)) {
-      let hashKey = el[keyfield] as string;
-      if (hashKey in hashmap) {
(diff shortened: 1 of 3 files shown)


jfmcquade (approved):
Seems sensible.

I came across the same issue when working on #2490 – I originally wanted to include  an `addQueryParamsToUrl()` function in a utils file, I think `string-utils`, but got the same error and wasn't able to track down the issue, so just put it in the component file in the end. In general do you think it would make sense to group the utils in `packages` in a way that made this distinction clearer?

On reflection it might make sense for me to put the query params function in `src/app/shared/utils/utils.ts`, but we also have other `utils` folders within the app code, like `src/app/shared/components/template/utils` for template utils.

chrismclarke:
> Seems sensible.
> 
> I came across the same issue when working on #2490 – I originally wanted to include an `addQueryParamsToUrl()` function in a utils file, I think `string-utils`, but got the same error and wasn't able to track down the issue, so just put it in the component file in the end. In general do you think it would make sense to group the utils in `packages` in a way that made this distinction clearer?
> 
> On reflection it might make sense for me to put the query params function in `src/app/shared/utils/utils.ts`, but we also have other `utils` folders within the app code, like `src/app/shared/components/template/utils` for template utils.

It's quite hard to clearly distinguish what scripts are made for node vs browser environment (particularly given some use 3rd party packages like danfoJS which claim to support both but have issues in browser). 

I think probably the most sensible thing to do is just not re-export modules from a single `index.ts` file, so that browser code is less likely to try import from sources that contain node code. This would also make it easier for tree-shaking, with the downside of being a bit more fragile to changes when methods moved from one file to another.
Given that this would be a mid-sized refactor probably best for now to just keep moving things as issues come up.

The other workaround would probably be to just include polyfills for methods like `os`, `fs`, `crypto` etc. in webpack, however I think that could lead to trying to use methods that aren't really supported by accident.

PR #2492, opened 2024-10-29
chore(android): upgrade gradle

jfmcquade:
PR Checklist

## Description

Upgrades the Android Gradle Plugin from version 8.3.1 to 8.7.1 in the android build files.

I was prompted to run this upgrade via Android Studio's AGP Upgrade Assistant when opening the project in Android Studio.

From the [release notes](https://developer.android.com/build/releases/gradle-plugin), I believe devs will need to upgrade their Android Studio to the latest version ("Ladybird").

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.3.1'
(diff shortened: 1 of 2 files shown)


chrismclarke (approved):
Tried to run the android test build but didn't seem to trigger (not sure if gh actions issue generally or if we've broken the action for debug). In any case I expect minor gradle bump shouldn't cause many issues (and any it did would become obvious pretty quickly)

PR #2493, opened 2024-10-29
review: 2490

chrismclarke:
PR Checklist

## Description
Proposed changes for #2490. Main changes:

- Use signals instead of lifecycle methods
- Separate out value and parameter list parsing, include value validation function
- Handle videoId extract as part of general value parse to allow additional search params to be retained (e.g. video timestamp start)

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/cspell.config.yml
+++ b/cspell.config.yml
@@ -37,6 +37,7 @@
     "tmpl",
     "venv",
     "viewbox",
+    "youtube",
   ]
 # flagWords - list of words to be always considered incorrect
 # This is useful for offensive words and common spelling errors.
--- a/src/app/shared/components/template/components/youtube/youtube.component.html
+++ b/src/app/shared/components/template/components/youtube/youtube.component.html
@@ -1,11 +1,11 @@
-@if (src) {
+@if (src()) {
   <div class="youtube-container">
     <!-- The `iframe` property `allowFullscreen` must be set as a static attribute,
      see https://angular.dev/errors/NG0910 -->
-    @if (params.allowFullScreen) {
-      <iframe [src]="src" frameborder="0" allowfullscreen></iframe>
+    @if (params().allowFullScreen) {
+      <iframe [src]="src()" frameborder="0" allowfullscreen></iframe>
     } @else {
-      <iframe [src]="src" frameborder="0"></iframe>
+      <iframe [src]="src()" frameborder="0"></iframe>
     }
   </div>
 }
--- a/src/app/shared/components/template/components/youtube/youtube.component.ts
+++ b/src/app/shared/components/template/components/youtube/youtube.component.ts
@@ -1,13 +1,12 @@
-import { Component, OnInit } from "@angular/core";
+import { Component, computed } from "@angular/core";
 import { TemplateBaseComponent } from "../base";
-import { DomSanitizer, SafeResourceUrl } from "@angular/platform-browser";
-import {
-  getBooleanParamFromTemplateRow,
-  addQueryParamsToUrl,
-  getQueryParam,
-  extractTwoLetterLanguageCode,
-} from "src/app/shared/utils";
+import { DomSanitizer } from "@angular/platform-browser";
+import { extractTwoLetterLanguageCode, getBooleanParamFromTemplateRow } from "src/app/shared/utils";
 import { TemplateTranslateService } from "../../services/template-translate.service";
+
+interface ITemplateParams {
+  allow_full_screen?: string;
+}
 
 interface IYoutubeParams {
   /** TEMPLATE PARAMETER: allow_full_screen. Default true */
@@ -22,11 +21,7 @@ const YOUTUBE_VIDEO_ID_QUERY_PARAM = "v";
   templateUrl: "./youtube.component.html",
   styleUrls: ["./youtube.component.scss"],
 })
-export class YoutubeComponent extends TemplateBaseComponent implements OnInit {
-  params: Partial<IYoutubeParams> = {};
-  src: SafeResourceUrl;
-  videoId: string;
-
+export class YoutubeComponent extends TemplateBaseComponent {
   constructor(
     private domSanitizer: DomSanitizer,
     private templateTranslateService: TemplateTranslateService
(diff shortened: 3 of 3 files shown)


jfmcquade (approved):
Thanks, looks good I think. I've made some comments below. I think we want to get this component in ASAP, so if we're both happy then I don't think we should worry too much about further changes, but I do think it's worth thinking about the params setting pattern more generally if it's going to be something we're rolling out across all components.

## Signals pattern
I can see how the move to signals over lifecycle methods makes sense. I found it a bit trickier to follow at first, but I think that's just due to familiarity with the old way of setting component params. I suppose that this represents how we should structure components going forward, using a computed `params` signal and a `parseParams` method instead of calling a `getParams()` method on init. In that case, I think we could probably iterate a bit to make things clearer, e.g.:
- At some point, we may want to generalise the `parseParams` to a reusable function that includes type checking on two component-specific types: the raw template param values and the parsed component param values (e.g. `ITemplateParams` and `IYoutubeParams` as they're currently defined in this file). That could be quite neat to not have to manually write out `getBooleanParamFromTemplateRow()` etc within the component code, and just have these inferred from the types that the component defines. This could even be defined in the `TemplateBaseComponent` so that for a given component we can just pass in our types(/params objects) in the constructor and then refer to `params().myParam` without needing to call any parsing on params explicitly in the component file.
- The difference between a property of the component class that sits inside `params` and one that sits outside has always been slightly vague. It makes sense that generally we should use `params.allowFullScreen` in the case that `allowFullScreen` maps one-to-one to an authorable template parameter, but I know there are some cases where we apply some minor processing to a value before storing it in the `params` property, in which case it should possibly sit outside the `params` signal as a top-level property
- See inline comment regarding `parseParams()` argument
I think this could be a process of iteration as we introduce/refactor other components that use this pattern, so I'm happy with things as they are for now (possibly pending my inline comment suggestion)

## Retaining search params
Originally I was thinking that we wanted to remove any query params from the URL, on the basis that I thought they would likely be included by mistake by an author (who may not understand that the URL contains additional parameters). For example, if I copy a URL from a video in my "watch later" playlist, then it includes the additional params `&list=WL&index=3`, which we wouldn't want to include (although I don't think it would necessarily have an effect in this case). Then we could add template parameters for those query params that we do want to expose, e.g. `start_time`, requiring that the author explicitly sets them as parameters rather than passing them in via the input URL.

However, I think it probably does make sense to retain the query params passed by the author. We could always explicitly clear any that we wish to ignore (e.g. `list`) while allowing the useful ones to pass through (e.g. `t`).

## Youtube query params type safety
I made some changes to enable type safety when setting the youtube query params, see https://github.com/IDEMSInternational/open-app-builder/pull/2493/commits/df9a9d15af6dbe191ec8bf809142e9f3738d0b5b

chrismclarke:
Thanks @jfmcquade, I like the changes you've made to make query params more explicit. Agreed this should be good to go and can always iterate once merged. Few quick responses 


> ## Signals pattern
>  I suppose that this represents how we should structure components going forward, using a computed `params` signal and a `parseParams` method instead of calling a `getParams()` method on init. In that case, I think we could probably iterate a bit to make things clearer, e.g.:
> * At some point, we may want to generalise the `parseParams` to a reusable function that includes type checking on two component-specific types: the raw template param values and the parsed component param values (e.g. `ITemplateParams` and `IYoutubeParams` as they're currently defined in this file). That could be quite neat to not have to manually write out `getBooleanParamFromTemplateRow()` etc within the component code, and just have these inferred from the types that the component defines. This could even be defined in the `TemplateBaseComponent` so that for a given component we can just pass in our types(/params objects) in the constructor and then refer to `params().myParam` without needing to call any parsing on params explicitly in the component file.

Strongly agreed, a bit like the old react `mapStateToProps` handling it would be good to have a uniform way of parsing value and parameter lists. It would also make sense to have a default (overridable) method in the base component assuming each child component includes some sort of mapping/validator, e.g. a mapping like `{allow_fullscreen: {default:true, type:'boolean', name: 'allowFullScreen'}` could infer and output the correct parameters. Although type-checking the output would be a bit of a pain as the input names would get mapped (so maybe better to just stick to authored naming convention). Or maybe defining the other way round would be better, e.g. `paramMapping = {allowFullScreen: {template_parameter:'allow_fullscreen', default:true, type:'boolean')` - this would at least get the keys correct although getting it to infer the output boolean would again be tricky.

> * The difference between a property of the component class that sits inside `params` and one that sits outside has always been slightly vague. It makes sense that generally we should use `params.allowFullScreen` in the case that `allowFullScreen` maps one-to-one to an authorable template parameter, but I know there are some cases where we apply some minor processing to a value before storing it in the `params` property, in which case it should possibly sit outside the `params` signal as a top-level property
> * See inline comment regarding `parseParams()` argument
>   I think this could be a process of iteration as we introduce/refactor other components that use this pattern, so I'm happy with things as they are for now (possibly pending my inline comment suggestion)

Agreed, I think the main thing to keep in mind moving forwards is that now that there are new methods available don't feel too restricted to keep using old patterns (which may no longer fit quite so well). Hopefully something sensible will just start to emerge and then we can go back and refactor later (or as we move more components to external packages)

> ## Retaining search params
> Originally I was thinking that we wanted to remove any query params from the URL, on the basis that I thought they would likely be included by mistake by an author (who may not understand that the URL contains additional parameters). For example, if I copy a URL from a video in my "watch later" playlist, then it includes the additional params `&list=WL&index=3`, which we wouldn't want to include (although I don't think it would necessarily have an effect in this case). Then we could add template parameters for those query params that we do want to expose, e.g. `start_time`, requiring that the author explicitly sets them as parameters rather than passing them in via the input URL.
> 
> However, I think it probably does make sense to retain the query params passed by the author. We could always explicitly clear any that we wish to ignore (e.g. `list`) while allowing the useful ones to pass through (e.g. `t`).

I think it is pretty much only the `t` searchParam that's useful and could be manually provided by author param, although I still think probably no harm in keeping all params just so that authors have full control. E.g. there might also be various `rel` style params that authors may use if youtube allows referral source tracking, for example.

PR #2494, opened 2024-10-30
Chore: items code tidying

chrismclarke:
PR Checklist

## Description
Small amount of tidying for `items` and `data_items` (to support ongoing refactor work)

- Improve some variable naming
- Improve some types
- Move `DYNAMIC_PREFIX` definitions
- Update template item handling to pass arrays instead of hashmap
- Move item data pipe to own file

## Review Notes
I've checked `comp_data_items` and it looks to be working same as in `master` - example 3 not working in either but others appear to be (but good to double-check)

I've also checked `example_items` and `feature_items` and both appear to be working as intended. There's quite a lot of other debug sheets with the word items in, I haven't checked any of these as I'm not aware if any are still relevant or not.

There's a bit of a mix of changes here, if you'd like me to separate type-only from more functional let me know and I can refactor

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/data-models/flowTypes.ts
+++ b/packages/data-models/flowTypes.ts
@@ -362,7 +362,7 @@ export namespace FlowTypes {
     [key: string]: any;
   };
 
-  const DYNAMIC_PREFIXES_COMPILER = ["gen", "row"] as const;
+  const DYNAMIC_PREFIXES_COMPILER = ["gen", "row", "default"] as const;
 
   const DYNAMIC_PREFIXES_RUNTIME = [
     "local",
--- a/packages/data-models/flowTypes.ts
+++ b/packages/data-models/flowTypes.ts
@@ -339,7 +339,7 @@ export namespace FlowTypes {
     /** Keep a list of dynamic dependencies used within a template, by reference (e.g. {@local.var1 : ["text_1"]}) */
     _dynamicDependencies?: { [reference: string]: string[] };
     _translatedFields?: { [field: string]: any };
-    _evalContext?: { itemContext: TemplateRowItemEvalContext }; // force specific context variables when calculating eval statements (such as loop items)
+    _evalContext?: any; // force specific context variables when calculating eval statements (such as loop items)
     __EMPTY?: any; // empty cells (can be removed after pr 679 merged)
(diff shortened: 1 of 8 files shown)


packages/data-models/flowTypes.ts
```diff
@@ -362,7 +362,26 @@ export namespace FlowTypes {
     [key: string]: any;
   };
 
-  type IDynamicPrefix = IAppConfig["DYNAMIC_PREFIXES"][number];
+  const DYNAMIC_PREFIXES_COMPILER = ["gen", "row"] as const;
```
jfmcquade, line 365: Should "default" be here too?

src/app/shared/components/template/processors/item.ts
```diff
@@ -1,27 +1,28 @@
...
+
+interface ITemplateRowWithItemContext extends FlowTypes.TemplateRow {
+  _evalContext: { itemContext: IItemEvalContext }; // force specific context variables when calculating eval statements (such as loop items)
+}
 
 export class ItemProcessor {
   constructor(
-    private dataList: any,
-    private parameterList?: any
+    private dataList: FlowTypes.Data_listRow[] = [],
```
jfmcquade, line 13: Could this property be renamed to indicate more clearly that it represents an array of rows?

jfmcquade (approved):
Looks good, definitely tidier. I've added a couple of minor queries inline

packages/data-models/flowTypes.ts
chrismclarke: You're right, updated with [569234b](https://github.com/IDEMSInternational/open-app-builder/pull/2494/commits/569234b99d0a19553250a2f32aadde77d7328ef2)

src/app/shared/components/template/processors/item.ts
chrismclarke: Good idea, I can't tell you how unnecessarily confusing it's been having ambiguous names without type definitions or tests in the code. If you have at least 1 you can usually infer a certain amount, but definitely good to aim for all 3 where possible. Updated with 569234b

PR #2498, opened 2024-10-31
fix: text bubble component handles markdown

### LINKED ISSUE #2435: [BUG] Text bubble doesn't recognise markdown
**Describe the bug**
When using Markdown **bold** or *italics* text in a text bubble, this is not interpreted and shows instead as `**bold**` or `*italics*`. Same holds for numbered lists

**App version**
v0.16.34

**Deployment Sheet**
Some articles in PLH Kids TZ, such as [the_power_of_play_2](https://plh-kids-tz.web.app/template/article_play_the_power_of_play_2) and [using_books_5](https://plh-kids-tz.web.app/template/article_play_using_books_5)

**Debug Sheet**
[comp_text_bubble](https://docs.google.com/spreadsheets/d/1Yo8X4TrTW2lzEgusDplAQ-y-JJvQOIqXGNLVON1uvRo/edit?gid=569531329#gid=569531329)
image

**Screenshots**
image

image

jfmcquade:
PR Checklist

## Description

Text passed to as the `value` to the `text-bubble` component is now parsed as markdown, in line with other components like `text`.

## Dev notes

It makes sense to me that when we're rendering text as a standard paragraph, we should be calling an instance of the `text` component itself rather than replicating its logic. This PR implements a pattern, previously used in #2460, of
instantiating the text component and passing in an ad-hoc template row.

One issue with this is that we can't directly modify the styles of the child `text` component from the parent component (without using `ViewEncapsulation.none`, `::ng-deep` etc.), making it tricky to apply a minor styling change to the text as it appears in context. Some styling changes can be achieved through the `text` component's existing API, for example we could pass in `text_align: center` to the parameter list of the inline row in the parent component template. 

In the case of the `text-bubble` component, the issue is that the `text` component includes its own substantial top and bottom margins. When combined with the margin applied directly on the text-bubble container, the spacing was too large. But as we want to support the text bubble containing multiple child rows (which could instantiate arbitrary components), removing the default padding would not be sensible. So I've included a hack that consists of reducing the default padding of the text-bubble container, but applying additional padding in the case that it contains child rows. This leaves the styling unchanged for all the examples in the [component demo template](https://docs.google.com/spreadsheets/d/1Yo8X4TrTW2lzEgusDplAQ-y-JJvQOIqXGNLVON1uvRo/edit?gid=569531329#gid=569531329).

## Git Issues

Closes #2435

## Screenshots/Videos

[comp_text_bubble](https://docs.google.com/spreadsheets/d/1Yo8X4TrTW2lzEgusDplAQ-y-JJvQOIqXGNLVON1uvRo/edit?gid=569531329#gid=569531329)

Screenshot 2024-10-31 at 12 12 33

Screenshot 2024-10-31 at 12 13 02

### DIFF
--- a/src/app/shared/components/template/components/text-bubble/text-bubble.component.html
+++ b/src/app/shared/components/template/components/text-bubble/text-bubble.component.html
@@ -5,9 +5,9 @@
 >
   <div class="text-bubble" [attr.data-position]="params.speakerPosition">
     @if (_row.value) {
-      <p class="text">
-        {{ _row.value }}
-      </p>
+      <plh-tmpl-text
+        [row]="{ _nested_name: '', name: '', type: 'text', value: value() }"
+      ></plh-tmpl-text>
     }
     @for (childRow of _row.rows | filterDisplayComponent; track trackByRow($index, childRow)) {
       <plh-template-component
--- a/src/app/shared/components/template/components/text-bubble/text-bubble.component.scss
+++ b/src/app/shared/components/template/components/text-bubble/text-bubble.component.scss
@@ -48,12 +48,17 @@ $imageSize: 64px;
 }
 
 .text-bubble {
-  padding: var(--large-padding);
-  p {
-    line-height: inherit;
-    margin: 0;
-    font-size: var(--font-size-text-large);
-    color: var(--ion-color-primary);
+  padding: var(--small-padding) var(--large-padding);
+
+  // HACK: The `text` component already includes a substantial margin, but when the `text-bubble`
+  // component contains child rows rather than a `value` string, apply a margin around these rows.
+  plh-template-component {
+    &:first-child {
+      margin-top: var(--regular-padding);
+    }
+    &:last-child {
+      margin-bottom: var(--regular-padding);
+    }
   }
 }
 
--- a/src/app/shared/components/template/components/text-bubble/text-bubble.component.scss
+++ b/src/app/shared/components/template/components/text-bubble/text-bubble.component.scss
@@ -48,17 +48,12 @@ $imageSize: 64px;
 }
 
 .text-bubble {
-  padding: var(--small-padding) var(--large-padding);
+  padding: var(--large-padding) var(--large-padding);
 
-  // HACK: The `text` component already includes a substantial margin, but when the `text-bubble`
-  // component contains child rows rather than a `value` string, apply a margin around these rows.
-  plh-template-component {
-    &:first-child {
-      margin-top: var(--regular-padding);
-    }
-    &:last-child {
-      margin-bottom: var(--regular-padding);
-    }
+  // Override default margins of child text component:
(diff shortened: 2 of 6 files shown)


esmeetewinkel (approved):
Functional test passed

chrismclarke (changes requested):
Makes sense to me to render the text component as you've done for better consistency and markdown support.

Just to quickly check the issue you mentioned, if I understand correctly there is an inconsistency with margins applied to the text bubble child as, when rendered as text there is a margin and as a component there is not (?).

If this is the case, then would it make more sense to change the `plh-tmpl-text` component to have a host variable `--margin-top` that can be updated to override the behavior, e.g. adding styles and container class, along the lines of

`text.component.scss`
```css
:host{
--margin-top: 1rem;
}
.text-content{
margin-top:var(--margin-top)
}
```

And then just override from the `text-bubble.scss`
```css
plh-tmpl-text{
--margin-top: 0
}
```

That should make styling consistent, and then the `text-bubble` itself can set it's own padding/margin for rendered content

jfmcquade:
> If this is the case, then would it make more sense to change the `plh-tmpl-text` component to have a host variable `--margin-top` that can be updated to override the behavior, e.g. adding styles and container class, along the lines of

Thanks, I like this suggestion. I can't quite apply it as you've suggested because the margin on the `p` block of the text component is actually being set from a global class within `src/theme/_typography.scss`, and additionally we don't have direct access to the style rules on the `p` block within the text component, because it's rendered as `innerHTML`. This could probably be improved, but for now I've updated this PR to work within the current structure, see https://github.com/IDEMSInternational/open-app-builder/pull/2498/files/0ea2a9364a149f32466d6d923d755be3addad762..72d51ebefc41aa648822c710019a2f59dd653584 for the specific changes. The `text` component is unchanged when rendered independently, and now can have its styling tweaked via the `--margin-text` variable when it's called as the child of another component.

chrismclarke:
> Thanks @chrismclarke, I like this suggestion. I can't quite apply it as you've suggested because the margin on the `p` block of the text component is actually being set from a global class within `src/theme/_typography.scss`, and additionally we don't have direct access to the style rules on the `p` block within the text component, because it's rendered as `innerHTML`. This could probably be improved, but for now I've updated this PR to work within the current structure

Thanks for updating, yeah I think that's looking a lot cleaner overall. Yeah I can see how the global typography styles are difficult to work around (particularly given that without browsers still have their own default `p` styles). I'm guessing at some point in the future we may want to consider a total reset on global selector styles a-la [Tailwind Preflight](https://tailwindcss.com/docs/preflight), but given the immediate knock-ons probably not worth doing right now.

> The `text` component is unchanged when rendered independently, and now can have its styling tweaked via the `--margin-text` variable when it's called as the child of another component.
Feels like a good pattern to me

chrismclarke (approved):

PR #2502, opened 2024-11-04
ci: fix test-appetize action trigger

jfmcquade:
PR Checklist

## Description

Follow-up to #2200, this PR reinstates the PR label trigger such that the `test-appetize` action will be run if the `test - appetize` label is added to a PR.

As noted in #2200, an `APPETIZE_APP_KEY` secret must be added to the targeted content repo. This is the "Build Identifier", i.e. the last section of the URL path, `https://appetize.io/app/{buildId}`, see [official appetize docs](https://docs.appetize.io/platform/sharing-apps#build-identifier). This can only be obtained after an initial build has been uploaded manually to appetize. I have update the debug content repo accordingly.

## Git Issues

Closes #

### DIFF
--- a/.github/workflows/reusable-appetize.yml
+++ b/.github/workflows/reusable-appetize.yml
@@ -22,7 +22,9 @@ env:
 
 
 on:
- workflow_call:
+  pull_request:
+    types: [labeled, synchronize]
+  workflow_call:
 
 jobs:
 


.github/workflows/reusable-appetize.yml
```diff
@@ -31,7 +33,9 @@ jobs:
       uses: ./.github/workflows/reusable-android-build.yml
       secrets: inherit
       with:
-        branch: ${{ github.base_ref }} 
+        branch: ${{ github.ref }}
```
chrismclarke, line 36: The reusable branch input refers to the branch of the deployment repo to checkout, not the ref of the builder repo which is set via `env.APP_CODE_BRANCH`

So I don't think it's possible to get the reusable action working in this way when being called from the builder repo itself. Instead we would need to add a different action for the builder repo like we do for pr-preview

PR #2503, opened 2024-11-04
style: make border default gray for professional and plh_kids_kw themes

### LINKED ISSUE #2426: Remove navy outline from white components
**Details**
PLH has requested to remove the navy outline from white components for a cleaner look. This will apply both to the Kids KW and FacilitatorApp deployments. Discussed briefly with @jfmcquade who said this could be done on the theme level. See also issue #2101 about a PLH theme more generally.

This affects the components: 
- Audio player (already has an issue for new design #2147)
   image
- Accordion: 
   image
- Radio button grid: 
   image
- Combo box:
   image
   image

jfmcquade:
PR Checklist

## Description

- `professional` theme: Makes the default border colour grey/gray
- `plh_kids_kw` theme: Makes the default border colour grey/gray
- Minor refactor to default border styles: remove unused global style variables

## Review notes

There should be no changes to component styles for the `default` theme. Below is a comprehensive list of all the components that make use of the variable that has been updated by this PR. If you come across any components that you would expect to be changed in accordance with #2426 but are not affected by this PR, then they do not use the default variable and can be refactored as part of this PR or as a follow-up.

@FaithDaka I've applied these changes to the `plh_kids_kw` theme (see #2426 for justification). I imagine that any custom styling you've been applying as part of that theme or otherwise will override the changes I've made as part of this PR. But it's worth being aware of these changes and identifying whether you'll need to make any updates to your custom styles.

## Git Issues

Closes #2426

## Screenshots/Videos

Below is a comprehensive list of all affected components

| Component Name    | Before | After |
|-------------------|--------|-------|
| audio             | Screenshot 2024-11-04 at 16 58 30      | Screenshot 2024-11-05 at 12 19 18    |
| accordion         |  Screenshot 2024-11-04 at 16 59 17     | Screenshot 2024-11-05 at 12 20 21    |
| combo_box         | Screenshot 2024-11-04 at 16 59 44      | Screenshot 2024-11-05 at 12 35 27   |
| combo_box_modal   | Screenshot 2024-11-04 at 17 00 06      | Screenshot 2024-11-05 at 12 35 53    |
| display_grid      | Screenshot 2024-11-04 at 17 00 31      | Screenshot 2024-11-05 at 12 36 48    |
| number_selector   | Screenshot 2024-11-04 at 17 00 49      | Screenshot 2024-11-05 at 12 37 25   |
| radio_button_grid | Screenshot 2024-11-04 at 17 01 08      | Screenshot 2024-11-05 at 12 38 01   |
| radio_group       | Screenshot 2024-11-04 at 17 01 32      | Screenshot 2024-11-05 at 12 38 29    |
| slider            | Screenshot 2024-11-04 at 17 01 48      | Screenshot 2024-11-05 at 12 39 02    |
| text_area         | Screenshot 2024-11-04 at 17 02 09      |  Screenshot 2024-11-05 at 12 39 33    |
| text_box          | Screenshot 2024-11-04 at 17 02 27      | Screenshot 2024-11-05 at 12 40 05     |
| timer             | Screenshot 2024-11-04 at 17 02 46      | Screenshot 2024-11-05 at 12 40 32   |

### DIFF
--- a/src/theme/themes/professional.scss
+++ b/src/theme/themes/professional.scss
@@ -12,7 +12,8 @@
     $variable-overrides: (
       // BORDERS
       border-color-default: var(--ion-color-gray-200),
-      border-standard: 2px solid var(--border-color-default),
+      border-width-default: 1px,
+      border-standard: var(--border-width-default) solid var(--border-color-default),
       border-thin-standard: 1px solid var(--border-color-default),
       button-background-primary: var(--ion-color-primary-500),
       button-background-secondary: var(--ion-color-secondary),
--- a/src/theme/themes/professional.scss
+++ b/src/theme/themes/professional.scss
@@ -8,8 +8,11 @@
     $color-secondary: hsl(31, 100%, 57%); // #fa9529
     $page-background: white;
 
-    /** Authoring component overrides **/
+    /** Global and component variables **/
     $variable-overrides: (
+      // BORDERS
+      ion-border-standard: 2px solid var(--ion-color-gray-200),
+      ion-border-thin-standard: 1px solid var(--ion-color-gray-200),
       button-background-primary: var(--ion-color-primary-500),
       button-background-secondary: var(--ion-color-secondary),
       button-background-option: var(--ion-color-primary-800),
@@ -48,7 +51,7 @@
       ion-item-background: var(--ion-color-gray-light),
       task-progress-bar-color: var(--ion-color-green),
       // checkbox-background-color: white,
-      progress-path-line-background: var(--ion-color-gray-100),
+      progress-path-line-background: var(--ion-color-gray-100)
     );
     @include utils.generateTheme($color-primary, $color-secondary, $page-background);
     @each $name, $value in $variable-overrides {
--- a/src/theme/themes/professional.scss
+++ b/src/theme/themes/professional.scss
@@ -11,8 +11,9 @@
     /** Global and component variables **/
     $variable-overrides: (
       // BORDERS
-      ion-border-standard: 2px solid var(--ion-color-gray-200),
-      ion-border-thin-standard: 1px solid var(--ion-color-gray-200),
+      border-color-default: var(--ion-color-gray-200),
+      ion-border-standard: 2px solid var(--border-color-default),
+      ion-border-thin-standard: 1px solid var(--border-color-default),
       button-background-primary: var(--ion-color-primary-500),
       button-background-secondary: var(--ion-color-secondary),
       button-background-option: var(--ion-color-primary-800),
--- a/src/theme/themes/professional.scss
+++ b/src/theme/themes/professional.scss
@@ -12,8 +12,8 @@
     $variable-overrides: (
       // BORDERS
       border-color-default: var(--ion-color-gray-200),
-      ion-border-standard: 2px solid var(--border-color-default),
-      ion-border-thin-standard: 1px solid var(--border-color-default),
+      border-standard: 2px solid var(--border-color-default),
+      border-thin-standard: 1px solid var(--border-color-default),
       button-background-primary: var(--ion-color-primary-500),
       button-background-secondary: var(--ion-color-secondary),
       button-background-option: var(--ion-color-primary-800),
--- a/src/app/shared/components/template/components/accordion/accordion.component.scss
+++ b/src/app/shared/components/template/components/accordion/accordion.component.scss
@@ -1,7 +1,7 @@
 // Create overlapping effect
 ion-accordion {
   background: white;
-  border: 1px solid var(--ion-color-primary);
+  border: var(--ion-border-standard);
   border-radius: 10px;
   margin-top: -12px;
   padding-top: 8px;
--- a/src/app/shared/components/template/components/accordion/accordion.component.scss
+++ b/src/app/shared/components/template/components/accordion/accordion.component.scss
@@ -1,7 +1,7 @@
 // Create overlapping effect
 ion-accordion {
   background: white;
-  border: var(--ion-border-standard);
+  border: var(--border-standard);
   border-radius: 10px;
   margin-top: -12px;
   padding-top: 8px;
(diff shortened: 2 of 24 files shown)


chrismclarke (commented):
Am I right in thinking that most ionic-based components use the `ion-border-standard` variable, and so this PR mostly exposes that variable to the override system using an author-facing `--border-color-default` variable?

If so the code changes make sense to me, I guess there's the issue of any custom components that have border styles defined standalone, but with or without this PR those would be inconsistent so I think the change should overall be good. 

Although from a design-perspective I don't like the gray. It's the same colour as the drop shadow which breaks the illusion of raised components, leaving just slightly odd sized borders. I'm not sure if this decision is coming direct from design or not, but in any case I'd encourage the default value to have better contrast. See latest material design specs on elevation:
https://m3.material.io/styles/elevation/overview

jfmcquade:
> Am I right in thinking that most ionic-based components use the `ion-border-standard` variable, and so this PR mostly exposes that variable to the override system using an author-facing `--border-color-default` variable?
> 
> If so the code changes make sense to me, I guess there's the issue of any custom components that have border styles defined standalone, but with or without this PR those would be inconsistent so I think the change should overall be good.

That was my assumption, although having investigated, ionic doesn't actually use the variable at all. So I've renamed the theme variable(s) `--ion-border-*` -> `--border-*`. There's still a lot of inconsistency within the variable names (using the `ion-` prefix being one example, as this isn't reserved for variables that are consumed by ionic components), but I'm happy not to tidy that up as part of this PR. We never officially moved non-colour variables into the theming system (as you say, there's inconsistency in how custom components apply non-colour variables, and the variables in `src/theme/variables.scss` should be moved over to the theme level and standardised), so this PR makes a start at doing so. 

> Although from a design-perspective I don't like the gray. It's the same colour as the drop shadow which breaks the illusion of raised components, leaving just slightly odd sized borders. I'm not sure if this decision is coming direct from design or not, but in any case I'd encourage the default value to have better contrast. See latest material design specs on elevation: https://m3.material.io/styles/elevation/overview

| M2 | M3 |
|---------|
| lwykc8yv-2 | lwykckey-3 |

I agree, the gray border doesn't look very good. I've updated the border to be 1px rather than 2px and I think this makes a big difference (see updated screenshots in PR description). Now the border feels more like part of the drop shadow, and hopefully reinforces the illusion of raised components rather than breaking it.

The design request as I understand it was to "remove the border" (#2426), so I think they were imagining something broadly inline with the M2 specification. As we don't consistently apply drop-shadows, I think making the border a thin grey is the closest we can get with minimal changes. I do know what you mean though. Ultimately we still have a lot of inconsistency, but I do think the changes included here are an improvement visually.

The changes that @FaithDaka is making for the `plh_kids_kw` deployment feel more in line with the M3 specification in the docs you've linked to, and the plan is to eventually roll those styling changes out across other deployments (at least PLH ones), so I think that should definitely move us in the right direction.

esmeetewinkel (approved):

src/theme/themes/professional.scss
esmeetewinkel, line 1: @jfmcquade actually I would like to leave the `professional` theme as it is. It's used in Kids TZ and Teens TZ (where this change has not been requested)

chrismclarke:
> That was my assumption, although having investigated, ionic doesn't actually use the variable at all. So I've renamed the theme variable(s) `--ion-border-*` -> `--border-*`. There's still a lot of inconsistency within the variable names (using the `ion-` prefix being one example, as this isn't reserved for variables that are consumed by ionic components), but I'm happy not to tidy that up as part of this PR. We never officially moved non-colour variables into the theming system (as you say, there's inconsistency in how custom components apply non-colour variables, and the variables in `src/theme/variables.scss` should be moved over to the theme level and standardised), so this PR makes a start at doing so.

Ah, that makes a lot of sense (not from the code, but as a 'why things are how they are'). Yes I vaguely remember a time when developers would just use `ion` as a generic prefix for variables that they created, which is obviously a huge source of confusion. Glad you were able to identify this and continue to move things in a clearer direction.

> I agree, the gray border doesn't look very good. I've updated the border to be 1px rather than 2px and I think this makes a big difference (see updated screenshots in PR description). Now the border feels more like part of the drop shadow, and hopefully reinforces the illusion of raised components rather than breaking it.

Thanks for revisiting this, yes I think that small change actually makes a huge difference. 

> The design request as I understand it was to "remove the border" (#2426), so I think they were imagining something broadly inline with the M2 specification. As we don't consistently apply drop-shadows, I think making the border a thin grey is the closest we can get with minimal changes. I find it difficult to keep track of the latest design mock-ups, but looking at [this Figma](https://www.figma.com/design/CYnvCNWHbgMdXlqBeBnc4T/ParentApp-for-Kids?node-id=428-4559&node-type=frame&t=ijafah11RdOr227K-0), specifically the "Library Page Mockup - 1" frame, there is a thin grey border around the elements, coupled with a drop-shadow to achieve the desired effect. Likewise some examples from [this figma](https://www.figma.com/design/inDybEiRljIM4KRLyFhIxR/Laurie's-Playground?node-id=0-1&node-type=canvas&t=YJfdtJ2ZIm3WLHk6-0):

Yeah I'm personally not a huge fan of large drop shadows on rounded components anyway (feels a bit outdated, similar to a lot of the 2009/10 designs from https://www.toptal.com/designers/ui/button-design-dribbble-timeline). I think ionic themselves also started to move away from this style a few years back but our design defaults typically fallback to the early design drafts that were more like these. So nice if our defaults do start to better-reflect more common modern design practices.

>   

> I do know what you mean though. Ultimately we still have a lot of inconsistency, but I do think the changes included here are an improvement visually.

Agreed, and if we just keep making these minor changes as things get raised then I think over time hopefully most of the codebase should continue to improve without the need for a major overhaul. 

> The changes that @FaithDaka is making for the `plh_kids_kw` deployment feel more in line with the M3 specification in the docs you've linked to, and the plan is to eventually roll those styling changes out across other deployments (at least PLH ones), so I think that should definitely move us in the right direction.

chrismclarke (approved):

PR #2504, opened 2024-11-04
feat: display group variant box white

### LINKED ISSUE #2437: [FEATURE] Display group variants with no background
**What?**
I would like to have further variants of the gray / primary / secondary display groups where the background is transparent or white.

image

**Why?**
Requested by PLH Kids TZ

**How?**
I think I can figure out how to add variants following PR https://github.com/IDEMSInternational/open-app-builder/pull/2227 but I'm not sure what naming convention we'll want to use.

jfmcquade:
PR Checklist

## Description

Adds a new variant for the display group: `box_white`. The border colour and weight are determined by the new theme variables introduced in #2503, so currently the border colour is "primary" for the `default` theme and light grey for the `professional` theme. One knock-on is that the border weight of the other `box_*` variants is also now thinner for the professional theme, in line with the theme-level styling introduced in #2503.

## Git Issues

Closes #2437 

## Screenshots/Videos

| Theme: default | Theme: professional |
|------|------|
| Screenshot 2024-11-07 at 12 54 08 | Screenshot 2024-11-07 at 12 58 22 |

### DIFF
--- a/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
+++ b/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
@@ -52,7 +52,8 @@
     border-radius: var(--ion-border-radius-secondary);
     flex: 1;
     background-color: var(--background-color, transparent);
-    border: 2px solid var(--border-color, transparent);
+    border: var(--border-standard);
+    border-color: var(--border-color);
   }
 
   &[data-variant~="box_gray"] {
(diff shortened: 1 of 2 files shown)


esmeetewinkel (approved):
Looks good to me, thanks. Noted about the knock-on - I don't think it's an issue, nice to be consistent.

chrismclarke (commented):
Looks good to me, feels like `white` is a common enough requirement to sit outside of primary/secondary naming.

The other option would be something like `light` or `dark`, but I think that can be a bit confusing, and doesn't work well if planning for things like a dark mode where you want the main colours used to shift. In that case we may want to export a author variable to specify the default `box` background colour, which they could update depending on theme.

PR #2509, opened 2024-11-07
Review: 2501

chrismclarke:
PR Checklist

## Description

Targets branch of #2501 

Main changes:
- Simplify template-metadata service logic using signals
- Tracked currently locked orientation and only change if different value provided
- Add support for unlocking orientation
- Add angular utility function for route subscription

## Review Notes
I've also updated the demo template:
https://docs.google.com/spreadsheets/d/13KnDdmBicPDZaWFNnlIck7WNMLlM02HjD0-_UiqZ4z8/edit?gid=569531329#gid=569531329

image

For now I've kept enabled on web when not in production (local testing) to see clearly the console logs when orientation changes are being called, let me know if things work as expected to you and we can remove before pushing into production.

I haven't tested on a physical device, but using logs to see if appears to trigger as expected

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/services/template-metadata.service.ts
+++ b/src/app/shared/components/template/services/template-metadata.service.ts
@@ -1,10 +1,11 @@
-import { Injectable, OnDestroy, signal } from "@angular/core";
+import { computed, effect, Injectable, signal } from "@angular/core";
 import { SyncServiceBase } from "src/app/shared/services/syncService.base";
 import { TemplateService } from "./template.service";
 import { FlowTypes } from "src/app/shared/model";
-import { ActivatedRoute, NavigationEnd, Router } from "@angular/router";
-import { distinctUntilChanged, filter, map, Subject, takeUntil } from "rxjs";
-import { Capacitor } from "@capacitor/core";
+import { Router } from "@angular/router";
+import { toSignal } from "@angular/core/rxjs-interop";
+import { ngRouterMergedSnapshot$ } from "src/app/shared/utils/angular.utils";
+import { isEqual } from "packages/shared/src/utils/object-utils";
 
 /**
  * Service responsible for handling metadata of the current top-level template,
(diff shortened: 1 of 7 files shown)


src/app/shared/components/template/services/template-metadata.service.ts
```diff
@@ -13,64 +14,32 @@ import { Capacitor } from "@capacitor/core";
...
 
   constructor(
     private templateService: TemplateService,
     private router: Router
   ) {
     super("TemplateMetadata");
 
-    // Currently the only watched parameter is for screen orientation,
-    // which is only supported on native platforms
-    this.enabled = !Capacitor.isNativePlatform();
```
jfmcquade, line 29: I can see how setting the enabled state based on a child dependency isn't sensible. But I'm aware that this service adds logic that runs on every route change and will only have an effect on one deployment currently, which makes use of the screen orientation feature. Do you think it would be sensible to do something like run a check of all templates on init to see whether any make use of the `parameter_list` properties tracked by this service, and only enable it in the case that such a template exists?

Either way, I imagine the performance impact of this service is minimal, so I'm happy to merge as-is.

jfmcquade (approved):
Thanks, @chrismclarke, really nice to have these changes.

Unlocking the screen orientation by default rather than locking to portrait makes sense. And the signal refactoring is nice.

I will merge this into the PR branch and then run some testing on native platforms and update the PR description.

src/app/shared/components/template/services/template-metadata.service.ts
chrismclarke: Yeah as this has no UI re-rendering effect and that templates change at most once per few seconds I can't see this really doing much to impact performance, so would make sense to optimise as you've suggested (from the templating). We're already planning to do this for other modules/features, so that if not in use the entire service will be excluded from the build (no need to toggle in the service itself). This should hopefully keep the code a bit simpler.

Template analysis won't know anything about native/web devices, so good to keep those checks in other services

PR #2511, opened 2024-11-07
feat: set screen orientation on fullscreen pop-ups

jfmcquade:
PR Checklist

## Description

Following on from #2501, allows fullscreen pop-ups to display in landscape mode – i.e. when a template has an `orientation` value set in its `parameter_list`, e.g. `orientation: landscape`, the screen orientation will be set to the respective value when launching the template as a fullscreen popup, and reset when the pop-up is closed, whether through the "close" button or through emitting `completed`/`uncompleted`.

## Notes

Originally I had sought to find a solution from authoring within the current system, along the lines of:
```
click | screen_orientation: landscape;
click | pop_up:example_emit | fullscreen: true;
uncompleted | screen_orientation: portrait;
completed | screen_orientation: portrait;
```
But while this works for setting the orientation when launching the pop-up, for whatever reason `completed`/`uncompleted` events from the pop-up template to not trigger the action as expected. So it was still required to set the orientation to "unlock" within the `dismiss()` method of the popup component (see https://github.com/IDEMSInternational/open-app-builder/pull/2511/commits/5f210d100a972856de88b9173b625d977b9faa07), at which point it made sense to handle the screen orientation entirely within the component based on the template `parameter_list`, and remove the need for authoring the action explicitly.

Whilst still a hack, this approach means it doesn't rely on the screen orientation template action, which could still be removed as suggested [here](https://github.com/IDEMSInternational/open-app-builder/pull/2501#issuecomment-2462699823).

## Testing

Latest debug builds updated on Appetize:
- [Android](https://appetize.io/apps/android/international.idems.debug_app)
- [iOS](https://appetize.io/apps/ios/international.idems.debug-app)

## Git Issues

Closes #

## Screenshots/Videos

[feature_screen_orienation](https://docs.google.com/spreadsheets/d/13KnDdmBicPDZaWFNnlIck7WNMLlM02HjD0-_UiqZ4z8/edit?gid=569531329#gid=569531329)

Screenshot 2024-11-07 at 19 29 41

Android:

https://github.com/user-attachments/assets/5a9352ca-d065-41c4-abf2-54e1dc6a85f8

### DIFF
--- a/src/app/shared/components/template/components/layout/popup/popup.component.ts
+++ b/src/app/shared/components/template/components/layout/popup/popup.component.ts
@@ -2,6 +2,7 @@ import { Component, Input } from "@angular/core";
 import { ModalController } from "@ionic/angular";
 import { FlowTypes, ITemplateContainerProps } from "../../../models";
 import { TemplateContainerComponent } from "../../../template-container.component";
+import { ScreenOrientationService } from "src/app/shared/services/screen-orientation/screen-orientation.service";
 
 @Component({
   templateUrl: "./popup.component.html",
@@ -14,7 +15,10 @@ import { TemplateContainerComponent } from "../../../template-container.componen
 export class TemplatePopupComponent {
   @Input() props: ITemplatePopupComponentProps;
 
-  constructor(private modalCtrl: ModalController) {}
+  constructor(
+    private modalCtrl: ModalController,
+    private screenOrientationService: ScreenOrientationService
+  ) {}
 
   /**
    * When templates emit completed/uncompleted value from standalone popup close the popup
@@ -38,6 +42,11 @@ export class TemplatePopupComponent {
 
   dismiss(value?: { emit_value: string; emit_data: any }) {
     this.modalCtrl.dismiss(value);
+    // HACK - unlock screen orientation when dismissing a fullscreen pop-up. This allows screen orientation
+    // to be set on a fullscreen pop-up (and crucially, unset) despite it not being a top-level template
+    if (this.props.fullscreen) {
+      this.screenOrientationService.setOrientation("unlock");
+    }
   }
 }
 
--- a/src/app/shared/components/template/components/layout/popup/popup.component.ts
+++ b/src/app/shared/components/template/components/layout/popup/popup.component.ts
@@ -1,8 +1,12 @@
-import { Component, Input } from "@angular/core";
+import { Component, Input, OnInit } from "@angular/core";
 import { ModalController } from "@ionic/angular";
 import { FlowTypes, ITemplateContainerProps } from "../../../models";
 import { TemplateContainerComponent } from "../../../template-container.component";
-import { ScreenOrientationService } from "src/app/shared/services/screen-orientation/screen-orientation.service";
+import {
+  IScreenOrientation,
+  ScreenOrientationService,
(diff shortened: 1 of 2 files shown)


chrismclarke:
I feel like the proposed solution using actions is the way we want things to work, and a lot cleaner than trying to manually track popup template params in the same way we track normal params.

I'd say it would be best to figure out why the completed/uncompleted action isn't triggering, or whether there is a different action trigger that is linked instead that can be used. If not possible or not making sense then probably best to just add a hackUnlockOrientation method wherever the popup is triggered from to manually unlock on dismiss

jfmcquade:
> I feel like the proposed solution using actions is the way we want things to work, and a lot cleaner than trying to manually track popup template params in the same way we track normal params.

Hmm yeh it didn't seem too unpalatable to me to track the template params at the popup component level – as we're essentially passing in a whole template then handling the template metadata at this level doesn't seem obviously wrong 
 – apart from being independent of the existing template metadata service. Not ideal, but the pop-up feature already bypasses standard nav features in other ways.

> I'd say it would be best to figure out why the completed/uncompleted action isn't triggering, or whether there is a different action trigger that is linked instead that can be used. If not possible or not making sense then probably best to just add a hackUnlockOrientation method wherever the popup is triggered from to manually unlock on dismiss

Yes unfortunately we don't have a `dismissed` event trigger that would be emitted when a pop-up is dismissed. I considered adding one, or else firing `emit: uncompleted` on pop-up dismiss, but as mentioned the `completed`/`uncompleted` events don't seem to be caught at the level of the authored `pop_up` action anyway, so there's more to unpick. I'm out of office for the weekend from tomorrow so don't have time to investigate further.

Do you think the current implementation on this PR is too dodgy to incorporate? @esmeetewinkel needs to do a release tomorrow, but it's not a disaster if we don't have landscape mode for the storybooks this time round.

chrismclarke:
> Hmm yeh it didn't seem too unpalatable to me to track the template params at the popup component level – as we're essentially passing in a whole template then handling the template metadata at this level doesn't seem obviously wrong (apart from being independent of the existing template metadata service). Not ideal, but the pop-up feature already bypasses standard nav features in other ways.

> Yes unfortunately we don't have a `dismissed` event trigger that would be emitted when a pop-up is dismissed. I considered adding one, or else firing `emit: uncompleted` on pop-up dismiss, but as mentioned the `completed`/`uncompleted` events don't seem to be caught at the level of the authored `pop_up` action anyway, so there's more to unpick. I'm out of office for the weekend from tomorrow so don't have time to investigate further.

> Do you think the current implementation on this PR is too dodgy to incorporate? @esmeetewinkel needs to do a release tomorrow, but it's not a disaster if we don't have landscape mode for the storybooks this time round.

Oh whoops, I thought originally I was commenting on an issue, didn't even realise this was a PR, sorry about that. I'll take a better look at the code and see if I can get anything any tidier. I'd say probably not worth rushing this in right now if can be avoided as does feel like quite a lot of code to add to the popup service to handle something it really shouldn't have awareness of...

chrismclarke (changes requested):
@jfmcquade 
I've opened up an alternate PR #2514 which focuses more on just tracking the template name of the full screen popup (when exists), so that the rest of the metadata service can function unchanged. Let me know what you think

PR #2517, opened 2024-11-09
fix: seo service config

chrismclarke:
PR Checklist

## Description

Fix issue where seo service fails to initialise if deployment does not have any app header config set

## Dev Notes
The reason for the issue is that the seo-service was importing app config via the deployment config (which only includes config overrides), instead of app config service (which includes full merged config). So when accessing nested properties it would throw an error if no APP_HEADER_DEFAULTS had been overwritten.

Most deployments typically have app header overrides so would mostly go unnoticed, it was only when running the local deployment that the error was noticed. Searching the codebase this is the only service that I could see that tried to access app config in this way.

## Git Issues

Closes #

## Screenshots/Videos
**Source code search for other services accessing app_config** - only case is a comment in task-card component and the app-config service itself which handles merging full config values
image


jfmcquade (approved):
Good catch, looks good to me

PR #2520, opened 2024-11-12
Feat: deployment build optimisation

chrismclarke:
PR Checklist

**Future Follow-ups**
- [ ] Service optimisation/opt-out (where reasonable to configure/interpret)
- [ ] Markdown pipe refactor and opt-out
- [ ] Latex pipe refactor and opt-out #2530
- [ ] QR Code pipe refactor and opt-out #2530
- [ ] Font optimisation (currently bundles duplicate fonts, and for all themes)
- [ ] Documentation
- [ ] Consider additional types of optimisation, such as asset compression and deduplication

## Description
Add a new workflow `yarn workflow optimise_build` to remove unused code from deployment build. 
In it's first iteration there it performs the following optimisations:

- Remove code for any unused components
- Add manifest system to define 3rd party dependencies of components
- Remove imported 3rd party modules required by components (if not in use)
- Remove bundled 3rd party assets included in angular json

By default the optimisation workflow is run as part of any app build, however will only change files if specific deployment configuration chooses to opt in.

## Author Notes
**Opt In**
Deployments can opt in via a new `optimisation` config property. Currently defined as
```ts
config.optimisation = {
  components:{
    enabled:true
  }
}
```

**Advanced Use Cases**
There may be cases where some template components use other components within their core code, but where these implicit components are not referenced anywhere else and so removed from the build. E.g. the `task_card` component which uses the `task_progress_bar` component. 

If this happens the build will fail with an error such as
image

To fix this the optimisation can manually specify a list of components to include, e.g.
```ts
config.optimisation = {
  components:{
    enabled:true,
    implicit:["task_progress_bar"]
  }
}
```
Alternatively we have started tracking a list of these implicit dependencies within the code itself, so if you notice any we can always update the list to save having to manually specify at the deployment level.

**Local Diffs**
Any deployment that has optimisation included will change the default `angular.json` file that is included in the main build. These changes shouldn't be committed to the main repo (not an issue when running build on github action, just need to remember to discard the changes if running build locally)

**Misc Changes**
All generated reports now have a parent `data:` property used to improve consistency across reports for code checks (will show up a small diff when next syncing)

## Review Notes
- Update deployment config to opt-in to optimisation
- Recompile deployment via `yarn workflow deployment set [name]`
- Sync the deployment to update the reports (format changed slightly); `yarn workflow sync`
- Build via `yarn build`
- Review changed files listed in console logs

Additionally you can analyse the build bundle sizes by running the `yarn analyse` command (instead of `yarn build`)
(see example screenshots)

If wanting to just run the optimisation step can manually call `yarn workflow optimise_build` (this is included in the build command)

Tests can be run via
```sh
yarn workspace scripts test -t optimisers/components.spec.ts
```

## Dev Notes
**Review Support**
There's quite a lot of changes here, many of which are just tidying the existing lists of components. If a bit confusing let me know and I can split the functional from non-functional changes and open as separate PRs.

All the code to handle optimisation processes should be within the optimisation scripts code folder. Tests are also included to cover the main methods and use-cases.

**General Approach**
The process of optimising components is a bit complex given that components can also depend on imported modules or modifications to angular.json (assets, styles or scripts). In addition, many components have implicit dependencies, where they import another component into the HTML templates which in turn may have its own requirements. As a result of this, the following approach has been implemented to try and solve:

1. Create a clear separation from common components used by templating to render outputs (e.g. `button`), base components defined in templates however removed during processing and non-rendered (e.g. `set_field`), and a handful of `core` components that are required by specific hardcoded functionality (e.g. `text` which is used in not-found page redirect and other places)

2. Develop a manifest schema that can be use to define component dependencies across module imports, angular json modifications and implicit components required

3. Create a new optimisation workflow and deployment configuration to enable build-time changes that modify core code, following analysis of which components are/aren't used in the deployment sheets and their corresponding manifests. The optimisations center around identifying imports no longer required and commenting out (allowing angular to drop from the build when tree-shaking). These changes are written to separate `.deployment` files (e.g. `template.module.deployment.ts`) and overwritten during angular build

**Performance Gains**
The main result of the current set of optimisations is to reduce the amount of code that will be bundled into the compiled `main.js` file. This will have the biggest performance gains on web, where the main.js file will have to be fully downloaded before the app can start to initialise. This file is typically around 1.1MB, when served over a host with gzip, and reduced by around 25-50% depending on what components the deployment is using (e.g. 700kb for plh_kids_kw, mostly saved from opting out of pdf and odk components).

There are also some gains to android, however less significant as all files are included in the build and do not require internet to download. So the main improvement is just a slightly smaller overall bundles size when comparing the parsed bundle sizes before and after (3.85MB -> 2.44MB for plh_kids_kw). There will also be savings from bundled assets, e.g. `comp_pdf` bundles around 4.5MB of assets.  

There is unlikely to be much of a performance gain for general app speed/startup times as most of the blocking logic happens within service initialisation that isn't managed from components. This will be a future area of follow-up.

**Limitations**
There's noticeable bundle increase required to include dependencies for markdown (marked) and latex (katex), however as these dependencies are coded into pipes (instead of imported directly into components), it's not possible to remove (unless including duplicate dummy pipes). It would be good as a follow-up to refactor this to import directly into the component instead. Should also do the same for qr-code pipe.

The next larges bundle to remove would likely be intro.js, however as this is linked to actions we do not currently have a means to optimise. This could be a follow-up

It's not currently possible to opt out of components imported from the components package (e.g. plh components). As these cannot currently register additional 3rd party libs it is assumed they will be small in size so not a priority to remove from build. Likely a better solution in the future will be to move the additional component packages out of the main build entirely, and have some different means to import as requried per-deployment.

I had wanted to include the component manifest in the same `index.ts` file where the components were listed, however during to typescript interop issues this wasn't possible (scripts workspace is commonjs but src code is es2020, when importing from src also follows imports of individual components with `@component` decorators which is also not supported). So using a separate `manifest.ts` file for now and could look to unify in the future.

**Additional Notes**
- The decision to comment out code and write to a separate file, instead of just removing from original, is to provide slightly easier comparison when inspecting builds and prevent build breaking when switching between deployments that do and do not use optimisations.

- It wasn't possible to take the same approach to the `angular.json` file (e.g. `angular.deployment.json`) as the angular cli does not support pointing to custom file paths. Whilst it would be possible to keep a `angular.base.json` which is copied to `angular.json`, this would mean removing `angular.json` from git and no deployments working without the optimisation step carried out first (not a nice pattern).

- The option to provide deployment config with manual `implicit:[]` component definitions came before I added a more general component manifest approach, and so this can likely be removed in the future. I've kept in for now just in case we need to quickly fix emergent issues (but would still favour just updating the component manifests instead). It might also be useful in the future if we were to provide live-sync of sheets from an external source, so that an author could include components they no likely to be needed in the future

- The `display_group` component includes a number of implicit dependencies for highly specific use cases. It would be good to check if still in use and ideally refactor. For now there is a hardcoded hack to include these implicit dependencies. Reviewing other components it seems most are fairly pure (contain no other component references), however it might be good to get to a point where all either contain no other component references (beyond core `text` and a maybe a couple others), or are bundled together as a component pack to import. In the future I might just look at component prefixes, and if for example see any reference to `parent_` import all components within that namespace (even if not all used).

- It's not currently possible to optimise ionic components, however it would be possible if we migrate to using standalone angular components. This could be considered as a follow-up (ionic does have auto code-gen to migrate: https://ionicframework.com/docs/angular/build-options#standalone)

**Afterthought**
There's part of me that thinks we might be going about this all backwards. Instead of starting with a full set of components and dependencies and removing what isn't required, it would probably make more sense to implement some sort of plugin system where components themselves specify the changes to be made to angular.json, package.json and any other files in order to implement (and then only run that code if included in templates). As of now if we add new components that require changes to angular.json then we need to manually update optimisations to include code to revert.

But this would be a larger feature to implement, so perhaps the optimisation system will be ok for now (at least until we have deployments that need to install packages that we haven't included). In these cases I expect short term we just add more packages as required to core (e.g. charts lib, payment processor etc.), and review in future best way to make more extensible.

So for now I've included a component `manifest` which is being used to handle knock-on effects of opting out of components,  but in the future could also be used to generate the file updates required.

## Git Issues

Closes #

## Screenshots/Videos
**Before** - compiled main.js for non-optimised deployments
image

**After** - plh_kids_kw main bundle smaller after dropping dependencies for odk, pdf and a number of other components
image

**Example angular.json diff** - removes comp-pdf assets not in use and includes file overrides
image

**Example component index override** - unused components are commented out and not included in build
image

**Example module override** - unused modules are commented out and not included in build
image


esmeetewinkel (commented):
I tested this on the `plh_kids_tz` deployment. When running `yarn build` I'm getting an error 
```
ERROR! An error was encountered while executing
TypeError: Cannot read properties of undefined (reading 'map')
```

image

chrismclarke:
> I tested this on the `plh_kids_tz` and `plh_kids_kw` deployments. When running `yarn build` I'm getting an error
> 
> ```
> ERROR! An error was encountered while executing
> TypeError: Cannot read properties of undefined (reading 'map')
> ```
> 
> image

My bad, I had left a small note in the **misc notes** section but didn't make it clear in the review notes. A sync is needed on the deployment before build just to re-generate the deployment reports which have a slightly changed structure (should work after that). I've added another bullet the review notes

> - Sync the deployment to update the reports (format changed slightly); yarn workflow sync

esmeetewinkel (commented):
I tested on `plh_kids_tz`. I added 
```
config.optimisation = {
  components:{
    enabled:true
  }
}
``` 
to the config, i.e. not explicitly mentioning any components that should remain included. 

1. For some reason the reports didn't get updated when running `yarn workflow sync` (even though it didn't display any warnings/errors in the console). I then ran `yarn workflow report` instead after which I did see the report changes. 
2. Two lines were commented out in `src\app\shared\components\template\components\index.deployment.ts` that I believe are needed for this app: 

   ```
   // import { AccordionSectionComponent } from "./layout/accordion-section/accordion-section.component";
   // import { YoutubeComponent } from "./youtube/youtube.component";
   ```
   However, I then ran `yarn start` and found that both the accordion and youtube player work as expected. So I'm not sure how to interpret this?

**After:** 4.11MB
image

jfmcquade (approved):
Seems like a good solution to a very complex problem. I understand your points about potentially working from the other way round at some point, and including plugins/modules as required rather than excluding elements that are not used. But this still seems like a step in the right direction.

I think I understand the code enough to work with this feature going forward. I noticed that on the `debug` deployment, the `workshops_accordion` component was throwing an error as it has a dependency that wasn't declared in the manifest, I've added with https://github.com/IDEMSInternational/open-app-builder/pull/2520/commits/1bd38faeb36884527162db0776e99b5e23f9f62a.

I'm aware that this feature could potentially cause some issues that might be hard for a new developer to debug. I knew what might be causing the error I was seeing (which resembled the one screenshotted in the "**Advanced Use Cases**" section of the PR description) because I knew I was testing this particular feature and because I'd acquainted myself with the idea behind the new component manifest functionality. But I can imagine that for a developer working on the code for the first time, such an error would be extremely difficult to debug as it relies upon an understanding that the build contains a step to strip out some code (and knowing that the `index.deployment.ts` file is there to inspect). Some documentation would help, as noted in the suggested follow-ups, but I think it could still cause confusion. In any case, I don't see how that can be avoided – it would be difficult to present a dev-facing error message that distinguished between this kind of error due to a genuine code issue and one that was only present after the optimisation step. I just thought I'd flag here as I'm aware of the potential difficulty.

chrismclarke:
> to the config, i.e. not explicitly mentioning any components that should remain included. Two lines were commented out in `src\app\shared\components\template\components\index.deployment.ts` that I believe are needed for this app:
> 
> ```
> // import { AccordionSectionComponent } from "./layout/accordion-section/accordion-section.component";
> // import { YoutubeComponent } from "./youtube/youtube.component";
> ```
> 
> However, I then ran `yarn start` and found that both the accordion and youtube player work as expected. So I'm not sure how to interpret this?

The optimisation only happens when creating a production build via the `yarn build` command, when running as `yarn start` it instead uses a development build which doesn't have the optimisations. So this is good in terms that people shouldn't ever face issues when running locally related to optimisation, but with the trade-off that the production build might have issues not detected locally.

Whilst it would be possible to apply the same optimisations when running locally it would necessitate the author to restart the app and trigger the optimisations any time a new component is used when authoring, which isn't an ideal workflow either. So for now optimisation is more of an advanced feature which we definitely will need to test more carefully when implementing for the first time (e.g. adding screenshot test action to deployments before enabling to check all the templates look the same before and after)

chrismclarke:
> Seems like a good solution to a very complex problem. I understand your points about potentially working from the other way round at some point, and including plugins/modules as required rather than excluding elements that are not used. But this still seems like a step in the right direction.

Yeah, I think in particular by having a manifest it opens up work in either direction, so that's the most important part for sure.

> I think I understand the code enough to work with this feature going forward. I noticed that on the `debug` deployment, the `workshops_accordion` component was throwing an error as it has a dependency that wasn't declared in the manifest, I've added with [1bd38fa](https://github.com/IDEMSInternational/open-app-builder/commit/1bd38faeb36884527162db0776e99b5e23f9f62a).

Thanks for catching this, yes I haven't extensively gone through all the components to check and so wouldn't be surprised if a few more crop up over time. At least for this type of issue it's pretty obvious during optimised build.

> I'm aware that this feature could potentially cause some issues that might be hard for a new developer to debug. I knew what might be causing the error I was seeing (which resembled the one screenshotted in the "**Advanced Use Cases**" section of the PR description) because I knew I was testing this particular feature and because I'd acquainted myself with the idea behind the new component manifest functionality. But I can imagine that for a developer working on the code for the first time, such an error would be extremely difficult to debug as it relies upon an understanding that the build contains a step to strip out some code (and knowing that the `index.deployment.ts` file is there to inspect). Some documentation would help, as noted in the suggested follow-ups, but I think it could still cause confusion. In any case, I don't see how that can be avoided – it would be difficult to present a dev-facing error message that distinguished between this kind of error due to a genuine code issue and one that was only present after the optimisation step. I just thought I'd flag here as I'm aware of the potential difficulty.

Agreed, optimisations are an advanced feature which I expect most developers shouldn't really find themselves dealing much with. If enabling optimisation makes builds fail then that would be a good cue to open an issue for core dev team to look at (likely just a manifest entry missing), but I'd hope over time these types of issues become fewer as we retrofit the manifests and either become more systematic or add more tooling to check dependencies moving forwards.

I think it will be hard to have much control over compiler errors due to how much the angular compiler runs in a sandbox (it barely resembles the original webpack compiler, plus it's now all getting shifted over to esbuild so hard to know where a person might begin with handling compiler output beyond wrapping all calls in a custom child process)

PR #2530, opened 2024-11-15
Refactor: latex and qrcode pipes

chrismclarke:
PR Checklist

**TODO**
- [x] Builds on top of #2520, should merge after. Specific changes in [9c046fb](https://github.com/IDEMSInternational/open-app-builder/pull/2530/commits/9c046fb3ca8692f24e519a98768a9b6363e8ffcc)

## Description
Refactor `latex` and `qrcode` pipes to instead be called from their corresponding components.

## Review Notes
Confirm qrcode and latex example component sheets working as expected (looks correct to me)

## Dev Notes
As there are currently no other components that attempt to parse qrcode or latex values it makes sense to keep the code component-specific, and hence will be able to tree-shake in/out during an optimised build depending on component use.

I've moved the latex code to a util file so that if other components did want to implement latex they could still call the method (which would by default also be included in optimised build whenever the component is, without the need to define any implicit dependencies due to the direct import).

I haven't refactors the `markdown` pipe as it is currently in use by both button and text component which are included in every build regardless of optimisation (so will break the pipe for any build that tries to remove). It's also used by tile and simple checkbox components.

In the future likely the best way to remove would probably be some sort of parameter_list setting to specify `format: markdown`, which can be parsed in the template service before providing the value at the component level. We could then use report analysis to see if authors ever use, and remove formatter as necessary. Would likely also want to include a global option to enable by default on all components (if desirable). But all this is quite overkill for now, so just sticking to the easier-to-manage pipes.

## Git Issues

Closes #

## Screenshots/Videos

**Build Analysis** - qrcode and katex dependencies no longer appearing when compared to #2520 snapshot of plh_kids_kw build. Core bundle reduced additional 300kb/~10%
image


chrismclarke:
#2520 now merged so marking this as ready for review @jfmcquade

jfmcquade (approved):
Looks good to me. The QR code and Latex components seem to be functioning as expected.

I can see how the changes are required for better tree-shaking, but initially it felt like something was lost by removing the pipes (more so for `latex` than `qrcode`, as the latter is unlikely to have general application), as latex parsing feels like a perfect candidate for an Angular pipe. On reflection, I think where we actually want that flexibility and convenience is at the authoring level: it would be nice if text values themselves could be tagged to be parsed as latex/markdown etc. from templates. 

This is inline with what you've suggested in the PR notes as desirable in the future. It feels like ultimately whether a text string should be interpreted as latex is really metadata on the string itself. But without the ability to attach such metadata, I think specifying a text transform within a parameter list would make sense.

Issue #2507, opened 2024-11-06
[BUG] Custom content card has uneven padding


esmeetewinkel:
**Describe the bug**
A card with custom content (authored with `begin_task_card` and `end_task_card`) seems to have less padding on the right hand side than on the left. This is particularly obvious when displaying "centred" content (which appears off-centre). 

**App version**
v0.16.36 

**Deployment Sheet(s)**
PLH Kids TZ: module_landing [sheet](https://docs.google.com/spreadsheets/d/1hIT1V3xGZRFyD-k2heJibYr-UTqkkCc6eCvreiokNMk/edit?gid=452625780#gid=452625780) [preview](https://plh-kids-tz.web.app/template/module_landing_connect)
![image](https://github.com/user-attachments/assets/ecfb830b-e408-47b8-af11-fc559d5a7079)

**Debug Sheet(s)**
comp_task_card [sheet](https://docs.google.com/spreadsheets/d/1_RY8HbsATOAZU2EDXY5R9OOE_YBEbW0EjTP5zxYP2aA/edit?gid=569531329#gid=569531329) [preview](https://idems-debug.web.app/template/comp_task_card)
![image](https://github.com/user-attachments/assets/4f84071d-1ff6-4261-b07e-31874312403a)

jfmcquade:
Closed by #2543 (just confirmed by functional testing)

PR #2543, opened 2024-11-20
feat: apply item row operations, e.g. filter, on task progress bar data rows

### LINKED ISSUE #2537: [FEATURE] Ability to filter task groups
**What?**
This should make it possible for task groups to be set against a subset of a datalist. The possibility to add at least one filter condition. i.e. `group == [group_name]`

**Why?**
Currently is a task group is defined it needs to be the entire datalist. This adds complications as the subsets need to be created by generators. It also makes it very difficult to have multiple task progress bars i.e. When Module, Group, Article structure is used Module will use Group as a task group and Group will use Article. This means there is no way to see the progress of all Articles in a Group. 

Another use case is the Facilitator MX deployment. That wants to show progress per user group. A way this can be achieved is by using the completed field for all task groups and storing a separate completed_[user_group] field alongside. When the user group changes the completed field will be reset to the completed[user_group] value. This will be difficult across multiple datalists

jfmcquade:
PR Checklist

## Description

Makes the following changes to the `task_card` and `task_progress_bar` components:
- Both components can now take parameters matching the item data pipe operations, namely `shuffle`, `sort`, `filter`, `reverse` and `limit`. These operations will be applied to the data list rows of subtasks that make up the progress through the named task group. Note that rows within the targeted data list should be referenced with `@task_item` to avoid conflict with the `@item` context from any enclosing data-items loop.
  - E.g., in component parameter list, `filter: @task_item.id !== @local.excluded_task_id;`
  - The immediate use case is for `filter`, as requested by #2537
  - The other operations are unlikely to be useful: as the data rows (and their respective completion statuses) are just counted when passed to these components, the `shuffle`, `sort` and `reverse` operations will have no effect. However it makes sense to expose the same logic. It may make sense to just explicitly expose `filter` and `limit` as options, but currently all operations are technically applied.
- Exposes the `completed_column_name` to the task-card component
  - This parameter already existed on the `task_progress_bar`, but was not exposed to the `task_card` component.
- Fixes a styling issue where the padding around the content in the task card component was asymmetrical.
  - Highlighted in #2507 

The `filter` parameter has the same limitations:
- Does not support `_index`, `_id`, `_first`, `_last` metadata properties, since these are only applied in a data-items context
  - ID can be accessed just via `@task_item.id`
  - index can be accessed via `@task_item.row_index`
- Currently some string comparisons may not work as expected. Additional `"` marks may be needed around strings when using them in a filter.

The task-card and task-progress-bar components contain logic for handling task completion status via two methods:
1. Using fields
    - This should be considered legacy, but is still in active use by some deployments 
2. Using a data list column value that is updated by dynamic data updates

Further changes could be made to the code of these components to clarify this and potentially further separate the logic for each system, but that could be done in a follow-up PR.

## Testing

Examine the linked templates below and test in-app to ensure they behave as expected.

## Follow-ups/TODO

- [ ] Make breakout issue for evaluation of `@task_item` expressions (e.g. handling of string comparisons is not the same as when using `@item`)

## Git Issues

Closes #2537, #2507

### [comp_task_card](https://docs.google.com/spreadsheets/d/1_RY8HbsATOAZU2EDXY5R9OOE_YBEbW0EjTP5zxYP2aA/edit?gid=569531329#gid=569531329)

Screenshot 2024-12-02 at 17 23 43

Screenshot 2024-11-20 at 17 28 18

Screenshot 2024-11-20 at 17 28 43

### [comp_task_progress_bar](https://docs.google.com/spreadsheets/d/1KenOF4C7BjVmg07zONWQgRygVWMDaXCOGLOSfC5vx7g/edit?gid=2015154474#gid=2015154474)
Screenshot 2024-12-02 at 17 34 59

Screenshot 2024-12-02 at 17 38 48


ChrisMarsh82:
@jfmcquade if @item is used for the rows with the task_group_data it will become confusing when embedded in a begin_data_items. What does @item then refer to? Could we use @task_item instead?

ChrisMarsh82:
Although @task_item now works if @item is used it will look at @task_item instead.

chrismclarke (approved):
Thanks @jfmcquade 

I've added a minor tidying commit with [9483df8](https://github.com/IDEMSInternational/open-app-builder/pull/2543/commits/9483df841d9faf6bae231afd9182d36e2c322d50), otherwise I'm happy with the changes to the item handling/processing.

I haven't look too deeply into the specific task card/progress-bar code as I think this is mostly just a temporary workaround until we have a v2 task system so happy as long as functional tests passing

jfmcquade:
@ChrisMarsh82 Are you happy for this to be merged if I add a follow-up issue regarding using the filter inside another `@item` namespace?

ChrisMarsh82:
@jfmcquade at the moment I can't get the filter to work. It is possibly something I am doing. Happy for this to be merged and can add a new issue for the filter if there is one

jfmcquade:
I've added #2643 as a follow-up to handle using the operations inside a parent data-items loop.

@ChrisMarsh82 – this means you won't need to be building from this branch after merge

PR #2549, opened 2024-11-22
Review: pr 2351 proposed changes

chrismclarke:
PR Checklist

## Description

Targets PR #2351 branch

- Add override hierarchy to appConfig service to make it easier to track overrides coming from different levels (defaults, deployment, skin and template)

- Remove override revert logic from skin service (now handled by app config)

- Remove additional template-app-config service to instead include effects within existing template-metadata service

- Update tests

## Review Notes
I checked the `feature_app_layout_footer` and appeared to be working as demonstrated, but would be good to double check still functions as expected. I also checked the debug skin select and looks to be functioning as I would expect, but would be good to double check reverting defaults after skin change still works as expected (skin revert code removed from service but should still be handled by app config service)

## Dev Notes
I purposefully haven't included a level for config changes triggered by actions, as I'm not sure this would be handled in the same way as other overrides or not. Deployment, skin and template-level overrides are all quite similar in the way that they have clearly defined lifespans (persistent, until skin change, until template change). It's not quite clear how setting config from an action would work (would it persist until next action changes? or until template/skin change?). In any case I think we only need to deal with that once we have a clear use-case.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/services/app-config/app-config.service.spec.ts
+++ b/src/app/shared/services/app-config/app-config.service.spec.ts
@@ -86,10 +86,19 @@ describe("AppConfigService", () => {
     expect(appConfigSetSpy).toHaveBeenCalledTimes(1);
   });
 
-  it("reverts to initial config values when template override removed", async () => {
+  it("reverts to initial config values when all overrides removed", async () => {
     service.setAppConfig({ APP_FOOTER_DEFAULTS: { templateName: "template_footer" } }, "template");
     expect(service.appConfig().APP_FOOTER_DEFAULTS.templateName).toEqual("template_footer");
     service.setAppConfig({}, "template");
     expect(service.appConfig().APP_FOOTER_DEFAULTS.templateName).toEqual("mock_footer");
+  });
+
+  it("reverts to lower order config values when higher order override removed", async () => {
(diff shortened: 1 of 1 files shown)


chrismclarke:
> Great, thanks @chrismclarke. I think I was anticipating a solution along these lines to be much more complicated, but it was good to discuss it on the call and this implementation is actually much easier to understand than the current handling in my opinion – the `AppConfigService.configOverrides` array is a reasonably intuitive representation of the current overrides (and it can be logged out for debugging). And nice that the app config service handles the override order in a way that those services which set the respective overrides do not need to be aware of.
> 
> The functionality does seem to be as expected when testing with the feature example sheets (in combination with different skins), and the tests are passing, nice to have some more (I added one more). I will merge this and continue to make your suggested changes on the targeted PR.

Great, yeah I was pleasantly surprised too, although in hindsight makes a lot more sense when trying to keep track of multiple layers of overrides, instead of focusing on how to roll-back the current level of changes just keep track of the separate levels and you can always just recreate from the base. Always a good sign when you can remove a net 30 lines of code whilst still adding 20 lines of tests... now just need to do that with the rest of the codebase 😅

PR #2550, opened 2024-11-23
feat: nav stacks

jfmcquade:
PR Checklist

## Description

New feature: nav stacks.
- Launched via actions:
  - `nav_stack: add`: create a new nav stack on top of any that are currently open
    - Additional parameters:
      - `template` (required): the name of the template to display inside the stack
      - `title` (optional): a title to display in the stack's integrated header
      - `show_close_button` (default: `true`): set to `false` to hide the close button that is displayed in the top right of the stack's integrated header
    - Example action authoring:
      - `click | nav_stack: open | template: feature_nav_stack_child, title: Stack Title, show_close_button: false;`
  - 

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/feature/stack/components/stack/stack.component.ts
+++ b/src/app/feature/stack/components/stack/stack.component.ts
@@ -1,0 +1,20 @@
+import { Component, Input, OnInit } from "@angular/core";
+import { ModalController } from "@ionic/angular";
+// import { TemplateBaseComponent } from '../../../shared/components/template/components/base';
+
+/** The logic for stack open/dismiss exists in the stack-navigation service */
+@Component({
+  selector: "plh-stack",
+  templateUrl: "./stack.component.html",
+  styleUrls: ["./stack.component.scss"],
+})
+export class StackComponent {
+  @Input() templatename: string;
+
+  constructor(private modalCtrl: ModalController) {}
+
+  public close() {
+    console.log("close");
+    this.modalCtrl.dismiss();
+  }
+}
--- a/src/app/feature/stack/components/stack/stack.component.ts
+++ b/src/app/feature/stack/components/stack/stack.component.ts
@@ -1,6 +1,12 @@
 import { Component, Input, OnInit } from "@angular/core";
 import { ModalController } from "@ionic/angular";
-// import { TemplateBaseComponent } from '../../../shared/components/template/components/base';
+
+export interface IStackConfig {
+  templateName: string;
+  title?: string;
+  icon?: string;
+  showCloseButton?: boolean;
+}
 
 /** The logic for stack open/dismiss exists in the stack-navigation service */
 @Component({
@@ -9,12 +15,25 @@ import { ModalController } from "@ionic/angular";
   styleUrls: ["./stack.component.scss"],
 })
 export class StackComponent {
-  @Input() templatename: string;
+  private _config: IStackConfig = {
+    templateName: "",
+    showCloseButton: true,
+  };
+  // Merge input values with defaults, ignoring any `undefined` values
+  @Input()
+  set config(config: IStackConfig) {
+    this._config = {
+      ...this._config,
+      ...Object.fromEntries(Object.entries(config).filter(([_, v]) => v !== undefined)),
+    };
+  }
+  get config(): IStackConfig {
+    return this._config;
+  }
 
   constructor(private modalCtrl: ModalController) {}
 
   public close() {
-    console.log("close");
     this.modalCtrl.dismiss();
   }
 }
--- a/src/app/feature/stack/stack.service.ts
+++ b/src/app/feature/stack/stack.service.ts
@@ -4,6 +4,8 @@ import { IStackConfig, StackComponent } from "./components/stack/stack.component
 import { SyncServiceBase } from "src/app/shared/services/syncService.base";
 import { TemplateActionRegistry } from "src/app/shared/components/template/services/instance/template-action.registry";
 
+interface StackModal extends HTMLIonModalElement {}
+
 @Injectable({
   providedIn: "root",
 })
@@ -54,25 +56,43 @@ export class StackService extends SyncServiceBase {
   }
 
   /**
-   * Create a new stack modal and push it to openStacks array.
+   * Create and present a new stack modal and push it to openStacks array.
    * Await and remove from openStacks array on dismiss
    */
   private async pushStack(stackConfig: IStackConfig) {
+    const modal = await this.createStackModal(stackConfig);
+    await this.presentAndTrackModal(modal);
+    await this.handleModalDismissal(modal);
+  }
+
+  /**
+   * Creates a new stack modal instance
+   */
+  private async createStackModal(config: IStackConfig) {
     const modal = await this.modalCtrl.create({
       component: StackComponent,
-      componentProps: { config: stackConfig },
-      // update to this styling must be done in global theme scss as the modal is injected dynamically into the dom
+      componentProps: { config },
+      // Styling must be done in global theme scss as the modal is injected dynamically into the dom
       cssClass: "stack-modal",
     });
-    modal.present();
-    this.openStacks.push(modal);
-    const stackIndex = this.getStackIndex(modal);
+    return modal as StackModal;
+  }
+
+  private async presentAndTrackModal(modal: StackModal) {
+    const stackIndex = this.openStacks.length;
     modal.setAttribute("data-stack-index", stackIndex.toString());
-    // Set property for dynamic CSS calculations
     modal.style.setProperty("--stack-index", stackIndex.toString());
+
+    await modal.present();
+    this.openStacks.push(modal);
+  }
+
+  private async handleModalDismissal(modal: StackModal): Promise<void> {
     await modal.onWillDismiss();
     // Stack modal has already been closed by component, so just remove from array
-    this.openStacks.splice(stackIndex, 1);
+    const index = this.getStackIndex(modal);
+    if (index === -1) return;
+    this.openStacks.splice(index, 1);
   }
 
   private closeTopStack() {
--- a/src/app/feature/stack/stack.service.ts
+++ b/src/app/feature/stack/stack.service.ts
@@ -1,0 +1,67 @@
+import { Injectable } from "@angular/core";
+import { ModalController } from "@ionic/angular";
+import { StackComponent } from "./components/stack/stack.component";
+import { SyncServiceBase } from "src/app/shared/services/syncService.base";
+import { TemplateActionRegistry } from "src/app/shared/components/template/services/instance/template-action.registry";
+
+@Injectable({
+  providedIn: "root",
+})
+export class StackService extends SyncServiceBase {
+  openStacks: HTMLIonModalElement[] = [];
+
+  constructor(
+    private modalCtrl: ModalController,
+    private templateActionRegistry: TemplateActionRegistry
+  ) {
+    super("stack");
+    this.init();
+  }
+
+  private init() {
+    this.registerTemplateActionHandlers();
+  }
+
+  private registerTemplateActionHandlers() {
+    this.templateActionRegistry.register({
+      nav_stack: async ({ args, params }) => {
+        const [actionId] = args;
+        const childActions = {
+          push: async () => {
+            const { template } = params;
+            this.navStackPush(template);
+          },
+          pop: async () => {
+            this.navStackPop();
+          },
+        };
+        if (!(actionId in childActions)) {
+          console.error("app_update does not have action", actionId);
+          return;
+        }
+        return childActions[actionId]();
+      },
+    });
+  }
+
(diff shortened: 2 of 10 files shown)


chrismclarke (changes requested):
Just made a quick first pass, a few notes

**Folder Structures**
I can understand the motivation for making this a standalone module given the service + component nature of the code, although I think up till now the `feature` folder has more been used for content that also includes pages/routing. Although for now I can't really think of a more sensible place to put it, so I'd say lets keep where it is but register the module within `src\app\deployment-features.module.ts` as I'll use that entry point later for optimising features/modules. 

It should also be possible to register action handlers as part of the module declaration, and avoid importing more service code into the `app.config.ts`

**Naming Conventions**
I like the use of `nav_stack` for flow types, although it appears most components/services have dropped the `nav` part from the name and just gone for `stack`. I think `stack` isn't very indicative so would suggest refactoring the keep the `nav` mention throughout. 

Similarly you've inherited the `plh-` prefix in some of the components which isn't necessary for this broad-scoped feature. Given everything is still pretty inconsistent I'd recommend picking one of the alternatives (`template-` or `tmpl-`) prefixes in use 

**Demo**
Thanks for adding the debug sheets. There was a typo in the demo that I've corrected, but otherwise does a good job of indicating the current functionality. 

**General Approach**
I think the service code used to track the stacks functions much better than query params, and should hopefully make the feature much easier to extend in the future. I'll add a few minor comments in-line but I think generally looks to be shaping up well.

**UI**
I think it would be good to reduce the top margin/offset so not as much screen space is lost when adding multiple layers. I think ideally we actually want to hide the visible text from the previous navbar as this would now be slightly confusing within the context of the nested navigation open
**Current - 56px offset**
image

**Proposed - 16px offset (?)**
image


**Tests**
I think the service code has enough going on that at a minimum we should include some basic tests for managing functionality (mostly just the nav stack array manipulation)

src/app/feature/stack/components/stack/stack.component.ts
```diff
@@ -0,0 +1,38 @@
...
+  private _config: IStackConfig = {
+    templateName: "",
+    showCloseButton: true,
+  };
+  // Merge input values with defaults, ignoring any `undefined` values
+  @Input()
+  set config(config: IStackConfig) {
+    this._config = {
+      ...this._config,
+      ...Object.fromEntries(Object.entries(config).filter(([_, v]) => v !== undefined)),
```
chrismclarke, line 26: nit(non-blocking): angular input signals offer a nicer pattern than defining separate getter/setter input handlers (and intermediate variables), as they expose a `transform` operation that can be used to merge defaults with passed values:
https://angular.dev/guide/components/inputs#input-transforms

src/app/feature/stack/stack.service.ts
```diff
@@ -0,0 +1,114 @@
...
+    this.registerTemplateActionHandlers();
+  }
+
+  private registerTemplateActionHandlers() {
+    this.templateActionRegistry.register({
+      nav_stack: async ({ args, params }) => {
+        const [actionId] = args;
+        const childActions = {
+          open: async () => {
+            const { template, title, show_close_button } = params;
```
chrismclarke, line 33: nit(non-blocking): might be better to add default fallback values here instead of later filtering out the `undefined` values from the service code. E.g. 
```ts
 const { template, title, show_close_button=true } = params;
```

src/theme/_layout.scss
```diff
@@ -63,6 +64,37 @@ ion-modal.template-popup-modal {
...
+    position: absolute;
+    bottom: 0px;
+  }
+
+  ion-toolbar {
+    --background: var(--ion-color-background);
+    --color: var(--ion-color-primary);
+  }
+}
+
```
chrismclarke, line 97: nit(blocking): this feels like a lot of component-specific CSS in what should be a more general layout override. I know the styles need to be hoisted out of the component to apply at the injected modal level, although I think it would be worth trying to introduce a different pattern to do so.

How about we define global style overrides within a separate scoped scss file, e.g. `src\app\feature\stack\components\stack\stack.component.global.scss`, and then use the existing global deployment scss files (or a similar file within) to reference. E.g. 
```scss
@use(..\..\..\app\feature\stack\components\stack\stack.component.global.scss
```

jfmcquade: Good idea, I've refactored along these lines in https://github.com/IDEMSInternational/open-app-builder/pull/2550/commits/32650368240a12e732ecdb76c9b1185ef959fc74

jfmcquade:
Thanks for the review, responses to your comments below

> **Folder Structures** I can understand the motivation for making this a standalone module given the service + component nature of the code, although I think up till now the `feature` folder has more been used for content that also includes pages/routing. Although for now I can't really think of a more sensible place to put it, so I'd say lets keep where it is but register the module within `src\app\deployment-features.module.ts` as I'll use that entry point later for optimising features/modules.
> 
> It should also be possible to register action handlers as part of the module declaration, and avoid importing more service code into the `app.config.ts`

Done in https://github.com/IDEMSInternational/open-app-builder/pull/2550/commits/2eac9909313b1929fb8b6502f1467167f6e1c128 and https://github.com/IDEMSInternational/open-app-builder/pull/2550/commits/72d167a505cc51c198ad2942ae63970a4da2c2a0

> **Naming Conventions** I like the use of `nav_stack` for flow types, although it appears most components/services have dropped the `nav` part from the name and just gone for `stack`. I think `stack` isn't very indicative so would suggest refactoring the keep the `nav` mention throughout.
> 
> Similarly you've inherited the `plh-` prefix in some of the components which isn't necessary for this broad-scoped feature. Given everything is still pretty inconsistent I'd recommend picking one of the alternatives (`template-` or `tmpl-`) prefixes in use

Done in https://github.com/IDEMSInternational/open-app-builder/pull/2550/commits/2eac9909313b1929fb8b6502f1467167f6e1c128

> **UI** I think it would be good to reduce the top margin/offset so not as much screen space is lost when adding multiple layers. I think ideally we actually want to hide the visible text from the previous navbar as this would now be slightly confusing within the context of the nested navigation open **Current - 56px offset** image
> 
> **Proposed - 16px offset (?)** image

I like this suggestion, I agree with your reasoning. However, I've kept the large offset for the first nav-stack, as I think it makes sense to keep some of the content below visible. This is in keeping with the [designs](https://www.figma.com/design/CYnvCNWHbgMdXlqBeBnc4T/ParentApp-for-Kids?node-id=3-4585&node-type=canvas&t=lLigZRUWIBZfe8nX-0).
Screenshot 2024-11-26 at 17 18 25

Screenshot 2024-11-26 at 17 14 56

> **Tests** I think the service code has enough going on that at a minimum we should include some basic tests for managing functionality (mostly just the nav stack array manipulation)

Added in https://github.com/IDEMSInternational/open-app-builder/pull/2550/commits/8745cc9de6997d25db55055066aba03e6f30d926

> **Misc** I'm seeing a lot of console warnings related to accessibility, not sure if this is something we have control over or not...? image

Thanks for flagging, this should be resolved by https://github.com/IDEMSInternational/open-app-builder/pull/2550/commits/75d4c50eb4460ba9866df4320e4ecde3a843943f

chrismclarke (approved):
Thanks for all the updates @jfmcquade , the changes are looking good.

I've added a few commits to this branch, but otherwise I think should be good to go if you're happy with them.

8bdf7ee
In keeping with the dynamic data updates, actions have been separated to their own file. I've also updated the type definitions to allow actions to more easily pass the params type expected

3dda7af
Reorganises the service code to keep public methods together before private (for clarity). It also removes the async `handleModalDismiss` method and instead just binds the event listener to the modal as its created and without await (as otherwise the `pushNavStack` method will never actually resolve until after the modal is dismissed).  

c56db46
A few improvements to the tests. I appreciate the lengths you went to both in terms of covering all the main functionality as well as finding a way to interact with multiple levels of modal generation (controller + modal). The main change is to add tests that prioritise the public service methods, given that these are the ones called by actions, and removing more of the intermediate private calls. 

I also refactored the mock modal generation code to a function so that it could be called multiple times when creating distinct modals, included mock methods for set/get attributes, and added a more realistic binding for the `onWillDismiss` event (as previously the mock would resolve immediately once the method was triggered from frontend code, instead of returning a promise that resolves only when the modal's own dismiss event was triggered)

chrismclarke:
> UPDATE: Ah I've just seen this doesn't resolve the issue and the accessibility warning is still there. I can have another look tomorrow

No worries, I'm seeing it in quite a few other places in the code too so I'm guessing part of recent chrome updates to identify. So still good to resolve, but happy for it to be a follow-up or as required

chrismclarke:
@esmeetewinkel 
On the question of authoring, do the actions make sense to you? I.E.
```
nav_stack: open
nav_stack: close_top
nav_stack: close_all
```
I'm wondering if there should be a default `nav_stack: close`, although I don't know whether then the default behaviour would be to close all the stack or just the top...

src/app/feature/nav-stack/nav-stack.service.ts
```diff
@@ -0,0 +1,83 @@
...
+  public async pushNavStack(navStackConfig: INavStackConfig) {
+    const modal = await this.createNavStackModal(navStackConfig);
+    await this.presentAndTrackModal(modal);
+  }
+
+  public async closeAllNavStacks() {
+    // Close nav-stacks in reverse order
+    for (let index = this.openNavStacks.length - 1; index >= 0; index--) {
+      await this.closeNavStack(index);
+    }
```
chrismclarke, line 32: nit(non-blocking): I'm not sure if this level of UX is necessary/helpful... if closing all stacks I don't think we necessarily need to do them one at a time and in order, might be better to just use the modalCtrl to close all modals and reset the array manually (if not automated) - or map the items to a Promise.all so they can work in parallel

jfmcquade: I found that if I do, for example, 
```ts
  public async closeAllNavStacks() {
    await Promise.all(this.openNavStacks.map((_, index) => this.closeNavStack(index)));
  }
```
Then the test for `closeAllNavStacks()` fails, with the length of the `openNavStacksArray` still being 1 (even with an artificial `_wait` time added to the test. I'm not quite sure what this is due to, potentially just something in the test setup, but I haven't been able to resolve it with various async configurations, so this brute force approach seemed acceptable.

jfmcquade:
Great, thanks for all the updates, @chrismclarke.

I've added one more commit to rename a `stack` instance to `nav-stack` that slipped through my original regex net: https://github.com/IDEMSInternational/open-app-builder/pull/2550/commits/089f4aab8b62b1fab76696134f3d7967dddb77ba

> > UPDATE: Ah I've just seen this doesn't resolve the issue and the accessibility warning is still there. I can have another look tomorrow
> 
> No worries, I'm seeing it in quite a few other places in the code too so I'm guessing part of recent chrome updates to identify. So still good to resolve, but happy for it to be a follow-up issue if blocking on this PR
> 
> It may also be something for ionic to resolve (in which case we still need to move forward with the v8 update) `https://github.com/ionic-team/ionic-framework/issues/29773`

Ah yeh, it does appear from that thread that ionic should be working on it directly. I've removed my ineffective workaround (https://github.com/IDEMSInternational/open-app-builder/pull/2550/commits/cbd6a3a1f02add5084370d6ce8652ed5b78dbd4a). I'm happy to live with the warning for now, but good to be aware of it, and looking forward to making the ionic upgrade at some point soon.

src/app/feature/nav-stack/nav-stack.service.ts
chrismclarke: Ah, the problem is for those tests we were still manually adding the navStack modal to the array and calling dismiss, instead of programmatically generating them via the `pushNavStack` method (which also adds the event bindings for dismiss). I think part of the reason for this is to have direct access to the modal spy that would have been generated.

I've added [a50d10c](https://github.com/IDEMSInternational/open-app-builder/pull/2550/commits/a50d10c151de037d430cb3cd69580d28e8a62129) to fix this - now the `pushNavStack` method returns the created modal (to make it easier to keep reference to it), and changed the service code to first push the nav stacks before dismissing. 

I've also updated the promise to call the specific modal `dismiss` function instead of looking up its index (the `openNavStacks` already store the full modal so no need to retrieve). Look to be working for me now, so if you're happy I'd say good to go pending any input from @esmeetewinkel

esmeetewinkel:
> On the question of authoring, do the actions make sense to you? I.E.
> 
> ```
> nav_stack: open
> nav_stack: close_top
> nav_stack: close_all
> ```
> 
> I'm wondering if there should be a default `nav_stack: close`, although I don't know whether then the default behaviour would be to close all the stack or just the top...

I agree, not clear what `close` should do so better to be specific with `close_top` and `close_all`

esmeetewinkel (approved):
Functional test passed (noted that browser back-button navigation is on the future to-dos)

PR #2551, opened 2024-11-25
hotfix: plh components loading from external package

jfmcquade:
PR Checklist

## Description

Fixes an issue where the components defined in the PLH components package (`packages/components/plh`) were not available in the app. On `master`, these components defined in packages cannot be instantiated from templates (no error is thrown, the component row is just not rendered), and do not appear in the list available on the `/component` page.

Based on when this issue was first observed (by @FaithDaka a couple of days ago), this is likely a regression introduced by #2520. I haven't been able to track down why those changes would cause the issue, but it seems to be likely due to some circular imports between our different module and component declaration files that is resolved by this PR.

I've marked is at a `hotfix` because it seems to move us in the wrong direction with regards to eventually being able to declare external packages containing components and other dependencies outside of the main codebase. I wasn't able to resolve the issue from the package level. However, as the packages were already not independent of the core code (we have to explicitly import and consume the `PLH_COMPONENT_MAPPING` object in the core code, for example), this doesn't add a new dependency.

@FaithDaka, you should be able to merge this PR's feature branch into your working branch in order to continue working on the PLH components' styling.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/components/demo/index.ts
+++ b/packages/components/demo/index.ts
@@ -1,16 +1,11 @@
-import { NgModule, Type } from "@angular/core";
+import { Type } from "@angular/core";
 
 import { DemoBasicComponent } from "./basic/basic.component";
 import type { ITemplateRowProps } from "src/app/shared/components/template/models";
-import { CommonModule } from "@angular/common";
-import { IonicModule } from "@ionic/angular";
 
-@NgModule({
-  imports: [CommonModule, IonicModule],
-  exports: [DemoBasicComponent],
-  declarations: [DemoBasicComponent],
-})
-export class DemoComponentsModule {}
+export { DemoBasicComponent };
+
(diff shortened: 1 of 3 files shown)


chrismclarke (approved):
Thanks @jfmcquade 

Yeah this makes sense to me, although even with previous implementation the component-specific pages would never show up for component module pages given the way that the component page template imports the hardcoded component list (so not a regression, likely never working). As for why components previously rendered and now don't - weird but don't think there's much benefit in diving too deeply into.

Longer term if we want component packages to have the same functionality as core components (specifically debug page entries) then yeah that would require a larger refactor to setup a system for registering component pages (a bit like we do with actionListRegistry). As for optimisation, this needs more work anyway so I'm not too worried about the changes (would likely become a lot easier if we refactor all packaged components to standalone).

But for now I think the change makes sense and good to get things working

PR #2556, opened 2024-11-26
chore: remove gdrive sync credentials

chrismclarke:
PR Checklist

## Description

Following #2544

 - remove dev credentials from gdrive-tools package
 - remove encrypted credentials from scripts/config package
 - remove scripts used to decrypt config credentials
 - update main readme

## Author Notes
In order to sync content from google drive a special `credentials.json` file is required to authorize the platform to act on behalf of users. This file is sensitive so we had previously kept it encrypted in the repo, and provided users with a `private.key` file used to decrypt (different from the deployment private keys), and instructed to run `yarn scripts config decrypt` command.

This PR completely removes the encrypted config (google service terms prohibits doing this), so now instead whenever onboarding a new developer they will need to be provided with the `credentials.json` file directly (instead of private key used to decrypt). This file will need to be re-shared any time the credentials are revoked and recreated (shouldn't be too often)

Existing users should be able to continue syncing with their previously decrypted credentials (local copy remains), unless it's decided to replace these (would require notifying all developers)

## Dev Notes
I don't think the dev credentials were ever actually in use (platform seems to default to credentials in the scripts/config workspace). Once merged should be able to safely remove the dev credentials from cloud console (believe its in the `plh-teens-app1` project). We may also want to consider rotating the other credentials

## Git Issues

Closes #

## Screenshots/Videos
**Example - error message observed if credentials are not available**
image


chrismclarke:
(updating to also remove config scripts credentials)

chrismclarke:
@jfmcquade - going to merge here as I can see Ian has already emailed out new credential instructions so would be good to have the readme updated and decrypt workflow removed

jfmcquade (approved):
Makes sense and looks good

PR #2558, opened 2024-11-26
Kuwait Theme: Progress Wheel

### LINKED ISSUE #2538: [FEATURE] Kuwait Theme - Progress Wheel Component
**What?**
Create a progress wheel component according to the Kuwait design theme.

**Why?**
The wheel design upgrades the design from the current progress bars and shows the percentage completed.

**How?**
- [x] Create a progress wheel component
- [ ] Test the component with data from the component sheet

**Other details**

FaithDaka:
PR Checklist

## Description

Adds a progress wheel which displays task completion progress as a percentage.

- New variant for `task_progress_bar` component: `wheel`.
  - As this component didn't previously have variants, the default variant has been named `bar`
- New parameter for `task_progress_bar` component: `wheel_title`
  - The title string to display 

## Notes

The `wheel` variant is available to all themes, not just the `plh_kids_kw` theme, however it has been optimised for this theme. Further optimisation (e.g. exposing colour parameters as theme-level variables) will likely be required if we wish to use this component in other deployments.

## Git Issues

Closes #2538

## Screenshots

[comp_task_progress_bar](https://docs.google.com/spreadsheets/d/1KenOF4C7BjVmg07zONWQgRygVWMDaXCOGLOSfC5vx7g/edit?gid=2015154474#gid=2015154474) template:

| Design    | New variant |
|---------------|----------------|
|     | Kuwait theme: Screenshot 2024-12-03 at 11 49 24 | 
| | Default theme: Screenshot 2024-12-03 at 11 48 30 |


esmeetewinkel (commented):
Looking great! I was wondering if it would make sense to add the gradient design that's in the Mexico app in this PR too, or that that will be a follow-up PR (I don't mind either way). 

image

jfmcquade (changes requested):
This looks really nice, and the logic is nice and straightforward.

Currently, you've exposed the `task_total` and `task_progress` parameters to set the completion percentage of the component. This would mean that an author would have to add their own logic for evaluating which tasks have been completed and pass that to the component. Instead, the component should be able to take two parameters and calculate the completion percentage itself: 
1. `data_list` – the name of a data list containing rows that relate to tasks that can be completed
2. `completed_column` – the name of the column within that data list that stores the completed status of the task row

From these values, the component can then look up the data list with that name, look at the completed column and compare the total number of tasks to the number that have been completed to calculate the total progress itself.

For this reason, I think it would make sense for this component to be a variant of the existing `task-progress-bar` component, which already contains this logic. The `task-progress-bar` component does contain some other logic to support a legacy system for handling tasks, but that can be ignored. 

In order to implement this wheel as a variant of the `task-progress-bar` component, you'll need to do the following:
- Configure the `task-progress-bar` component to have variants, e.g. `bar` and `wheel`, where `bar` is default.
  - `src/app/shared/components/template/components/audio/audio.component.ts` could be a good example of how to implement variants with a default
  - Likewise, the `audio.component.html` template can be used as an example – you'll want to wrap most of the existing template code of `task-progress-bar.component.html` in an `@if(params.variant.includes("bar"))` block, and add your code from `progress-wheel.component.html` in another `@if()` block underneath.
- Move most of your component logic into `task-progress-bar.component.ts` and refactor to use the existing properties
  - i.e. `subtasksCompleted` and `subtasksTotal` can be used to calculate the completion percentage in your `getStrokeOffset()` method

FaithDaka:
Thanks @jfmcquade 
The changes have been made and PR description updated.

src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
```diff
@@ -85,6 +89,9 @@ export class TmplTaskProgressBarComponent
   standalone: boolean = false;
   useDynamicData: boolean;
   private dataQuery$: Subscription;
+  radius = 16; // Radius of the circle
```
jfmcquade, line 92: Could be good to add a comment above these three properties to indicate that they relate to the "wheel" variant

src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.ts
```diff
@@ -147,6 +158,14 @@ export class TmplTaskProgressBarComponent
     return (this.subtasksCompleted / this.subtasksTotal) * 100;
   }
 
+  getStrokeOffset(): number {
```
jfmcquade, line 161: Again, would be good to have a comment here (JSDoc style) to indicate that this method relates to the "wheel" variant. It should also be explicitly tagged as `public`.

jfmcquade (changes requested):
Thanks for refactoring, I think the logic fits in pretty well to the task progress bar component. I really like the animation when updating the progress on the debug template.

I have a couple of further comments (plus some minor suggestions inline):
1. I think that the component styles should be moved to `src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.scss`, out of the theme-level overrides (this will also allow you to merge master in without merge conflicts). It's a grey area, but I think it makes sense to expose that variant to other deployments, even though the styling is quite theme-specific for now.
2. On initial load, the component shows briefly with "10000%" completion and the animation glitches. I think you'll need to add a timeout to only display the progress once it has been calculated. See demo:

https://github.com/user-attachments/assets/b8facd63-3bee-4183-aa60-e6136b7b1cda

FaithDaka:
> Thanks for refactoring, I think the logic fits in pretty well to the task progress bar component. I really like the animation when updating the progress on the debug template.
> 
> I have a couple of further comments (plus some minor suggestions inline):
> 
> 1. I think that the component styles should be moved to `src/app/shared/components/template/components/task-progress-bar/task-progress-bar.component.scss`, out of the theme-level overrides (this will also allow you to merge master in without merge conflicts). It's a grey area, but I think it makes sense to expose that variant to other deployments, even though the styling is quite theme-specific for now.
> 2. On initial load, the component shows briefly with "10000%" completion and the animation glitches. I think you'll need to add a timeout to only display the progress once it has been calculated. See demo:
> 
>  progress.wheel.glitch.mov

@jfmcquade 
I've transferred the wheel styling to the component `scss` file and it will use an `ion-secondary` colour so that the colour changes per theme, otherwise everything else is specific to the Kuwait theme.

For the glitching, I've added the `getStrokeOffset` method to the `ngOnInit` and made 0 the default value while the percentage gets calculated. Let me know if the way it works now makes sense. The `timeout` function wasn't effective on my end.

jfmcquade (approved):
Thanks, @FaithDaka 

I've added https://github.com/IDEMSInternational/open-app-builder/pull/2558/commits/02d27798a90903173af4e5c8af81bb69ec7631bd with couple of minor changes:
- Now the progress wheel fills up from empty on initial load (previously the wheel would start full and then animate to its partially completed state, which looked odd to me)
- Minor code tidying

I think I'll be happy to merge this PR if @esmeetewinkel agrees that things are looking good

esmeetewinkel:
@FaithDaka Could the PR description be updated to reflect the latest? 

Currently the linked sheet doesn't demonstrate the task based functionality.

jfmcquade:
@FaithDaka I've updated the PR description to reflect the latest updates.

@esmeetewinkel this PR should be ready for functional testing

esmeetewinkel (approved):
Functional test passed. The animation for adding progress looks great! 

I removed the spreadsheet [component_plh_progress_wheel](https://docs.google.com/spreadsheets/d/1aXtJi0oCxsHISq6m-xwJi_R_BSpUj4wHQlO_g4LUjSY/edit?gid=2015154474#gid=2015154474) which I believe is now redundant

PR #2561, opened 2024-11-26
Hotfix: Module List Item getParams()

### LINKED ISSUE #2560: [BUG] PLH module list item not able to read is_locked from local variable
**Describe the bug**
Currently, the parameter `is_locked` on the `plh_module_list_item` component works when the string `true` or `false` is supplied directly in the parameter list. This should also work when this string comes from a local variable. 

**App version**
v0.17.0

**Debug Sheet(s)**
Compare the [web preview](http://localhost:4200/template/comp_plh_module_list_item?debugMode=true) and [sheet](https://docs.google.com/spreadsheets/d/1Tq8Ib_ZcJsABW9U9XbQxgnmtv_uayXk28DXoa94j1aE/edit?gid=569531329#gid=569531329) for the `comp_data_items` template

**Screenshots and videos**
image
image

FaithDaka:
PR Checklist

## Description

- Allows `is_locked` param to be read from a local variable.

Fixes #2560

### DIFF
--- a/packages/components/plh/plh-kids-kw/components/module-list-item/module-list-item.component.ts
+++ b/packages/components/plh/plh-kids-kw/components/module-list-item/module-list-item.component.ts
@@ -1,11 +1,7 @@
 import { Component, OnInit } from "@angular/core";
-import { FlowTypes } from "packages/data-models";
 import { TemplateBaseComponent } from "src/app/shared/components/template/components/base";
 import { TemplateTranslateService } from "src/app/shared/components/template/services/template-translate.service";
-import {
-  getBooleanParamFromTemplateRow,
-  getStringParamFromTemplateRow,
-} from "src/app/shared/utils";
+import { getParamFromTemplateRow, getStringParamFromTemplateRow } from "src/app/shared/utils";
 
 interface IModuleListItemParams {
   /* TEMPLATE PARAMETER: "module_image_asset". The image attached to the module */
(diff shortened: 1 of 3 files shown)


chrismclarke (changes requested):
I feel like you took the right approach the first time using `getBooleanParamFromTemplateRow`, but for whatever reason that method isn't doing a very good job of converting strings to the right boolean representation.

Not sure if this is something you're already aware of @jfmcquade, but instead of the workaround you've made @FaithDaka would you be able to experiment if still working if you update the underlying method to parse the string in a better way (we've already got other utilities for parsing boolean values). E.g.

Before
```ts
export function getBooleanParamFromTemplateRow(
// ...
  return params.hasOwnProperty(name) ? params[name] === "true" : _default;
}
```

After
```ts
export function getBooleanParamFromTemplateRow(
// ...
  return params.hasOwnProperty(name) ? parseBoolean(params[name]) : _default;
}
```

jfmcquade:
Thanks, @FaithDaka. I agree that your original approach was correct and that the issue lies with the underlying `getBooleanParamFromTemplateRow()` method. I'm surprised this hasn't come up previously.

I've made a new PR for clarity, that will close this PR and the associated issue: #2563. Your component-level workaround should no longer be required.

esmeetewinkel:
Closed by https://github.com/IDEMSInternational/open-app-builder/pull/2563

PR #2564, opened 2024-11-27
feat: inline header or footer, via 'sticky' display group param

jfmcquade:
PR Checklist

## Description

Allows for inline template authoring of elements that stick to the top or bottom of the page.

This is achieved with a new param for the display group:
- `sticky`
  - Possible values
    - `top`
      - The display group element will stick to the top of the main content (below any app header). Scrollable content will scroll behind it
    - `bottom`
      - The display group element will stick to the bottom of the main content (above any app header). Scrollable content will scroll behind it

Also introduces a new parameter for the `title` component:
- `variant: header`
  - This centres the text and adds padding so that the title takes up space to be suitably used as a header

This feature can also be used to set the header/footer inside a nav-stack (#2550). See below for demo. This PR adds a new parameter for the `nav_stack: open` action, which should be used in this case (to hide the default nav-stack header):
- `header: false` (default `true`)
  - Set to `false` to hide the default nav-stack header, e.g. when using an inline sticky header on the target template.
  - NB, the template header should then handle its own nav-stack close button if required.

## Usage notes
- These display groups must sit at the top and bottom of the template respectively (this is because a dummy element is created on the page in their place so that content doesn't disappear behind them unless scrolled).
- A sticky display group centres its content by default
- Currently no specific padding or margins are applied to the content in a sticky display group. The thought is that any such styling should be applied to the enclosed content itself.
- New components will be required to achieve certain visual use cases, for example the component added in #2546, and the blue module header in this image (from [these designs](https://www.figma.com/design/CYnvCNWHbgMdXlqBeBnc4T/ParentApp-for-Kids?node-id=3-4585&node-type=canvas&t=pmsntYi3xK3pYCmK-0)):
    Screenshot 2024-11-27 at 16 52 43

## Git Issues

Closes #

## Screenshots/Videos

[feat_display_group_sticky](https://docs.google.com/spreadsheets/d/1tY8sa8ocDl552gIUAP263zgl3JQVcIeS4foaKk7nlMk/edit?gid=1992624740#gid=1992624740)

Screenshot 2024-11-28 at 17 33 27

https://github.com/user-attachments/assets/d71bb08d-f03b-4064-ac04-e2e7adbcf2c0

Integration with nav-stacks:

https://github.com/user-attachments/assets/73f3dbe9-c19b-4012-bc61-8f18b422c890


esmeetewinkel (changes requested):
I tried this out on the `plh_kids_kw` deployment (with same theme selected) and found that some components don't disappear properly behind the sticky header.

[Screenity video - Nov 29, 2024.webm](https://github.com/user-attachments/assets/4b0c8913-6630-43a0-a8c7-51d333ab608e)

jfmcquade:
> I tried this out on the `plh_kids_kw` deployment (with same theme selected) and found that some components don't disappear properly behind the sticky header.

Thanks @esmeetewinkel, this should be fixed now as of https://github.com/IDEMSInternational/open-app-builder/pull/2564/commits/9fd016c468ed4c7693d2406882b1df00f3d23d35

Screenshot 2024-12-02 at 15 42 00

chrismclarke (changes requested):
I think the changes to the nav-stack and title components make a lot of sense and overall the functionality is really nice. 

The main issue I have is how it 3x's the amount of code in the display component, which tells me it really should be it's own component. I'm not sure how best to manage this given from an authoring perspective it probably does make sense to maintain a single `display_group` type, but possibly could be a child component rendered from the display component with some shared styles/classes. I'll see if I can draft a PR to refactor

https://github.com/IDEMSInternational/open-app-builder/pull/2600/files

src/app/shared/components/template/components/layout/display-group/display-group.component.ts
```diff
@@ -45,4 +73,38 @@ export class TmplDisplayGroupComponent extends TemplateBaseComponent implements
...
+        // In the case of a sticky header, the top padding/margin of the main app content and template container need to be accounted for,
+        // now that the display group sits at the very top of the content window outside of the main content and template container
+        if (this.params.sticky === "top") {
+          containerPadding = this.getTotalContainerTopPadding();
+        }
+        const entry = entries[0];
+        this.stickyHeight.set(entry.contentRect.height - containerPadding);
+      });
+
+      this.resizeObserver.observe(this.displayGroupWrapper.nativeElement);
```
chrismclarke, line 91: when calling `observer` there should also be some handling for when to `unobserve` to avoid hanging listeners/memory leaks. 
It might be a case that chrome auto-handles this when element destroyed, however I'm not sure if this would still be the case for chrome on android or the ios webview handler, so probably best to manually enforce
https://stackoverflow.com/questions/67581868/should-i-call-resizeobserver-unobserve-for-removed-elements

chrismclarke (approved):
Approved following work in #2600

chrismclarke:
I think this should now be good to go, given that the review block from @esmeetewinkel has also been resolved

Issue #2489, opened 2024-10-28
[BUG] Carousel component can't contain dynamic data list


esmeetewinkel:
**Describe the bug**
The carousel component can render content form a `begin_items` `end_items` loop, but not from a `begin_data_items` `end_data_items`. 

**App version**
v0.16.36

**Deployment Sheet(s)**
Want to use this structure in the PLH Kids TZ deployment in the Resource Centre ([activity_carousel](https://docs.google.com/spreadsheets/d/1BoC8Zide9G0WNwqkVA2VYAEsE47kSUrBeFUJ8TJI7f0/edit?gid=1139990492#gid=1139990492)) and in the Library ([lib_books](https://docs.google.com/spreadsheets/d/1Jt_JwuTkbSC79WXNhUjgPUFiewdRwRSKUvkZuUxwU3w/edit?gid=1491860017#gid=1491860017)).

When using dynamic data for this, we can immediately store the "favourite" and "accessed timestamps" on the relevant row of the dynamic data list.

**Screenshots and videos**
![image](https://github.com/user-attachments/assets/52a8b36b-6556-45d1-8026-23228f95d11b)

jfmcquade:
@chrismclarke, you've been looking at code in this area a lot recently, do you have an idea of how difficult this would be to fix?

As I understand it, this issue is due to the fact that rows within a standard `items` loop are processed at the template row level and are rendered at the same level as standard template rows in a sense, whereas rows within a `data-items` loop are further processed by a specific `data-items` component and are rendered inside a container associated with that component. The carousel logic is such that it renders a new "slide" for each of its child rows, and in the case of a `data-items` loop, the immediate child is the data items component itself, so the whole container gets put in a single slide. 

The same issue, where the `data-items` loop can't be used to render a list of component-specific elements, would apply to other components that loop over their child rows to render each one in a certain way (e.g. `accordion`, `display-group`, `animated-slides` etc.), I think we just haven't encountered these use cases yet. 

For that reason I think a fix would ideally involve a change at the `data-items` level rather than at the level of these parent components. However, I'm not sure how we could rework the `data-items` component to add its children to the DOM in a way that didn't wrap them in a single parent component element.

esmeetewinkel:
Same holds for `display_grid` component (works for `items` but with `data_items` it just shows the items in a single column).

![image](https://github.com/user-attachments/assets/1458711d-e602-4c7b-9def-9febf4076159)

[...]

PR #2568, opened 2024-11-29
refactor: data items logic separation

chrismclarke:
PR Checklist

## Description

A common incompatibility between `items` and `data_items` is the way in which `items` render a the list of generated child rows instead of itself, whilst `data_items` render a single container row which handles data subscriptions, rendering generated child rows nested one level deeper. This makes it impossible to interop with components that take items as children of the component (e.g. carousel), as data_items will always render a single child item and items multiple.

As a partial workaround, this PR refactors the data_items logic to a standalone service, so that individual components can use directly to render the output children as required. This is likely a temporary measure, but could form the basis of a future implementation where the template container handles all data_item subscriptions and injects the generated items automatically within the child components as required.

This is part of the work to a solution required for #2489

It is designed to be followed up by https://github.com/IDEMSInternational/open-app-builder/pull/2570 which specifically handles the `carousel` component (although will need a little bit of reworking following some additional changes made to this pr)

## Author Notes
Previously the [comp_data_items](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892) sheet had an example of setting multiple items from within a loop via the `set_items` action (and filtering the data_items list to only show a single entry). This example has been removed and the functionality marked as deprecated, as it is better to just call `set_data` from outside of the item loop

## Review Notes
A small number of changes have been made to core code however all examples should still be working from [comp_data_items](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892) 

Additional tests have been added and some old tests removed which were a bit hard to follow (likely copy-pasted from specific debug templates)

## Dev Notes
The code shows a large number of changes, however most of these are the result of copy/paste with minor edits. Where larger code issues identified for refactor, comments have been left for future follow-up issues

New signals have been added to the base component for the full `row` value, named `rowSignal()`, and the child `row.rows` as `rows()`. The `row` has been renamed as many existing components use the `row` name directly and so would require a larger refactor.  Similarly a few component did reference `rows` directly, but there weren't as many so these have been refactored.
The main motivation for this is to allow the `data_items` component to more easily react to incoming changes.

The pattern used to manage subscribing to data changes and preparing child rows for render is a slightly complicated as the input data list name and data pipe operators come in as signals, but require processing that includes async code. Signals don't easily support `async` (no such thing as `asyncComputed`), so instead the signal is converted into an observable (which has patterns for processing async) and then converted back into a signal. Whilst a little complex under the hood, it means in terms of the code no need to manually subscribe/unsubscribe from data and call specific processing functions on change

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade (approved):
Nice to have the logic extracted out of the component code and it makes sense as a step towards handling item subscriptions at a higher level.

I concur that the examples in [comp_data_items](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892) all seem to still function as expected.

Agreed that the observable/signal conversion is a bit hard to follow, but it seems like a good solution to handling `async` with the signal implementation.

esmeetewinkel (approved):
Functional test passed. 

> Previously the [comp_data_items](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892) sheet had an example of setting multiple items from within a loop via the set_items action (and filtering the data_items list to only show a single entry). This example has been removed and the functionality marked as deprecated, as it is better to just call set_data from outside of the item loop

Noted, don't think I had used this anywhere yet.

PR #2583, opened 2024-12-02
hotfix: processTemplateWithoutRender issue

jfmcquade:
PR Checklist

## Description

A hotfix for an issue where any templates processed without being rendered (e.g. startup templates, which are processed on app start without rendering), would throw an error. This is a regression issue introduced by #2531. See [this mattermost thread](https://chat.idems.international/all/pl/pgm7d7ccepfb3qottn76frboco) for an identification of the issue.

## Dev Notes

Error log on master, for `plh_kids_tz` deployment:

Screenshot 2024-12-02 at 14 04 05

The issue arises when we try to instantiate a template container component programmatically, outside of Angular's component lifecycle, as we do in the `TemplateProcessService.processTemplateWithoutRender()` method (which is called to handle lifecycle action templates). In that case, the `effect()` in the component's constructor, introduced in #2531, throws an error.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/services/instance/template-process.service.ts
+++ b/src/app/shared/components/template/services/instance/template-process.service.ts
@@ -57,7 +57,10 @@ export class TemplateProcessService extends SyncServiceBase {
     this.container = new TemplateContainerComponent(
       this.templateService,
       this.templateNavService,
-      this.injector
+      this.injector,
+      undefined,
+      undefined,
+      true
     );
     // this.container.name = this.container.name || this.templatename;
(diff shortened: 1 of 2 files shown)


chrismclarke (changes requested):
Interesting issue, didn't realise how angular signals interop with injection contexts but it makes sense. 

Nice workaround to provide a static option within the constructor, although I found a slightly simpler solution from the signal code as it contains options to manually specify the injector, which we happen to be passing anyways when creating standalone. 

I've opened up an alternate pr #2588 to do this, which will supercede this PR if you're happy with the change

PR #2588, opened 2024-12-02
fix: template container error alt

chrismclarke:
PR Checklist

## Description

Alt proposal for #2583, with additional fix to container input error logs

## Dev Notes
The `templatename` input has been changed from `input.required` to regular `input` as it's not set when calling headless. 

Ideally this could be reverted back in the future if we can decouple the render/processing logic from the container component to allow it to truly be called in a headless way (added TODO comment in the code). 

## Git Issues

Closes #2583

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/template-container.component.ts
+++ b/src/app/shared/components/template/template-container.component.ts
@@ -43,7 +43,7 @@ export class TemplateContainerComponent implements OnInit, OnDestroy {
   /** unique instance_name of template if created as a child of another template */
   @Input() name: string;
   /** flow_name of template for lookup */
-  templatename = input.required<string>();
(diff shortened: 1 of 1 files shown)


jfmcquade (approved):
Great, I agree that this is a nicer solution than #2583, I'm happy to approve and merge. 

I did wonder if removing `required` from the input signal might make the template name evaluate twice for each display component (once as `undefined`, then as the input value), but a debug log confirms that this isn't the case – which makes sense in terms of how inputs should be handled.

It would be quite nice to make use of the `input.required()` signal, but as you note, this would probably only be at the point of decoupling the processing logic from the container component – good to have the TODO in the code at least.

PR #2589, opened 2024-12-03
Kuwait Theme: Add Background Image to Display Group

### LINKED ISSUE #2579: [FEATURE] Add background image attribute to display group
**What?**
Additional property for the display group to have background images

**Why?**
This makes the component robust and usable for a larger use-cases

**How?**

- [ ] Add `background-image` attribute to `display_group` comp sheet
- [ ] Refactor component code to feature a background image
- [ ] Test with an example

**Other details**

FaithDaka:
PR Checklist

## Description
Adds a `background_image_asset` to `feature_display_group` component. [Link](https://docs.google.com/spreadsheets/d/1ARbNRGDer5vj9qSpRMZFrMkYifGkH3TLtDVp72YbaqU/edit?usp=sharing)
Also includes display group styling according to the Kuwait theme.

## Author Notes
New parameter list:
- `background_image_asset`: Adds a background image to the display group.
- `background_image_position`: Positions the background image. Default `top center`.
  - The possible values for this parameter are those of the CSS `background-position` property, see "Property Values" table [here](https://www.w3schools.com/cssref/pr_background-position.php). This means that, for example, the x and y positioning of the  background image can be set via, e.g., `background_image_position: right bottom`

## Further additions

By @jfmcquade:

1. New display group variant, `inline-padding`
    - This applies the app-level content padding to the display group content
    - This is required as otherwise when using the display group with `sticky` set to `top` or `bottom`, there is no left or right padding around the content, as demonstrated in [example_inline_header_footer](https://docs.google.com/spreadsheets/d/1D7BtGET428mBHMRIZbiDY_sCHPnAd1g7F44Jhs-3b1Y/edit?gid=1921555013#gid=1921555013):
      Screenshot 2024-12-06 at 18 34 05
    - I did consider alternatives to an explicit variant, e.g. always applying this padding when the display-group is sticky; or always applying this padding when there is a background image set. But these seemed too restrictive on those two scenarios, as we may want to use them without applying padding in this way.

2. New `style` param value: `background_solid`
    - Setting `style: background_solid` on the display group means that the This was previously the default when setting a display group to `sticky` (#2564), but we agreed that the stickiness should be independent of styles, so this style must now be used in order to prevent the display group from having a transparent background when setting `sticky`.

See examples in [feature_display_group](https://docs.google.com/spreadsheets/d/1ARbNRGDer5vj9qSpRMZFrMkYifGkH3TLtDVp72YbaqU/edit?gid=1231713195#gid=1231713195)

## Git Issues

Closes #2579

## Screenshots
| Design        |  Kuwait Style       |
|----------|-------------|
|  |  |
|  |  |


FaithDaka:
@jfmcquade I tested the concept of adding a `margin` to the text using the `style_list` column but given that it requires one to set two values (for top-bottom and left-right) then we don't get the desired effect. 
It will be helpful to discuss the need for a `spacer` component such that space can always be added between components.

jfmcquade:
> @jfmcquade I tested the concept of adding a `margin` to the text using the `style_list` column but given that it requires one to set two values (for top-bottom and left-right) then we don't get the desired effect. It will be helpful to discuss the need for a `spacer` component such that space can always be added between components.

Thanks for looking into it. Does it not work with `margin-top`? In theory we should be able to pass any valid CSS to the `style_list` column so I would have thought that setting `margin-top: 48px`, for example, could work

FaithDaka:
Oh, I was using this format: `margin: 50px 0px 0px 0px`. Using `margin-top` works fine @jfmcquade

jfmcquade (changes requested):
I think the background image options look good.

The changes to the `box_primary` etc. variants should be pulled into a separate PR I think. I'm not sure whether we want to overwrite the `box_*` variants for the KW theme or offer new variants. Presumably we're not planning on using the default `box_*` stylings and so overwriting them may make sense, perhaps @esmeetewinkel can confirm.

Would be good to have those changes broken out into a separate PR for that reason, but if we're sure we're happy to override then they can stay with this PR, but the title and description should be updated.

jfmcquade:
Hi @FaithDaka, I have broken out the display group styling for the `box_*` variants to another PR, see #2604.

I've also added some additional changes with https://github.com/IDEMSInternational/open-app-builder/pull/2589/commits/f7c6ecfef6e5c2c4c3be2b0e34157a8b085443cc to enable the styling of your designs to be achievable. I will update the PR description with what I've added.

jfmcquade (approved):

src/app/shared/components/template/components/layout/display-group/display-group.component.html
```diff
@@ -3,7 +3,13 @@
   @if (params.sticky === "top") {
     <div [style.height.px]="sticky.height()"></div>
   }
-  <tmpl-display-group-sticky [position]="params.sticky" #sticky>
+  <tmpl-display-group-sticky
+    [position]="params.sticky"
+    [style.--sticky-background]="
+      params.stickyBackground === 'transparent' ? 'transparent' : 'var(--ion-background-color)'
+    "
```
chrismclarke, line 10: nit(blocking): assigning styles in this way looks quite messy. Given the sticky host still renders the child templated content then the child content should be the place to handle this. I'm not sure if maybe I've missed a use case here?

chrismclarke (changes requested):
Definitely a nice feature to have and good way to address the design need for wavy backgrounds (assume can just use an SVG background image now)

See on minor (blocking) issue inline, I'll take a quick look to see if I can propose changes to fix.

Appreciate the example sheets as well, @jfmcquade - it would be good to merge with other display_group examples into the `feature_display_group` sheet

chrismclarke:
I've tested removing all references to `stickyBackground` and `--sticky-background` and I can't see any difference on my side, so guessing it might be an issue related to theming. @jfmcquade or @FaithDaka would one of you be able to check  

I also noticed a minor issue where the background image would be applied to all display groups even if not specified (with a URL to an undefined image). I've updated with the above changes as part of #2612)

jfmcquade:
> I also noticed a minor issue where the background image would be applied to all display groups even if not specified (with a URL to an undefined image). I've updated with the above changes as part of #2612)

Thanks, makes sense

> I've tested removing all references to `stickyBackground` and `--sticky-background` and I can't see any difference on my side, so guessing it might be an issue related to theming. @jfmcquade or @FaithDaka would one of you be able to check

I've added a comment to #2612 as to why something along those lines is needed, and have pushed an alternative implementation, let me know what you think.

@FaithDaka I've noticed an additional issue: on wider screen sizes, the limits of the image are clearly visible. Whilst we don't make an effort to support desktop screen dimensions, we should at least be supporting tablets and larger phones. I can't think of a simple way to resolve this, can you?
Screenshot 2024-12-10 at 12 41 54

chrismclarke:
> @FaithDaka I've noticed an additional issue: on wider screen sizes, the limits of the image are clearly visible. Whilst we don't make an effort to support desktop screen dimensions, we should at least be supporting tablets and larger phones. I can't think of a simple way to resolve this, can you? 

I think there's a couple levels of issue, some easier to fix than others.
For the immediate alignment issue, it will be partially solved using the `backround_image_position` author variable exposed as the header can be fixed `right top;` which I think looks better

image

But then you have the issue of whether the sticky group should max the max content width that the main templated content does or not (as currently all sticky content 100% width). 

image

I think the most sensible default behaviour would be to keep the sticky header as full width (so background image can be relative to entire page), but all child components set with the same `max-width: var(--content-max-width);` otherwise assigned at the template-container level. End result would look like

image

I'm not sure if it would be best to include that as part of this pr or a separate one

FaithDaka:
@jfmcquade 
From @chrismclarke suggestion it makes sense to change the html layout such that the outer `div` with the `background-asset` spreads 100% while the inner content maintains a `max-width`. But also the outer `div` would need a `max-width` otherwise if the web-view is projected using a large TV then the background image would still spread to fill the screen and fail to maintain its appearance.

chrismclarke (approved):
Changes from #2612 now included which I think address the only issues I had, @FaithDaka you might be interested to note the changes particularly how to assign a solid background colour when using a combination of text and image backgrounds. 

A couple specific sticky header/footers with background images have also now been included in the general [feature_display_group](https://docs.google.com/spreadsheets/d/1ARbNRGDer5vj9qSpRMZFrMkYifGkH3TLtDVp72YbaqU/edit?gid=1231713195#gid=1231713195) sheet 

(sidenote - really all the sheets in this google sheet document should be split into standalone component sheets)

@jfmcquade - I've retagged you for final review just so that you can sign-off on the changes merged from #2612

esmeetewinkel (changes requested):
I'm struggling a little to understand from the PR description how the different parameters (are intended to) work. 

**Clarifications:**
1. I'm understanding that `background_image_position` can be one of top / center / bottom (vertical position), _but can also contain left / right (horizontal position)_. Default is top + horizontally centred. If this is correct can this be added tot he PR description?

2. The description of `style: background_solid` in the PR notes is cut off. This seems to be the default - I don't observe any difference when I do / don't include this parameter. I don't understand how to achieve a transparent background, as in the design:
image

**Bugs:**
1. When the `background_image_position` is set to `top right` the image doesn't remain glued to the top on narrow screens. 
   
   https://github.com/user-attachments/assets/98fb1fb2-20fd-41e4-9987-e0e9c58a88aa

2. The `background_image_position` left / right should flip in RTL and the background image should be flipped. 
   Current view in RTL:
   image

jfmcquade:
> **Clarifications:**
> 
> 1. I'm understanding that `background_image_position` can be one of top / center / bottom (vertical position), _but can also contain left / right (horizontal position)_. Default is top + horizontally centred. If this is correct can this be added tot he PR description?

Good point, I've clarified in the PR description:
- `background_image_position`: Positions the background image. Default `top center`
  - The possible values for this parameter are those of the CSS `background-position` property, see "Property Values" table [here](https://www.w3schools.com/cssref/pr_background-position.php). This means that, for example, the x and y positioning of the  background image can be set via, e.g., `background_image_position: right bottom`

> 2. The description of `style: background_solid` in the PR notes is cut off. This seems to be the default - I don't observe any difference when I do / don't include this parameter. I don't understand how to achieve a transparent background, as in the design:

Well spotted. I think in various discussions and sources of changes something got lost, and the default solid background colour wasn't removed properly. I've rectified this and you can see the difference between the sticky top display group with `style: background_solid` and the sticky bottom display group with the default background colour as transparent:

Screenshot 2025-01-03 at 10 52 51

Note that this does mean the default is changing for the sticky display group. If there are other places in the app that use a sticky display group, they may need to have `style: background_solid` set on them in order to preserve the behaviour of having a solid background.

jfmcquade:
> **Bugs:**
> 
> 1. When the `background_image_position` is set to `top right` the image doesn't remain glued to the top on narrow screens.

The display group at the top of the [debug template](https://docs.google.com/spreadsheets/d/1ARbNRGDer5vj9qSpRMZFrMkYifGkH3TLtDVp72YbaqU/edit?gid=1231713195#gid=1231713195) previously just had `background_image_position: right`. Updating this to `background_image_position: right top` resolves the issue:

Screenshot 2025-01-03 at 10 59 18

> 2. The `background_image_position` left / right should flip in RTL and the background image should be flipped.

This is proving quite difficult to implement, because of the way the CSS background images work. I think it would make sense for me to merge this PR as-is if possible, then I'll work on this as a follow-up.

> 3. On some screen ratios (e.g. Moto G4 and Samsung Galaxy S8+) I'm seeing a faint blue line underneath the image

I believe this is an issue due to the fact that source image is not a true svg, but rather a png image embedded in an svg. @FaithDaka, can you supply the image file as a true vector svg? Or if that isn't available, could you provide the asset as a png? This doesn't block this pR being merged, but would be good to have for the production deployment.

esmeetewinkel:
> I believe this is an issue due to the fact that source image is not a true svg, but rather a png image embedded in an svg. @FaithDaka, can you supply the image file as a true vector svg? Or if that isn't available, could you provide the asset as a png? This doesn't block this pR being merged, but would be good to have for the production deployment.

No need for @FaithDaka to provide these images, as they have been changed and I'm getting the new versions from the illustrator directly.

esmeetewinkel (approved):

PR #2594, opened 2024-12-03
feat: dismiss top nav-stack on back (alt)

jfmcquade:
PR Checklist

## Description

Alternative implementation of #2569.

The top nav-stack will be dismissed when navigating back via the hardware/browser back button. Back navigation should continue from the base template as expected.

## Testing

From the [feat_nav_stack](https://docs.google.com/spreadsheets/d/1qZNmJre8eqVqQpBO8w3iMHJC84W0QOAnryRLL5tQKZg/edit?gid=569531329#gid=569531329) template, functionally test opening and closing nav stacks, along with navigating back, in every combination.

## Dev notes

Because the "back"(/"popstate") event cannot have its default behaviour prevented, handling a back navigation like this is somewhat difficult. Whilst the history can be manipulated to a degree, the difficulty is in navigating in such a way that we can continue, as expected, once we get to the top-level template that initially launched the first nav-stack. 

Initially (in #2569), I implemented behaviour to navigate back and then forward again upon each nav-stack dismissal. However that led to some screen flickering and didn't seem ideal. This current PR achieves the same result by adding and removing history entries upon presenting and dismissing each modal respectively.

The compromise here is that the `closeAllNavStacks()` method did not effectively remove all the nav-stack related history states: when the modals are closed in parallel, the dismiss handlers do not have a chance to remove their respective history states. I also tried awaiting the promise of all of them closing in parallel and then removing the number of history states required to return to the state before any nav-stacks were opened, but this caused the browser to throttle (via a console warning) and the page to crash (including when setting a timeout of `100`ms between each `history.back()`. So I've reverted to closing the nav-stacks in sequence, as in [this comment discussion](https://github.com/IDEMSInternational/open-app-builder/pull/2550#discussion_r1859354943). 

### Alternative approaches/compromise
- Close all nav-stacks when the user performs a back action
  - Would still require the custom back handler introduced by this PR, but would simplify the logic and allow all nav-stacks to close simultaneously
- Do not override default "back" behaviour (navigating to previous page relative to the top-level template that launched the first nav-stack), but just dismiss any open nav-stacks on navigation

## Git Issues

Closes #

## Screenshots/Videos

https://github.com/user-attachments/assets/3be5498d-2b56-459f-9c5d-459ef8399dfb


esmeetewinkel:
Please re-add me when code is signed off so that I can functionally test

jfmcquade:
@chrismclarke It would be good if you could review, I'm not sure whether my implementation is hacky or not. If we're not happy with this approach, I think we'll want to put in something in the short term so that nav-stacks dismiss when nav actions are triggered (without trying to intercept the top-level template navigation), as currently the navigation happens but the nav-stacks remain open

chrismclarke:
@jfmcquade - took a first pass at this and can follow what's going on and it does seem to work for the most part (with the limitations mentioned), although I'm wondering whether this might start to get a lot more complicated if we also were to add navigation within a stack. I think any time you start to interfere with browser history you're likely to create code that is somewhat difficult to maintain. 

I think my longer term suggestion would be for nested stacks to actually use routing so that we can still use the native back button functionality (once we've decided how navigation within a stack would work alongside navigation of multiple stacks), and in the short term actually just focusing on intercepting android device back button when stacks are open to use them to close top stack (and ignore any routing for now).

Does that sound like a sensible approach to you? If so I think I'd be tempted to take the half of this PR that tidies up code and nav stack tracking, and just add a small method to intercept browser back button when stacks are open.

jfmcquade:
> @jfmcquade - took a first pass at this and can follow what's going on and it does seem to work for the most part (with the limitations mentioned), although I'm wondering whether this might start to get a lot more complicated if we also were to add navigation within a stack. I think any time you start to interfere with browser history you're likely to create code that is somewhat difficult to maintain.

That makes sense, as I was working on this I couldn't help but feel like I was replicating some of the more confusing pop-up code that I was trying to work on with #2329.

> I think my longer term suggestion would be for nested stacks to actually use routing so that we can still use the native back button functionality (once we've decided how navigation within a stack would work alongside navigation of multiple stacks), and in the short term actually just focusing on intercepting android device back button when stacks are open to use them to close top stack (and ignore any routing for now).

I think this is a good suggestion, my only caveat is that intercepting the device back button to close the top nav-stack isn't trivial, as the default back button behaviour, which cannot be prevented, needs to be handled. 

I've opened #2655 which I think represents a satisfactory short term solution, as you've proposed.

PR #2597, opened 2024-12-04
Fix: parameter list dynamic keys

chrismclarke:
PR Checklist

## Description

Fix issue raised in #2513 where dynamic references cannot be used as part of action_list parameters.

Includes additional fix to address issue where previously setting a dynamic field could break loading the same data_list on next template visit

## Review Notes
[feature_data_actions](https://docs.google.com/spreadsheets/d/1nREviZhw2HG1YQy6KA1S9PU4Q36RRVHdwSYf3TMbduk/edit?gid=1992624740#gid=1992624740) updated to include an explicit example of setting a variable field to a specific value
```
click | set_data | 

_list_id: feat_data_actions_list,
@local.number_field: 20
```

The fix also includes a minor update to the parser which should highlight diffs where previously this case was/wasn't addressed properly. I attempted making before/after content prs for the debug repo however the diff isn't showing correctly. I expect it's because this case doesn't currently exist in the debug repo and so git doesn't know how to show the empty PR. Could be worth re-checking after debug content repo updated
https://github.com/IDEMSInternational/app-debug-content/pull/124

## Author Notes
Previously trying to set something like `@local.some_field: some value` would store the update as-is (with a field-name prefixed by `@local`. This could still technically save, however when reloading the data_list it would try to infer the schema from saved data, and throw an error because it detected data fields that started with a forbidden `@` character.

This update also includes a fix to always infer the data_list schema from the initial sheet data (including any `@schema` overrides specified), so it should not cause further issue.

## Dev Notes
The fix for dynamic parameter list needed a fix at both the parser and template level as both were designs to only process an object's values and not the keys. The template fix is quite hacky, but I couldn't find a simpler workaround within the existing system (overdue for an overhaul). The code for the changes can be seen in [b462f16](https://github.com/IDEMSInternational/open-app-builder/pull/2597/commits/b462f168f03c80c7ffa387b09af0b6fac8d96d65)

The fix for schema inference is in [8026f4b](https://github.com/IDEMSInternational/open-app-builder/pull/2597/commits/8026f4b3342c1eb45da16b7498c0207eb3606fe3), it was included in this branch in case any other users may have accidentally put their local data_lists into a bad state. It also includes handling the data_list `_metadata` updated in #2529 but not previously included in the frontend code (lingering TODO item)

There's also a few minor comment and type updates just to clarify some of the existing code when looking at it, and a handful of commits at the end to add a basic test case and fix up some broken tests that I came across

## Git Issues

Closes #

## Screenshots/Videos

[Screenity video - Dec 3, 2024.webm](https://github.com/user-attachments/assets/595a2958-8998-48fd-952e-11532dc903f4)


jfmcquade (approved):
Looks good. Very helpful to have the commits separated out and clearly explained.

The somewhat hacky solution to evaluating objects' keys at the template level makes sense, and the relevant functionality on linked feature template is working as expected.

esmeetewinkel (approved):
Functional test passed.

PR #2600, opened 2024-12-05
refactor: display group sticky component

chrismclarke:
PR Checklist

## Description

Targets branch of #2564 

Rework to move all sticky display logic to separate component

## Review Notes
To confirm if still functions identically to target branch pr

## Dev Notes
This creates a child `display-group-sticky` component that is rendered within the main `display-group` component, with the sole purpose of tracking content height for use in the included padding offsets.

Most of the code was copy-pasted, although I wasn't sure whether it still needed all the logic to calculate additional padding, given that it is now still rendered inside the main display-group component (I think maybe not). I also wasn't sure if the CSS z-index that references nav-stacks is required (commented out, but should be removed if not needed).

Although it declares a new `display_group_sticky` component, that is only to support the current optimisation system (only named templating components can be marked as implicit dependencies)

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/feature/nav-stack/components/nav-stack/nav-stack.component.global.scss
+++ b/src/app/feature/nav-stack/components/nav-stack/nav-stack.component.global.scss
@@ -8,8 +8,6 @@ ion-modal.nav-stack-modal {
   --box-shadow: 0 10px 15px -3px rgb(0 0 0 / 0.1), 0 4px 6px -4px rgb(0 0 0 / 0.1);
   --backdrop-opacity: 0.5;
 
-  z-index: calc(10 + var(--nav-stack-index));
-
   // Override defalut ion-modal behaviour of hiding backdrop for subsequent modals after first
   &:not([data-nav-stack-index="0"]) {
     --backdrop-opacity: 0.2 !important;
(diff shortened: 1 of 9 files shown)


jfmcquade:
> I wasn't sure whether it still needed all the logic to calculate additional padding, given that it is now still rendered inside the main display-group component (I think maybe not). I also wasn't sure if the CSS z-index that references nav-stacks is required (commented out, but should be removed if not needed).

While the more fiddly logic template container margin can be removed (thankfully), I do think we still need to calculate and account for the default padding around the whole `ion-content` element. I have reinstated with https://github.com/IDEMSInternational/open-app-builder/pull/2600/commits/8f5be5fd0796596527ebb20242329d07bd958c08

Here's why I think it's needed:

We apply padding around the `ion-content` container like this:

Screenshot 2024-12-05 at 11 57 52

This is applied around our template content, including the inline placeholder for the sticky header, like this:

Screenshot 2024-12-05 at 11 53 40

In order for the spacing between the sticky display group and the rest of the template content to match how it would be if the display group was not sticky, we want to effectively bump up the main content to account for that top padding, like this:

Screenshot 2024-12-05 at 12 12 37

One way to do that is to reduce the height of the placeholder component by the size of that top padding, which is what https://github.com/IDEMSInternational/open-app-builder/pull/2600/commits/8f5be5fd0796596527ebb20242329d07bd958c08 accomplishes:

Screenshot 2024-12-05 at 12 24 58

An argument could be made that this isn't necessary: without it there is effectively padding below the sticky header before the main content starts that matches the padding of the main `ion-content`, which could be seen as desirable. But I think this padding looks wrong and the gap looks overly large (especially when in a `nav-stack`), and it makes sense to me that the spacing should match how the content would display if the display group was not sticky.

jfmcquade (approved):

chrismclarke:
Thanks for the clear explanation, makes a lot of sense given we're using `fixed` positioning and not `absolute` or `sticky`, so happy with the changes. I'll get this merged in and hopefully the rest can follow

PR #2601, opened 2024-12-05
Kuwait Theme: Completion modal

### LINKED ISSUE #2540: [FEATURE] Kuwait Theme - Completion Modal
**What?**
Create a module completion component according to the Kuwait theme design.

**Why?**
This modal congratulates a user after they have completed a module and links to a moment of play activity.

**How?**
- [x] Create a module completion component. This component activates after a user clicks the trophy icon and completes the module.
- [x] Test the component with data from the component sheet

**Other details**

FaithDaka:
PR Checklist

## Description
Styling added for the completion modal page `plh_completion_modal`. [Link](https://docs.google.com/spreadsheets/d/1Uw0QcLJfb2PzvKLSdxXdJojx-zvbUfw12dl-Sy6KrC8/edit?usp=sharing)

## Git Issues

Closes #2540

### DIFF
--- a/packages/components/plh/index.ts
+++ b/packages/components/plh/index.ts
@@ -1,19 +1,27 @@
 import { PlhParentPointCounterComponent } from "./parent-point-counter/parent-point-counter.component";
 import { PlhParentPointBoxComponent } from "./parent-point-box/parent-point-box.component";
 import { PlhModuleListItemComponent } from "./plh-kids-kw/components/module-list-item/module-list-item.component";
+import { PlhCompletionModalComponent } from "./plh-kids-kw/components/completion-modal/completion-modal.component";
 
-export { PlhParentPointCounterComponent, PlhParentPointBoxComponent, PlhModuleListItemComponent };
+export {
+  PlhParentPointCounterComponent,
+  PlhParentPointBoxComponent,
+  PlhModuleListItemComponent,
+  PlhCompletionModalComponent,
+};
 
 export const PLH_COMPONENTS = [
   PlhParentPointCounterComponent,
   PlhParentPointBoxComponent,
   PlhModuleListItemComponent,
+  PlhCompletionModalComponent,
 ];
 
 export const PLH_COMPONENT_MAPPING = {
   parent_point_counter: PlhParentPointCounterComponent,
   parent_point_box: PlhParentPointBoxComponent,
   plh_module_list_item: PlhModuleListItemComponent,
+  plh_completion_modal: PlhCompletionModalComponent,
 };
 
 export type PLHComponentName = keyof typeof PLH_COMPONENT_MAPPING;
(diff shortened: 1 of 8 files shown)


FaithDaka:
The modal "image" will be a `.gif`. It was mentioned that we could use Lottie files?
I have a file available for the gift box animation
[Gift-box-animation.json](https://github.com/user-attachments/files/18020681/New-file.json)

jfmcquade (changes requested):
I think ultimately this should be a display group variant rather than a standalone component, but I appreciate that the styling of the background image is quite custom, so I think I'm happy with this being a standalone `plh` component for now.

In line with @esmeetewinkel's comment [here](https://github.com/IDEMSInternational/open-app-builder/issues/2540#issuecomment-2518167943), you're right to not include any modal logic in the pop-up as it will just be displayed within a standard pop-up. Possibly the component name could be changed from "modal", but as it's custom for a specific deployment I don't think that's too important. I do think that it will have to be displayed in a full-screen pop-up, as our non-fullscreen pop-up does not support custom styling like (we would want to apply a gradient going to the very edges of the container and add the background "confetti" image with an offset). So I think displaying this component, with its own rounded corners, within a fullscreen pop-up is probably the best option for now.

I would suggest making the following changes:
- The height of the component should be the full container height rather than a set value, so that it always fills the screen but doesn't 
- The `gift_image_asset` doesn't need to be passed into the component params, it can just be authored in the child template rows, like the title and button
  - This will also enable the use of a lottie animation, via the existing `lottie_animation` component.
  
So essentially this component should be seen as a container with a background gradient applied and background image at the top with an offset, that spans the height of the screen and vertically centres its child rows.

FaithDaka:
@jfmcquade This is the outcome of `height:100%` for the outer div


I think a min-height is still required to achieve the styling

FaithDaka:
Also to note: Setting the `background_image_asset` as the component's `background-image` cuts the top offset off.

jfmcquade:
> I think a min-height is still required to achieve the styling

Yeh, makes sense that you'd need a min-height. Perhaps this should be calculated relative to the viewport height, e.g. `90vh`?

> Also to note: Setting the `background_image_asset` as the component's `background-image` cuts the top offset off.

You're right, so I guess this does need to be passed into an `img` component as you had before.

FaithDaka:
> > I think a min-height is still required to achieve the styling
> 
> Yeh, makes sense that you'd need a min-height. Perhaps this should be calculated relative to the viewport height, e.g. `90vh`?
> 
> > Also to note: Setting the `background_image_asset` as the component's `background-image` cuts the top offset off.
> 
> You're right, so I guess this does need to be passed into an `img` component as you had before.

Changes made @jfmcquade

jfmcquade:
I've updated the styling to make the component taller, and also added negative margins to make it appear better when used within a fullscreen pop-up (the intended use-case). This is a bit of a hack, but this component only has that one specific use-case.

In fullscreen pop-up:
Screenshot 2024-12-06 at 12 59 54

In standalone template:
Screenshot 2024-12-06 at 13 01 28

jfmcquade (approved):

PR #2603, opened 2024-12-06
hotfix: nav-stack dismissal

jfmcquade:
PR Checklist

## Description

Provides a temporary fix to https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/issues/88. This allows for a working web preview to be demonstrated on 09/12/24, but the bug should be investigated further.

The workaround is achieved by implementing the following:

- Nav-stacks now only dismiss when the dismiss is triggered in one of the following ways:
  - programmatically through the nav-stack service code
    - including action calls to `nav_stack: close_top` and `nav_stack: close_all`
  - via the "close" button on the nav-stack
  - via clicking the "backdrop" over the page behind the nav-stack
  - when the browser navigates back
    - this is newly implemented behaviour, borrowing logic from #2594 

This represents an exhaustive list of the pre-existing expected triggers to close nav stacks, however by making this explicit in the code, the bug that causes the nav-stack to close unexpectedly, as documented in https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/issues/88, is avoided.

Also includes
- Minor refactoring of nav-stack service code
- Removal of debug logs from task code and answer-list utils

## Git Issues

Closes #

## Screenshots/Videos

Article navigation launched from [module_landing](https://docs.google.com/spreadsheets/d/1nvyc-d_noz-W7guO6x_33c6cCXMXpexGt5HYf5T5J50/edit?gid=452625780#gid=452625780), working as expected. Compare to behaviour on [live web preview](https://plh-kids-kw-uat1.web.app/template/module_landing_selfcare?nav_parent=home_screen&nav_parent_triggered_by=task_card).

https://github.com/user-attachments/assets/ced8b560-6d66-49bf-9789-bcf455308dbc

### DIFF
--- a/src/app/feature/nav-stack/components/nav-stack/nav-stack.component.ts
+++ b/src/app/feature/nav-stack/components/nav-stack/nav-stack.component.ts
@@ -17,6 +17,6 @@ export class NavStackComponent {
   constructor(private modalCtrl: ModalController) {}
 
   public close() {
-    this.modalCtrl.dismiss();
+    this.modalCtrl.dismiss("", "close-button");
   }
 }
--- a/src/app/feature/nav-stack/nav-stack.service.ts
+++ b/src/app/feature/nav-stack/nav-stack.service.ts
@@ -2,16 +2,24 @@ import { Injectable } from "@angular/core";
 import { ModalController } from "@ionic/angular";
 import { INavStackConfig, NavStackComponent } from "./components/nav-stack/nav-stack.component";
 import { SyncServiceBase } from "src/app/shared/services/syncService.base";
+import { TemplateNavService } from "src/app/shared/components/template/services/template-nav.service";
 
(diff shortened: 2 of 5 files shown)


chrismclarke (changes requested):
Taking a closer look at the core issue, I think the likely cause is the `audio` component, which has an `onDestroy` hook that automatically closes the top modal (assuming the top modal is one that it had created to display transcript). However if it has not opened a transcript then the modal it closes will be the nav stack. 

I'll make a separate PR to address that, and check for any other components that might have similar issue, and put a block on this PR for now

jfmcquade:
Thanks for looking into it, @chrismclarke, I've approved and merged #2605 and added a debug template for clarity.

Indeed this PR borrows logic from #2594 and can be ignored

PR #2605, opened 2024-12-06
fix: audio modal close trigger

chrismclarke:
PR Checklist

## Description

Fix issue where audio component would accidentally close any parent nav_stack it is presented in
Alt fix to #2603

## Dev Notes
The only other component that I could see that interacts directly with modal logic is the `combo_box`, so it might also be worth investigating if similar problems appear there, although it looks like the logic is setup to dismiss from inside the created modal so I think shouldn't be able to accidentally close the wrong one. 

## Git Issues

Closes #2603

## Screenshots/Videos

https://github.com/user-attachments/assets/924fa60a-b7cb-4c66-952f-dea2a29c1938

### DIFF
--- a/src/app/shared/components/template/components/audio/audio.component.ts
+++ b/src/app/shared/components/template/components/audio/audio.component.ts
@@ -92,6 +92,8 @@ export class TmplAudioComponent
   hasStarted: boolean = false;
   /** @ignore */
   trackerInterval: NodeJS.Timeout;
+  /** Track any opened transcript modal to close on destroy */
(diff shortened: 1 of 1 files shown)


jfmcquade (approved):
Great detective work! I couldn't figure out what the problematic templates had in common, but this makes complete sense.

It seems to address the issue from functional testing, and I can clearly see what the issue was with the code and how this resolves it, so I'm going to merge this PR so we can generate a web preview from `master`.

PR #2607, opened 2024-12-09
feat: add display group variant, 'header'

jfmcquade:
PR Checklist

## Description

Includes #2608, which should be merged first for clear diffs.

Adds a new variant for the `display_group` component: `header`.

This is intended to be used in conjunction with the `sticky: top` param to position the content of the header to be 

NB, the [plh_kids_kw deployment repo ](https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content) has temporarily been configured to target this PR branch when building. This can be changed after this PR is merged.

## Git Issues

Closes #

### DIFF
--- a/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
+++ b/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
@@ -42,6 +42,13 @@
   margin: 1em 0 0 0;
 }
 
+.display-group-wrapper[data-variant~="header"] {
+  width: 100%;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 var(--small-padding);
+}
+
 .display-group-wrapper {
   &[data-variant~="box_gray"],
   &[data-variant~="box_primary"],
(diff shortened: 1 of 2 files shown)


ChrisMarsh82 (approved):

chrismclarke (changes requested):
I don't think the naming of this variant really makes much sense to me. A `header` is more of a specific component/part of UI than a general style. So Either this should be a `header` component, or should be renamed to be more indicative of what it is.

As for the specific style changes this variant introduces, again I'm not entirely convinced that a variant is the right mechanism to apply white-space only changes. 

I would probably suggest one (or both) of the following alternatives:

1. Create a `header` or `toolbar` component which is designed to be a space-between flex layout with default reduced padding

2. Pass the `header` variant at the deployment level and migrate the core scss to the deployment override so that they can specify their own custom header variant which might make more sense in the context of the deployment as opposed to the core repo.

jfmcquade:
Thanks @chrismclarke, I understand your reservations. The use-case that led to me rushing to make this PR is now addressed by this alternative PR: #2614. I think that it still could be useful to have a dedicated `toolbar` component, as you suggest, but as of #2614 it is no longer strictly necessary.

PR #2610, opened 2024-12-09
feat: expose main app content padding as app config param

jfmcquade:
PR Checklist

## Description

Exposes a new property for customising the padding around the main app content: `config.app_config.LAYOUT.main_content_padding`. This takes a string value representing a value in a standard CSS length unit, e.g. "20px", "2rem", etc.

As a property off `app_config`, this value can be set at 3 different levels (with setting at each subsequent level overriding any previous values):
- In the deployment config
- In a skin
- In a template's parameter list (as of #2531)

## Notes

I was unsure about how to name and nest this property, and am open to alternative suggestions. The `LAYOUT` namespace could potentially be expanded to include what is currently in `APP_HEADER_DEFAULTS` and `APP_FOOTER_DEFAULTS`

## Git Issues

Closes #

### DIFF
--- a/packages/data-models/appConfig.ts
+++ b/packages/data-models/appConfig.ts
@@ -117,6 +117,10 @@ const APP_FOOTER_DEFAULTS: { templateName: string | null } = {
   templateName: null,
 };
 
+const LAYOUT = {
+  main_content_padding: "24px",
+};
+
 const APP_SIDEMENU_DEFAULTS = {
   enabled: true,
   title: "App",
@@ -205,6 +209,7 @@ const APP_CONFIG = {
   APP_UPDATES,
   ASSET_PACKS,
   FEEDBACK_MODULE_DEFAULTS,
+  LAYOUT,
   NOTIFICATIONS_SYNC_FREQUENCY_MS,
   NOTIFICATION_DEFAULTS,
   SERVER_SYNC_FREQUENCY_MS,
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -42,7 +42,10 @@
     <ion-split-pane when="lg" contentId="main" [disabled]="!sidebarRouter.isActivated">
       <div style="display: flex; flex-direction: column; height: 100%; width: 100%" id="main">
         <plh-main-header></plh-main-header>
-        <div class="route-container">
+        <div
+          class="route-container"
+          [style.--main-content-padding]="layoutConfig().main_content_padding"
+        >
           <ion-router-outlet id="main-content"></ion-router-outlet>
         </div>
         @if (footerDefaults().templateName; as footerTemplateName) {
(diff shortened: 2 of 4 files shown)


packages/data-models/appConfig.ts
```diff
@@ -117,6 +117,10 @@ const APP_FOOTER_DEFAULTS: { templateName: string | null } = {
   templateName: null,
 };
 
+const LAYOUT = {
+  main_content_padding: "24px",
+};
```
chrismclarke, line 122: I think the namespace makes sense to me, and agreed would be nice to migrate more layout variables over time from their somewhat verbose alternatives.

I might question how necessary the `main` part of `main_content_padding` is - do we have other specific content that we are assigning padding for which isn't otherwise namespaced? Similarly is `content` indicative (given that content appears in lots of places) - would `page_padding` be more relevant or are we applying this padding in more places than just the template page ion-content? 

I think this is half an authoring preference question, half a dev one

chrismclarke (approved):
Overall looks good to me. See one (non-blocking) comment inline, but otherwise I think makes sense.

I also added one commit to change how the style variable is passed to the route container (find the `[style.--some-var]` pattern a little clunky looking so instead have added an intermediate ngStyle)
[de24b06](https://github.com/IDEMSInternational/open-app-builder/pull/2610/commits/de24b06040faaf8726a92377028967820b595377)

I also gave a quick functional test and all working as expected, although as the debug content is still pending sync I just manually updated the skin config in the main config.ts file as a temporary workaround, which then could recompile via `yarn workflow deployment set debug` without manual contents deletion. 

```ts
SKINS.debug.appConfig!.LAYOUT = {page_padding : '0px'}; // <-- added this locally on L75

config.app_config.APP_SKINS.defaultSkinName = SKINS.defaultSkin.name
```

packages/data-models/appConfig.ts
jfmcquade: I agree that `page_padding` is better, I've renamed the property and updated the PR description (the outstanding [debug content repo PR](https://github.com/IDEMSInternational/app-debug-content/pull/126) is now out of date, but I'll wait until this PR is finalised before updating).

@esmeetewinkel let us know if you have any thoughts about naming this property

esmeetewinkel: Happy with `page_padding`

esmeetewinkel (approved):
Functional test passed. 

To use this however on the `plh_kids_kw` deployment for the `module_landing` screens, I come again across the issue (tracked [here](https://github.com/IDEMSInternational/open-app-builder/issues/2630)) that I cannot add template metadata to generated templates. See template [module_landing](https://docs.google.com/spreadsheets/d/1nvyc-d_noz-W7guO6x_33c6cCXMXpexGt5HYf5T5J50/edit?gid=452625780#gid=452625780) and generator [gen_module_landing](https://docs.google.com/spreadsheets/d/1nvyc-d_noz-W7guO6x_33c6cCXMXpexGt5HYf5T5J50/edit?gid=369775764#gid=369775764).

PR #2612, opened 2024-12-09
review: 2589 tidying

chrismclarke:
PR Checklist

## Description

Targets branch of #2589 

- Removes sticky display group specific styling (should be inherited within main display-group wrapper)
- Restricts background image styles to only include when image provided

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/components/layout/display-group/display-group.component.html
+++ b/src/app/shared/components/template/components/layout/display-group/display-group.component.html
@@ -26,14 +26,7 @@
     [style.marginBottom.px]="params.offset"
     [ngSwitch]="type"
     [style]="_row.style_list | styleList"
-    [style.backgroundImage]="
-      params.backgroundImageAsset
-        ? 'url(' + (params.backgroundImageAsset | plhAsset) + ')'
-        : undefined
-    "
-    [style.backgroundPosition]="
-      params.backgroundImageAsset ? params.backgroundImagePosition : undefined
-    "
+    [ngStyle]="backgroundImageStyles"
   >
     <!-- Default Layout -->
     <ng-container *ngSwitchDefault>
--- a/src/app/shared/components/template/components/layout/display-group/display-group.component.html
+++ b/src/app/shared/components/template/components/layout/display-group/display-group.component.html
@@ -3,11 +3,7 @@
   @if (params.sticky === "top") {
     <div [style.height.px]="sticky.height()"></div>
   }
-  <tmpl-display-group-sticky
-    [position]="params.sticky"
-    #sticky
-    [attr.data-param-style]="params.style"
-  >
+  <tmpl-display-group-sticky [position]="params.sticky" #sticky>
     <ng-container *ngTemplateOutlet="content"></ng-container>
   </tmpl-display-group-sticky>
   @if (params.sticky === "bottom") {
--- a/src/app/shared/components/template/components/layout/display-group/display-group.component.html
+++ b/src/app/shared/components/template/components/layout/display-group/display-group.component.html
@@ -3,7 +3,11 @@
   @if (params.sticky === "top") {
     <div [style.height.px]="sticky.height()"></div>
   }
-  <tmpl-display-group-sticky [position]="params.sticky" #sticky>
+  <tmpl-display-group-sticky
+    [position]="params.sticky"
+    #sticky
+    [attr.data-param-style]="params.style"
+  >
     <ng-container *ngTemplateOutlet="content"></ng-container>
   </tmpl-display-group-sticky>
   @if (params.sticky === "bottom") {
--- a/src/app/shared/components/template/components/layout/display-group/display-group.component.html
+++ b/src/app/shared/components/template/components/layout/display-group/display-group.component.html
@@ -3,13 +3,7 @@
   @if (params.sticky === "top") {
     <div [style.height.px]="sticky.height()"></div>
   }
-  <tmpl-display-group-sticky
-    [position]="params.sticky"
-    [style.--sticky-background]="
-      params.stickyBackground === 'transparent' ? 'transparent' : 'var(--ion-background-color)'
-    "
-    #sticky
-  >
+  <tmpl-display-group-sticky [position]="params.sticky" #sticky>
     <ng-container *ngTemplateOutlet="content"></ng-container>
   </tmpl-display-group-sticky>
   @if (params.sticky === "bottom") {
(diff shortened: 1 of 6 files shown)


src/app/shared/components/template/components/layout/display-group/display-group.component.html
```diff
@@ -3,7 +3,11 @@
   @if (params.sticky === "top") {
     <div [style.height.px]="sticky.height()"></div>
   }
-  <tmpl-display-group-sticky [position]="params.sticky" #sticky>
+  <tmpl-display-group-sticky
+    [position]="params.sticky"
+    #sticky
+    [attr.data-param-style]="params.style"
```
jfmcquade, line 9: Because there is an additional layer of nesting when rendering a sticky display group, we need to re-apply the style params here in order to make them available as selectors on the `:host` element within the sticky display group component.

jfmcquade:
Thanks, @chrismclarke 

I believe we do want something like the background styling that you removed, I'll explain why:

When setting a display group to sticky, the display group should have a background colour that matches the app background colour by default. This means we get functionality like [feat_display_group_sticky](https://docs.google.com/spreadsheets/d/1tY8sa8ocDl552gIUAP263zgl3JQVcIeS4foaKk7nlMk/edit?gid=434509236#gid=434509236), where making a display group sticky means that the main content scrolls behind the whole display group as a block by default:

Screenshot 2024-12-10 at 12 30 42

Otherwise, as the standard display group doesn't have a background colour by default, we get an undesirable overlap of elements:

Screenshot 2024-12-10 at 12 31 20

But we also want the ability to set this background colour to be transparent. That is required for the footer in the designs referenced in #2589.

| Without transparent background | with transparent background (desired behaviour) |
|-----|-----|
| Screenshot 2024-12-10 at 12 04 13 | Screenshot 2024-12-10 at 12 19 59 |

That's the justification for the css variable that you removed, although I agree that the implementation could have been better.

Another way to handle this, aligned this with our plans for global style params (#1971), is to use a param like `style: background-transparent`. That's what I've implemented in https://github.com/IDEMSInternational/open-app-builder/pull/2612/commits/283a8cc1b617d4f68f7b94766f36c7dab5adffae. As the default background colour for a non-sticky display group is already transparent, adding this style param has no effect unless the display group is sticky.

(Footnote: if we want to explicitly set the background colour to be the default app background colour, we would need a way for authors to refer to this, as they can `background-primary`, `background-secondary` etc., as we want to avoid requiring the author to write `var(--ion-color-background)` or `ion-color-background`. The word "background" is the key work in the ionic colour variable naming convention, but if we're appending it to `background-*` then we might need to use something like `background-default`. But there isn't a use case for that at the moment, as long as sticky display groups get that background colour applied by default)

chrismclarke:
The use case makes sense, thanks for clarifying. However in that case I really don't think the sticky layout should be imposing any specific colour styling (it's a layout element, not a design one). 

If the default behaviour is for a sticky to display transparent, then the authors can still decide a background colour to use (I assume there's easy enough ways to set a primary/secondary colour as required). Ideally the same variables used to set the background colour on a regular display group should have effect in the sticky version

jfmcquade:
> The use case makes sense, thanks for clarifying. However in that case I really don't think the sticky layout should be imposing any specific colour styling (it's a layout element, not a design one).
> 
> I'd be much more in favour of the default handling to be transparent sticky and authors to actively set a background colour variable as required, as opposed to trying to predict the colour of the content that is below it (could sit on top of a variety of content types/styles, or have good reason to be distinct).

Yes, good point. I've updated so that setting the display group to `sticky` doesn't apply a background colour by default, and exposed a style for setting the background colour to the default app background colour. When this PR is merged into the target PR branch, I'll update the description of #2589 to include these changes, and mark the PR as breaking to indicate that existing templates may need to manually set the sticky display group background colour to preserve the current styling, although I'm not sure that that authoring pattern is currently being used anywhere.

> I think to move forward this PR adding an extra `primary` or `bg-primary` style would make sense and author along with the sticky display group

This is where we get into a bit of complication with naming conventions...
I've named the style `background-page` currently. I think there's good reasons we can't use `background-primary` in this case: `background-primary` should be reserved for setting the background colour to the base primary colour, or else a different shade of that colour interpreted contextually. Whilst it's true that the page background colour of the app is derived from the primary theme colour by default, it can also be overridden so is independent (as is the case on our non-default themes, which tend to have a `white` background colour) – indeed the three source colours that can be provided when generating a theme are `primary`, `secondary`, and `page-background`.

So we need a way to refer to that theme-level page background colour when applying it to components` backgrounds. `style: background-background` is no good, and `style: background-page-background` seems verbose, so I went for `style: background-page`. I'd be opposed to using another name though, some others I considered:

- background-app
- background-default
- background-surface
- background-theme
- background-standard

But as we already use `page-background` when supplying a colour to theme generation, `background-page` seemed like the best option.

I also wouldn't be opposed to changing this to `bg-page`. I don't believe we've previously exposed any property containing "bg" as a shorthand for "background", but we're setting a precedent here that we don't currently have implemented. 

> This would need more sensible ways of passing styles, which I think would also be desirable. E.g. we've introduced `box_primary` and `box_secondary` which probably would make sense as a separate `box` styling and `primary` or `bg-primary` styles. 

Agreed. I think in order to capture the current styling, this would need to be something like:
```
variant: box;
style: background-primary-200 border-primary-500;
```
or possibly
```
variant: box;
style: background-primary-light border-primary;
```

But I don't think that needs to implemented now.

chrismclarke:
@jfmcquade 
Actually I made enough additional changes I think it would be good for you to take one final look before merging

See commits:

a5a4b3906005963b912d422aa5cb70e06e7ef139

- Update CSS variable and style list names

[1f5a46a](https://github.com/IDEMSInternational/open-app-builder/pull/2612/commits/1f5a46af936ec6928942be0df6f7000bebdbda42)

- Add `background-size: contain` property
- Refactor background CSS extraction to separate function and ngStyle variable to apply slightly more cleanly

I also updated the [feature_display_group](https://docs.google.com/spreadsheets/d/1ARbNRGDer5vj9qSpRMZFrMkYifGkH3TLtDVp72YbaqU/edit?gid=1231713195#gid=1231713195) demo you linked, to improve some of the sizing on smaller devices (wrapping text in max 40vw to prevent overflowing image too much, and adding min-height to the display_group to reduce the need for arbitrary top margin in the text component)

image

chrismclarke:
Oh and totally unrelated, but I noticed the image currently used for the background is really a png image that's just be embedded directly into an SVG file as base64 data. This isn't the nicest example, do we have a pure svg form? If not I'd probably say better to use as a .png file to not obscure the fact that it is not vector scalable

jfmcquade:
@chrismclarke thanks for making the changes, I think they generally look good.

I'm not totally convinced by `background_colour` as a style name. For one thing it contains the region-specific spelling of "colour", which would be nice to avoid. But I'm also not sure that it's clearly extendable with `background_color_primary`, `background_color_white` etc. I agree that `background_page` isn't very clear in isolation either, but I can see how it's using a syntax for adding a style to certain feature of the component: `<feature>_<property-to-apply>`. This is somewhat inline with the discussion on #1971, although that is quite dated now.

Is the idea that setting `style: background_colour` is specifying that we want this component to have _a_ background colour, which will be the app page background colour by default? That does actually make sense on reflection. 

In any case, I'm happy to merge as-is as we're not rolling out a global system just yet. Apologies for prolonging this derailment, but it's hard not to engage...

Another minor point:

> albeit I think the default colour is the 500 weight so can probably use border-primary instead of border-primary-500 and use a different weight only when needed.

Currently the `primary` colour is the exact source colour provided to generate the theme from, which might sit anywhere on the scale based on how the scale is generated (I think that's the case anyway...). This isn't a great system as we don't have any restrictions on how light or dark that source colour might be. We can always interpret `primary` within a given component as a certain shade, but we do offer authors the ability to reference the primary colour as provided, which is probably desirable in the case of specific brand colours, for example.

PR #2614, opened 2024-12-10
fix: sticky display group custom styles

### LINKED ISSUE #2613: [BUG] Styles authored in style_list do not apply to sticky display group as expected
**Describe the bug**
Custom styles authored through the `style_list` column on a display group with the `sticky` param set do not apply as expected.

See the template [debug_dg_sticky_style_list](https://docs.google.com/spreadsheets/d/1fwibHu1p6mOTGAwVb0cpi3IloiiKjwxqyidfj32zXMs/edit?gid=1422144003#gid=1422144003) (designed to mimic the real-world use case on the [KW app home screen](https://docs.google.com/spreadsheets/d/1cNFMnElaJFjenwQCGt1K7m6fT7pGX69SiIBdu6mu-as/edit?gid=687714871#gid=687714871)).

Screenshot 2024-12-10 at 15 22 06

Screenshot 2024-12-10 at 15 20 38

jfmcquade:
PR Checklist

## Description

Custom styles, authored through the `style_list` column, were not having the expected effect when applied to a display group with the `sticky` param set (see #2613). This PR resolves this by removing disabling flexbox on the sticky display group container. This means that authors may need to apply additional custom styling in order to achieve the desired result when authoring a sticky display group.

## Dev notes

Sticky display groups have an additional level of nesting vs their non-sticky counterparts, since the display group component renders a child `tmpl-display-group-sticky` element. After spending more time than I'd like to admit wrestling with CSS inheritance to try and conditionally set defaults on the `.display-group-wrapper` element depending on whether the display group is sticky, I realised that it was the fact that the sticky display group component was setting `display: flex` on its host container that was causing the inconsistencies in applying the authored styles. 

## Author notes

I think the original motivation for enabling flexbox by default on the sticky display group's container was that when authoring an inline header/footer, we likely want to position the content centrally. However, it seems sensible to remove this default so that custom styles are applied in the same way for sticky/non-sticky display groups, and the more fine-tuned styling can be left to authors themselves, provided they have the tools to do so.

It may still be useful to have a dedicated `toolbar` component that can be used in such cases (as suggested in [this PR comment](https://github.com/IDEMSInternational/open-app-builder/pull/2607#pullrequestreview-2489987223)). This would be an alternative to manually authoring `justify-content: space-between` and setting the padding for such cases.

This PR supersedes #2607, which contains a workaround for the same use-case . After the current PR is merged, the [KW app home screen tempalte](https://docs.google.com/spreadsheets/d/1cNFMnElaJFjenwQCGt1K7m6fT7pGX69SiIBdu6mu-as/edit?gid=687714871#gid=687714871) can be updated to reflect the `parameter_list` and `style_list` in the [debug_dg_sticky_style_list](https://docs.google.com/spreadsheets/d/1fwibHu1p6mOTGAwVb0cpi3IloiiKjwxqyidfj32zXMs/edit?gid=1422144003#gid=1422144003). 

## Git Issues

Closes #2613, #2607

## Screenshots/Videos

[debug_dg_sticky_style_list](https://docs.google.com/spreadsheets/d/1fwibHu1p6mOTGAwVb0cpi3IloiiKjwxqyidfj32zXMs/edit?gid=1422144003#gid=1422144003)

Screenshot 2024-12-10 at 15 31 14

Screenshot 2024-12-10 at 15 31 50


src/app/shared/components/template/components/layout/display-group/sticky/display-group-sticky.component.scss
```diff
@@ -1,10 +1,8 @@
 // When the display group is "sticky", style it as an inline header/footer
 :host {
-  display: flex;
```
chrismclarke, line 3: question: have you checked whether offsets still calculate correctly with this property removed. I expect we might need the default display to be set as `block` for the `#sticky.contentHeight()` property to evaluate correctly

chrismclarke (approved):

esmeetewinkel (approved):
Functional test passed

PR #2625, opened 2024-12-14
Feat: add_data action

chrismclarke:
PR Checklist

## Author Notes
Adds support for a new `add_data` action which minimally takes the `_list_id` and key-value pairs to set. See examples in 
[feat_data_actions_edit](https://docs.google.com/spreadsheets/d/1nREviZhw2HG1YQy6KA1S9PU4Q36RRVHdwSYf3TMbduk/edit?gid=1583208946#gid=1583208946)

Any data that has been created using an `add_data` action will automatically have a metadata property `_user_created` which can be used to help identify or sort data in the list.

There is also a `remove_data` action which could be called directly (passing the _list_id and item _id), or more concisely there is also a `remove_item` which infers both based on the current item in a `data_items` loop. Currently any data can be removed (not just user_created), however any initial sheet data will be restored on app load (not intended to be removable)

By default data is sorted by the original index of the row from the sheet. When new data items are added they receive a generated row index as if appending rows to the original sheet  (takes the number of existing rows and adds 1 each time). This should mean all user generated content will appear at the bottom of data_lists, unless manually sorted by some other field. The caveat here is if the source data is later updated to include more entries, there is no easy way of identifying that the user_generated data now needs to be offset - Iso the user_generated data will be preserved in previous order with new data interspersed. 

It is therefore recommended when using `add_data` to ideally use with an empty initial list, to avoid confusion between hardcoded sheet data and user_generated content. It might be worth discussing in the future if there might be better ways to automatically store all user_generated content to separate tables.

In the example sheets I had wanted to provide example of adding new rows that incremented the number field each time, however there wasn't an easy way to refer back to the initial data list and calculate it's length whilst adding new rows. It might be good to include a syntax for `@list.length` or similar source data_list properties in the future, there are some hacky workarounds for now but not ideal.

## Review Notes
See examples in [feat_data_actions_edit](https://docs.google.com/spreadsheets/d/1nREviZhw2HG1YQy6KA1S9PU4Q36RRVHdwSYf3TMbduk/edit?gid=1583208946#gid=1583208946)

Tests can be run from scripts noted in test files

## Dev Notes

The code for `set_data` and `reset_data` has been slightly refactored to move into a dedicated data-actions subfolder, the core methods have had minor changes to decouple some of the parsing logic but should still function in same way (tests still passing)

The `persistedMemory` and `reactiveMemory` db adapters have both been updated. This was to help resolve a point of confusing when working on them related to how metadata fields are handled. There had been a method in the main service to replace all metadata into a child `APP_META` property, however this only repopulated during query and not regular get operations. Also it was somewhat confusing as the `persistedMemory` adapter stores all data within a single `data` property which can support keys with underscore values, so doesn't require the `APP_META` workaround. So now the metadata workaround is handled directly within the `persistedMemory` adapter. I'm not actually sure how important this is as it appears there are also methods in place to strip away the metadata anyway, so a bit confusing what is and isn't needed

I wasn't sure how best to handle ids for author-generated content. Whilst there might be some benefit to authors providing a naming convention, implementing it in a sensible way felt quite challenging. So for now any user_generated content has an id generated using the uuid v4 spec. 

`remove_data` and `reset_data` actions haven't been split into their own files and test suites as the logic to implement is very basic and the tests typically depend on first triggering an add/update data action, so tests have been included in those corresponding suites instead.

## Git Issues

Closes #

## Screenshots/Videos
[Screenity video - Dec 14, 2024.webm](https://github.com/user-attachments/assets/de847a90-77a0-44a2-8afa-ac1cce72f2be)


src/app/shared/services/dynamic-data/actions/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    const { _id, _list_id } = params;
+    return this.service.remove("data_list", _list_id, [_id]);
+  };
+
+  public add_data: IActionHandler = async ({ params }: { params?: IActionAddDataParams }) => {
+    const parsedParams = this.parseActionParams(params);
+    return addDataAction(this.service, parsedParams);
+  };
+
+  /** Parse action parameters to ensure list id provided and  */
```
jfmcquade, line 46: typo

src/app/shared/services/dynamic-data/actions/index.ts
```diff
@@ -0,0 +1,86 @@
...
+      if (!parsed._list_id) {
+        console.error(params);
+        throw new Error("[Data Actions] could not parse list id");
+      }
+      return parsed;
+    }
+
+    // throw error if args not parsed correctly
+    console.error(params);
+    throw new Error("[set_data] could not parse params");
```
jfmcquade, line 59: worth logging the params maybe?

src/app/shared/services/dynamic-data/actions/index.ts
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Any params provided by templating system will already be partially parsed to replace dynamic references
+ * such as @local or @field. This also converts @item references to `this.item`
+ * Revert the item changes and also convert string parameter values to number where required
+ *
+ * TODO - this method should be removed and `items` made to update their own `this` context
+ **/
+function hackParseTemplatedParams<T extends IActionBaseParams>(params: T) {
+  const parsed = {} as T;
+  // HACK - un-parse @item references that the templating system converts to `this.item`
```
jfmcquade, line 72: this is a bit painful but I see the point

src/app/shared/services/dynamic-data/actions/set_data.action.ts
```diff
@@ -0,0 +1,101 @@
...
+import { FlowTypes } from "packages/data-models";
+import {
+  coerceDataUpdateTypes,
+  evaluateDynamicDataUpdate,
+  isItemChanged,
+} from "../dynamic-data.utils";
+
+interface IActionSetDataOperatorParams {
+  // TODO - ideally use same itemPipe operators
+  // (although filter will need to be updated to include dynamic context)
```
jfmcquade, line 13: I believe you've addressed this now so this can possibly be updated?

src/app/shared/services/dynamic-data/actions/set_data.action.ts
```diff
@@ -0,0 +1,101 @@
...
+  let items = await firstValueFrom(ref);
+  if (items.length === 0) {
+    const msg = `[Update Fail] no doc exists\ndata_list: ${_list_id}\n_id: ${_id}`;
+    throw new Error(msg);
+  }
+  // NOTE - RXDB doesn't support querying by index or pagination so still retrieve all and then reduce
+  if (typeof _index === "number") {
+    const targetItem = items[_index];
+    if (!targetItem) {
+      const msg = `[Update Fail] no doc exists\ndata_list: ${_list_id}\n_index: ${_index}`;
```
jfmcquade, line 67: message should state that no _row_ exists

jfmcquade (approved):
Looks good. Quite a lot going on, but nicely organised. I've looked through all the logic although haven't picked over the fine details. A few very minor comments inline.

Issue #2611, opened 2024-12-09
[FEATURE] Expose more header/footer customisation options to app config; rationalise naming of associated properties


jfmcquade:
**What?**
Currently, the options for customising the header and footer via app config properties are limited.

The header config has the following properties:
```ts
const APP_HEADER_DEFAULTS = {
  title: "App",
  collapse: false,
  colour: "primary" as IHeaderColourOptions,
  // The "compact" variant reduces the header height and removes the title
  variant: "default" as IHeaderVariantOptions,
  // default only show menu button on home screen
  should_show_menu_button: (location: Location) =>
    activeRoute(location) === APP_ROUTE_DEFAULTS.home_route,
  // default show back button on all screens except home screen
  should_show_back_button: (location: Location) =>
    activeRoute(location) !== APP_ROUTE_DEFAULTS.home_route,
  // on device minimize app when back button pressed from home screen
  should_minimize_app_on_back: (location: Location) =>
    activeRoute(location) === APP_ROUTE_DEFAULTS.home_route,
};
```

and the footer:

```ts
const APP_FOOTER_DEFAULTS: { templateName: string | null } = {
  templateName: null,
};
```

Minimally, we require to be able to set the background colour of the footer (see #2508), and to supply a template to display in the header (possibly required for #2535, although that could also be achieved with a display group with `sticky: top`).

It would make sense to bring the two sets config options in line with each other as much as possible. It would also be a good opportunity to rename these sets of parameters, e.g. from `APP_HEADER_DEFAULTS` -> `HEADER`.

**How?**
It would be good to still support the current syntax but mark it deprecated and provide a warning in the console. Otherwise, this will be a breaking change that will require updates to all deployment configs.

**Other details**

Following #2531, it is possible to set app config properties at the template-level.

PR #2631, opened 2024-12-16
feat!: expose additional header and footer configuration props to app config

jfmcquade:
PR Checklist

## Description

Exposes additional properties for customising header and footer to app config. This allows for further template-level customisation of header and footer, as per #2531. 

The new properties are:
- Header properties (inside `app_config.APP_HEADER_DEFAULTS` object)
  - `hidden` (boolean)
    - hide/show the header
  - `template` (string)
    - The name of a template to display in the header
  - `back_button.hidden`
    - hide the back button
  - `menu_button.hidden`
    - hide the menu button
- Footer properties (inside `app_config.APP_FOOTER_DEFAULTS` object)
  - `background` (`"primary" | "secondary" | "none"`)
    - Sets the background colour of the footer
    - This is equivalent to the `APP_HEADER_DEFAULTS.colour` property, but I've used the name "background" as this seems preferable. See TODOs below for notes around bringing these two sets of properties into alignment

## Breaking Changes

This PR removes three config properties from the `APP_HEADER_DEFAULTS` property (which each took a function as their value):
- `should_show_menu_button`
- `should_show_back_button`
- `should_minimize_app_on_back`

This is technically a breaking change, however I don't believe any deployments actually overrode these properties so there should be no knock-ons from these changes.

## TODOs

Originally I planned to also rationalise the naming of the header and footer config properties, see #2611. However, as agreed on a call, doing so in a way that supports the current, would-be legacy, syntax is not trivial, and is not a priority at this moment.
 
## Git Issues

Closes #

## Screenshots/Videos

See the various templates in the [feature_app_layout](https://docs.google.com/spreadsheets/d/1Z46gTluTPeffqpiAZMapYo0imm59GU03j0sW9UChvPM/edit?gid=3937162#gid=3937162) workbook:

Screenshot 2024-12-19 at 16 27 37

https://github.com/user-attachments/assets/110e850f-bf5b-4695-a601-268348497225


packages/data-models/appConfig.ts
```diff
@@ -83,13 +83,15 @@ const APP_ROUTE_DEFAULTS = {
...
-export type IHeaderColourOptions = "primary" | "secondary" | "none";
+export type IHeaderFooterBackgroundOptions = "primary" | "secondary" | "none";
 export type IHeaderVariantOptions = "default" | "compact";
 
 const APP_HEADER_DEFAULTS = {
-  title: "App",
   collapse: false,
-  colour: "primary" as IHeaderColourOptions,
+  colour: "primary" as IHeaderFooterBackgroundOptions,
+  show: true,
```
chrismclarke, line 92: nit(blocking): would make more sense to set `hide: true` as a way of hiding the header given the default value of `show` is true

packages/data-models/appConfig.ts
```diff
@@ -83,13 +83,15 @@ const APP_ROUTE_DEFAULTS = {
...
+  template: null as string | null,
```
chrismclarke, line 93: nit(non-blocking): I'm not quite sure what casting the `null` to string really achieves here. If it's an optional property could just leave as `undefined`, unless `null` specifically differentiates from `undefined` in another meaningful way I'm missing here

chrismclarke: Oh, actually I see it's because we're defining as a const and then deriving types from it. Now that the list of defaults is getting quite long and specific I think it would be better to first define the header config type definition and then assigning the defaults from the type (save all the additional type-casting)

src/app/app.component.html
```diff
@@ -41,13 +41,15 @@
     <!-- Main content: shows in split-pane when sidebar route active -->
     <ion-split-pane when="lg" contentId="main" [disabled]="!sidebarRouter.isActivated">
       <div style="display: flex; flex-direction: column; height: 100%; width: 100%" id="main">
-        <plh-main-header></plh-main-header>
+        @if (headerConfig().show) {
+          <plh-main-header></plh-main-header>
```
chrismclarke, line 45: nit(non-blocking): given that the header might appear/disappear on a per-template basis it might be easier to use a `style.display` or `style.visibility` attribute to handle show/hide (assuming the header host is rendered as block or flex)

src/app/app.component.ts
```diff
@@ -51,8 +51,9 @@ import { TemplateMetadataService } from "./shared/components/template/services/t
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class AppComponent {
-  sideMenuDefaults = computed(() => this.appConfigService.appConfig().APP_SIDEMENU_DEFAULTS);
+  headerConfig = computed(() => this.appConfigService.appConfig().APP_HEADER_DEFAULTS);
   footerDefaults = computed(() => this.appConfigService.appConfig().APP_FOOTER_DEFAULTS);
```
chrismclarke, line 55: nit(non-blocking): I can see the existing code interchanges between config and defaults quite liberally - I know we've also been migrating config to move away from default to config, so might make sense to also use the term here (not v. important though)

src/app/shared/components/header/header.component.html
```diff
@@ -14,15 +14,16 @@
...
-        *ngIf="!(headerConfig().variant === 'compact')"
-        style="text-align: center"
-        routerLink="/"
-      >
+    @if (headerConfig().template) {
+      <plh-template-container
+        [templatename]="headerConfig().template"
+        [ignoreQueryParamChanges]="true"
+      ></plh-template-container>
+    } @else if (!(headerConfig().variant === "compact")) {
```
chrismclarke, line 22: nit(non-blocking): feels like a bit of a clunky if statement. So the first case is to show a template if the user includes a template, and the second case is to show the title text if the user both has not selected a template nor set the variant as compact.... I'm not sure if there is a way to tidy up the negation a bit, or better means to handle what to display.

I'm also slightly confused as to what is designed to be displayed if the author has specified a compact variant with a title?

src/app/shared/components/header/header.component.ts
```diff
@@ -69,11 +69,10 @@ export class headerComponent implements OnInit, OnDestroy {
     );
     effect(
       () => {
+        const config = this.headerConfig();
```
chrismclarke, line 72: nit(blocking): it's not really clear why we're triggering the `handleRouteChange` method whenever the config changes, and not when a navigation change event is triggered. 

If there is a side-effect from the routeChange method that is linked to the config it would be better to separate this out with a second effect hook and/or a computed property from the headerConfig that actually controls when it needs to be changed.

If there is a good reason to trigger this whenever any header config changes code comments should explain and ideally tests added to ensure the functionality isn't accidentally regressed in the future

src/app/shared/services/app-config/app-config.service.ts
```diff
@@ -100,12 +100,25 @@ export class AppConfigService extends SyncServiceBase {
       return;
     }
 
+    // some app config properties should be functions, but may be defined as boolean values,
+    // e.g. in template parameter_list to disable a certain property on that template
```
chrismclarke, line 104: nit(blocking): slightly tricky issue here, I can see how the introduction of different config overloads (function or boolean) could create a bit of an issue. 

I think we either need to solve this in one of three ways

1) Accept we want this dual functionality, update the type definitions to specify `boolean | function` and add a method to `castBooleanOrFunction` to handle. This code should be called from where the specific properties are used (e.g. the back-button component) to keep highly specific properties out of the general config service

2) Unify around a single type, insisting either authors define functions or defaults use booleans

3) Use separate properties, one that accepts a function and one that accepts boolean, and add logic where these are accessed to identify which one to use and handle accordingly (can likely assume evaluating both statements assuming sensible defaults are used)

chrismclarke (changes requested):
Adding the new properties sounds good to me, and think the trade-offs you've made regarding legacy/modern naming conventions seems reasonable.

I've added a number of minor in-line comments as I think there's probably a few areas that would be good to tidy up a bit

src/app/shared/components/header/header.component.html
jfmcquade: If the `compact` variant is specified then no title is displayed, even if the author has provided one.

I've improved this logic slightly in https://github.com/IDEMSInternational/open-app-builder/pull/2631/commits/d5fe6c086d2ab2eba22220b839662152b25d942f, which I think, makes it clearer

jfmcquade:
Thanks, @chrismclarke, I've acted on your suggestions. 

@esmeetewinkel I think this should now be ready for a functional review

esmeetewinkel (changes requested):
> I tested on the PR preview link. For me the `hidden` properties are not working (not the hidden back button nor the hidden header). I can see they do work in the screen video in the PR description - did maybe something get lost when you renamed this property?

Thanks for testing. The issue was that I hadn't updated the debug content repo to reflect the changes tot he debug sheets that reflected the updated syntax that I changed since this PR was opened. I've now committed those changes to the debug content repo's `main` branch with https://github.com/IDEMSInternational/app-debug-content/pull/134

Testing on the updated PR preview link, the behaviour is now as expected (I tested in an incognito window to avoid caching issues):
Screenshot 2024-12-23 at 17 26 16
Screenshot 2024-12-23 at 17 26 20

src/app/shared/services/skin/skin.service.spec.ts
```diff
@@ -32,6 +32,8 @@ const MOCK_SKIN_2: IAppSkin = {
 
 const MOCK_APP_CONFIG: Partial<IAppConfig> = {
   APP_HEADER_DEFAULTS: {
+    template: null,
+    show: true,
```
esmeetewinkel, line 36: Should this be 
```suggestion
    hidden: false,
```
?

chrismclarke:
> I tested on the PR preview link. For me the `hidden` properties are not working (not the hidden back button nor the hidden header). I can see they do work in the screen video in the PR description - did maybe something get lost when you renamed this property?
> 
> Getting this console error: image
> 
>  Screenity.video.-.Dec.20.2024.mp4

That's weird. Might be worth trying to run locally in case there's something from the debug config that needs updating which hasn't yet been merged into the debug repo

chrismclarke (changes requested):
Thanks @jfmcquade , the config updates look a lot nicer to me and testing the debug templates you added looked good to me.

I've proposed one small tidying PR targeting this branch, 
https://github.com/IDEMSInternational/open-app-builder/pull/2648

Once resolved and the issue flagged by @esmeetewinkel checked then I'd say should be good to go

jfmcquade:
> I tested on the PR preview link. For me the hidden properties are not working (not the hidden back button nor the hidden header). I can see they do work in the screen video in the PR description - did maybe something get lost when you renamed this property?

Thanks for testing. The issue was that I hadn't updated the debug content repo to reflect the changes tot he debug sheets that reflected the updated syntax that I changed since this PR was opened. I've now committed those changes to the debug content repo's main branch with https://github.com/IDEMSInternational/app-debug-content/pull/134

Testing on the updated PR preview link, the behaviour is now as expected (I tested in an incognito window to avoid caching issues):

Screenshot 2024-12-23 at 17 26 20

Screenshot 2024-12-23 at 17 26 16

Issue #2636, opened 2024-12-17
[BUG] Variable changes not propagating to templates nested inside data-items


jfmcquade:
**Describe the bug**
Changes to values of (local) variables are not properly propagating to templates which are nested inside a `data-items` component.

**App version**
First identified by @esmeetewinkel as a bug in unreleased `v0.17.0` on the `plh_kids_kw` deployment within the [article](https://docs.google.com/spreadsheets/d/1gfiZARKRfqyvaRg8AP20bi2D4K8NrCxmnW-3cMNhX2I/edit?gid=403467348#gid=403467348) template.

**Debug Sheet(s)**
[debug_data_items_local](https://docs.google.com/spreadsheets/d/14XXwZ7R_7qIBo5xKYuxnPVQbQidI--PPJ7TVvZxwy1k/edit?gid=1990453574#gid=1990453574)

On current `master`, dynamic changes to the local variable are not reflected in the nested template (to see the template working as expected, without the bug, checkout an older commit, e.g. `cad4962`).

<img width="600" alt="Screenshot 2024-12-17 at 15 41 34" src="https://github.com/user-attachments/assets/1266b913-bae5-4cf6-ac18-5155900f525a" />

<img width="280" alt="Screenshot 2024-12-17 at 15 36 02" src="https://github.com/user-attachments/assets/150f1a1c-f9d4-4e8a-b3a8-35971788fd35" />

**Additional context**
Using `git bisect` I've narrowed down the problematic commit to [this one](https://github.com/IDEMSInternational/open-app-builder/commit/95574546782186e507cf13a0ea9f5e54fe13fd2f#diff-d8552418615ba5964f90623a1d7bca321897bedf58c8f3118b60ea1af0a76da5R82), which makes sense, although I haven't yet been able to find a solution. It's worth noting that an equivalent bug was observed previously and resolved by this #2221 (and it's follow-up #2223).

esmeetewinkel:
Screen videos of issue on `plh_kids_kw` deployment: 

**On `master`** can't navigate beyond second screen
https://github.com/user-attachments/assets/aa43bff6-3aa2-4b97-8705-a0e49f322399

**On `before_issue_2636` (checking out commit `cad4962`)** works as expected

https://github.com/user-attachments/assets/e3e0d31e-1f15-4503-94e1-30d4a6bc3d59

PR #2637, opened 2024-12-17
spike: data-items nested change detection

jfmcquade:
PR Checklist

## Description

WIP solution to #2636.

When rendering `row.rows` in the data-items component template, the value of the nested row does update dynamically when the local variable changes. But this change is not reflected in the value rendered in the title component.

[debug_data_items_local](https://docs.google.com/spreadsheets/d/14XXwZ7R_7qIBo5xKYuxnPVQbQidI--PPJ7TVvZxwy1k/edit?gid=1990453574#gid=1990453574):
Screenshot 2024-12-17 at 18 26 11

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/components/data-items/data-items.component.html
+++ b/src/app/shared/components/template/components/data-items/data-items.component.html
@@ -1,7 +1,8 @@
-@for (row of itemRows(); track row._nested_name) {
+@for (row of itemRows(); track trackByRow($index, row)) {
   <plh-template-component
     [row]="row"
     [parent]="parent"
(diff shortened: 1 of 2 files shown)


chrismclarke:
Thanks for creating the debug sheet and initial investigation. I came to the same conclusion as you, that actually everything looks correct from the populated data but for some reason the nested template isn't recompiling as expected.

I've tried to track down the cause and apply a fix in https://github.com/IDEMSInternational/open-app-builder/pull/2639

PR #2644, opened 2024-12-20
feat: add support for combo_box child data_items

chrismclarke:
PR Checklist

## Description

Add support for providing a `data_items` list to a combo_box to generate answer options

## Author Notes
See updated example in [comp_combo_box](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=569531329#gid=569531329)

Note - if passing a data_list it is expected that the list has columns for `name` and `text`, same as hardcoded lists. The `data_items` loop should not contain any child rows

If using `data_items` list there is not quite the same safeguards in place as with some of the hardcoded lists (e.g. filtering out incomplete item rows) - this is due to the assumption that data_lists are less prone to authoring error than long lists of hardcoded variables. 

There isn't currently any integration between the system used by combo_boxes to add custom items and the ability for data_lists to add new items (i.e. adding a custom combo box item will not make any changes to the data_list). If this functionality is required it should be prioritised with a follow-up issue

## Dev Notes
The core combo_box code has been left unchanged as much as possible (although could definitely do with a bit of tidying/refactor), the main change was to pass the generated `answerOptions` directly to the modal (save generating in multiple places), and add the binding to subscribe to data_lists. The input `template` variable was also removed as it appeared to not be in use

Additional knock-on was required to allow passing back the data_items as a pure item list, as typically a data_items loop expects child template rows to generate as looped template item rows. 

## Git Issues

Closes #

## Screenshots/Videos

[Screenity video - Dec 19, 2024 (1).webm](https://github.com/user-attachments/assets/84015daa-0aad-4cf9-bfa0-9cd3c4d1f2f8)

### DIFF
--- a/src/app/shared/components/template/components/combo-box/combo-box-modal/combo-box-modal.component.html
+++ b/src/app/shared/components/template/components/combo-box/combo-box-modal/combo-box-modal.component.html
@@ -8,7 +8,7 @@
         <div
           class="container-radio"
           [class.checked-radio]="form.get('answer').value === radio.name"
-          *ngFor="let radio of valuesFromListAnswers"
+          *ngFor="let radio of answerOptions()"
         >
           <label>
             <input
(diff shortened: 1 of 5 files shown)


jfmcquade (approved):
Great, the demo is working for me and the code looks good

esmeetewinkel (approved):
Functional test passed. I added to your example some dynamic data, and can confirm that shows as expected. 

https://github.com/user-attachments/assets/966794fd-7203-4ab9-bfa0-ec977d7c01be

I think this is ready to merge - @jfmcquade / @chrismclarke Can one of you resolve the conflicts and then merge this?

PR #2648, opened 2024-12-20
refactor: header defaults and functions

chrismclarke:
PR Checklist

## Breaking Changes
Header config removes override for functions `should_show_menu_button`, `should_show_back_button` and `should_minimize_app_on_back`. All of these functions were by default linked to whether on app home page or not, and typically not overridden by deployments. Now that there are more specific parameters such as `back_button.hidden: true` that can be used to handle specific functionality

NOTE - probably good to mention this in the target PR if merging, although I'm not aware of any deployments that modify these functions

## Description

Targets branch of #2631 

- Remove some app header config defaults where they can be implied (`undefined` as falsy default)
- Refactor menu/back button home-specific side-effects to be included in component code (remove authoring override)

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/data-models/appConfig.ts
+++ b/packages/data-models/appConfig.ts
@@ -89,44 +89,27 @@ export type IHeaderVariantOptions = "default" | "compact";
 
 interface IAppConfigHeader {
   back_button: {
-    hidden: boolean;
+    hidden?: boolean;
   };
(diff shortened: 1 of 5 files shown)


jfmcquade (approved):
Thanks, much neater

PR #2655, opened 2024-12-24
chore: dismiss any open nav stacks when navigating back

jfmcquade:
PR Checklist

## Description

Currently, when any nav-stack pop-ups are open and the user navigates back using the in-browser/device back button, the nav-stacks remain open and the page below them navigates back to the previous template. Whilst ultimately we should be handling routing within nav-stacks, this PR offers a short term solution: in the case described, any open nav-stacks will be dismissed, and the use will be left on the top-level template page that originally launched the nav-stack(s).

Also includes some tidying of the nav-stack code.

## Testing

From the [feat_nav_stack](https://docs.google.com/spreadsheets/d/1qZNmJre8eqVqQpBO8w3iMHJC84W0QOAnryRLL5tQKZg/edit?gid=569531329#gid=569531329) template, functionally test opening and closing nav stacks, along with navigating back, in every combination.

## Dev notes

These changes continue the discussion from #2594 ([comment](https://github.com/IDEMSInternational/open-app-builder/pull/2594#issuecomment-2561025239)). I had originally thought that I would just dismiss the nav-stacks and allow the navigation to leave the user on the previous page from history (as we do with non-nav-stack pop-ups), however it seemed trivial to add a line to programmatically navigate forwards in history after the back action, leaving the user on the page that launched the nav-stack(s). This does result in some screen flickering, but I think that navigating via the device back button in this way can be considered reasonably niche, and remaining on the same top-level template seems like a better user experience. But if there's a good reason not to do this it can be removed to leave the rest of the functionality intact.

## Git Issues

Closes #2594

## Screenshots/Videos


chrismclarke (changes requested):
@jfmcquade 
When you get a chance could you check out https://github.com/IDEMSInternational/open-app-builder/pull/2656 and confirm whether that pr would be sufficient to replace this one, or whether we want to merge some of the code

PR #2656, opened 2024-12-27
Feat: nav stack back native

chrismclarke:
PR Checklist

## Description
Intercepts hardware back-button presses (android only), to close the top nav stack (if open)

## Author Notes
The work initially included handling back-button press in browser, but quickly became over-complicated and messy. As such pressing the browser back button will still not close nav stacks and just navigate the underlying template. It's expected that users on browsers will close nav stacks using the close button or clicking outside of the stack. 

On android, however, the hardware back button functions independently the the rest of the navigation system, and so it is easier to intercept hardware back-button presses and add custom functionality. In this case pressing the hardware back button will close the top nav-stack if open. 

## Review Notes
The feature is only enabled on android, so requires local build/run on an android emulator to test

## Dev Notes
The code started from #2594 to keep the improved handling of tracking nav stack arrays, although the custom navigation code was removed and tracking array converted to signals.

A new service has been added to better-encapsulate the custom back-button interception. This could later be extended if adding further functionality on web.

Testing on device I noticed that by default capacitor/angular appears to disable default back-button navigation behaviour when a custom interceptor added, meaning that repeated back button presses appear to close stack, followed by nav-back and finally minimise. I think this is the behaviour we want, although just to note that this hasn't been explicitly coded in this PR (could be done if we notice things not functioning as expected on particular devices or in future capacitor versions).

I'm not sure if this would supercede/close #2655 or not - both PRs were worked on at a similar time in parallel so unclear whether there is still functionality in 2655 that would be nice to have or not

## Git Issues

Closes #2594
Closes #2569

## Screenshots/Videos

[2024-12-27 10-50-10.webm](https://github.com/user-attachments/assets/6627612b-e814-48dd-8036-8f76544b0894)


jfmcquade (commented):
Thanks, @chrismclarke. Your changes seem good, I like the use of signals and nice to have the back button handling logic separated out within the nav-stack feature folder.

> Testing on device I noticed that by default capacitor/angular appears to disable default back-button navigation behaviour when a custom interceptor added, meaning that repeated back button presses appear to close stack, followed by nav-back and finally minimise. I think this is the behaviour we want, although just to note that this hasn't been explicitly coded in this PR (could be done if we notice things not functioning as expected on particular devices or in future capacitor versions).

Noted.

I've added one commit, https://github.com/IDEMSInternational/open-app-builder/pull/2656/commits/1b40fc3beb5fefb65405f4d1f34c7ce3dc048ced, to make the `closeAllNavStacks()` method close all open nav stacks simultaneously (closing them sequentially was a compromise to support the logic in #2594).

In terms of other changes from #2655, there is still the issue, as you note, of users pressing the browser back button and the underlying template navigating without closing any open nav-stacks. Whilst it's true that we can expect users to close nav-stacks via in-app means, I think that it would be good to handle this behaviour if it's not too difficult to do so. 

I've opened #2659 which proposes incorporating the back button listener logic from #2655 into the new nav-stack-back service. Let me know if you think that's desirable or if the logic is too messy to include.

chrismclarke:
> I've added one commit, [1b40fc3](https://github.com/IDEMSInternational/open-app-builder/commit/1b40fc3beb5fefb65405f4d1f34c7ce3dc048ced), to make the `closeAllNavStacks()` method close all open nav stacks simultaneously (closing them sequentially was a compromise to support the logic in #2594).

Thanks for adding that back, sounds good

> In terms of other changes from #2655, there is still the issue, as you note, of users pressing the browser back button and the underlying template navigating without closing any open nav-stacks. Whilst it's true that we can expect users to close nav-stacks via in-app means, I think that it would be good to handle this behaviour if it's not too difficult to do so.
> 
> I've opened #2659 which proposes incorporating the back button listener logic from #2655 into the new nav-stack-back service. Let me know if you think that's desirable or if the logic is too messy to include.

I'm still not entirely convinced the forward nav is the best way to proceed, so I think might be better to address separately as a follow-up. There could be some angular-specific handling for router event interception/handling examples below:
https://stackoverflow.com/questions/42808818/is-it-possible-to-stop-router-navigation-based-on-some-condition

So I'd say if happy with the state of this PR might be best to review/merge as is, then retarget #2659 as a spike PR on master and iterate from there

jfmcquade (approved):

jfmcquade:
> So I'd say if happy with the state of this PR might be best to review/merge as is, then retarget #2659 as a spike PR on master and iterate from there

OK, sounds good.

@esmeetewinkel I've removed the request for your review as I don't think you'll be able to test this on Android. To summarise our converstion:
- This PR handles the Android hardware back button (including the newer swipe gesture) such that it will close the top nav-stack without navigating back on the underlying template
- Handling the browser back button is more complicated and has not yet been implemented. Issue: #2665. So for now, the browser back button will navigate back on the underlying template without closing the nav-stack(s). I believe this is the same behaviour as we have for pop-ups.

PR #2659, opened 2024-12-30
review: PR 2656 proposed changes

jfmcquade:
PR Checklist

## Description

Incorporates logic originally from #2655 to handle closing all nav stacks when the browser back button is pressed.

## Git Issues

Closes #

## Screenshots/Videos


chrismclarke (changes requested):
The forward nav is a neat solution, although I'm wondering if there might still be ways to avoid the flicker if looking more directly into the angular router. 

There might be some examples in https://stackoverflow.com/questions/42808818/is-it-possible-to-stop-router-navigation-based-on-some-condition

So for now I think hold off integrating browser-back behaviour as a lower-priority issue pending possible smoother fix

PR #2645, opened 2024-12-20
fix: rtl language config is applied to app launch templates

jfmcquade:
PR Checklist

- [x] PR title descriptive (can be used in release notes)

## Description

Fixes a bug where templates launched as part of app launch actions would not have the app's RTL config applied to them.

The template [debug_launch_actions](https://docs.google.com/spreadsheets/d/1zofHkaZ9NMcUgM7U7fPXxclVw4qXN8crHkXcQf-HCr8/edit?gid=1541644091#gid=1541644091) has been updated such that launch actions can be tested from the debug deployment. On first app launch, you will be prompted to select a language. This will not apply on subsequent app launches, until the app data is reset.

### Recreating the issue

On `master`, run a sync to pull the latest changes to [debug_launch_actions](https://docs.google.com/spreadsheets/d/1zofHkaZ9NMcUgM7U7fPXxclVw4qXN8crHkXcQf-HCr8/edit?gid=1541644091#gid=1541644091). Run the app and confirm that the bug documented in #2615 is present for the language select template that is displayed on launch (you may need to reset the app data if this template is not displayed)

### Testing the resolution

Now checkout this PR feature branch and perform the same test. The issue should be resolved.

## Git Issues

Closes #2615

## Screenshots/Videos

Updated [debug_launch_actions](https://docs.google.com/spreadsheets/d/1zofHkaZ9NMcUgM7U7fPXxclVw4qXN8crHkXcQf-HCr8/edit?gid=1541644091#gid=1541644091)

Screenshot 2024-12-20 at 12 51 19

https://github.com/user-attachments/assets/dd1a32a5-ec6e-4ebb-aacf-2f62f2af3c75

PR #2660, opened 2024-12-30
fix: pop-up close button position for RTL languages

### LINKED ISSUE #2618: [BUG] Pop-up close icon should sit on the left in RTL
**Describe the bug**
When a  RTL language is selected, the pop-up close button (both in a regular and full-screen pop up) should sit in the top left rather than the top right. This also applies to the transcript pop up in the audio player. 

**App version**
v0.17.0

**To reproduce**
Launch regular and full screen pop ups on `example_pop_ups` with language set to `en_rtl`

**Deployment Sheet(s)**
e.g. `plh_kids_kw` audio player pop ups

**Debug Sheet(s)**
`example_pop_ups` [template](https://docs.google.com/spreadsheets/d/1K9JZ9Glnj73CFJmpfGIRa5dRA2TBxSsJmaguV6GL8Pc/edit?gid=402992224#gid=402992224) and [web preview](https://idems-debug.web.app/template/example_pop_ups) with 

**Screenshots and videos**
image   image

jfmcquade:
PR Checklist

## Description

Handles repositioning the pop-up close button when the app language direction is RTL. Changes applied to both fullscreen and non-fullscreen pop-ups.

## Git Issues

Closes #2618

## Screenshots/Videos

Pop-ups launched from [example_pop_ups](https://docs.google.com/spreadsheets/d/1K9JZ9Glnj73CFJmpfGIRa5dRA2TBxSsJmaguV6GL8Pc/edit?gid=402992224#gid=402992224):

Screenshot 2024-12-30 at 12 35 08

Screenshot 2024-12-30 at 12 35 39

Screenshot 2024-12-30 at 12 35 55

### DIFF
--- a/src/app/shared/components/template/components/layout/popup/popup.component.html
+++ b/src/app/shared/components/template/components/layout/popup/popup.component.html
@@ -6,9 +6,16 @@
   [attr.data-variant]="props.variant ? props.variant : null"
 >
   <div class="popup-container">
-    <div (click)="dismiss()" class="close-button" fill="clear" *ngIf="props.showCloseButton">
-      <ion-icon slot="icon-only" name="close"></ion-icon>
-    </div>
+    @if (props.showCloseButton) {
+      <div
+        (click)="dismiss()"
+        class="close-button"
+        fill="clear"
+        [attr.data-language-direction]="templateTranslateService.languageDirection()"
+      >
+        <ion-icon slot="icon-only" name="close"></ion-icon>
+      </div>
+    }
     <div class="popup-content" [attr.data-fullscreen]="props.fullscreen ? true : null">
       @if (props.popupText) {
         <!-- Use a text component row to display popup text -->
--- a/src/app/shared/components/template/components/layout/popup/popup.component.scss
+++ b/src/app/shared/components/template/components/layout/popup/popup.component.scss
@@ -21,6 +21,10 @@
     .close-button {
       top: 10px;
       right: 10px;
+      &[data-language-direction~="rtl"] {
+        right: unset;
+        left: 10px;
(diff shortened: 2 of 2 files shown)


esmeetewinkel (approved):
Functional test passed on PR preview link

chrismclarke (changes requested):
Nice to have things displaying correctly, although I think it would be good if possible to find a solution that doesn't require detecting ltr/rtl so much at the component level. 

I first attempted passing down a CSS custom property from the app component that defined an `--is-rtl` property and conditionally style using the new(ish) container style queries syntax to apply CSS conditionally based on the inherited property:
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_size_and_style_queries

This did work, although I then realised it's probably better to just use insets in a similar way to what you did on #2661 as insets should automatically change with text direction

So I've opened https://github.com/IDEMSInternational/open-app-builder/pull/2664 targeting this branch. Let me know what you think @jfmcquade

jfmcquade:
Thanks @chrismclarke, I've merged your suggestions with the caveat that the pop-up component does need to set its own `dir` attribute, as it is sometimes rendered outside of the main app component, see #2645.

> I first attempted passing down a CSS custom property from the app component that defined an `--is-rtl` property and conditionally style using the new(ish) container style queries syntax to apply CSS conditionally based on the inherited property: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_size_and_style_queries

This is a good pattern that I may implement going forwards if it's appropriate for outstanding RTL issues.

chrismclarke (approved):

PR #2663, opened 2024-12-30
feat: data_items triggered actions

chrismclarke:
PR Checklist

## Description

Add a custom `data_changed` action trigger that updates in sync with data_items loops.
This can be used to set local or field values depending on data queries

**Follow-up Tasks**
- [ ] Add support for triggered actions when data_items loop includes child rows
- [ ] Update data_actions RFC with reference to changes made here

## Author Notes

The action list is defined in the normal way, with the exception that the `@items` variable will contain the value of the associated item list

| type             | value    | action_list                                                 | parameter_list          |
|------------------|----------|-------------------------------------------------------------|-------------------------|
| begin_data_items | my_items | data_changed \| set_local : total_completed : @items.length | filter: @item.completed |
| end_data_items   |          |                                                             |                         |

See examples in [component_data_items](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892)

**NOTE 1** - If the begin/end data_items has inner rows used for templating, the list of `@items` will include the loop of generated rows. E.g. If a data_list has 3 rows and there are 2 child text/buttons in the loop the list will represent 6 items.

So for now it is recommended to leave any data_items loops free of inner data which returns just the raw data_list data (will be fixed in a future PR).

**NOTE 2** - A new `data_changed` trigger has been included instead of using the existing `changed` as the action is processed differently (includes extra evaluation step for items context), and may be further differentiated in the future as needs arise (easy to make changes in isolation).

**NOTE 3** - The action hasn't been tested with more complex use cases such/mixed contexts, e.g. 
```
data_changed | set_local : @local.field_name_look_up : @items.length
``` 
So for now it's designed more around setting specific locals/fields based on the list values, however could be further developed as required

## Review Notes
See examples in [component_data_items](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892)

Tests can be run from the updated specs

## Dev Notes
It's noted in a couple places that there is currently a limitation due to the way the data-items service only emits a stream of generated loop item rows (with fallback to data_list rows if no inner child rows exist). I'm planning to address this in a follow-up that provides different emits for the list rows vs generated item rows, but it has a few knock-ons so I figured best to omit from this PR

When the data_items action trigger a local/field update that automatically triggers reprocessing template rows, which in turn triggers the data_items subscription to update. This resulted in an infinite loop of data updates and action triggers. The issue has been resolved by updating the service subscription to include `distinctUntilChanged` which allows passing our custom equality checker function that will only re-emit when data has changed. It would also usually be possible to include within signal/computed code, however the `toSignal` operator does not allow this (and figured it makes sense to include at the service level anyways)

## Git Issues

Closes #

## Screenshots/Videos

https://github.com/user-attachments/assets/4743a902-40bc-4701-9d94-307c5c30c0d9

### DIFF
--- a/packages/data-models/flowTypes.ts
+++ b/packages/data-models/flowTypes.ts
@@ -391,6 +391,7 @@ export namespace FlowTypes {
     | "changed"
     | "click"
     | "completed"
+    | "data_changed"
     | "info_click"
     | "nav_resume" // return to template after navigation or popup close;
     | "sent" // notification sent
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/flowParser.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/flowParser.ts
@@ -4,7 +4,7 @@ import { IConverterPaths, IFlowHashmapByType, IParsedWorkbookData } from "../../
 import { arrayToHashmap, groupJsonByKey, IContentsEntry } from "../../utils";
 import BaseProcessor from "../base";
 
-const cacheVersion = 20241203.0;
+const cacheVersion = 20241230.0;
 
 export class FlowParserProcessor extends BaseProcessor<FlowTypes.FlowTypeWithData> {
   public parsers: { [flowType in FlowTypes.FlowType]: Parsers.DefaultParser } = {
--- a/packages/scripts/src/commands/app-data/convert/utils/app-data-action.utils.ts
+++ b/packages/scripts/src/commands/app-data/convert/utils/app-data-action.utils.ts
@@ -27,6 +27,7 @@ export function parseAppDataActionString(actionString: string): FlowTypes.Templa
     changed: true,
     click: true,
     completed: true,
+    data_changed: true,
     uncompleted: true,
     nav_resume: true,
     sent: true,
--- a/src/app/shared/components/template/components/base.ts
+++ b/src/app/shared/components/template/components/base.ts
@@ -25,7 +25,9 @@ export class TemplateBaseComponent implements ITemplateRowProps {
   rowSignal = signal<FlowTypes.TemplateRow>(undefined, { equal: isEqual });
   value = computed(() => this.rowSignal().value, { equal: isEqual });
   parameterList = computed(() => this.rowSignal().parameter_list || {}, { equal: isEqual });
-  actionList = computed(() => this.rowSignal().action_list || [], { equal: isEqual });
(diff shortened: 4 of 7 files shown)


esmeetewinkel (commented):
Thanks. Reading through the notes and seeing the screen video this definitely addresses some of the use cases I have. One clarification: this action triggers _when the template is rendered_ as well as whenever the data changes, right? That's what I need (but I thought the `changed` action only triggers when there is a change, not when the template is rendered - I might be wrong - so the naming is slightly confusing).

I somehow can't get this to work (like in your screen video) neither on the web preview nor when running locally from this branch - both the examples show as "undefined".


https://github.com/user-attachments/assets/34bade58-aa07-46d4-aaa4-6ea76cdf261d

chrismclarke:
> Thanks. Reading through the notes and seeing the screen video this definitely addresses some of the use cases I have. One clarification: this action triggers _when the template is rendered_ as well as whenever the data changes, right? That's what I need (but I thought the `changed` action only triggers when there is a change, not when the template is rendered - I might be wrong - so the naming is slightly confusing).

Yes it does trigger on template load due to the fact that the data goes from being `undefined` (initial data takes time to load) to the first value loaded. The actions are not triggered when data is undefined, but will trigger once for the first load. It would be possible to have separate triggers for initial load and subsequent changes (e.g. `data_loaded` and `data_changed`), however I figured simpler for now to just keep one action trigger

> I somehow can't get this to work (like in your screen video) neither on the web preview nor when running locally from this branch - both the examples show as "undefined".

Weird. Web preview won't work for sure as it won't have the latest content synced, however locally should. Might be worth checking you also have the latest content synced locally as I remember I had similar undefined states with an earlier content version too

esmeetewinkel (approved):
Functional test passed. I can confirm this works when first running a content sync of the debug deployment on this PR branch (I hadn't understood that the parsed jsons would be different when syncing from this PR branch, makes sense it didn't work when I tried before...)

image

esmeetewinkel:
Merging now since needed imminently and functional test passed. Agreed with @jfmcquade he'll do code review post merge.

jfmcquade (commented):
Looks good. Appreciate the current limitations, but great to have this functionality

PR #2664, opened 2024-12-31
review: pop-up-button-alt

chrismclarke:
PR Checklist

## Description
Targets branch of #2660 

Prefers use of inset instead of specific rtl/ltr detection and modifications

## Review Notes
To confirm fixes the same issue as target pr

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade (approved):
Thanks, @chrismclarke, I agree this is a nicer solution. However we do need the explicit setting of the ionic `dir` attribute in this component, see #2645, I've reinstated with https://github.com/IDEMSInternational/open-app-builder/pull/2664/commits/fd15e1f3ca0c20370045e29d54105d0436fb1cbb

PR #2666, opened 2024-12-31
Feat!: Auth user db sync

chrismclarke:
PR Checklist

## Breaking Changes
The contact field `_app_auth_user` has been renamed to `auth_user_id` and is no longer formatted as json but instead just an id. I'm not sure if this protected field was included in any templates beyond debug ones. It has been updated in [example_google_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705)

The current `app_user_id` remains unchanged

## Description
Adds an `auth_users` endpoint to the server that can be used to retrieve a user by a given `auth_user_id`. It also creates an explicit `auth_user_id` column within the `app_users` table to make it more efficient to retrieve (instead of looking at contact_fields)

Updates server sync code to include `auth_user_id` when available

These server updates are required to then enable lookup and restore of a given user by their auth_id (instead of device id)

**Follow-ups**

- [ ] Deploy changes to server
- [ ] Add frontend code to lookup auth user from db and restore as required

## Author Notes
The updates here are mostly at the server level and will not work until the server is updated post-merge.

I had thought about exposing action triggers to better-enable authors to update templates based on auth events (e.g. `auth_signed_in` and `auth_signed_out`), however this started to get a little bit complicated so for now I've manually updated it so whenever a user signs in the `server_sync` action is immediately triggered to ensure the signed in user profile is backed up to the db. If wanting to have more fine-grained support at an authoring level can create a follow-up issue.

## Review Notes
If testing endpoints locally will need locally running postgres and API
```sh
yarn workspace api start
```
A local deployment will also need to be set to target the locally running endpoint
```ts
config.api.endpoint = 'http://localhost:3000'
```

Alternatively can build API and run in docker container following instructions from the `api` and `server` package readmes

Once running performing a user sync from the `debug_sync_id` template or `/user` debug page. 
You should see the user populated without an `auth_user_id`

Sign_in either from the _example_google_auth_ template and trigger sync and should see db populated with `auth_user_id` property

Examples in screenshots below

## Dev Notes
The `app_auth_user` property has been removed in favour of `auth_user_id` which is namespaced better and also stores a simple string representation instead of JSON. 

There is a small amount of placeholder code for the start of the user profile restore frontend feature, although for now this is still a work-in-progress and can be ignored for review

## Git Issues

Closes #

## Screenshots/Videos
**Example sync non-auth user to API running locally**
image

**Note that the response now includes an auth_user_id (null)**
image

**Example sync post-login - includes `auth_user_id` response from server**
image

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "api",
-  "version": "1.5.1",
+  "version": "1.6.0",
   "scripts": {
     "prebuild": "rimraf dist",
     "build": "nest build",
--- a/packages/api/spec-export.json
+++ b/packages/api/spec-export.json
@@ -130,6 +130,39 @@
         ]
       }
     },
+    "/auth_users/{auth_user_id}": {
+      "get": {
+        "operationId": "AuthUsersController_findOne",
+        "summary": "Get auth user profile",
+        "parameters": [
+          {
+            "name": "auth_user_id",
+            "required": true,
+            "in": "path",
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
+            "name": "x-deployment-db-name",
+            "in": "header",
+            "description": "Name of db for deployment to populate",
+            "schema": {
+              "type": "string",
+              "default": "plh"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": ""
+          }
+        },
+        "tags": [
+          "Auth Users"
+        ]
+      }
+    },
     "/contact_fields": {
       "get": {
         "operationId": "ContactFieldController_findAll",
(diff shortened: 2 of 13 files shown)


esmeetewinkel (approved):
Looks good, I'm not concerned about the breaking change - don't think any deployment is using this yet.

I'll leave testing to @jfmcquade

jfmcquade (approved):
I've been having an issue with my local postgres installation after making some local changes to where I install packages. I haven't been able to run the api locally and test, and need to finish for the day. 

The code and description look good so I'm happy for this to be merged to unblock the continuing work.

PR #2669, opened 2025-01-02
Kuwait Theme: Map View Styling

### LINKED ISSUE #2667: [FEATURE] Kuwait Theme - Map View
**What?**
1. Handle locked status of modules in map view. This should be similar to the way we handle the locked status in the list view.
2. Rework the map view so that the path is more curvy and compressed.
3. Display image behind the map view. This relates to PR [https://github.com/IDEMSInternational/open-app-builder/pull/2589](https://notifications.googleapis.com/email/redirect?t=AFG8qyVBbec3qfjmTS-TVnuH12HiK_eYYsiuQx-aPCFHyuL2P1SpMH3qp7Yu53QeboT81v0nsFrTAnK0NCqFi2VGq4mzaR7tto4VIxTZH-yeR5JIJybZbxZb5iQHtuFqRiBSTeqNiFDluoIm6XnV4lbGeH2dDcInuB89hUOAUiSRt8sj1ArC7YlVz-OPLTkpbYsSk2IJx03152GB6NXPOKUvicYiTpYNEangF_kc4MMOEG8cCcONYRaDxsHLMHCIKwfsRsk&r=eJzLKCkpKLbS10_PLMkoTdJLzs_V93Rx9Q32zCtJLcpLLMnMz0vM0c8vSM3TTSwo0E0qzcxJSS3SLyjNydE3MrWwBAD68hcL&s=ALHZ2r5-f2rhCHUapgicj6UgMSLr) but might need some tweaking once we have (2).
4. Page scroll starting at bottom rather than top (and order modules accordingly). This is a nontrivial piece of work - I think it might be best to stick to top-to-bottom scroll for now.

**How?**

- Style locked modules on map-view
- Style curvy paths

**Other details**

FaithDaka:
PR Checklist

## Description
- Styling support added for a curved map path

## Author Notes
- The Kuwait map design has paths that wind a lot more than the `wavy` variant accommodates. A new variant, `curved`, is added in this update to support curvier paths.
- The task card now includes an additional attribute, `locked_image_asset`, to accommodate the locked icon that appears on top of the module.

Attributes added to [sheet](https://docs.google.com/spreadsheets/d/14BBKG07nqUWDmcscdUMv1AKYSWGBUPCNDT-VcSERHyU/edit?usp=sharing):
- `variant`: "curved"
- `locked_image_asset`

## Git Issues

Closes #2667 

## Screenshots

To visualise the map path, the color stroke here is dark

Otherwise the map path is a light color to contrast with the proposed background image

A `completed` module is highlighted with a blue border while an `in_progress` module is highlighted with a yellow border

### DIFF
--- a/src/app/shared/components/template/components/progress-path/progress-path.component.ts
+++ b/src/app/shared/components/template/components/progress-path/progress-path.component.ts
@@ -5,7 +5,7 @@ import { TemplateTranslateService } from "../../services/template-translate.serv
 
 interface IProgressPathParams {
   /** TEMPLATE_PARAMETER: "variant". Default "wavy" */
-  variant: "basic" | "wavy";
+  variant: "basic" | "wavy" | "curved";
 }
 
 // HACK - hardcoded sizing values to make content fit reasonably well
@@ -27,7 +27,7 @@ const SIZING = {
 })
 export class TmplProgressPathComponent extends TemplateBaseComponent implements OnInit {
   private params: Partial<IProgressPathParams> = {};
-  private pathVariant: "basic" | "wavy";
+  private pathVariant: "basic" | "wavy" | "curved";
 
   public svgPath: string;
   public svgViewBox: string;
@@ -47,14 +47,14 @@ export class TmplProgressPathComponent extends TemplateBaseComponent implements 
     this.params.variant = getStringParamFromTemplateRow(this._row, "variant", "wavy")
       .split(",")
       .join(" ") as IProgressPathParams["variant"];
-    this.pathVariant = this.params.variant.includes("basic") ? "basic" : "wavy";
+    this.pathVariant = this.params.variant;
   }
 
   /**
    * Generate a base SVG segment used to connect 2 progress items together
    * Roughly a horizontal line and smooth bend, adjusted for sizing
    */
-  private generateSVGPath(variant: "basic" | "wavy" = "wavy") {
+  private generateSVGPath(variant: "basic" | "curved" | "wavy" = "wavy") {
     // arbitrary values used to make base width/height fit
     const { widthPx, xOffset, yOffset, textContentHeight } = SIZING;
 
@@ -87,7 +87,14 @@ export class TmplProgressPathComponent extends TemplateBaseComponent implements 
     c 48,0 72,64 48,${viewboxHeight - yOffset - 4}
     `.trim();
 
-    this.svgPath = variant === "basic" ? basic() : wavy();
+    const curved = () =>
+      `
+    M ${xOffset},${yOffset}
+    c 0,140 280,-80 252,160 1
+    v 80
+    `.trim();
+
+    this.svgPath = variant === "basic" ? basic() : variant === "wavy" ? wavy() : curved();
     this.svgViewBox = `0 0 ${widthPx} ${viewboxHeight}`;
     this.contentHeight = `${textContentHeight}px`;
   }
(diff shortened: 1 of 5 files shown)


FaithDaka:
For your review @jfmcquade @esmeetewinkel

jfmcquade:
The new path style looks nice. However, it doesn't reflect the design and so won't line up with the background image in the same way. I think that will have to be addressed in the future as a follow-up.

I've tidied and clarified some of the logic with https://github.com/IDEMSInternational/open-app-builder/pull/2669/commits/b7263f4dfb9307f73731886a629d4b9e368f12fe. 

Additionally, I think we'll need to handle the "locked" status differently. @esmeetewinkel – currently, this PR sets the "locked" status is through the same logic that determines the completion status of the task group: namely, if the task group is "not started" (i.e. none of its constituent tasks are completed), then the task group is considered "locked". However, this doesn't allow us to handle unlocking task groups. If the task group is not clickable when it is locked, then there I don't believe there is a way for the user to begin that task group. So I think we need to handle the "locked" status via a dedicated parameter, and handle the logic elsewhere (e.g. in a dedicated `locked` column that gets updated through template logic). Does that sound right to you @esmeetewinkel? If so, I can update this PR.

jfmcquade (changes requested):
The new path style looks nice. However, it doesn't reflect the design and so won't line up with the background image in the same way. I think that will have to be addressed in the future as a follow-up.

I've tidied and clarified some of the logic with https://github.com/IDEMSInternational/open-app-builder/pull/2669/commits/b7263f4dfb9307f73731886a629d4b9e368f12fe. 

Additionally, I think we'll need to handle the "locked" status differently. @esmeetewinkel – currently, this PR sets the "locked" status is through the same logic that determines the completion status of the task group: namely, if the task group is "not started" (i.e. none of its constituent tasks are completed), then the task group is considered "locked". However, this doesn't allow us to handle unlocking task groups. If the task group is not clickable when it is locked, then there I don't believe there is a way for the user to begin that task group. So I think we need to handle the "locked" status via a dedicated parameter, and handle the logic elsewhere (e.g. in a dedicated `locked` column that gets updated through template logic). Does that sound right to you @esmeetewinkel? If so, I can update this PR.

esmeetewinkel:
> (e.g. in a dedicated locked column that gets updated through template logic). Does that sound right to you @esmeetewinkel?

Yes that sounds right

jfmcquade (approved):
OK @esmeetewinkel, I've updated to expose an explicit `locked` param. See final example in updated [comp_progress_path](https://docs.google.com/spreadsheets/d/14BBKG07nqUWDmcscdUMv1AKYSWGBUPCNDT-VcSERHyU/edit?gid=569531329#gid=569531329) template

Screenshot 2025-01-03 at 19 09 23

Screenshot 2025-01-03 at 19 09 28

Screenshot 2025-01-03 at 19 08 17

esmeetewinkel (commented):
@FaithDaka / @jfmcquade Is this PR superseded by https://github.com/IDEMSInternational/open-app-builder/pull/2698?

jfmcquade:
> @FaithDaka / @jfmcquade Is this PR superseded by #2698?

Good question, #2698 builds on this PR but doesn't include the changes to the `progress_path` component (it adds a new `plh_progress_path` component). I think it would be best to keep this PR open but convert to draft, then reassess whether the changes are desirable after #2698 is closed.

Issue #2459, opened 2024-10-08
[FEATURE] Retrieve account data through login


esmeetewinkel:
**What**
Enable users to retrieve their account data through logging in.

**Why?**
Required for [PLH Kids Kuwait](https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/issues/53). 

>Story: As a user, if I lose or break my device, I can purchase a new device and login to my ParentApp account in order to pick-up my progress where I left off.
>Users of ParentApp Kuwait need to be able to login to the app so that their progress is automatically saved. They should be able to create an account with their Google ID. Logging in should not be mandatory, but available in settings to allow users to save their progress.

esmeetewinkel:
I added a follow-up issue #2765, closing here since PR #2671 addresses the core of this issue

PR #2671, opened 2025-01-03
feat: user profile restore

chrismclarke:
PR Checklist

## Description
- Add support for profile restore from signed in user
- Update user debug page

## Author Notes
This PR adds the core methods and updates the user debug screen for testing purposes. Follow-up discussion and work will be required to determine how best to expose custom templating to manage user profile restoration

The user debug page can be found at `/user`

## Review Notes
The required server updates have been deployed so all should work when running locally from this branch. See demo below for example using 2 devices and restoring profile from first onto second.

## Dev Notes
One additional knock-on was how best to handle protected fields, which are by default not imported/restored. I've opted for manually specifying a list of fields which we do want to restore (e.g. app theme, skin etc.), however in the future we may want a better solution. Also would be good to check the list to see if it makes sense.

## Git Issues

This addresses the core of #2459, however does not close all the related-issues mentioned (not sure whether to close or not...)

## Screenshots/Videos

**Demo - Profile Restore**

Uses one regular browser tab and one incognito to mimic multiple devices. Initial device is used with app, logged in and synced. Second device logs in and imports data to continue from where first device left off

_Note 1:_ There is an additional profile for restore from previous testing

_Note 2:_ Ignore the infinite mirrors. Had to use full desktop capture with OBS to show both windows and running on a single monitor :'(

https://github.com/user-attachments/assets/41009135-f30d-4def-ac9c-4731aec8643d


src/app/shared/services/auth/auth.service.ts
```diff
@@ -53,12 +66,20 @@ export class AuthService extends AsyncServiceBase {
...
+    // TODO - get type-safe return types using openapi http client
+    const authEntries = await firstValueFrom(
+      this.http
+        .get(`/auth_users/${authUser.uid}`, { responseType: "json" })
+        .pipe(map((v) => (v as IServerUser[]) || []))
+    );
+
+    const currentUserId = this.localStorageService.getProtected("APP_USER_ID");
+    const restoreProfiles = authEntries
+      .filter((v) => v.app_user_id !== currentUserId)
```
jfmcquade, line 80: I can see how there might be a use case for restoring from the same ID (e.g. after factory resetting a single device). But it makes sense to focus on the requested use case for now

jfmcquade (approved):
Looks good, and I was able to complete the same user flow as in your demo. 

I can see how there is further discussion and work needed to expose the functionality to templating. But as I understand it, if we include a sign-in option currently, then add the ability to restore from the server based on auth ID later, then the feature should be available for users who have signed in as soon as they update to the latest app version

esmeetewinkel (approved):
Looks good

esmeetewinkel:
I closed https://github.com/IDEMSInternational/open-app-builder/issues/2459 and added a follow-up issue [#2765](https://github.com/IDEMSInternational/open-app-builder/issues/2765)

PR #2677, opened 2025-01-03
fix: display group handles RTL language direction

jfmcquade:
PR Checklist

## Description

Follow-up to #2589 
- Now, when a display group has a background image set, the image will be flipped horizontally when the language direction is RTL (as requested in [this comment](https://github.com/IDEMSInternational/open-app-builder/pull/2589#pullrequestreview-2511570773)).
  - We may want to expose a parameter to handle this behaviour, as some images may not be intended to be mirrored
- Additionally, fixes an undocumented RTL issue where the contents of a display group would be aligned to the left of the screen when the language was RTL.

## Git Issues

Closes #

## Screenshots/Videos

[feature_display_group](https://docs.google.com/spreadsheets/d/1ARbNRGDer5vj9qSpRMZFrMkYifGkH3TLtDVp72YbaqU/edit?gid=1231713195#gid=1231713195):

LTR:

Screenshot 2025-01-03 at 16 57 27

RTL: 

Screenshot 2025-01-03 at 16 57 06

### DIFF
--- a/src/app/shared/components/template/components/layout/display-group/display-group.component.html
+++ b/src/app/shared/components/template/components/layout/display-group/display-group.component.html
@@ -23,6 +23,7 @@
     [attr.data-param-style]="params.style"
     [attr.data-rowname]="_row.name"
     [attr.data-variant]="params.variant"
+    [attr.data-has-background-image]="params.backgroundImageAsset ? true : false"
     [style.marginBottom.px]="params.offset"
     [ngSwitch]="type"
     [style]="_row.style_list | styleList"
--- a/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
+++ b/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
@@ -28,6 +28,19 @@
   background-color: var(--ion-background-color);
 }
 
+// HACK: Flip background image (but not child content) when language direction is RTL
+:dir(rtl) {
+  .display-group-wrapper {
+    justify-content: flex-end;
+    &[data-has-background-image] {
+      transform: scaleX(-1);
+      > * {
+        transform: scaleX(-1);
+      }
+    }
+  }
+}
+
 // Default spacing within display groups. Adapted from `template-component.scss`
 .display-group-wrapper[data-param-style~="row"] > plh-template-component:not([data-hidden="true"]) {
   margin: 0;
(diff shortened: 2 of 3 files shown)


chrismclarke (changes requested):
I think perhaps this needs to be handled differently as I can imagine cases where the background image includes text that doesn't want to be flipped horizontally (e.g. logos/icons). 

Could we not just use the asset language override system to include a flipped version for specific language codes?

esmeetewinkel:
> Could we not just use the asset language override system to include a flipped version for specific language codes?

I think that might not work exactly, since also an image that is e.g. aligned to the top right would need to become aligned to the top left(?)

jfmcquade:
> I think perhaps this needs to be handled differently as I can imagine cases where the background image includes text that doesn't want to be flipped horizontally (e.g. logos/icons).
> 
> Could we not just use the asset language override system to include a flipped version for specific language codes?

Hmm yes I did consider this but dismissed it for the reason @esmeetewinkel mentions. It's quite a tricky problem to solve, but we do have an immediate use case where it is required.

One option would be to extend this PR's implementation with an exposed property, such as `background_image_flip_on_rtl` or similar, to toggle whether or not the mirroring is applied.

One messy option would be to author the change manually: i.e. author two sets of the display group in question, with only one displaying conditionally based upon the current app language.

esmeetewinkel:
I don't have time to review now - but would be good to test this carefully on the `plh_kids_kw` deployment prior to merge because I - at some point - saw that display groups were no longer swapping as they should for RTL. Not sure this was the culprit...

chrismclarke:
> > Could we not just use the asset language override system to include a flipped version for specific language codes?
> 
> I think that might not work exactly, since also an image that is e.g. aligned to the top right would need to become aligned to the top left(?)

It should still be possible to change the alignment based on language code, either via an override template or some sort of calc statement reading from `@fields._app_language`. Alternatively we could add a new contact field that specifically tracks whether the current language is RTL (e.g. `@fields._app_language_rtl`) which could be used with condition columns to provide multiple combinations of asset + alignment. Or if wanting a single asset can provide a raw CSS override to flip in the same way this PR tries to hardcode - `transform: scaleX(-1);`

In either case I still think better to handle from the authoring, as doing it in code raises more questions about why, for example, all images aren't flipped horizontally when in rtl mode and just the background image. Better to at least be consistent

esmeetewinkel:
@jfmcquade shall we close this PR favouring the workaround that @chrismclarke suggested?

jfmcquade:
> @jfmcquade shall we close this PR favouring the workaround that @chrismclarke suggested?

Yes I think so, it's true that we can't really apply this mirroring universally as it's all content-specific, so it makes sense that it should be handled at a content authoring level. If we decide further tools for doing so would be helpful, we can consider those independently of these proposed changes.

PR #2683, opened 2025-01-06
fix: ios browser nav-stack height

### LINKED ISSUE #2680: [BUG] nav-stack close button not visible on iOS Safari
**Describe the bug**
We received a user report that the close button is missing on a fullscreen pop-up on ios.

After some investigation, it appears that the issue is with nav-stacks, specifically when the web app is accessed via the iOS browser (Safari) – the issue is not present when running the native app.

This issue is due to the fact that the height of the nav-stack is calculated with reference to the values of `--ion-safe-area-top` and `--ion-safe-area-bottom` CSS variables that Ionic exposes (which in turn are based on iOS's `env(safe-area-inset-*)` CSS variables when on iOS, see [ionic code](https://github.com/ionic-team/ionic-framework/blob/0873dc2ffbc623439a74926937a75f54a28b9709/core/src/css/core.scss#L241)). When using the Safari browser on iOS, these values are set to 0px, despite the browser including large UI elements that potentially obscure the content.

https://forums.developer.apple.com/forums/thread/699415

Running the app in the emulator's Safari browser, note that the ``--ion-safe-area-*` variables are set to 0px, and the top of the nav-stack element disappears off the top of the viewable area:

Screenshot 2025-01-06 at 16 22 33

Fullscreen pop-ups themselves are unaffected:

Screenshot 2025-01-06 at 17 14 00

**App version**

**To Reproduce**

**Deployment Sheet(s)**

**Debug Sheet(s)**

**Screenshots and videos**
User-provided screenshot:

fullscreen pop-up ios

**Additional context**

jfmcquade:
PR Checklist

## Description

Fixes an issue where the top of a nav-stack could be cut off when accessing the app via Safari on iOS.

## Testing

iOS appetize [link](https://appetize.io/app/ios/international.idems.debug-app?device=iphone14pro&osVersion=16.2).

## Git Issues

Closes #2680 

## Screenshots/Videos

I've included screenshots of the fullscreen pop-up component, as this is the other component that makes use of the CSS property that has been updated in this PR. The behaviour of fullscreen pop-ups is unchanged as shown here.

|   | Native app | Safari |
|---|---|---|
| nav-stack | nav-stack native app | nav-stack safari |
| fullscreen pop-up | fullscreen pop-up native app | fullscreen pop-up safari |

### DIFF
--- a/src/app/feature/nav-stack/components/nav-stack/nav-stack.component.global.scss
+++ b/src/app/feature/nav-stack/components/nav-stack/nav-stack.component.global.scss
@@ -20,7 +20,7 @@ ion-modal.nav-stack-modal {
   &::part(content) {
     border-radius: var(--border-radius) var(--border-radius) 0 0;
     position: absolute;
-    bottom: 0px;
+    bottom: calc(0px + var(--ion-safe-area-bottom, 0));
   }
 
   ion-toolbar {
--- a/src/theme/variables.scss
+++ b/src/theme/variables.scss
@@ -18,8 +18,10 @@
 
 /** Ionic CSS Variables **/
 :root {
-  // The total height of the central safe area of the device viewport
-  --safe-area-height: calc(100vh - var(--ion-safe-area-top, 0) - var(--ion-safe-area-bottom, 0));
+  // The total height of the central safe area of the device viewport.
+  // Use 100% rather than 100vh to handle browsers with toolbars,
+  // see https://github.com/IDEMSInternational/open-app-builder/issues/2680
+  --safe-area-height: calc(100% - var(--ion-safe-area-top, 0) - var(--ion-safe-area-bottom, 0));
 
   // Margins //
   --row-margin-top: 1em; // used in downstream calculations, e.g. sticky display-group


chrismclarke (approved):

esmeetewinkel (approved):
I don't think I have a way to functionally test on Safari on iOS. 

I reviewed the Appetize link, where I noticed that the stack cannot be closed by clicking above/outside the stack (on the app header). Is that the intended behaviour? I'm pretty sure that does (or did use to) work on web preview.

jfmcquade:
> I don't think I have a way to functionally test on Safari on iOS.

Good point, adding a build on appetize doesn't help with this. I can publish a web preview then you can open an emulator in appetize and navigate to the link in Safari manually.

> I reviewed the Appetize link, where I noticed that the stack cannot be closed by clicking above/outside the stack (on the app header). Is that the intended behaviour? I'm pretty sure that does (or did use to) work on web preview.

You're right, I think this is an additional bug on iOS, I'll make an issue

esmeetewinkel:
> Good point, adding a build on appetize doesn't help with this. I can publish a web preview then you can open an emulator in appetize and navigate to the link in Safari manually.

Ah yes that works. Functional test passed on Safari on Appetize link.

image

PR #2688, opened 2025-01-07
fix: iOS fullscreen popups

jfmcquade:
PR Checklist

- [x] PR title descriptive (can be used in release notes)

## Description

Fixes an issue where the content of fullscreen pop-ups may not all be visible when accessing the web app via Safari on iOS. Whilst I haven't been able to recreate this issue more generally, it's possible that it was also occurring when accessing the native iOS app, which should also be fixed by these changes.

## Testing

As these changes refactor some of the styling for pop-ups in general, we should test every pop-up type for every platform. I have attempted this in the videos embedded below. In each case, we should be ensuring that all the content is visible (particularly relevant for scrolling content) and that all the spacing looks correct – e.g. the space around the non-fullscreen pop-up element; the positioning of the close button.

Pop-up types:
- Available via [example_pop_ups](https://docs.google.com/spreadsheets/d/1K9JZ9Glnj73CFJmpfGIRa5dRA2TBxSsJmaguV6GL8Pc/edit?gid=402992224#gid=402992224):
  - pop-up with short content
  - pop-up with long, scrollable content
  - fullscreen pop-up with short content
  - fullscreen pop-up with long, scrollable content
- Available on initial app launch (after reset)
  - Fullscreen pop-up launched as part of launch actions

Platforms:
- Desktop browser
- iOS (in theory should test different device types with different notches/cut-outs etc.) - [appetize](https://appetize.io/app/ios/international.idems.debug-app?device=iphone14pro&osVersion=16.2)
  - native app
  - web app via Safari
- Android - [appetize](https://appetize.io/app/android/international.idems.debug_app?device=pixel7&osVersion=13.0)
  - native app
  - web app via Chrome

## Dev notes

Even with live reloading, the developer experience of developing a web app for iOS is very frustrating. There appear to be differences in how CSS is interpreted depending on whether the app is accessed through Safari or the bundled Capacitor app. This includes the fact that the safe area does not get set correctly in Safari (see #2680), but extends to more general issues too. Sometimes, some elements would display incorrectly, but would be fixed when I enabled and disabled a CSS rule via the desktop Safari developer panel linked to the app in my emulator. Sometimes, the safe area would be incorrect until I quit and relaunched the emulator.

Hopefully these issues are related to the emulator itself, and would not be present for an end user.

## Git Issues

Closes #2687

### Desktop browser

https://github.com/user-attachments/assets/360fe484-1975-4302-bc58-0f84dfcf1604

### iOS (local emulator)
[...]

PR #2693, opened 2025-01-08
fix: handle safe area when app header is hidden

### LINKED ISSUE #2692: [BUG] iOS safe area is not respected when header is hidden
**Describe the bug**
On iOS, when the app header is hidden (i.e. `app_config.APP_HEADER_DEFAULTS.hidden` is `true`), the top of the content is cut off behind the top unsafe area of the screen.

**App version**
Present in v0.17.0. Likely introduced by #2631 but not tested on iOS

**To Reproduce**

**Deployment Sheet(s)**

**Debug Sheet(s)**
[feat_app_layout_no_header](https://docs.google.com/spreadsheets/d/1Z46gTluTPeffqpiAZMapYo0imm59GU03j0sW9UChvPM/edit?gid=1912123117#gid=1912123117), [feat_app_layout_inline_header](https://docs.google.com/spreadsheets/d/1Z46gTluTPeffqpiAZMapYo0imm59GU03j0sW9UChvPM/edit?gid=618720041#gid=618720041)

**Screenshots and videos**

Screenshot 2025-01-07 at 17 48 01

**Additional context**

Ionic handles the device safe area natively, to an extent. For example, padding is applied to certain elements to account for the top and bottom unsafe areas ([docs](https://ionicframework.com/docs/theming/advanced#safe-area-padding)). This includes the `<ion-toolbar>` element that we use in the header and footer, so the header automatically adjusts to fill the unsafe area at the top of the screen on devices with a notch, for example. However, as this padding is only applied to certain elements and not the content as a whole, when we don't render the app header with its included `ion-toolbar`, there is no automatic handling of the safe area.

A possible approach to resolution is adding the following to `capacitor.config.ts`, see https://stackoverflow.com/a/74106265:
```ts
  ios: {
    contentInset: "always"
  },
```
Also worth trying setting padding-top on all content to `var(--ion-safe-area-top)` when heading is hidden

jfmcquade:
PR Checklist

## Description

Fixes an issue where content would spill out of the "safe area" ("the section [of the display] that is not covered by the device's notch, status bar, or other elements that are part of the device's UI and not the app's" – [docs](https://ionicframework.com/docs/theming/advanced#safe-area-padding)) when the app header is hidden.

These changes should not have an effect on the web app accessed through Safari on iOS, as no safe area is applied in this case (see #2688).

#2631 added functionality to hide the app header and the issue has likely been present since then. Really we should be doing more testing on iOS in general (and multiple iOS devices), but this is time consuming and difficult to prioritise when deadlines are tight. In future, we should at least make sure to do additional testing on iOS for any changes that may affect interactions with the safe area.

## Testing

Test [feat_app_layout_inline_header](https://docs.google.com/spreadsheets/d/1Z46gTluTPeffqpiAZMapYo0imm59GU03j0sW9UChvPM/edit?gid=618720041#gid=618720041), and other linked templates

Appetize links:
- [ios](https://appetize.io/apps/ios/international.idems.debug-app)
- [android](https://appetize.io/apps/android/international.idems.debug_app)

## Dev notes

Ionic handles the device safe area natively, to an extent. For example, padding is applied to certain elements to account for the top and bottom unsafe areas ([docs](https://ionicframework.com/docs/theming/advanced#safe-area-padding)). This includes the `<ion-toolbar>` element that we use in the header and footer, so the header automatically adjusts to fill the unsafe area at the top of the screen on devices with a notch, for example. However, as this padding is only applied to certain elements and not the content as a whole, when we don't render the app header with its included `ion-toolbar`, there is no automatic handling of the safe area.

I've opted to handle this by explicitly adding a spacer element when the header is hidden, taking its height from the exposed `--ion-safe-area-top` CSS variable. An alternative approach would be to configure the app content to be restricted to the safe area at the `CapacitorConfig` level (see [this answer](https://stackoverflow.com/questions/73897759/content-displayed-over-the-safe-area-ios-capcitor-angular/74106265#74106265)):

```ts
  ios: {
    contentInset: "always"
  }
```

Whilst this might be more robust, from some quick testing it seems to be over-active: for example, the footer becomes excessively large, including double the amount of necessary padding. So if we did want to go down this route, we'd need to make changes to how we currently handle the safe area.

## Git Issues

Closes #2692

## Screenshots/Videos

Screen recordings taken using local iOS emulator.

`debug` deployment (including new [feat_app_layout_inline_header](https://docs.google.com/spreadsheets/d/1Z46gTluTPeffqpiAZMapYo0imm59GU03j0sW9UChvPM/edit?gid=618720041#gid=618720041) template:

https://github.com/user-attachments/assets/fb8a915c-52e6-4362-b642-c865db6c6007

`plh_kids_kw` deployment:

https://github.com/user-attachments/assets/23e8d962-7c48-4762-995e-eef6e87b4fba

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -44,6 +44,12 @@
         <plh-main-header
           [style.display]="headerConfig().hidden ? 'none' : 'block'"
         ></plh-main-header>
+        <!-- Add spacer element when header is hidden to account for safe area (native status bar, notch etc.) -->
+        <div
+          class="safe-area-spacer"
+          [style.display]="headerConfig().hidden ? 'block' : 'none'"
+          [style.height]="'var(--ion-safe-area-top, 0)'"
+        ></div>
         <div class="route-container" [ngStyle]="routeContainerStyle()">
           <ion-router-outlet id="main-content"></ion-router-outlet>
         </div>


src/app/app.component.html
```diff
@@ -44,6 +44,12 @@
         <plh-main-header
           [style.display]="headerConfig().hidden ? 'none' : 'block'"
         ></plh-main-header>
+        <!-- Add spacer element when header is hidden to account for safe area (native status bar, notch etc.) -->
+        <div
+          class="safe-area-spacer"
+          [style.display]="headerConfig().hidden ? 'block' : 'none'"
+          [style.height]="'var(--ion-safe-area-top, 0)'"
+        ></div>
```
chrismclarke, line 52: nit(non-blocking): I think it might be a little cleaner to bundle all this logic within the `plh-main-header` component so that it can handle itself either rendering content when config enables, or just placeholder safe-area padding when not. I think the `ion-header` component might do this by default (essentially what you are recreating when hiding it), so it might be possible always render and just show/hide inner content based on config and still get the correct padding adjustments without styling

chrismclarke:
> #2631 added functionality to hide the app header and the issue has likely been present since then. Really we should be doing more testing on iOS in general (and multiple iOS devices), but this is time consuming and difficult to prioritise when deadlines are tight. In future, we should at least make sure to do additional testing on iOS for any changes that may affect interactions with the safe area.

I think the biggest challenge here is with issues that are immediately obvious (i.e. from a home screen), vs not immediate (e.g. requiring navigation within app and triggering something like a full-screen template popup). For any app release that goes to internal testing/testflight it's a reasonable expectation that a tester spends a few minutes to download the update and verify things look ok on the surface before promoting to production. Internally we should probably have some sort of checklist of requirements to satisfy before promoting to production (which may be slightly different for web/android/ios due to different capacities). This may look something like:

- [ ] Available to testers
- [ ] Installs and opens on device
- [ ] Any developer-flagged native issues checked (e.g. when applying core updates)
- [ ] Testers notified and open to feedback for 1-week (or whatever time period we think makes a sensible default, ideally not same-day)

For slightly more complex scenarios we're likely going to need automated testing to support some minimal user journeys that include full screen popups in a similar way to how we use visual testing on web for core components.

Might be good to make a follow-up issue for this.

chrismclarke (approved):
see minor (non-blocking) comment inline, as I can't test on any ios devices and my only android emulator that includes a notch (pixel 3 xl) doesn't seem to be affected in same way not much else I can do to test/review, but looks reasonable either way.

src/app/app.component.html
jfmcquade: That makes sense. I did consider this approach but thought that handling the colour of the toolbar element might be tricky. In fact I think it makes sense to simply set the colour to "none" in the case that the header is hidden at the app config level. I've made an alternative PR with this approach here: #2701

jfmcquade:
Thanks @chrismclarke. I've made an alternative PR that takes the approach suggested in your inline comment, see #2701. I think I do prefer that approach but don't have a strong opinion so would be happy to merge either version.

> Might be good to make a follow-up issue for this.

See #2700

chrismclarke:
close in favour of #2701

PR #2698, opened 2025-01-09
[FEATURE] Kuwait theme: Map view progress path

### LINKED ISSUE #2667: [FEATURE] Kuwait Theme - Map View

FaithDaka:
PR Checklist

## Description
- New component created for a Kuwait curved map path

## Author Notes
- The Kuwait map design has paths that wind a lot more than the `wavy` variant accommodates. A new component, `plh_progress_path`, has been created to manually style the map path and its modules.
- The [Component Sheet](https://docs.google.com/spreadsheets/d/130xy0FbCqgVMg8astiR5FBzuUMfVBHtg9k6wmweE1BA/edit?usp=sharing) which remains similar to the current `progress_path`
- Support added to lock task cards with attributes: 
`locked` - which is a boolean that marks a task card as locked
`locked_image_asset` - which is the image that appears on top of the locked task card

## Git Issues

Closes #2667


src/theme/themes/plh_kids_kw/_overrides.scss
```diff
@@ -1153,4 +1210,93 @@ body[data-theme="plh_kids_kw"] {
       }
     }
   }
+
+  // Plh Progress Path
+  plh-plh-progress-path {
```
jfmcquade, line 1215: nit(blocking): All this styling should live in `packages/components/plh/plh-kids-kw/components/plh-progress-path/plh-progress-path.component.scss`. As the component is already scoped to the deployment, I think that's the best place for the styling

jfmcquade (changes requested):
Thanks @FaithDaka, it looks great, very close to your designs. 

A question about the implementation: currently on smaller screens, the width gets cut off and becomes scrollable. How hard do you think it would be to implement a degree of responsiveness so that the whole component scaled down on narrower screens?


https://github.com/user-attachments/assets/73c9e4d2-05ab-49cc-8af0-291155817de5


I've also added some inline comments with suggestions for tidying up the code

FaithDaka:
> Thanks @FaithDaka, it looks great, very close to your designs.
> 
> A question about the implementation: currently on smaller screens, the width gets cut off and becomes scrollable. How hard do you think it would be to implement a degree of responsiveness so that the whole component scaled down on narrower screens?
> 
@jfmcquade Scaling it down proved quite difficult (not impossible) given that the path must wind and curve in certain places, so scaling the image and path down required the path to be manually re-created (which was tedious and very manual) and for the task cards to also be scalable in size for it to all fit on the small screen. 
The `overflow-y:scroll` seemed like a good compromise for this version.

In summary: The component can be scaled but it would wind and curve in a places it's not supposed to.

src/theme/themes/plh_kids_kw/_overrides.scss
jfmcquade: I think that's true for theme-based styling of a component that exists already, but in this case the component has no general styling currently, so there's no sense it which it is flexible across themes.

Also FYI, I've just pushed a minor update to this PR's branch and am going to target this branch for a build as per Laurie's email request, so pushing further commits to this branch today if possible

jfmcquade:
> @jfmcquade Scaling it down proved quite difficult (not impossible) given that the path must wind and curve in certain places, so scaling the image and path down required the path to be manually re-created (which was tedious and very manual) and for the task cards to also be scalable in size for it to all fit on the small screen. The `overflow-y:scroll` seemed like a good compromise for this version.
> 
> In summary: The component can be scaled but it would wind and curve in a places it's not supposed to.

Thanks, yes I appreciate that it's very difficult to have the elements scale effectively. I think ultimately this will need to be discussed with the PLH team in terms of how much time we want to invest in refining this screen. I do think that as it is the first screen users will see, it probably is worth further work. @esmeetewinkel, tagging you here for reference for discussions when you're back.

As well as the issue on small screens, the component also doesn't scale well to large screens:

Screenshot 2025-01-09 at 16 01 14

Again, there's a question of how much we want to invest in supporting large screen sizes like tablets. The only solution I can think of right now would be to have the illustration extend way beyond its current bounds and have the component display the cropped centre of the image by default, so that on larger screens the whitespace is filled with the background illustration at least.

jfmcquade:
> @jfmcquade For wider screens, we need a custom stretched-out illustration because the main issue is the winding of the path. If the path was allowed to wind over certain elements like the people or over the buildings or camel etc, then the image and path would scale dynamically. 

Yes I agree that we would need a custom illustration to handle large screen sizes. I'm imagining that this would still have the same size content in the centre, but the edges would be extended. That way the path could remain the same size too and not intersect with the illustrations, as you say. 

> The module cards are also positioned absolutely on the path so if its size changes, those values have to change.

Indeed, this would need to be changed if we're implementing some scaling to react to different screen size.

On reflection, if we're saying that the path should scale to exactly match the image, wouldn't it make sense to just draw the path on the image itself and save it as a single asset? The path and image are already fully customised to match each other. Then the issue becomes scaling the position of the module cards so that they have fixed positions with respect to the background, combined with allowing the whole component to scale down with screen width to a sensible minimum (e.g. 300px).

jfmcquade:
> I can't get the component template to work on the debug app when running locally. Could someone confirm this is working on their end?

It is working for me, you just need to make sure you set the theme to `plh_kids_kw`. I have suggested [above](https://github.com/IDEMSInternational/open-app-builder/pull/2698#discussion_r1908698329) that the styling be moved to the component level so that it doesn't rely on the theme (as the component is a custom deployment component anyway)

Screenshot 2025-01-15 at 12 32 30


> The example in the component template uses an `items` loop to render the module cards, but this will need to be a `data_items` loop since the completion / locked data is dynamic. Should that work equally?

Good point, I've updated the component code to handle data-items child rows with https://github.com/IDEMSInternational/open-app-builder/pull/2698/commits/d3c4746de254785585f4dbcf29deef1d878b6a38, and have updated the [comp_plh_progress_path](https://docs.google.com/spreadsheets/d/130xy0FbCqgVMg8astiR5FBzuUMfVBHtg9k6wmweE1BA/edit?gid=603128459#gid=603128459) template accordingly. (FAO @FaithDaka, you'll want to pull my changes before continuing work on this branch)

> How does the component pick which module to "highlight"? As said I've not been able to reproduce, but from the screenshot/example provided I don't understand how this is deduced. From an authoring perspective, I'm happy for the logic for this to sit in the spreadsheet, in which case `highlighted` could just be a boolean property on the component, much like `locked`.
> 
> I think the tasks are redundant here, all the logic can just be done through templating.

Currently, the "highlighted" style is just applied to any task group that is "in progress". On reflection, if we are relying on task evaluation then this should probably be updated apply to those tasks that are "not started" – this should work for this use case since there should only ever be one task group at a time that is "not started" and not "locked".

However, alternatively we could also expose an explicit `highlighted` param to authoring. This would definitely make sense, my only reservation is doing so in a way that doesn't override the existing "highlighted" logic (which I believe is still in use for some deployments) – the code gets quite confusing when we're supporting both patterns, but it would of course be possible.

esmeetewinkel:
> However, alternatively we could also expose an explicit highlighted param to authoring. This would definitely make sense, my only reservation is doing so in a way that doesn't override the existing "highlighted" logic (which I believe is still in use for some deployments) – the code gets quite confusing when we're supporting both patterns, but it would of course be possible.

My preference would be for this component not to use the task system at all. To prevent confusion we can just pick a different word than highlighted for this property, e.g. featured /  accentuated / prominent / accent / spotlight

jfmcquade:
> My preference would be for this component not to use the task system at all. To prevent confusion we can just pick a different word than highlighted for this property, e.g. featured / accentuated / emphasised / prominent / accent / spotlight

In that case I think it would make sense to use a standalone component instead of a task card with `variant: circle`, e.g.  `plh_module_button` or similar. 

The code and styling could be copied over from the existing task card component, but all the task logic could be stripped out and just the following parameters exposed:
- `image_asset`
- `locked_image_asset`
- `locked` (boolean)
- `highlighted` (boolean)

How does that sound, @FaithDaka?

jfmcquade:
FAO @esmeetewinkel and @FaithDaka, I've added code (https://github.com/IDEMSInternational/open-app-builder/pull/2698/commits/140840772c5f4478f8cd8fef776b9ecf688a64cc) to handle scrolling the content to the bottom middle of the page upon load of this component. I think this does quite a good job of making the component more usable and feel complete (coupled with removing the page padding at the template level). It would still be preferable to have the component adapt responsively to different screen sizes, but I think this is an imporvement.

https://github.com/user-attachments/assets/d61d1e9d-b16f-4a81-b8fd-26b9577ac794

FaithDaka:
@jfmcquade With this update, the background image is not scrollable and the task-cards have been adjusted in size and position to take up minimal space on smaller screens.


https://github.com/user-attachments/assets/b9d61f3a-3ad5-477e-8c7c-e992580a6546

jfmcquade:
Thanks @FaithDaka, I think the responsiveness is definitely an improvement so I think we should get this merged when possible and add any issues as follow-ups. I've added various points of code tidying in https://github.com/IDEMSInternational/open-app-builder/pull/2698/commits/6e0a76a315c3f12ad46d9fe773dcb5cbbd480589

TODO: I'm going to add a `highlighted` param to replace the auto-calculated progress/highlighted statuses. I'll do so after lunch and then tag this PR as ready for functional testing

jfmcquade (approved):

esmeetewinkel:
Functional test passed. I've tried implementing this in the Kuwait deployment, but ran into some further [issues](https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/issues/145), I believe because `@calc` doesn't have access to dynamic data. 

However, I think we can merge this PR and address those issues separately. @jfmcquade could you resolve the conflicts on this branch? Thanks!

esmeetewinkel (approved):

PR #2701, opened 2025-01-09
fix: handle iOS safe area when app header is hidden (alt)

### LINKED ISSUE #2692: [BUG] iOS safe area is not respected when header is hidden

jfmcquade:
PR Checklist

## Description

Alternative implementation of #2693. See that PR for context.

## Git Issues

Closes #2692, #2693

## Screenshots/Videos
Screen recordings taken using local iOS emulator.

`debug` deployment (including new [feat_app_layout_inline_header](https://docs.google.com/spreadsheets/d/1Z46gTluTPeffqpiAZMapYo0imm59GU03j0sW9UChvPM/edit?gid=618720041#gid=618720041) template:

https://github.com/user-attachments/assets/97cace71-1087-44e0-9255-aee7bb20d14d

`plh_kids_kw` deployment:

https://github.com/user-attachments/assets/bf6fd2de-5fd0-4450-bf62-1d6a066d3a42


chrismclarke (approved):
Thanks @jfmcquade 
I think from a code side this looks much nicer than #2693, so assuming it still functions as expected I say lets plan to merge this branch instead.

esmeetewinkel (commented):
@jfmcquade Not sure how to test this PR (but happy to do so if you think there's need to). Did you create Appetize links for this PR too?

jfmcquade:
> @jfmcquade Not sure how to test this PR (but happy to do so if you think there's need to). Did you create Appetize links for this PR too?

As we're pressed for time I think I'll merge this, I'm confident it should resolve the issue. I've made a note to share appetize links for this PR at some point

Issue #2704, opened 2025-01-13
[BUG] Unable to use togglebar within data_items loop


ChrisMarsh82:
**Describe the bug**
Toggle bar does not work within a data_items as it cannot set its own value.

Example error: 
```
template-action.service.ts:344 row "data_items_54.display_group_1.toggleparent_1" not found 
Setting local variable which does not exist {key: 'data_items_54.display_group_1.toggleparent_1', value: true} TODO
```

**Deployment Sheet(s)**
[edit_report ](https://docs.google.com/spreadsheets/d/1pqNDT__R83Co4cyz-E-Y5wiw0i020Bg7HnQ5t_wHqrA)

((https://docs.google.com/spreadsheets/d/1pqNDT__R83Co4cyz-E-Y5wiw0i020Bg7HnQ5t_wHqrA))

**Additional context**
Need to be able to create a list of parents from a datalist that is added to dynamically. From this parent list will need to create a toggle for each parent to define whether they have attended or not. This data then needs to be captured in a report. i.e. a column in a datalist which holds a list of all parent ids who attended.
![image](https://github.com/user-attachments/assets/c7b02960-6361-4b95-a448-0eb4063b01d0)

jfmcquade:
Thanks, @ChrisMarsh82. 

@chrismclarke could you advise on this?

On inspection of the code, this is a case of a more general issue that a component cannot set its own value (its row value) if within a data-items loop. This is related to the issue of not being able to access a local variable that is set inside a data-items loop from outside of that same loop within a template. 

The data-items loop is, in some sense, being treated as a separate nested template (see https://github.com/IDEMSInternational/open-app-builder/blob/45b435106f0e67c03b88a9931b828ba2d293c540/src/app/shared/components/template/services/instance/template-row.service.ts#L285), which means that its child rows are not being added to the map of rows that is accessible for setting and getting from the top-level template.

I'm not sure how this should be handled. In theory the `templateRowMap` could be reactive to dynamic changes in the data-items child rows, although I haven't thought through how that would work. 

I've opened #2705 with hacky workaround, intended more as an investigation into the issue rather than a proposed solution.

The pattern here could also be relevant: https://github.com/IDEMSInternational/open-app-builder/blob/45b435106f0e67c03b88a9931b828ba2d293c540/src/app/shared/components/template/services/instance/template-row.service.ts#L321, since we're concerned with the difference between referencing a row by its name and its nested name.

[...]

PR #2705, opened 2025-01-13
spike: fix component.setValue inside data-items

jfmcquade:
PR Checklist

## Description

Investigation around #2704.

Allows a toggle bar inside a data items loop to set its own value.

## Git Issues

Closes #

## Screenshots/Videos

[debug_toggle_bar](https://docs.google.com/spreadsheets/d/1GeClVPEv2Trom59KvVnTXRwNyVbWCzKDl8-M7wmbNlc/edit?gid=1681985300#gid=1681985300)

Screenshot 2025-01-13 at 16 59 16

Screenshot 2025-01-13 at 16 59 45

### DIFF
--- a/src/app/shared/components/template/services/instance/template-action.service.ts
+++ b/src/app/shared/components/template/services/instance/template-action.service.ts
@@ -165,8 +165,23 @@ export class TemplateActionService extends SyncServiceBase {
         const selfField = arg.split(".")[1];
         arg = this.container?.templateRowMap[action._triggeredBy?._nested_name]?.[selfField];
       }
+
       return arg;
     });
+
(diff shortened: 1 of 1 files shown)


chrismclarke:
Interesting, yes you're right, because the data_items are detached from the parent container, the templateRowMap has no awareness of them and so can't respond to set_local actions correctly. 

Ideally we would probably want to intercept those set_local actions within the loop and convert them to set_item, however doing so would also mean having to reverse-engineer what property the component is reading from the item loop when assigning a value. So for now I think just ignoring the set_local action is fine, relying on the data_items list as source of truth. 

I think probably better to handle this from data_items component itself, I've opened up https://github.com/IDEMSInternational/open-app-builder/pull/2706 to do so and will close here

PR #2706, opened 2025-01-13
fix: data-items set local

### LINKED ISSUE #2704: [BUG] Unable to use togglebar within data_items loop

chrismclarke:
PR Checklist

## Description

Fix issue where `set_local` operations cannot be used inside data-items to set values of list items. This also applies to components that track there own internal state, such as toggle_bar, slider and checkbox

Also adds support for self-referenced values in actions via `this.value` (as supported outside of data_items loops)

## Author Notes
It should still be possible to use `set_local` for values that are outside the data_items loop, such as the demo example that sets the name of the input data_list.

## Review Notes
See [component_data_items](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892)

**Example 1** - now includes inline toggle

**Example 6a and 6b** - newly added to test and demo updating text fields through text_box input

## Dev Notes
There already existed a method for a given component to intercept/alter triggered actions, so using that I've set it to ignore any set_value operations on local variables that are not tracked within the parent (i.e. all data-items child rows which are processed independently). 

We still want to allow inner components to process set_local updates for any sibling managed outside of the data_items loop (allow green arrow, prevent red arrow)

image

During testing it also became apparent that any actions that require reference to the component row updated value would not work (again, something that is handled as a one-time operation in parent containers). So additional action list mapping is included to handle the case as the actions are triggered. In the future it would probably make sense to handle this as the default (not just in action lists), as statically assigning all the action values from variable isn't necessary unless an action is actually triggered, but beyond the scope of this PR.

## Git Issues

Closes #2705
Closes #2704

## Screenshots/Videos
**Demo 1 - Item Toggle** - toggle used within data items loop, keeping local state in sync with items loop. Also includes demo that uses set_local within items loop to assign a sheet-level local variable (outside of items loop)

[Screenity video - Jan 13, 2025.webm](https://github.com/user-attachments/assets/91f0e498-b1e6-4e31-9423-3308f814a4a6)

**Demo 2 - Text Edit** - two examples, first directly updates items when text input changes. Second uses an intermediate variable to track edit mode, only updating data on save. Also shows bug at end where tex_box only updates after click outside of box and not directly clicking on button after (could create issue, not related specifically to data-items)

https://github.com/user-attachments/assets/cfccbf41-913e-4149-85b5-98e2fe1f99f0


jfmcquade (changes requested):
Thanks, @chrismclarke. Nice that the action interceptor was already there and great to have updated new/examples on the data items component template.

I've noticed a couple of issues.

# 1
The first is that components in the same template outside of the data-items loop are also affected by the action interceptor (since the interceptor is at the level of the parent template, which is shared). In practice this means that self-referencing `set_local` actions are ignored for components at the top-level of the template, if the template contains a data-items loop. See example recording of the [debug_toggle_bar](https://docs.google.com/spreadsheets/d/1GeClVPEv2Trom59KvVnTXRwNyVbWCzKDl8-M7wmbNlc/edit?gid=1681985300#gid=1681985300) template (I'm logging out the ignored action that meets the condition of being a self-update):


https://github.com/user-attachments/assets/33ee0a09-ba7a-4b1b-9037-f8f8abfc1e27

# 2
The second issue is some odd behaviour with Example 6b of the [comp_data_items](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892) template. The values of the items seem to get conflated. I haven't investigated further, and I'm not sure if this is a fundamental issue or an issue with the authoring pattern of the example.

https://github.com/user-attachments/assets/825ec1fb-a277-4cfb-96da-d3096f09e17a

chrismclarke:
> I've noticed a couple of issues.
> 
> # 1
> The first is that components in the same template outside of the data-items loop are also affected by the action interceptor (since the interceptor is at the level of the parent template, which is shared). In practice this means that self-referencing `set_local` actions are ignored for components at the top-level of the template, if the template contains a data-items loop. See example recording of the [debug_toggle_bar](https://docs.google.com/spreadsheets/d/1GeClVPEv2Trom59KvVnTXRwNyVbWCzKDl8-M7wmbNlc/edit?gid=1681985300#gid=1681985300) template. I'm logging out the ignored action that meets the condition of being a self-update – you can see that the self-update action of the toggle-bar inside the items loop gets ignored as expected, but the self-update action of the top-level toggle-bar also gets ignored, meaning its associated local variable is not updated as expected:

Thanks for adding the debug sheet for this. I could see a few authoring options with the sheet (creating local variable in the loop, but not using it with the toggle bar (not sure what the purpose of `my_local` was. I've tidied up the example but could still see issues with the way the interceptor was being processed. The main issue was the fact that template containers could only ever store a single action interceptor, which would get applied to all triggered actions. This would additionally cause conflict if being used alongside other components that include interceptors (a few examples from some legacy nav actions). 

As such I've added https://github.com/IDEMSInternational/open-app-builder/pull/2706/commits/3226adc9bfd149ab55faaf492480a15205b88e88 which provides a new mechanism for piping multiple interceptors, which can also be scoped to just child components of a given action. I haven't removed the original handling (just marked as deprecated) to avoid too many immediate knock-ons.

Additionally I noted the challenge with determining whether the `set_local` action was targeting itself or another value (I think comparing the component nested name with the action triggered_by nested name did work, although I found it a bit confusing whilst debugging) - as such I've refactored to include a specific `set_self` action for tracking this behaviour (not intended for use by authors).

I think these combinations should have resolved the debug case, but would appreciate if you could also check to make sure it matches what you expect.

> # 2
> The second issue is some odd behaviour with Example 6b of the [comp_data_items](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892) template. The values of the items seem to get conflated. I haven't investigated further, and I'm not sure if this is a fundamental issue or an issue with the authoring pattern of the example.
> 

As mentioned on the call, I believe this is an issue that's always existed and not specific to data-items loops. I've created a separate issue if we want to investigate this further https://github.com/IDEMSInternational/open-app-builder/issues/2711

jfmcquade (approved):
Thanks @chrismclarke.

Using a new `set_self` action makes sense to me. I think I may prefer `update_value` as a name, but it doesn't make explicit that the value in question is for the component/row itself, so I think `set_self` probably is better in the context of the other action names.

I appreciate your updates to the [debug_toggle_bar](https://docs.google.com/spreadsheets/d/1GeClVPEv2Trom59KvVnTXRwNyVbWCzKDl8-M7wmbNlc/edit?gid=1681985300#gid=1681985300) template, the functionality is clearer now and the behaviour is indeed what I expect.

I was trying to get a better grasp of the conditions under which `set_local` is now supported inside data-items loops, and made a new debug template: [debug_data_items_set_local](https://docs.google.com/spreadsheets/d/10S6JZ9ASRHsvRNQ8foe_tc8urt0KWLeoA8GX7oPLRAM/edit?gid=1681985300#gid=1681985300). To be clear (FAO @ChrisMarsh82 and @esmeetewinkel), the behaviour of this debug sheet is identical on `master` and this PR branch:
- `data-items` can read the value of a local variable that is set outside of the data-items loop
- `data-items` can update the value of a local variable that is set outside of the data-items loop via `set_local`
- `data-items` can read the value of a local variable that is set inside of the data-items loop

Screenshot 2025-01-15 at 16 31 36


https://github.com/user-attachments/assets/84b69393-753b-4770-a6b5-3f7eab3e7870

chrismclarke:
> Using a new `set_self` action makes sense to me. I think I may prefer `update_value` as a name, but it doesn't make explicit that the value in question is for the component/row itself, so I think `set_self` probably is better in the context of the other action names.

Sounds good, also less likelihood it might conflict with other action names in the future. 


> I was trying to get a better grasp of the conditions under which `set_local` is now supported inside data-items loops, and made a new debug template: [debug_data_items_set_local](https://docs.google.com/spreadsheets/d/10S6JZ9ASRHsvRNQ8foe_tc8urt0KWLeoA8GX7oPLRAM/edit?gid=1681985300#gid=1681985300). 

Thanks for this, useful to have

> To be clear (FAO @ChrisMarsh82 and @esmeetewinkel), the behaviour of this debug sheet is identical on `master` and this PR branch:
✅ `data-items` can read the value of a local variable that is set outside of the data-items loop 
✅ `data-items` can update the value of a local variable that is set outside of the data-items loop via `set_local` 
✅ `data-items` can read the value of a local variable that is set inside of the data-items loop 
❌ `data-items` cannot update the value of a local variable that is set inside of the data-items loop via `set_local` 
❌ `data-items` cannot set a local variable that is accessible to the template outside of the data items loop

Thanks for the breakdown. Of the two non-supported cases, I would probably argue that creating new locals inside of loops would be most useful (or some other method of creating temporary variables linked to item loops, to support cases like draft text changes before peristing with save). Although I'd say may start to get a little messy until we have a bigger refactor to the container templating system to add a better way to store and react to local variable changes. 

I feel like the notion of creating a variable inside a loop that is then accessed outside of the loop possibly doesn't make as much sense, it goes against typical [variable scope](https://github.com/IDEMSInternational/open-app-builder/pull/2706) (assuming the loop is treated like a function, variables from the inner scope shouldn't bleed out), and also unless an author is careful it would be possible to create many repeated variables (e.g. `my_var` vs `my_var_{item.id}`). Also then if an author needs to access outside they would need to know the suffix of the instance they need. I think probably better at supporting temporary columns on the data_list would also cover this case, so an author could set a `text_draft` for each item, read from outside as part of a different items loop/query, but not sync to server. I'm not sure whether these columns should get removed when the container is destroyed, or persisted but not synced to server.  


 
> I think possibly the PR description/title could be updated for clarity around this, referring to the specific case of `set_local` referred to?

Updated

PR #2708, opened 2025-01-14
Created new facilitator_mx style based on kids_kw

ChrisMarsh82:
@esmeetewinkel This has the Kids-kw css overrides


esmeetewinkel:
@ChrisMarsh82 I would like to merge this so that we can add the theme to the debug deployment for testing. Do you have any objections?

ChrisMarsh82:
Happy to merge. Can do it later tonight

esmeetewinkel (approved):
@ChrisMarsh82 Note also my commit that changes the theme colours. Following the designs it makes most sense to me if the primary and secondary colours are these two.

However some of the other colours based off these might have gone funny - I think we can address such issues as follow-ups (or maybe @FaithDaka can help out too).

Issue #2709, opened 2025-01-14
[BUG] Padding applied to display group through style_list is applied twice


jfmcquade:
**Describe the bug**
When setting the padding on a display group through the authored `style_list`, the padding value is applied twice.

Indeed for display groups, all styling is applied at two levels, which only sometimes causes issues: once on the component container (as all style_list properties are) and again on a child of the display group component (in accordance with the custom handling of styles for the display group component).

**App version**
_Specify the version of the app where this bug was first identified._

**To Reproduce**
_Describe the steps to reproduce this bug._

**Deployment Sheet(s)**
_Specify the sheet name and include a link where the bug appears if appropriate._

**Debug Sheet(s)**
[feat_app_layout_padding](https://docs.google.com/spreadsheets/d/1Z46gTluTPeffqpiAZMapYo0imm59GU03j0sW9UChvPM/edit?gid=2064945031#gid=2064945031)

<img width="800" alt="Screenshot 2025-01-14 at 15 49 58" src="https://github.com/user-attachments/assets/b5e36eaa-56ff-4229-a04a-41c7abe0a9ff" />

**Additional context**
_Add any other context about the problem here._

esmeetewinkel:
Is it possible that padding also gets applied twice when added on the style list of the text component?

Example on a template without inline padding (sorry for not making a debug sheet):
![Image](https://github.com/user-attachments/assets/1d124d3c-44b1-4665-bb4c-25d7004fd718)
![Image](https://github.com/user-attachments/assets/bee18618-a1dc-4458-af8f-22aa6653fd0a)
vs
![Image](https://github.com/user-attachments/assets/e04a0e2e-00dc-4328-9975-916cf4d098d3)
![Image](https://github.com/user-attachments/assets/b2e8e83d-c169-4643-9ece-ba6d2e783f12)
![Image](https://github.com/user-attachments/assets/a0271c98-8a08-402a-985b-fb70050ced42)

PR #2710, opened 2025-01-14
chore: expose horizontal and vertical values for app config layout padding; fix display group inline_padding variant 

jfmcquade:
PR Checklist

## Description

- New `app_config` properties: `LAYOUT.page_padding.horizontal` and `LAYOUT.page_padding.vertical`
  - These replace the previous single `LAYOUT.page_padding` property. This was only used in debug and WIP sheets, so I haven't considered this to be a breaking change.
  - Separating out these properties makes it much easier to assign them to ionic's four `--padding-*` css variables. It also makes sense from an authoring perspective, as the main use case for adjusting the padding is to do so for the horizontal axis only.
- Updates `inline_padding` variant of the display group component
  - Previously, the size of the padding was read from the same ionic padding variables that are set from the app config level, meaning that if the page padding was set to 0, the padding applied to the `inline_padding` variant would also be 0. 
  - This variant was originally introduced for convenience when dealing with sticky display groups, which ignore page padding by default (see #2589).
  - In theory, this variant shouldn't be necessary as it should accomplish the same thing as a single line of custom styling (namely `padding-inline: 24px`, for example). However, as well as providing a convenient alternative to adding custom styling, the custom styling approach is currently broken, see #2709.
  - In resolving #2709, it seems difficult to avoid unwanted knock-ons, since there may be production deployments which make use of the styling at both levels. We need a way to style padding-free pages immediately, therefore it seems like an acceptable short-term solution to leverage the existing `inline_padding` variant.
    - An alternative would be to set the custom `padding-inline` to be half the desired value, knowing that it will be applied twice on the rendered component

## Git Issues

Closes #

## Screenshots/Videos

[feat_app_layout_padding](https://docs.google.com/spreadsheets/d/1Z46gTluTPeffqpiAZMapYo0imm59GU03j0sW9UChvPM/edit?gid=2064945031#gid=2064945031):

Screenshot 2025-01-14 at 15 54 48

Screenshot 2025-01-14 at 15 54 02

Screenshot 2025-01-14 at 15 33 23

### DIFF
--- a/packages/data-models/appConfig.ts
+++ b/packages/data-models/appConfig.ts
@@ -128,7 +128,10 @@ const APP_FOOTER_DEFAULTS = {
 };
 
 const LAYOUT = {
-  page_padding: "24px",
+  page_padding: {
+    horizontal: "24px",
+    vertical: "24px",
+  },
 };
 
 const APP_SIDEMENU_DEFAULTS = {
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -59,7 +59,8 @@ export class AppComponent {
   public routeContainerStyle = computed(() => {
     const { page_padding } = this.layoutConfig();
     return {
-      "--page-padding": page_padding,
+      "--page-padding-horizontal": page_padding.horizontal,
+      "--page-padding-vertical": page_padding.vertical,
     };
   });
 
--- a/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
+++ b/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
@@ -52,8 +52,7 @@
 
 .display-group-wrapper[data-variant~="inline_padding"] {
   width: 100%;
-  padding-left: var(--padding-start);
-  padding-right: var(--padding-end);
+  padding-inline: var(--default-page-padding);
 }
 
 .display-group-wrapper {
--- a/src/theme/deployment/_overrides.scss
+++ b/src/theme/deployment/_overrides.scss
@@ -5,11 +5,11 @@ plh-button > ion-button {
 
 /// Ensure all pages have padding
 ion-content {
-  $default-padding: 24px;
-  --padding-top: var(--page-padding-vertical, #{$default-padding});
-  --padding-bottom: var(--page-padding-vertical, #{$default-padding});
-  --padding-start: var(--page-padding-horizontal, #{$default-padding});
-  --padding-end: var(--page-padding-horizontal, #{$default-padding});
+  --default-page-padding: 24px;
+  --padding-top: var(--page-padding-vertical, var(--default-page-padding));
+  --padding-bottom: var(--page-padding-vertical, var(--default-page-padding));
+  --padding-start: var(--page-padding-horizontal, var(--default-page-padding));
+  --padding-end: var(--page-padding-horizontal, var(--default-page-padding));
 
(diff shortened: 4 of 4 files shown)


chrismclarke (changes requested):
I'm not whether this totally makes sense to me, but possible that I've missed something with how pages are laid out. 

So if I understand the PR, by providing horizontal and vertical sub-padding properties authors can more easily differ the amount of horizontal and vertical padding, particularly to only implement horizontal padding. 

In my experience, vertical padding is always needed on a page that has readable content (i.e. not a full-screen image/map or similar), and I expect the only reason it might not be included is because we have components elsewhere that automatically include margins or padding themselves (likely inconsistent depending on components used). So I still think it makes sense that default option should be to apply a single value for both horizontal and vertical padding.

What might make more sense, however, is if we just allow the string to represent any valid CSS, so that: 
`page_padding: "24px"` sets equal padding
`page_padding: "24px 12px"` sets different vertical and horizontal padding
`page_padding: "24px 0 0 24px"` sets each individual top, right, bottom, left values.

As for whether this system would then line-up nicely with ionic variables, would need to check  exactly what variables are exposed and decide whether worth using, applying our own custom CSS, and/or if there would be need to include lightweight script in the parser to separate out a CSS string into individual values (probably not worthwhile if other options available)

Let me whether there's something I've overlooked and if still keen to proceed with separate horizontal/vertical values, in which case I can do a close review of the code as-is

jfmcquade:
> I'm not whether this totally makes sense to me, but possible that I've missed something with how pages are laid out.
> 
> So if I understand the PR, by providing horizontal and vertical sub-padding properties authors can more easily differ the amount of horizontal and vertical padding, particularly to only implement horizontal padding.
> 
> In my experience, vertical padding is always needed on a page that has readable content (i.e. not a full-screen image/map or similar), and I expect the only reason it might not be included is because we have components elsewhere that automatically include margins or padding themselves (likely inconsistent depending on components used). So I still think it makes sense that default option should be to apply a single value for both horizontal and vertical padding.
> 
> What might make more sense, however, is if we just allow the string to represent any valid CSS, so that: `page_padding: "24px"` sets equal padding `page_padding: "24px 12px"` sets different vertical and horizontal padding `page_padding: "24px 0 0 24px"` sets each individual top, right, bottom, left values.
> 
> As for whether this system would then line-up nicely with ionic variables, would need to check exactly what variables are exposed and decide whether worth using, applying our own custom CSS, and/or if there would be need to include lightweight script in the parser to separate out a CSS string into individual values (probably not worthwhile if other options available)
> 
> Let me whether there's something I've overlooked and if still keen to proceed with separate horizontal/vertical values, in which case I can do a close review of the code as-is

Thanks @chrismclarke. Yeh the original intention was that the `page_padding` param could use the CSS shorthand syntax for all 4 values in that way, but there was a bug in the way it was interpreted to set the CSS variables used by ionic. 

I've opened #2720 which replaces this PR and preserves the existing `page_padding` param in the `app_config`.

chrismclarke:
> Thanks @chrismclarke. Yeh the original intention was that the `page_padding` param could use the CSS shorthand syntax for all 4 values in that way, but there was a bug in the way it was interpreted to set the CSS variables used by ionic.
> 
> I've opened #2720 which replaces this PR and preserves the existing `page_padding` param in the `app_config`.

Sounds good, I've reverted this to draft and can close following functional test and merge of #2720

PR #2716, opened 2025-01-16
Mexico Theme: Refactor Theme-level Color Palette

### LINKED ISSUE #2713: [FEATURE] plh_facilitator_mx theme updates
**What?**
The `plh_facilitator_mx` theme was updated with the overrides from the `plh_kids_kw` theme in #2708. There are some additional changes that need to be made:
- The primary and secondary theme colours need to be adjusted to reflect the designs
  - Additional component-level overrides may need to be applied to ensure all components display properly
- The typography should be adjusted to reflect the designs
  - As the designs use the default typeface, Roboto, this is likely just a case of removing the theme-level typography overrides

**Other details**
[`plh_facilitator_mx` figma](https://www.figma.com/design/o1qvGOcpHDHaJ6BZvW2SGu/UI-Design-Kit?t=ANasrfCLccZl6lp6-0)

FaithDaka:
PR Checklist

## Description
The current `utils.generateTheme` method creates a large contrast for shades of green. The lower numbers (50-500) are too neon while the higher numbers (600-900) are sparsely dark. 
This theme refactor overrides the primary color palette with the actual design values, to create a cohesive palette.

Typography style file also removed as it's not needed.

## Git Issues

Closes #2713

## Screenshots
| Before | After |
|--------|------|
|  |   |
|  |  |
|  |  |


jfmcquade (approved):
Thanks @FaithDaka, I'm happy with this. Ideally our automatic theme colour generation should be flexible enough to handle different types of colours, but in the meantime it makes sense that designers may want to override the default generated values.

PR #2719, opened 2025-01-16
chore: new 'dropdown' variant of combo-box

### LINKED ISSUE #2718: [FEATURE] Combo-box component: dropdown variant
**What?**
A variant of the combo-box component that opens a dropdown menu from which to select options, as opposed to launching a modal.

Image

**Why?**
Requested for the `plh_facilitator_mx` deployment

**How?**
The quickest way would likely be to leverage the [ion-select](https://ionicframework.com/docs/api/select) component, with `interface` set to `popover`. This may not allow for a full level of customisation, but should be sufficient for now.

**Other details**

jfmcquade:
PR Checklist

## Description

Adds a new variant, `dropdown`, of the combo-box component. The authoring API of the component should remain unchanged for this variant.

# TODO
Follow-up:
#2732

## Dev notes

Using the [ion-select](https://ionicframework.com/docs/api/select) was convenient for functionality, but styling was quite fiddly as the component uses the shadow dom and only exposes limited styling options. I think I've got the styling to a reasonable point, but if we require further precise customisation in the future then we may need to build a custom component (minimally a custom button that launches the `ion-popover` element programmatically).

## Git Issues

Closes #2718

## Screenshots/Videos

Screenshot 2025-01-16 at 17 06 07

`plh_facilitator_mx` theme:

https://github.com/user-attachments/assets/b6a4c47d-ccdc-468d-a963-40262d4d13d4

`default` theme:

https://github.com/user-attachments/assets/c6ee099b-a1ed-4c8b-a7f6-9791e9906b03


src/app/shared/components/template/components/combo-box/combo-box.component.ts
```diff
@@ -22,6 +22,7 @@ export class TmplComboBoxComponent
   extends TemplateBaseComponent
   implements ITemplateRowProps, OnInit, OnDestroy
 {
+  public variant: "modal" | "dropdown" | string;
```
chrismclarke, line 25: nit(non-blocking): is there good reason to accept generic `string` type? Looking at the html nothing will displayed in this case

chrismclarke (approved):
nit(non-blocking): Given the code additions pretty much double the amount of CSS and HTML I think likely they might better suited for either a new component, or a child component that still gets rendered from the `combo-box` core but with code in a `combo-box-dropdown` subfolder (using input props to pass just what is needed). 

However I appreciate a refactor might be a bit time-consuming if required for an upcoming release so happy to approve as-is for now.

ChrisMarsh82:
@jfmcquade This displays the 'name' and not 'text'. Despite the naming name = id and text = display name

jfmcquade:
> @jfmcquade This displays the 'name' and not 'text'. Despite the naming name = id and text = display name

Thanks, resolved with https://github.com/IDEMSInternational/open-app-builder/pull/2719/commits/0ad0a4a904f05ea3249fc5a7994673655c26f878

PR #2720, opened 2025-01-16
fix: app config page padding property is properly interpreted as shorthand

jfmcquade:
PR Checklist

## Description

- Fixes an issue where the `app_config` property `LAYOUT.page_padding` would not be correctly interpreted as a shorthand CSS property for setting the padding on multiple sides, e.g. `24px 0` or `24px 32px 10px 24px`
- Updates `inline_padding` variant of the display group component
  - Previously, the size of the padding was read from the same ionic padding variables that are set from the app config level, meaning that if the page padding was set to 0, the padding applied to the `inline_padding` variant would also be 0. 
  - This variant was originally introduced for convenience when dealing with sticky display groups, which ignore page padding by default (see #2589).
  - In theory, this variant shouldn't be necessary as it should accomplish the same thing as a single line of custom styling (namely `padding-inline: 24px`, for example). However, as well as providing a convenient alternative to adding custom styling, the custom styling approach is currently broken, see #2709.
  - In resolving #2709, it seems difficult to avoid unwanted knock-ons, since there may be production deployments which make use of the styling at both levels. We need a way to style padding-free pages immediately, therefore it seems like an acceptable short-term solution to leverage the existing `inline_padding` variant.
    - An alternative would be to set the custom `padding-inline` to be half the desired value, knowing that it will be applied twice on the rendered component

## Dev notes

Because ionic components use the shadow dom, we must rely on the exposed CSS variables to customise the styling. In the case of padding, the shadow element `.inner-scroll` gets padding applied through 4 exposed CSS variables: `--padding-top`, `--padding-bottom`, `--padding-start` and `--padding-end`, see screenshot below. In order to set these variables from our authored `page_padding` value, which defines all 4 via in shorthand syntax, we must manually extract the different values.

Screenshot 2025-01-16 at 17 47 08

## Git Issues

Closes #2710 

## Screenshots/Videos

[feat_app_layout_padding](https://docs.google.com/spreadsheets/d/1Z46gTluTPeffqpiAZMapYo0imm59GU03j0sW9UChvPM/edit?gid=2064945031#gid=2064945031):

Screenshot 2025-01-16 at 19 11 36

Screenshot 2025-01-14 at 15 54 02

Screenshot 2025-01-16 at 19 16 30

Screenshot 2025-01-16 at 19 16 39

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -43,6 +43,7 @@ import { LocalStorageService } from "./shared/services/local-storage/local-stora
 import { DeploymentService } from "./shared/services/deployment/deployment.service";
 import { ScreenOrientationService } from "./shared/services/screen-orientation/screen-orientation.service";
 import { TemplateMetadataService } from "./shared/components/template/services/template-metadata.service";
+import { getPaddingValuesFromShorthand } from "./shared/components/template/utils";
 
 @Component({
   selector: "app-root",
@@ -58,8 +59,12 @@ export class AppComponent {
 
   public routeContainerStyle = computed(() => {
     const { page_padding } = this.layoutConfig();
+    const { top, right, bottom, left } = getPaddingValuesFromShorthand(page_padding);
     return {
-      "--page-padding": page_padding,
+      "--page-padding-top": top,
+      "--page-padding-start": left,
+      "--page-padding-bottom": bottom,
+      "--page-padding-end": right,
     };
   });
 
--- a/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
+++ b/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
@@ -52,8 +52,7 @@
 
 .display-group-wrapper[data-variant~="inline_padding"] {
   width: 100%;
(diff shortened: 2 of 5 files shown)


chrismclarke (approved):
Thanks @jfmcquade , really nice utility to have and think it's a lot nicer casting the css-style values to the custom ionic properties.

Appreciate having the tests as well, looks like it covers a sensible number of cases and the outputs are as I would expect. 

There could possibly be a case made for extracting these at the compiler level, although given there is nothing in the naming convention to indicate the config value is _special_ in some sort of way, and the calculation is incredibly lightweight also happy to keep as-is.

esmeetewinkel (approved):
Functional test passed

esmeetewinkel:
Post merge of this PR (I should have noticed when testing):

When the page padding is provided in the format `app_config.LAYOUT.page_padding: 0;` I now get unlimitedly many errors
image

Writing instead `app_config.LAYOUT.page_padding: 0 0;` resolves the issue.

Maybe worth adding this case to the tests?

PR #2721, opened 2025-01-16
chore: make pr preview action secrets explicit

chrismclarke:
PR Checklist

## Description

As mentioned in [GH Actions Docs](https://docs.github.com/en/actions/sharing-automations/reusing-workflows), secrets can only be inherited by reusable workflows if originating from the same github org. 

>Workflows that call reusable workflows in the same organization or enterprise can use the inherit keyword to implicitly pass the secrets.
```yml
jobs:
  call-workflow-passing-data:
    uses: octo-org/example-repo/.github/workflows/reusable-workflow.yml@main
    with:
      config-path: .github/labeler.yml
    secrets: inherit
```

As such when calling actions such as the pr-preview, the optional `DEPLOYMENT_PRIVATE_KEY` variable would not be passed if the workflow were triggered from a content repo outside of the IDEMSInternational org

This PR makes the required secrets explicit for reusable pr-preview action

**For Follow-up**
Any actions that call the reusable-build action should also be updated to include named secrets. The jobs that call the reusable build action themselves should be able to keep `secrets: inherit` as both reusable actions are within the same org. 

## Review Notes
I tested passing the secrets and targeting this action branch from the kids-app-kw content repo, and can confirm action now able to pass deployment private key, however the key seems to be formatted incorrectly so action still fails (but with a different issue)
https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/pull/141

[Example Action](https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/actions/runs/12816533430/job/35737689274?pr=141)

## Dev Notes
I did see one github discussion about cross-org secrets, but it didn't seem to get much activity so I assume not something actively planned at this time `https://github.com/orgs/community/discussions/65766`

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/.github/workflows/reusable-deploy-pr-preview.yml
+++ b/.github/workflows/reusable-deploy-pr-preview.yml
@@ -29,6 +29,8 @@ on:
     secrets:  # Declare secrets you expect to receive
       FIREBASE_SERVICE_ACCOUNT:
         required: true
+      DEPLOYMENT_PRIVATE_KEY:
+        description: Provide private key if deployment uses encrypted config
 
 jobs:
   build:    


jfmcquade:
Thanks, @chrismclarke. 

The issue with the [PR on the `plh_kids_kw` content repo](https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/pull/141) was just that the wrong secret was being passed through, see [this amendment](https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/pull/141/commits/9be8aae1b85ad74e86646a258ea9561f90bf8c48). The action running on that PR branch is [working now](https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/actions/runs/12825944257/job/35764912137?pr=141) so I'm happy that the changes across both PRs are what's needed to fix the issue.

I will merge this PR, and follow up with the following:
- [ ] Make the `DEPLOYMENT_PRIVATE_KEY` secret explicit for the other relevant reusable actions
- [ ] Update the relevant actions on the `plh_kids_kw` content repo to pass the `DEPLOYMENT_PRIVATE_KEY` secret to the reusable actions

jfmcquade (approved):

PR #2727, opened 2025-01-17
fix: allow firebase auth to be enabled independent of crashlytics

jfmcquade:
PR Checklist

## Description

Currently the firebase service initialises if and only if a firebase config is provided at the deployment config level (`config.firebase.config`) _and_ at least one feature is explicitly `enabled` in the firebase deployment config property (e.g. `config.firebase.crashlytics.enabled`). This PR removes the second condition, so that the firebase service initialises if and only if a firebase config is provided.

This means that, for example, the `auth` feature can be enabled with firebase as a provider, without explicitly setting any further properties of the firebase deployment config beyond providing a `firebaseConfig` file.

## Git Issues

Closes #

## Screenshots/Videos


chrismclarke (changes requested):
Following https://github.com/IDEMSInternational/open-app-builder/pull/2634 the auth config was updated to assume enabled if a provider was included. I think likely we can do the same of initialising firebase if a config is provided and avoid confusion of default `enabled` states which in some cases are `true` and others are `false`

I might even be tempted to suggest offering a `disabled` override if there is good reason for an author to provide config but not enable the functionality (more just for firebase config which encompassed multiple services, and likely just crashlytics as the auth is configured within the auth provider)

Do you think that sounds sensible? Ideally I want a way to avoid overuse of `enabled: true` and `enabled: false` settings

jfmcquade:
Thanks @chrismclarke, that definitely sounds sensible (I was a bit rushed in making this PR initially so hadn't thought it through). I've updated the PR and description.

> I might even be tempted to suggest offering a disabled override if there is good reason for an author to provide config but not enable the functionality (more just for firebase config which encompassed multiple services, and likely just crashlytics as the auth is configured within the auth provider)

I haven't acted on this as I wasn't quite sure where you're suggesting the `disabled` property to be: at the level of `config.firebase.disabled` or `config.firebase.crashlytics.disabled`? I'm assuming the former, but as this fix is needed urgently I've kept the changes minimal for now.

esmeetewinkel (approved):
Functional test passed

esmeetewinkel:
@chrismclarke Could you confirm whether your requested changes have been addressed? From my perspective this is ready to merge.

chrismclarke:
Sorry for the delay, this one has been on my backlog for a while and aware it has some conflicts with #2742
I'll take a look now and try get it merged in

jfmcquade:
Thanks @chrismclarke, the changes look good so I'm happy to merge

PR #2733, opened 2025-01-20
feat: combo-box component disabled state

jfmcquade:
PR Checklist

## Description

Adds a new disabled state for the combo-box component. 
- The combo-box is disabled if manually set to disabled via a new `disabled` parameter (i.e. `disabled = true`
- Else the combo-box is automatically disabled if no answer list is provided, or if the answer list is empty.
- When the combo-box is disabled, it will display any text provided to a new parameter, `disabled_text`

Also fixes an issue where the displayed text in the combo box may not update correctly to reflect the value.

## Testing

Every configuration of the combo-box should be tested to ensure no functionality is broken. The [comp_combo_box](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=569531329#gid=569531329) component was quite confusing, I've updated it and am satisfied that each configuration seems to be functioning as expected.

## Git Issues

closes https://github.com/ParentingForLifelongHealth/plh-facilitator-app-mx-content/issues/111 (alongside content change already made to [menu_task_flow_base](https://docs.google.com/spreadsheets/d/1qG_NlXNGEhXg6y2nJ2UdnfnsGcoazAtURkCIesOLnzA/edit?gid=482991517#gid=482991517) template.

## Screenshots/Videos

[comp_combo_box](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=569531329#gid=569531329):

Screenshot 2025-01-20 at 13 34 33

Screenshot 2025-01-20 at 13 34 05

### DIFF
--- a/src/app/shared/components/template/components/combo-box/combo-box.component.html
+++ b/src/app/shared/components/template/components/combo-box/combo-box.component.html
@@ -2,9 +2,9 @@
   <button
     class="open-combobox margin-t-regular"
     (click)="openModal()"
-    [class.placeholder-style]="(!_row.value && placeholder) || prioritisePlaceholder"
-    [class.disabled]="disabled()"
     [ngClass]="{
+      disabled: disabled(),
+      'placeholder-style': (!_row.value && placeholder) || prioritisePlaceholder,
       'with-value': _row.value,
       'no-value': !_row.value
     }"
--- a/src/app/shared/components/template/components/combo-box/combo-box.component.html
+++ b/src/app/shared/components/template/components/combo-box/combo-box.component.html
@@ -3,22 +3,24 @@
     class="open-combobox margin-t-regular"
     (click)="openModal()"
     [class.placeholder-style]="(!_row.value && placeholder) || prioritisePlaceholder"
+    [class.disabled]="disabled()"
     [ngClass]="{
       'with-value': _row.value,
       'no-value': !_row.value
     }"
-    [disabled]="_row.disabled"
+    [disabled]="disabled()"
   >
-    {{ (text && !prioritisePlaceholder ? text : placeholder) | translate }}
+    {{ displayText() | translate }}
   </button>
 } @else if (variant === "dropdown") {
   <div class="combo-box-container dropdown">
     <ion-select
       [class.hasValue]="!!_row.value"
-      [placeholder]="(text && !prioritisePlaceholder ? text : placeholder) | translate"
+      [placeholder]="displayText() | translate"
       (ionChange)="setValue($event.detail.value)"
       interface="popover"
       labelPlacement="stacked"
+      [disabled]="disabled()"
     >
       @for (answerOption of answerOptions(); track answerOption.name) {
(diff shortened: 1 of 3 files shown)


src/app/shared/components/template/components/combo-box/combo-box.component.html
```diff
@@ -3,22 +3,24 @@
     class="open-combobox margin-t-regular"
     (click)="openModal()"
     [class.placeholder-style]="(!_row.value && placeholder) || prioritisePlaceholder"
+    [class.disabled]="disabled()"
```
chrismclarke, line 6: nit(non-blocking): we currently have 3 separate class bindings alongside an ngClass binding. Might be tidier to use a single computed property that defines all classes from within the `ts` file and just passes into html

jfmcquade: I tried using a single computed property, but since some of the classes depend on the non-signal `_row.value`, the values don't update properly. For now I've just consolidated them into a single `ngClass` binding, see https://github.com/IDEMSInternational/open-app-builder/pull/2733/commits/6f594741734d421ac9931bf0203628be537d8e3f

src/app/shared/components/template/components/combo-box/combo-box.component.ts
```diff
@@ -48,28 +50,39 @@ export class TmplComboBoxComponent
...
+        if (this.answerOptions().length > 0 && this._row.value) {
+          const selectedAnswer = this.answerOptions().find((x) => x.name === this._row.value);
+          if (!selectedAnswer) {
+            this.customAnswerSelected.set(true);
+          } else {
+            this.answerText.set(selectedAnswer?.text || "");
+          }
+        }
+      },
+      { allowSignalWrites: true }
```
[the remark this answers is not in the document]
jfmcquade: Ah, interesting. I think I'd read that `allowSignalWrites` was becoming `true` by default, but I hadn't looked into linked signals yet

src/app/shared/components/template/components/combo-box/combo-box.component.html
chrismclarke: There is a separate `rowSignal()` signal that can be used generally instead of `_row`, and then a more fine-grained `value()` signal that can be used instead of `_row.value` (see full list on `src\app\shared\components\template\components\base.ts`), but in either case not a huge issue

jfmcquade: I did try implementing this, but I found that the `value()` signal didn't seem to update as expected, nor trigger my computed property that referenced it to update. Logging out the value of `rowSignal().value` and `value()` suggested that the former was updating with updating the latter. After spending more time than I meant to trying to resolve it, I opted to keep the logic in the template for now (and delete some WIP code I accidentally committed to this PR), see #2744

PR #2736, opened 2025-01-21
[Bug Fix] Kuwait Theme: Center align list item text

### LINKED ISSUE #2717: [BUG] KUWAIT: Text alignment on plh_module_list_item
**Describe the bug**
Text on `plh_module_list_item` is centred in general but left (or right for RTL) aligned when it wraps on multiple lines. I think it would look better if it was always centred.

**App version**
v0.17.1

**Debug Sheet(s)**
[comp_module_list_item](https://idems-debug.web.app/template/comp_plh_module_list_item)

**Screenshots and videos**
Image   Image

FaithDaka:
PR Checklist

## Description
Center alignment for `module-list-item` text

## Git Issues

Closes #2717


jfmcquade (approved):

esmeetewinkel (approved):
Functional test passed.

I can see there is still an inconsistency when there is no image provided (left aligned when text doesn't wrap, centre aligned when it does). However, since that's not a use case we have at the moment and this component is only specific to one use case I'm happy to approve as-is.

image
image

PR #2737, opened 2025-01-21
[FEATURE] Kuwait Theme: Radio group styling

### LINKED ISSUE #2641: [BUG] KW radio group styling
**Describe the bug**
The buttons in the `radio_group` have different heights on some screen widths.

Current vs design: 
image             image

**App version**
v0.17.0 

**Debug Sheet(s)**
`comp_radio_group` [template](https://docs.google.com/spreadsheets/d/1qfatsiHKJ8sCBcJ8oqo3InLrjzT_a2Qxvw6RyxVMTxY/edit?gid=569531329#gid=569531329)

**Screenshots and videos**

[Screenity video - Dec 18, 2024 (1).webm](https://github.com/user-attachments/assets/edd60269-e146-400d-b0dd-1ffda4efd37d)

**Additional context**
One might consider making this a variant of the `radio_button_grid` instead, since that component is trying to replace the `radio_group`. Although this design is not a grid, so maybe that doesn't make sense...

FaithDaka:
19/02/25 UPDATED by @jfmcquade:

PR Checklist

## Description

Adds additional parameter options for the `radio_button_grid` component:
- New variant: `flex`
  - Arranges the items in a "flexbox" layout
  - Parameters `item_width` (taken to be _minimum_ item width) and `grid_gap` are still applied for this variant
- New style: `secondary`
  - Applies secondary colour styling to the items
  - Style variables are exposed to theming

As these are independent of each other, they can be set in any combination. E.g. an instance of the component with the default variant (grid) can have `style: secondary` applied.

At the bottom of the [demo sheet](https://docs.google.com/spreadsheets/d/1yHLQHxZQuxs34gBU-ny95ZWs9e8-HybC4IAeFxwpcVw/edit?gid=569531329#gid=569531329) is an example of the configuration to achieve the requested use case in #2641.

## Git Issues

Closes #2641

## Screenshots

[comp_radio_button_grid](https://docs.google.com/spreadsheets/d/1yHLQHxZQuxs34gBU-ny95ZWs9e8-HybC4IAeFxwpcVw/edit?gid=569531329#gid=569531329)

Screenshot 2025-02-19 at 10 14 29

Demo by theme:

| default | professional | pfr | plh_facilitator_mx | early_family_math | plh_kids_kw | plh_kids_tz |
|---|---|---|---|---|---|---|
| Screenshot 2025-02-19 at 11 43 41 | Screenshot 2025-02-19 at 11 43 53 | Screenshot 2025-02-19 at 11 44 04 | Screenshot 2025-02-19 at 11 44 15 | Screenshot 2025-02-19 at 11 44 25 | Screenshot 2025-02-19 at 11 44 37 | Screenshot 2025-02-19 at 11 44 49|


src/app/shared/components/template/components/radio-button-grid/radio-button-grid.component.ts
```diff
@@ -27,6 +27,8 @@ interface IRadioButtonGridParams {
   grid_gap: string;
   /* TEMPLATE PARAMETER: "variant". The style variant of the button_grid */
   variant?: "default" | "circle-icon";
+  /* TEMPLATE PARAMETER: "style". Defines the display style of the buttons. Default is grid */
+  style?: "grid" | "flex";
```
jfmcquade, line 31: nit(blocking): I think these should be variants really as opposed to styles. In that case, the easiest way to handle `grid` as the default is probably just to treat it as default in the styling, i.e. scope the current grid styling to apply by default, and override it with the flex styling in the case that `[data-variant~="flex"]`.

@chrismclarke do you agree that this should be a variant? In some sense these properties are orthogonal (in theory we could support `circle-icon` and `flex` in combination, for example, but that should be achievable with variants by just setting both, e.g. `variant: circle-icon flex`

jfmcquade (changes requested):
The code looks good, nice to have the differences scoped entirely to the CSS styles. 

See my inline comments: I think the `flex`/`grid` should be set through the `variant` rather than `style` param. But if there's not time to fix before a release is needed, I'm happy to merge as-is

FaithDaka:
Changes made @jfmcquade

src/app/shared/components/template/components/radio-button-grid/radio-button-grid.component.scss
```diff
@@ -77,3 +77,69 @@ $text-color: var(--radio-button-font-color, var(--ion-color-primary));
...
+    align-items: center;
+
+    overflow: hidden;
+    border-radius: 8px;
+    padding: 0px;
+    padding: 12px 14px;
+    margin: 4px 0;
+
+    max-width: calc(7.5rem * 0.9);
+    min-width: calc(6rem * 0.9);
```
jfmcquade, line 108: Where are these values coming from? And do we need to set a `max-` and `min-` width if we've already set the width explicitly above?

src/app/shared/components/template/components/radio-button-grid/radio-button-grid.component.scss
```diff
@@ -77,3 +77,69 @@ $text-color: var(--radio-button-font-color, var(--ion-color-primary));
...
+      border: 2px solid var(--ion-color-secondary-600);
+      padding: 8px 12px;
+    }
+  }
+  .item-image-container {
+    margin-bottom: 4px;
+    img {
+      max-width: 75px;
+      min-width: 60px;
+      max-height: 52px;
```
jfmcquade, line 123: nit(non-blocking): It would be nice to have these as dynamic values rather than explicit values – either percentages or calculated from the `$item-width` variable suggested above.

src/app/shared/components/template/components/radio-button-grid/radio-button-grid.component.scss
```diff
@@ -77,3 +77,69 @@ $text-color: var(--radio-button-font-color, var(--ion-color-primary));
...
+    width: 100%;
+    display: flex;
+    align-items: flex-end;
+    justify-content: center;
+
+    height: 46px;
+    overflow: hidden;
+    p {
+      color: var(--ion-color-gray-900);
+      font-size: 16px;
```
jfmcquade, line 137: nit(non-blocking): This should read from a variable. We already have a font size variable defined as 16px, `var(--font-size-text-small)`, defined in `src/theme/variables.scss`

jfmcquade (changes requested):
Thanks for making the updates. I've added some minor comments inline, but I've also noticed an issue with the [comp_radio_button_grid](https://docs.google.com/spreadsheets/d/1yHLQHxZQuxs34gBU-ny95ZWs9e8-HybC4IAeFxwpcVw/edit?gid=569531329#gid=569531329) demo template: the text of the second button, which wraps over three lines, is obscured by the image:

Screenshot 2025-01-23 at 14 54 14

esmeetewinkel:
I'll review this when code review has passed, @FaithDaka could you check me again when @jfmcquade has approved?

src/app/shared/components/template/components/radio-button-grid/radio-button-grid.component.scss
FaithDaka: These are responsiveness values carried over from the old radio-group styling file. The width is explicit but the item should still stretch or shrink according to the length of text.

FaithDaka:
Changes made @jfmcquade

esmeetewinkel:
@jfmcquade thanks. Just trying to understand how I would use this: So the `flex` parameter puts the items in a flex box, does that mean I can then apply any flex box properties (e.g. anything mentioned [here](https://css-tricks.com/snippets/css/a-guide-to-flexbox/)) through the style list? 

A specific example (which is the current use case): Would there be a way to achieve something resembling the design: 
image
where the box width adapts depending on the text length (which will be different when translated). In your screenshots all boxes have the same width, even though text lengths vary. Is this something like `flex-grow`?

jfmcquade:
> @jfmcquade thanks. Just trying to understand how I would use this: So the `flex` parameter puts the items in a flex box, does that mean I can then apply any flex box properties (e.g. anything mentioned [here](https://css-tricks.com/snippets/css/a-guide-to-flexbox/)) through the style list?

Whilst the `flex` variant does put the items into a flexbox, it's more that `flex` is being used here as a name, it doesn't actually expose the flex-related properties to authoring.

> A specific example (which is the current use case): Would there be a way to achieve something resembling the design: image where the box width adapts depending on the text length (which will be different when translated). In your screenshots all boxes have the same width, even though text lengths vary. Is this something like `flex-grow`?

Ah thanks for highlighting, it's a simple change to make the width adaptive rather than fixed, like this:

Screenshot 2025-02-19 at 11 14 03

Potentially both options could be available, exposed via a parameter like `item_width_fixed: false`. But as this would be pretty much achievable via the `grid` variant and as we don't have a use case I think it makes sense to just make this the default for the `flex` variant. I will update the PR description with new images

jfmcquade:
@esmeetewinkel I've updated the PR and description following our conversation, should now be ready for testing

esmeetewinkel (changes requested):
Thanks! 

- When only images or only text are provided, the button height should adapt to the content. 
- I [had understood](https://docs.google.com/spreadsheets/d/1yHLQHxZQuxs34gBU-ny95ZWs9e8-HybC4IAeFxwpcVw/edit?gid=712189259#gid=712189259&range=C3) `item-width` is a min width, but it seems to behave more like a max width in the flex case (or at least setting is causes narrower buttons). 

**Screenshot in `debug` deployment:**
image

**Screenshots in `plh_kids_kw` deployment:**
(without specifying `item_width`)
image

(specifying `item_width: 96px`)
image

esmeetewinkel (approved):
Thanks for making those final changes. Looks very much like the design to me!

image image

image

PR #2741, opened 2025-01-21
Fix: data items translations

### LINKED ISSUE #2725: [BUG] Translation for text in data items does not appear
**Describe the bug**
When a `text` component (or potentially any component) appears directly within a data items loop, the translation does not appear. This does appear when a subtemplate is used. 

**App version**
v0.17.1

**Debug Sheet(s)**
[debug_translation_data_items](https://docs.google.com/spreadsheets/d/1Bwh8eb2B43AAlDND4q0b4E-EXgnd6IRjmIFIf1Jbk1c/edit?gid=650194570#gid=650194570)

**Screenshots and videos**

Image

Image

chrismclarke:
PR Checklist

## Description
- Fix issue where data_items loops do not translate text from source data_lists
- Update tests

## Author Notes
The default template parser has various fallback methods that would attempt to translate data_list columns regardless of whether they had been marked for translation or not. This PR does not change default/legacy behaviour, but does add specific method for use within data_items loops that expects any data_list to include `my_column::eng` syntax for any translatable columns in the data_list 

## Review Notes
See video below for [debug_translation_data_items](https://docs.google.com/spreadsheets/d/1Bwh8eb2B43AAlDND4q0b4E-EXgnd6IRjmIFIf1Jbk1c/edit?gid=650194570#gid=650194570) sheet

Have also manually checked `comp_data_items` still behaving as expected, but would also be good to check any deployments where issue first detected.

## Dev Notes
The added tests for the translation system are not comprehensive, more just to test the specific data_list functionality added.

I opted to add an entirely new method to the `templateTranslateService` instead of using existing, as the existing methods are all setup to translate rows after data_list converted to item rows (looking at child row `value` column), instead of the initial list. For the case of data-items loops this is undesirable as it would mean re-translating every generated item row where values are included from the data_list (current behaviour for items loops, although at least these don't reprocess as often as data-items might). 

I found myself tidying up the `hackParseDataList` function a little, as when it was first copy-pasted the data used was a hashmap, and on refactor to use arrays a lot of the names/methods hadn't been updated in a very clear way. I didn't notice any knock-ons from this (but might be good to quickly sense-check)

## Git Issues

Closes #2725

## Screenshots/Videos

[Screenity video - Jan 21, 2025.webm](https://github.com/user-attachments/assets/0fa69329-813a-4cb0-a1be-1f5d403a8f10)

### DIFF
--- a/src/app/shared/components/template/services/template-translate.service.ts
+++ b/src/app/shared/components/template/services/template-translate.service.ts
@@ -140,7 +140,6 @@ export class TemplateTranslateService extends AsyncServiceBase {
     const [{ _translatedFields }] = rows;
     if (!_translatedFields) return rows;
     const translateKeys = Object.keys(_translatedFields);
-    console.log("translate list rows", JSON.parse(JSON.stringify(rows)));
     // translate rows
     return rows.map((row) => {
       for (const key of translateKeys) {
@@ -187,7 +186,7 @@ export class TemplateTranslateService extends AsyncServiceBase {
     return translated;
   }
 
-  subscribeToAppConfigChanges() {
+  private subscribeToAppConfigChanges() {
     this.appConfigService.appConfig$.subscribe((appConfig: IAppConfig) => {
       this.appLanguages = appConfig.APP_LANGUAGES;
       this.appLanguagesMeta = appConfig.APP_LANGUAGES_META;
--- a/src/app/shared/components/template/services/template-translate.service.ts
+++ b/src/app/shared/components/template/services/template-translate.service.ts
@@ -134,7 +134,7 @@ export class TemplateTranslateService extends AsyncServiceBase {
    * NOTE - this method is currently only used by data-items as regular processor translates after parsing
    * into template, where translateRow method can be used
    */
-  translateDataListRows(rows: FlowTypes.Data_listRow[] = []) {
+  public translateDataListRows(rows: FlowTypes.Data_listRow[] = []) {
     if (rows.length === 0) return rows;
     // use first row to identify list of fields for translation
     const [{ _translatedFields }] = rows;
--- a/src/app/shared/components/template/services/template-translate.service.ts
+++ b/src/app/shared/components/template/services/template-translate.service.ts
@@ -128,6 +128,33 @@ export class TemplateTranslateService extends AsyncServiceBase {
   }
 
   /**
+   * Translate all rows from a data_list. Uses `_translatedFields` metadata to determine which columns
+   * require translation, extracted from names with suffix `::eng`
+   *
+   * NOTE - this method is currently only used by data-items as regular processor translates after parsing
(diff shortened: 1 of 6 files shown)


esmeetewinkel (commented):
Functional test passed on `debug` and `plh_kids_kw` deployments (where the issue was first observed). 

> This PR does not change default/legacy behaviour, but does add specific method for use within data_items loops that expects any data_list to include `my_column::eng` syntax for any translatable columns in the data_list

Noted

src/app/shared/components/template/services/template-translate.service.ts
```diff
@@ -127,6 +127,32 @@ export class TemplateTranslateService extends AsyncServiceBase {
...
+    if (rows.length === 0) return rows;
+    // use first row to identify list of fields for translation
+    const [{ _translatedFields }] = rows;
+    if (!_translatedFields) return rows;
+    const translateKeys = Object.keys(_translatedFields);
+    // translate rows
+    return rows.map((row) => {
+      for (const key of translateKeys) {
+        const sourceText = row[key];
+        // TODO - confirm whether potential cases that list includes non-text entries
```
jfmcquade, line 147: I can't parse this comment, is the thought that we might want to translate non-text entries, specifically lists? If so, I agree – I think columns named `*_list` will be arrays at this point, so if that's a pattern we want to support then we may need to handle it explicitly.

@esmeetewinkel, do you see using a `*_list` column in a data list to define a list within a single entry as a feature that we currently support (and would want the constituent list entries to be translated)?

jfmcquade (approved):
Looks good, I've added one query comment inline

jfmcquade (approved):
Looks good, I've added one query comment inline

src/app/shared/components/template/services/template-translate.service.ts
esmeetewinkel: > [@esmeetewinkel](https://github.com/esmeetewinkel), do you see using a *_list column in a data list to define a list within a single entry as a feature that we currently support (and would want the constituent list entries to be translated)?

I do use `data_list` columns ending in `_list` already (for adding "tags", e.g. [here](https://docs.google.com/spreadsheets/d/1zMzxZ_GF4i9m0EpA3InD8kZSo8vxAQEmHnAKS8ou7Qk/edit?gid=0#gid=0&range=G1)). I've not yet used these for translatable things, but I could imagine that being desired in the future (e.g. specifying the entries of a combo box in a list)

chrismclarke: Yeah that's basically the sidenote here, I think probably fine to leave unsupported for now and probably over time move away a bit more from the pattern of inline arrays (probably better to handle with data_lists that would have better translation support). If we need support for inline array translation in the future in the future we can open an issue to address

Issue #2196, opened 2024-02-07
[FEATURE] Fix tests and incorporate into CI pipeline


jfmcquade:
From #2080:

## TODO

- [ ] Fix or remove all component test specs to include basic row value if parsing parameter_list in ngOnInit
- [ ] Fix service injection issues, creating new mock services to inject where required
- [ ] Fix other test issues
- [ ] Add ng test to web-build action tests
- [ ] Add support for ng test coverage
- [ ] Consider migrating component tests to cypress or other runner (or keeping component tests but adding back e2e)
- [ ] Start to add more meaningful component spec tests (could be follow-up)

## Description
Angular has a component and service testing system which is included during new component and service creation, however only partially used for some of the newer services added to the platform. 

This PR aims to fix and remove legacy code to allow the testrunner to function, and include any working tests within existing github actions testing pipelines. 

https://angular.io/guide/testing

## Git Issues

Closes #

## Screenshots/Videos
**Significant number of test issues to fix**

![image](https://github.com/IDEMSInternational/parenting-app-ui/assets/10515065/74490caf-5db9-49ca-b76e-8e35d1d0d3a7)

![image](https://github.com/IDEMSInternational/parenting-app-ui/assets/10515065/bae94720-dcb3-4b6c-9fe2-2be3a4afd045)

chrismclarke:
As a quick update on the current state of ng tests, 163 specs, 65 failures, 1 pending
![Image](https://github.com/user-attachments/assets/688e5b2c-88b6-4226-9790-58c79c9ce0f8)

I'm aware that angular is in the process of deprecating karma, however would still be good to get the entire suite passing with karma before trying to migrate to any other test runner to make it easier to identify which issues are legacy and which are migration

As as summary of the types of issues see table below

| reason                                                                                                                                                                                                                                                | specs                                                                                                                                                                                             |
| ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Failed: R3InjectorError(DynamicTestModule)[TaskService -> AppConfigService -> DeploymentService -> InjectionToken Application Configuration -> InjectionToken Application Configuration]:                                                             | CarouselComponent, TmplTaskCardComponent                                                                                                                                                          |
[...]

PR #2742, opened 2025-01-22
Fix: ng tests

chrismclarke:
PR Checklist

## Description
Provide a number of fixes across various ng components and services

## Review Notes
I don't think we need all tests passing before merge, better to not let this get too stale over time as will build up conflicts with ongoing prs and tests. But feel free if wanting to push any extra commits here if wanted. Otherwise further test fixes can be included in a follow-up

Should note that most of the code changes are just to stub/import mock providers and such. There are a couple minor changes to update some mocked services, but as these aren't currently integrated into CI anyway any accidental side-effect test breakages can just be resolved during ongoing work to general fixes.

## Dev Notes
See list below of summary of common issues and fixes implemented. 

**Pipe 'ExamplePipe' not found**
`declarations: [ExamplePipe]`

**Can't bind to 'routerLink'**
```ts
imports: [RouterModule]
```

**ion-content is not a known element**
```ts
imports: [IonicModule]
```

**tmpl-some-component is not a known element**
Any referenced child component importing child component into declarations, e.g. 
```ts
declarations: [MainComponent, TmplSomeComponent]
```
If this has significant knock-ons (e.g. child component has lots of dependencies, can also just mock and use that in declaration instead
```ts
@Component({ selector: "tmpl-some-component", template: "<div></div>" })
class MockTmplSomeComponent {}
```

**No provider for InjectionToken Application Configuration!**
usually side-effect from dependent DeploymentService
```ts
{ provide: DeploymentService, useValue: new MockDeploymentService({}) }
```

**No provider for HttpClient!**
usually side-effect from dependent AppDataService
```ts
{ provide: AppDataService, useValue: new MockAppDataService() }
```
or if using directly import testing module
```ts
imports: [HttpClientTestingModule]
```

**No provider for ModalController!**
If dependent of parent service use either mock or stub of parent service. 

If required by component/service then stub as required
```ts
{ provide: ModalController, useValue: {} },
```

**Cannot read properties of undefined (reading 'parameter_list')**
Component init populates data from parameter_list, set in `beforeEach` hook with mocked row values as required
```ts
fixture.componentRef.setInput("row", MOCK_ROW);
await fixture.whenStable();
component = fixture.componentInstance;
```

**Misc Errors**
There were a handful of errors related to child dependencies which were being somewhat tricky/problematic. In these cases I've simply marked the default test as skipped for now (using the `xit` shorthand), so that it can be addressed in the future
```ts
xit("should create", () => {
    expect(component).toBeTruthy();
  });
```

### Code Changes
The only frontend code changes that I'm aware of:
- make `firebase` config optional in type definitions (although default value `{config: null}` retained for now to avoid knock-ons)

## Git Issues
Progress towards #2196 (but can only close once all tests passing and integrated into CI)
Closes #

## Screenshots/Videos
Current status. Should not that number of fails is not truly indicative of number of issues to resolve as many components/services have multiple issues to fix but fail when first encountered (so is a lower bound)

image


chrismclarke:
Another update, now all passing (except 4 which I've marked as skipped).
I think this should be good for now, individual test suites can be improved over time and when working on specific components. I'll look into adding to CI for next steps

image

chrismclarke:
@jfmcquade 
I've just updated, resolved conflicts with recently merged PRs, and ensured all current tests are passing

image

Given that this will just continue to fall out of sync with any other tests introduced I think it would be good to merge here (as mentioned previously, doesn't need a formal review as the PR is focused just on fixing and unblocking broken test suites)

I'll probably go ahead and just enable admin override to merge (as would be good to have tests passing to compare some pending code changes against), feel free to add a post-merge review when you have the chance

chrismclarke:
I've reverted the firebase config changes made in this pr to fix broken tests (now handled in #2727), so the only non spec changes here should be the minor changes to `src/app/feature/theme/components/css-variable-table/css-variable-table.component.ts`

As such I'm going to go ahead and merge to unblock ongoing work, and make a follow-up PR to integrate the test suite into PR testing

jfmcquade (approved):
I've read through the info (including the repeated requests for review, sorry!) and it all looks good to me.

Exciting to have the tests passing and to move towards integrating them into CI! Also really nice to have the different types of fix spelled out in the description, thanks.

PR #2748, opened 2025-01-27
chore: PLH Kids TZ theme

esmeetewinkel:
PR Checklist

## Description

This PR creates a theme specifically for the PLH Kids TZ deployment. Currently, it is just a copy of the `professional` theme but with 
- the secondary colour adapted from orange to yellow
- the progress bar colour adapted from green to blue

## Screenshots/Videos

Design vs this PR: 
image  image

## Comments
1. @jfmcquade the button text on these is required to be a dark shade of the primary colour, I think `colour-primary-800` or `900` would work. Would that be easy to add to this theme? 

   Perhaps relevant, you commented [here](https://github.com/IDEMSInternational/open-app-builder/pull/2259#issuecomment-2034922270) that the button text colour is auto computed to be white or black (I actually expected the highest contrast on this yellow shade to be black).
image image

2. @jfmcquade I think the [white footer](https://github.com/IDEMSInternational/open-app-builder/pull/2508) is currently only available to the `plh_kids_kw` theme. Could this be made available to the `plh_kids_tz` theme too?


esmeetewinkel:
I created content PR https://github.com/ParentingForLifelongHealth/plh-kids-app-tz-content/pull/159. Following discussion with @jfmcquade, updates on my two comments above: 

1. This will be done as a follow-up to this PR, issue here.
2. I have used the regular app footer (as opposed to the Kuwait specific footer) with a config property `config.app_config.APP_FOOTER_DEFAULTS.background = "none"`  for now. This shows the icon when selected and shades it slightly when not selected. Only remaining issue would be to allow this footer component (as an alternative) to take two icon input, one for selected and one for not selected. Issue here.

30ann (approved):

PR #2750, opened 2025-01-27
hotfix: action parameter self-reference concatenation

### LINKED ISSUE #2745: [BUG] set_data action on "changed" trigger does not have latest row value

jfmcquade:
PR Checklist

## Description

Follow-up to #2749.

Adds a workaround to handle a specific use case required on the `plh_facilitator_mx` deployment but not currently supported: authoring an action with a parameter value that uses the current value of its row in a concatenation. From [menu_task_flow](https://docs.google.com/spreadsheets/d/1qG_NlXNGEhXg6y2nJ2UdnfnsGcoazAtURkCIesOLnzA/edit?gid=1703276911#gid=1703276911):
```
changed | set_data |
_list_id: sub_items,
completed: @item.@local.completed_prefix+this.value;
```
Here the `this.value` reference is expected to be a string which will then be concatenated with the `@local.completed_prefix` value to ultimately reference a particular property on the current item, equivalent to authoring, for example, `completed: @item.completed_1`.

The included workaround is quite brittle: it requires the specific authoring syntax of `+this.value` (with no spaces before or after the `+` symbol) in order to make the substitution.

## Git Issues

~Provides a hacky workaround for #2745, but does not resolve the underlying issue.~
Closes #2745

## Screenshots/Videos


ChrisMarsh82 (approved):
This fixes my issue in facilitator MX

chrismclarke (changes requested):
Just to put a blocker on this as I believe it builds on top of #2749 and so shouldn't be merged until after

jfmcquade:
@chrismclarke I've tagged you for review again. The changes are much clearer after the updates to #2749

chrismclarke (approved):
@jfmcquade 
I've pushed 3bb2c316b8217d103675754d5fe04cf4397d1d8d to tidy up the code a bit and add a test case. I've modified the replacer function to look for `{this.value}` as I think this is more likely to be supported by default with future changes (so should be able to remove workaround at future stage). I've also updated the #2753 TODO items to include check against this case

I think this should reasonably address the issue flagged in #2745 so have marked this PR to close it, although if you think something is still not covered feel free to re-open or create a new issue to capture what parts are still missing

I've also updated the debug sheet to minimally capture what is now working (as well as what is not)
[debug_set_data_changed](https://docs.google.com/spreadsheets/d/1dR6v2rq_1EVCC07YVfuaSTKJOuZs666ykCsceFmOKks/edit?gid=1422144003#gid=1422144003)

https://github.com/user-attachments/assets/074688a0-2423-407b-b970-6125c2d2ce23

jfmcquade:
Thanks, @chrismclarke, the updates look good. I agree that it makes sense for this PR to close #2745.

@ChrisMarsh82 – since the MX deployment is targeting this branch, I've already updated the syntax in the [menu_task_flow](https://docs.google.com/spreadsheets/d/1qG_NlXNGEhXg6y2nJ2UdnfnsGcoazAtURkCIesOLnzA/edit?gid=1703276911#gid=1703276911) template to reflect the changes (cell D14). I believe that was the only case of the `+this.value` syntax, but correct me if not.

ChrisMarsh82 (approved):
Checked on MX

Issue #2622, opened 2024-12-12
[BUG] "now" datetime not re-calculated when used as a value in an action


jfmcquade:
**Describe the bug**
There are various ways to get a timestamp representing the current time from within a template (e.g. `@calc(new Date())`, `@calc(Date().now)`, `@calc(now())` (a custom exposed calc function), etc.). When these values are referenced in local variables and used in calculations or rendered within a component (e.g. displayed as `text`), the value of "now" seems to update to reflect the time at which the value was retrieved.

However, when using a value calculated from the current datetime is used as the value in a template action, e.g. `set_local: my_local_var: @calc(now())`, then the value does not get set as expected. It seems that the first time the action is fired, the value is set as expected, but on subsequent firings, the value does not update. Debug logs show that for every time the action is fired after the first time for the current initialisation of the template, the value received as the argument/parameter of the action remains the value of "now" at that first initial firing.

In fact, I think the value of `now()` that gets passed to the action is the timestamp for when the template was first initialised, which suggests that it is being evaluated only once on template init. This seems to be the case even when the `now()` call is within another local variable being referenced in the action value, which is strange.

**App version**
0.17.0

**Debug Sheet(s)**
Debug template: [debug_date_now](https://docs.google.com/spreadsheets/d/1VKoE4XTH4spXKQahoBGoLuurqeGMt-4sK6bvla445h4/edit?gid=314433546#gid=314433546)

<img width="800" alt="Screenshot 2024-12-12 at 16 49 46" src="https://github.com/user-attachments/assets/e0b2d037-3dcc-4223-999e-814aeb70c076" />

Demo below. Note that while the text component instantiated on row 5 updates to reflect the changes made to the `calculated_date` local variable, these changes are not reflected when using that local variable as the value to set as the value of `output_date`.

PR #2753, opened 2025-01-27
spike(action_lists)!: evaluation standalone

chrismclarke:
PR Checklist

## Breaking Changes
 **action_list evaluates using newer templated data system** 
This may lead to some unexpected behaviour if any action_lists include complex statements that mix variable references, e.g. `click | set_field : some_field : hello this.value`. I think these types of patterns had always been discouraged/not worked anyways (with preference of setting intermediate local variables instead), but just to be aware of some small potential for knock-on. Unfortunately we don't have many tests for action_list edge cases so might be good to check against existing debug sheets. 

**action_list values references now auto-update**
Previous workaround to specify `this.value` in action_lists to handle receiving updated value for triggered actions are no longer required. The syntax is still supported and not marked as deprecated (no plans to remove), but any documentation that refers to this could likely be updated.

## Description
Action lists are currently evaluated as part of main template container processing, meaning that all row action params and args are evaluated prior to the row that triggers them from rendering. There are 2 main drawbacks to this:

1) Actions need to be parsed even if they are never actually triggered. This isn't too expensive an operation, just lacking a bit in optimisation.

2) Actions can become stale if data changes between render and trigger. This typically happens if an is triggered by a component change (parsed action will have value before change), or if multiple actions are triggered in series where the first action manipulates data required for the second action.

This PR aims to detach action_list processing from the main templating system and instead only evaluate at the time of being triggered. This should address both issues 

Additionally it evaluates actions using the new app data templating system, which has better support for 

**TODO**
- [ ] Tests (requires previous pr)
- [ ] Resolve potential conflicts with outstanding PR

**Follow-up**
- [ ] Consider refactoring all dynamic evaluation to similar format and remove from runtime
- [ ] Dynamic evaluator registry
- [ ] Handle `this.` workarounds

## Dev Notes
parser vs runtime

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/data-models/flowTypes.ts
+++ b/packages/data-models/flowTypes.ts
@@ -338,7 +338,7 @@ export namespace FlowTypes {
 
   const DYNAMIC_PREFIXES_COMPILER = ["gen", "row", "default"] as const;
 
-  const DYNAMIC_PREFIXES_RUNTIME = [
+  export const DYNAMIC_PREFIXES_RUNTIME = [
     "local",
     "field",
     "fields",
@@ -354,6 +354,8 @@ export namespace FlowTypes {
     ...DYNAMIC_PREFIXES_COMPILER,
     ...DYNAMIC_PREFIXES_RUNTIME,
   ] as const;
+
+  export type IDynamicPrefixRuntime = (typeof DYNAMIC_PREFIXES_RUNTIME)[number];
 
   export type IDynamicPrefix = (typeof DYNAMIC_PREFIXES)[number];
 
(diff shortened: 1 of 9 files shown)


chrismclarke:
@jfmcquade 
I've still got a bit more work to do on this PR but I was wondering if I could ask you to take a first pass for initial impressions and whether you think anything could do with some additional explanation/tidying/refactor?

jfmcquade:
Thanks @chrismclarke. I've only had a chance for a quick look through, but the changes definitely make sense from a high level, it seems right that actions should evaluate only at the point of being triggered. I should get a chance to look in a bit more detail tomorrow.

In terms of timelines, because of the potential for knock-ons with action handling and the critical status of two deployments currently being worked on (`plh_kids_kw` and `plh_facilitator_mx`), I think it probably makes sense to wait until those have been released before merging these changes. Although saying that, it would be good to discuss with @esmeetewinkel and @ChrisMarsh82 if they expect any knock-ons from the breaking changes you've highlighted. I've added this PR as a discussion point for our meeting tomorrow, we can decide how much depth we need to go into as a group.

jfmcquade:
NB, this should resolve #2622

PR #2757, opened 2025-01-28
feat: make 'disabled' parameter of button component dynamic

jfmcquade:
PR Checklist

## Description

Enables the button component’s disabled/enabled state to dynamically reflect updates to the `disabled` parameter.

Additionally, this PR removes support within the button component for the row-level `disabled` property, which is considered deprecated. See follow-up #2761 to roll out this change to other components.

## Dev notes

As we've discussed previously (for example, [this comment thread](https://github.com/IDEMSInternational/open-app-builder/pull/2493#issuecomment-2446930347) on #2493), it would make sense to embrace angular signals more fundamentally in how we handle component parameters. This PR doesn't make any progress towards achieving this more generally, but ultimately it would likely make sense for all our authorable parameters to be parsed to computed/signal values that react to updates.

## Git Issues

Closes #

## Screenshots/Videos

Updated [comp_button](https://docs.google.com/spreadsheets/d/1OmgZICjM5EMT1KgLOU_ovDljRF_SPlpQAgkKWPrNX0s/edit?gid=569531329#gid=569531329):

Screenshot 2025-01-28 at 12 37 24

https://github.com/user-attachments/assets/b2a7f582-85c1-4827-b7ae-e74a4700bfcb

### DIFF
--- a/src/app/shared/components/template/components/button/button.component.ts
+++ b/src/app/shared/components/template/components/button/button.component.ts
@@ -51,11 +51,13 @@ interface IButtonParams {
 })
 export class TmplButtonComponent extends TemplateBaseComponent implements OnInit {
   params: Partial<IButtonParams> = {};
-  disabled = computed(
-    () =>
-      getBooleanParamFromTemplateRow(this.rowSignal(), "disabled", false) ||
-      this.rowSignal().disabled
-  );
+  disabled = computed(() => {
+    // TODO: is disabling by row still supported generally?
+    const disabledByRow = this.rowSignal().disabled;
+    if (disabledByRow) return true;
+    const disabledByParam = getBooleanParamFromTemplateRow(this.rowSignal(), "disabled", false);
+    return disabledByParam;
+  });
   /** @ignore */
   variantMap: { cardPortrait: boolean };
 
--- a/src/app/shared/components/template/components/button/button.component.ts
+++ b/src/app/shared/components/template/components/button/button.component.ts
@@ -1,4 +1,4 @@
-import { Component, ElementRef, OnInit } from "@angular/core";
+import { Component, computed, ElementRef, OnInit } from "@angular/core";
 import {
   getStringParamFromTemplateRow,
   getBooleanParamFromTemplateRow,
@@ -51,6 +51,11 @@ interface IButtonParams {
 })
 export class TmplButtonComponent extends TemplateBaseComponent implements OnInit {
   params: Partial<IButtonParams> = {};
+  disabled = computed(
+    () =>
+      getBooleanParamFromTemplateRow(this.rowSignal(), "disabled", false) ||
+      this.rowSignal().disabled
+  );
   /** @ignore */
   variantMap: { cardPortrait: boolean };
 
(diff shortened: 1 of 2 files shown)


src/app/shared/components/template/components/button/button.component.ts
```diff
@@ -51,6 +51,11 @@ interface IButtonParams {
 })
 export class TmplButtonComponent extends TemplateBaseComponent implements OnInit {
   params: Partial<IButtonParams> = {};
+  disabled = computed(
+    () =>
+      getBooleanParamFromTemplateRow(this.rowSignal(), "disabled", false) ||
+      this.rowSignal().disabled
```
chrismclarke, line 57: nit(blocking): I found myself a bit confused while making sense of this before I realised that we have both disabled within the parameter_list as well as disabled within the row. Can you remember why we have both cases and which should take precedence if there is a mismatch? In the replaced code it looks like the param value actually gets updated to reflect the row value, which I think no longer happens here, so would be good to confirm whether this is desirable or not

Once determined could you write this as a more direct function for clarity. E.g. if row takes precedence
```ts
const disabledByRow = this.rowSignal().disabled
if(disabledByRow) return true
const disabledByParam =  getBooleanParamFromTemplateRow(this.rowSignal(), "disabled", false)
return disabledByParam 
```

chrismclarke (changes requested):
Always happy to move towards more signal calls, although could you just check the one inline comment to clarify some expected behaviour?

src/app/shared/components/template/components/button/button.component.ts
jfmcquade: Good point, I wasn't too sure about the row-level disabling but intended to just preserve the existing functionality. I've updated to preserve the precedence and add clarity as suggested with https://github.com/IDEMSInternational/open-app-builder/pull/2757/commits/cf271bdc5fc999fc47db56fffcd2738b1269c6de

In terms of why we have both cases, I don't think I've come across the row-level `disabled` property before. I can see it is consumed by various (older) components:

Screenshot 2025-01-29 at 11 46 36

For all of these components listed, they do not support a `disabled` param, so currently the only way to disable them is via the dedicated row-level property.

But the only example of an authored `row.disabled` property within the `debug` deployment sheets is in [example_workshops_accordion](https://docs.google.com/spreadsheets/d/1gfnWmr59DF0ZsuSuBh0K1o9fMqEg-z6Qa7eJNsiB-rQ/edit?gid=1595229908#gid=1595229908).

I've added it to the discussion points for the meeting later today – it would be good to check whether the authored `disabled` column should still be supported. It may make sense to refactor these components to consume a `disabled` param instead and drop support for row-level disabled property.

jfmcquade:
@chrismclarke – Following our discussion, I've updated this PR to remove support within the button component for the row-level `disabled` property. I've added #2761 as a follow-up, to refactor other components to use a `disabled` component param as opposed to the row-level property.

@esmeetewinkel this PR is ready for functional testing.

chrismclarke:
Apologies @jfmcquade - I actually made a refactoring PR targetting this but forgot to tag you for review
It's #2760 if you have a chance to look

esmeetewinkel (commented):
I'll test on refactor branch instead

jfmcquade (commented):
Merging as these changes have been tested on #2760

esmeetewinkel (approved):

PR #2758, opened 2025-01-28
fix: text box value bug

### LINKED ISSUE #2759: [BUG] (MX) Text box shows old value when adding additional parents
**Describe the bug**
When adding subsequent parents after the second, the "first_name" text box displays the value of the previous first name that was added

[add_parent](https://docs.google.com/spreadsheets/d/14iUeYWUHIly3tyz61Wgz_tlPUVAqkI8aXFnlsrUA1ZY/edit?gid=293189857#gid=293189857) template:

https://github.com/user-attachments/assets/b3965382-4de4-4901-9bf6-a4d2d6882a4a

jfmcquade:
PR Checklist

## Description

Fixes a bug where the value of the text-box would sometimes not update as expected.

## Testing

This is quite a difficult issue to create a debug template for, since really we want direct access to the component's `value()` signal to check that it updates correctly, which is how I was debugging as a dev. I think it should suffice to confirm that the issue as observed is now resolved.

## Dev notes

Technically, this PR resolves a general issue with components' `value()` computed property not always being updated as expected via the `setValue()` component method (#2706 is relevant).

In debugging the issue, I observed that the component's `value()` computed property would not update as expected, even though `rowSignal().value`, on which it depends, was updating as expected. I observed that removing the `{ equal: isEqual }` option from the definition of the `rowSignal()` signal appeared to resolve the issue. I believe the issue to be as follows:

- When we update the value in `setValue()` via `this._row.value = value;`, we're mutating the `_row` object directly, but since rowSignal is using `isEqual` for comparison, it ends up comparing the same object reference with itself, returning a false positive, and preventing the changes from propagating.

At first I thought it would make sense to update the `isEqual()` function to handle such cases of self reference, but this proved difficult. I tried updating the first check to be `if (a === b && typeof a !== "object") return true;`, as well as updating the checks on object literals. I couldn't get these to work.

However, after some consideration I think that the change in the current PR are preferable: essentially creating a new object reference when updating the value of `rowSignal()`.

## Git Issues

Closes #2759 

## Screenshots/Videos

`plh_facilitator_mx` [add_parent](https://docs.google.com/spreadsheets/d/14iUeYWUHIly3tyz61Wgz_tlPUVAqkI8aXFnlsrUA1ZY/edit?gid=293189857#gid=293189857) template:

https://github.com/user-attachments/assets/f3da9b3d-d445-47cd-8c7e-a03725dc6b19

### DIFF
--- a/src/app/shared/components/template/components/base.ts
+++ b/src/app/shared/components/template/components/base.ts
@@ -36,7 +36,7 @@ export class TemplateBaseComponent implements ITemplateRowProps {
    **/
   @Input() set row(row: FlowTypes.TemplateRow) {
     this._row = row;
-    this.rowSignal.set(row);
+    this.rowSignal.set({ ...row });
   }
 
   /**
--- a/src/app/shared/components/template/components/base.ts
+++ b/src/app/shared/components/template/components/base.ts
@@ -36,6 +36,7 @@ export class TemplateBaseComponent implements ITemplateRowProps {
    **/
   @Input() set row(row: FlowTypes.TemplateRow) {
     this._row = row;
+    // take shallow clone to still be able to detect changes if this._row directly modified
     this.rowSignal.set({ ...row });
   }
 
@@ -70,6 +71,7 @@ export class TemplateBaseComponent implements ITemplateRowProps {
   async setValue(value: any) {
     // HACK - provide optimistic update so that data_items interceptor also can access updated row value
     this._row.value = value;
+    this.rowSignal.update((v) => ({ ...v, value }));
 
     const action: FlowTypes.TemplateRowAction = {
       action_id: "set_self",


ChrisMarsh82 (approved):
Tested on MX app and fixes issue

chrismclarke (approved):
I had left a bit of a longer comment but it seems to have got lost somehow. But to sum up, really interesting issue, and I think you're definitely on the right track with your assessment.

I was able to put together a more minimal debug sheet which seems to replicate the issue (text box only wipes after first edit and not second)
[debug_text_box](https://docs.google.com/spreadsheets/d/1B9X6lvSaLUHh7XmuNkJbK2Pup1H__jGf_DfvPQvLL7w/edit?gid=1422144003#gid=1422144003)

**Demo Before**
https://github.com/user-attachments/assets/0fc652f5-c72b-4c30-9c21-bf274098477e

Taking a (much) deeper dive it turns out not so much related to the isEqual check, but the fact that the signal value doesn't get updated properly when performing setValue (it updates the _row property but then as you noted the signal comparison is comparing against references to the already updated value).

I found that as well as your included fix I need to also ensure the signal updates as expected during a `setValue` action, and have pushed [b7fd84b](https://github.com/IDEMSInternational/open-app-builder/pull/2758/commits/b7fd84b669cb52f03d4a9af9c00608d2810ab2df) to this PR

**Demo After**
https://github.com/user-attachments/assets/a7a50858-b1c1-4318-b0a0-15ce2338a4fc

I think it should now be good to go, but please double check the referenced issue is still fixed following my changes

chrismclarke:
> Great, thanks for looking further into it, @chrismclarke. Good to have a debug sheet. I can confirm that with your additional changes the bug on the `plh_facilitator_mx` deployment is still resolved, so I'm happy to merge.
> 
> I'm obviously a fan of signals, but I do find in cases like this I find it hard to debug when the reactivity appears to be configured correctly. For example, I had previously tried a change along the line of adding `this.rowSignal.update((v) => ({ ...v, value }));` but found that it didn't resolve the issue in isolation. And likewise I'm not sure why both our changes are needed in order to resolve the issue on your debug sheet, when only the first was sufficient to resolve the issue as observed on the MX deployment. In any case, at a high level I can see how the changes resolve the issue we were seeing, and I do have a better understanding of the signal equality function now.

Yeah I really wish I could fully explain but the overall system is still just a little too complex, and the issue lies within that awkward intersection of variable processing that's passed through multiple services and components, as well as deep nested values and the quirks around javascript object references. 

In the debug sheet I created I added logs to the equality function and was able to realise it wasn't even being triggered (let alone evaluating) - although why for the mx deployment it would clear 2 text boxes but not the first is beyond me, not sure if there could  be some weird timing issue in the angular change detection. Could possibly have tried to examine a little more closely with https://angular.dev/tools/devtools

But in both cases I think if we proceed with move away entirely from the `_row` system and just use signals it should both simplify and provide a bit better observability in the future. More work definitely required as well to try and remove some of the complexity from the overall system too

PR #2760, opened 2025-01-29
Refactor: button disabled dynamic

chrismclarke:
PR Checklist

## Description
Targets branch of #2757

- Update parameter_list handling to update on change (not just init)
- Replace mapped params with signal

## Review Notes
Confirm example/test cases targetted in original PR branch still work

## Dev Notes
Looking through button code and proposed changes I didn't think it fully made sense to make just one single param responsive whilst keeping the others as is. 

I think moving forward it would make sense to move away from `ngOnInit` methods to process parameter_list variables, and instead use signals to compute component params whenever the author `parameter_list()` signal updates. 

This is a little awkward as for whatever reason we have a bunch of utility methods that take the entire row before parsing parameter list values to specific types (e.g. `getStringParameterFromRow`), but I also think these should likely also be deprecated and slowly moved away from. As a shorter term solution I've kept the methods in place and still pass the full row, just using the `parameter_list()` signal as a trigger to evaluate. 

It also may be worth considering whether we want to push for keeping the list of mapped params within a single `params()` signal (and a base method that can be extended to pass a mapping function from template to component params), or whether to leave to individual components and prefer to expose individual values/signals for each of the params used. I'd personally lean towards the latter (e.g. top-level `this.disabled`, `this.variant`, `this.textAlign` etc.) to allow for more performant change detection and minimal re-rendering if we ever get to a stage where parameter list updates don't force entire row re-renders, but short term I don't think it makes much difference.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/components/button/button.component.html
+++ b/src/app/shared/components/template/components/button/button.component.html
@@ -1,33 +1,33 @@
-<ng-container [ngSwitch]="variantMap.cardPortrait">
+<ng-container [ngSwitch]="variantMap().cardPortrait">
   <!-- Default variant -->
   <ion-button
     *ngSwitchDefault
-    [class]="'full standard medium' + ' ' + params.style + ' ' + params.buttonAlign"
-    [disabled]="disabled()"
+    [class]="'full standard medium' + ' ' + params().style + ' ' + params().buttonAlign"
+    [disabled]="params().disabled"
     (click)="triggerActions('click')"
-    [attr.data-param-style]="params.style"
-    [attr.data-variant]="params.variant"
+    [attr.data-param-style]="params().style"
+    [attr.data-variant]="params().variant"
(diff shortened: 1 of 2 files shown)


esmeetewinkel (approved):
Functional test passed

jfmcquade (approved):
Thanks @chrismclarke, definitely nice to have to have the responsiveness rolled out further. I'll merge this along with the PR for the targeted branch.

Thinking about future patterns for handling component params, in terms of a single `params()` signal vs individual signals for each param, I agree that we should probably favour the more performant, latter option. But I did like the idea of having a base level mapping function that could potentially parse each template param to a component param base on its type as specified in the component's defined interface, without needing to handle each param explicitly. It would be nice if we could achieve something like this, although I can't think how that would be possible with individual signals/computed properties as each would need to be explicitly defined as a property of the component class.

Issue #2766, opened 2025-01-30
[FEATURE] Intermediate variables in data items


esmeetewinkel:
**What?**
Ability to have intermediate variables in data items loops. 

**Why?**
To enable authoring patterns like the one below

`plh_kids_kw` [module_path](https://docs.google.com/spreadsheets/d/1nvyc-d_noz-W7guO6x_33c6cCXMXpexGt5HYf5T5J50/edit?gid=14877797#gid=14877797) template
![Image](https://github.com/user-attachments/assets/be5405b9-8723-47d1-99c5-b5b78b9e7cd1)

Currently the local variables display as `undefined`

PR #2767, opened 2025-01-31
Refactor: eval context typing

chrismclarke:
PR Checklist

**TODO**
- [x] Would benefit from having ng test suite passing from #2742 (but can be tested after as part of follow-up work in #2753)

## Description
As part of ongoing work in both #2753 and future PR to address #2766, improve some of the type-safety and handling of template variables, in particular:

**Create a `TemplateRowEvalContext` type** 
This defines the variables used when evaluating dynamic contexts. This is designed to be compatible with changes proposed in #2753, where variables are namespaced by dynamic prefix, e.g. 
```ts
{ local: my_local: 'some_value' }
```
Previously this was typed as `any`, and used alternate naming conventions `itemContext` for item, and `templateRowMap` for local variables. 

**Tidy handling of templateRowMap**
The same variable was used both as part of calc context, and for parser operations that depended on evaluating parent rows. As such entire rows were included in calcContext, leading to some messy code in places (there had initially been the proposal to allow referencing any column from a local variable, e.g. `@local.my_local.action_list`, however this was never actually implemented so lots of code to always use default `value` column). The full row is still needed for template parsing, so separate variables are now tracked for dynamic evaluation vs general processing.

**Add template row value type**
As row values are defined in template sheets it is limited what data_type they can natively be (string, boolean, number, undefined). These types have been added to the general definition, and component code tidied a little to make it clear where specific types are being assumed (previously would not show lint warnings as value could be `any` type). 

In the future we may want to include more QC/validation to ensure components receive values in the correct data type, but for now this PR just makes the type updates

Whilst there are some cases where row values get converted into objects (e.g. begin_items which parses the value from a data_list), these are not included in the general type as really the code should probably not be overwriting initial values in such a way (better to store as some additional metadata field). So these cases currently re-cast values back to `any`, with some TODO notes (particularly for the case of items, likely the entire component will be removed at some point when safe to merge with data_items) 

## Git Issues

Closes #

## Screenshots/Videos


src/app/shared/components/template/services/instance/template-row.service.ts
```diff
@@ -244,11 +248,12 @@ export class TemplateRowService extends SyncServiceBase {
...
   ) {
-    const { _nested_name, _evalContext } = preProcessedRow;
+    const { _nested_name, _evalContext = {} } = preProcessedRow;
     // Evaluate row variables in context of current local state
-    const evalContext = {
+    // TODO - only extract local variables as needed
+    const evalContext: FlowTypes.TemplateRowEvalContext = {
       ..._evalContext,
-      templateRowMap: this.templateRowMap,
+      local: { ...this.templateRowMapValues, ..._evalContext.local },
```
chrismclarke, line 256: NOTE- this is a feature change which slipped into this PR from the follow-up work. It shouldn't be too significant, just allowing additional local values to be included in dynamic evaluation context. This will be used by data_items to create intermediate variables without the need for uniqueness across the template (each data item row processes with access to its own value for the created local variable)

jfmcquade (approved):
Thanks @chrismclarke, appreciate you breaking this work out and giving an explanation. I like the new context variable namespacing, and the typing makes things clearer in lots of places.

PR #2768, opened 2025-02-01
Feat/share multiple

### LINKED ISSUE #2764: [FEATURE] Ability to share both text and image
**What?**
The `share` template action should support sharing both text and image. 

Support for sharing text and image is theoretically supported by the [Capacitor Share API](https://capacitorjs.com/docs/apis/share) (in particular when the image is provided as a URL). My research suggests that this should be supported on Android, but [may not be on iOS](https://stackoverflow.com/a/55688479), with a suggestion that  although I can't find a definitive up-to-date source. In any case, the interpretation of the shared content is app-specific, so we'll have to make sure through testing that whatsapp supports sharing both text and image in this way.

**Why?**
Users would like to be able to share a text and image combination to whatsapp through a single action.

**How?**
The `share` action syntax will need to be updated, from the current, arg-list based syntax, e.g.:
```
click | share: text: @local.shareable_text
```
to a param-based syntax:
```
click | share | text: @local.shareable_text, image: @local.image
```

This should be possible without dropping support for the current syntax.

**Other details**

jfmcquade:
PR Checklist

## Description

Allows authoring of sharing multiple data types at once via the platform's share API, e.g. an action to share both a text and image to be sent via WhatsApp (https://github.com/ParentingForLifelongHealth/plh-facilitator-app-mx-content/issues/169).

Refactors the `share` template action to use a new syntax. Previously, the share action use the syntax of a list of arguments, with the ability to specify a single data type, e.g. `share: file: path/to/image.png`. Now, a key/value pair parameter based syntax is supported, bringing this action in line with more recently added ones, as well as enabling support for multiple data types via a single action. E.g. `share | file: path/to/image.png, text: @local.some_text`.

| Parameter.  | Value | Notes |
|------|---|---|
| text | A string of text to share | |
| image | A path to an image, in usual authoring format | |
| url | A URL string | |
| title | A string of text | This will be interpreted by some apps, e.g. it will populate the subject line of a new email when shared to an email client, but does not have wide support. For example, it appears to be ignored when sharing to WhatsApp |
| dialog_title | A string to display as the title of the native "share" dialog that pops up | This is stated in [Capacitor's docs](https://capacitorjs.com/docs/apis/share#shareoptions) to be supported only on Android. However from my testing, it appears that recent versions of Android do not support it. I haven't been able to confirm support on older Android versions |

## Testing

Testing this functionality proved a bit difficult, considering the different moving parts and combinations of platform and target app for sharing.

We use [Capacitor's Share API](https://capacitorjs.com/docs/apis/share) to interface with native share functionality on Android and iOS. As sharing files on web does not seem to be supported via Capacitor Share, I've used the [Web Share API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API) directly on web. Once the data has been shared to the device-dependent sharing feature, this has to then interface with the specific API of the targeted app. Different apps receiving the shared data will interpret it differently, and of particular interest to our use cases is WhatsApp. Whilst I couldn't find definitive up-to-date information, there was [some suggestion](https://stackoverflow.com/a/55688479) that WhatsApp could not receive both text and image on iOS. Fortunately, it seems that information is inaccurate or outdated, as I was able to share text and image from the debug app to WhatsApp on iPhone.

In order to test the full sharing pipeline (specifically including sharing to WhatsApp), I needed to test on real devices with WhatsApp installed, as this isn't possible through appetize. From my testing, the functionality worked as expected (see screenshots below).

For additional functional testing of this PR, I think using appetize is sufficient. We should test that the share feature works as expected when sharing to _some_ app. 
- The Android emulator includes the default "Messages" app, which can be seen as analogous to WhatsApp.
- The iOS emulator does not include a messaging app. The only option is to share to the "Reminders" app, which prompts a warning that images cannot be shared without a premium account (showing at least that an image is attempting to be shared).

As long as the data is coming through as expected, we can consider the functionality to be working. Indeed, we have no control over how WhatsApp consumes the data it receives from the native share feature, so it's sufficient to confirm that we are handing over the data correctly. As stated, I was able to confirm on both Android and iOS that the feature worked as expected.

### Testing links
- [iOS appetize](https://appetize.io/app/ios/international.idems.debug-app?device=iphone14pro&osVersion=16.2)
- [android appetize](https://appetize.io/app/android/international.idems.debug_app?device=pixel7&osVersion=13.0)
- [PR Web Preview link](https://plh-teens-app1--pr2768-feat-share-multiple-5paiigcw.web.app/)
- [Link to join internal testers of the Debug App on Google Play](https://play.google.com/apps/internaltest/4700768283442703619) (must be logged in with an account that is registered in the list of IDEMS Beta Testers group)
- I can invite you to test the iOS app via TestFlight on request

## Git Issues

Closes #2764
Relevant content issue, which can be closed once the authoring has been updated accordingly: https://github.com/ParentingForLifelongHealth/plh-facilitator-app-mx-content/issues/169

## Screenshots/Videos

[feature_share](https://docs.google.com/spreadsheets/d/1VJFxbuUtKLiw9HucShQalalKcVpOKjryOzfKQlBHE_o/edit?pli=1&gid=960980632#gid=960980632) template:

Screenshot 2025-02-04 at 13 16 43

### Appetize Android:

https://github.com/user-attachments/assets/c38182cd-130f-4726-8711-3f5cb191e0c3

### Appetize iOS:

https://github.com/user-attachments/assets/80bca92c-6add-4b39-8634-0a5934c3fde1

### Native iOS

I haven't made a screen recording for privacy reasons as I was testing on someone else's phone, but this image shows the text and image successfully shared to whatsapp when passed to a single `share` action via the `text` and `file` params:

iphone whatsapp

## Native Android

Annoyingly, the image itself hasn't come through to the app build that I made using the [github action on the debug content repo](https://github.com/IDEMSInternational/app-debug-content/actions/runs/13135021750). There is an issue with assets being included properly, as is evident on the debug web preview, likely an issue with git lfs. However, the image and text do appear to be attempted to be shared correctly with WhatsApp, and I can confirm that they come through when sharing to an email client (Outlook)

android whatsapp

This was the result of sharing to Outlook via the first button on the tempalte, with all possible params populated:
android outlook


src/app/shared/services/share/share.service.ts
```diff
@@ -7,13 +7,20 @@ import { FileManagerService } from "../file-manager/file-manager.service";
...
+  title?: string;
+  dialog_title?: string;
+}
 
 @Injectable({
   providedIn: "root",
 })
 export class ShareService extends SyncServiceBase {
+  // Temporary local storage path on native devices for file being shared
+  localFilepath: string;
```
chrismclarke, line 23: nit(non-blocking): if this is not required by external components it should be marked as private. Also better to use `/** */` style comments for intellisense

src/app/shared/services/share/share.service.ts
```diff
@@ -30,90 +37,140 @@ export class ShareService extends SyncServiceBase {
...
-        if (!(actionId in childActions)) {
-          return await this.share({ text: args[0] });
+      share: async (action) => {
+        const { args, params } = action as { args: string[]; params: IShareActionParams };
+
+        if (params) {
+          await this.handleShare(params);
+        } else if (args) {
+          console.warn("[SHARE] Deprecated action syntax. Use `share | data_type: data` instead.");
+          await this.handleLegacyShare(args);
```
chrismclarke, line 47: nit(blocking): I think it would be better to map the legacy args to new params instead of triggering legacy methods if that's possible. This can be done outside of the if statement and then just have a check for params after

src/app/shared/services/share/share.service.ts
```diff
@@ -30,90 +37,140 @@ export class ShareService extends SyncServiceBase {
...
-        console.log("[SHARE] Content shared to", activityType);
-      } catch (error) {
-        this.handleShareError(error);
-      }
-    } else console.error(SHARE_NOT_SUPPORTED_ON_PLATFORM_ERROR_MESSAGE);
+  private async handleShare(options: IShareActionParams) {
+    // Rename `dialog_title` to `dialogTitle`
+    let parsedOptions = {
+      dialogTitle: options.dialog_title,
+      ...(({ dialog_title, ...rest }) => rest)(options),
```
chrismclarke, line 59: nit(blocking): there's a log happening in this inline statement. Could you extract first for better clarity? E.g.
```ts
const {dialog_title,...shareOptions} = options
const parsedOptions = {dialogTitle: dialog_title,...shareOptions}
```

src/app/shared/services/share/share.service.ts
```diff
@@ -30,90 +37,140 @@ export class ShareService extends SyncServiceBase {
...
+      const fileData = await this.getFileData(parsedOptions.file);
+      delete parsedOptions.file;
+      parsedOptions = { ...parsedOptions, ...fileData };
+    }
+    await this.share(parsedOptions);
   }
 
-  async shareFile(relativePath: string) {
-    let localFilepath: string;
+  private async share(options: ShareOptions | ShareData) {
```
chrismclarke, line 70: nit(non-blocking): given that the share method is called from `handleShare` shouldn't the type just be `IShareActionParams` (with `dialog_title` replaced by `dialogTitle`)? It might be a bit tidier to include this definition separately, e.g.
```ts
type IShareParams = Omit<IShareActionParams,'dialog_title'> & {dialogTitle?:string}
```

chrismclarke (changes requested):
I really like the move to params, think it gives the action a lot more flexibility. See a few minor nit-picks inline, but overall I think looking good.

jfmcquade:
Thanks @chrismclarke, good suggestions for code tidying.

ChrisMarsh82 (approved):

chrismclarke (approved):
Thanks for making the updates @jfmcquade , looks good to me!

esmeetewinkel (commented):
Looking good, appreciate the testing limitations. I tested where I could, working fine generally.

I did find when using the PR web preview on my windows computer to share (message + image) to whatsapp, that only the image was shared. Text does get shared in a message that has only text. Not sure if that is what you expected? 

[Screenity video - Feb 6, 2025.webm](https://github.com/user-attachments/assets/2de087db-b105-4d8e-8ad9-c78f7539c054)

jfmcquade:
> I did find when using the PR web preview on my windows computer to share (message + image) to whatsapp, that only the image was shared. Text does get shared in a message that has only text. Not sure if that is what you expected?

Yes, @ChrisMarsh82 also found some behaviour like this. It's not the behaviour I expected, but I don't think we have control over it. My understanding is that this lies with the app's integration into the platform's share API. In this case, the Web Share API's hand-off to WhatsApp Desktop on Windows. We know that both text and file are being sent by the app to the Web Share API, since variously the image and the text are shared when sharing to different target apps, so I think it's a case of WhatsApp Desktop never being able to receive both text and image from the API, and prioritising the image.

I'm aware that saying something isn't possible is a strong claim – @chrismclarke does this seem sensible? 

The fact that our current use case is covered, namely sharing text and image to whatsapp on Android and iOS means that I'm happy for this to be merged as-is. One thing we could consider would be providing a warning, either to users or authors, when sharing to different platforms to try and clarify what is possible (although it is hard to get clarity without exhaustively testing all combinations of platform, sharing data, and target app).

PR #2771, opened 2025-02-03
Fix: ci screenshots

chrismclarke:
PR Checklist

## Description

The visual screenshots test has been failing since last December, due to incompatibilities introduced with automated updates to github actions runner and outdated code dependencies

This PR aims to fix both the code to generate screenshots (applied on merge to master) and compare function (triggered by pr's labelled with `test - visual`)

**Potential Follow-ups**
- [ ] Create issue to automate generate to trigger ahead of test-visual compare action

- [ ] Consider adding better support to define generation parameters for deployments (e.g. setting specific indexeddb fields, which are currently hardcoded for debug deployment)

- [ ] Further improvements to outputs (e.g. include diff size in filenames and order from most to least changed)

## Author Notes
Once merged authors should start being able to use the `test - visual` action again to compare against master branc

Previously base screenshots would be generated on every merge to master branch. This isn't ideal as it ends up generating lots of screenshots that are never used for comparison, and also fails to take into account authoring changes which may be the cause of differences (author changes wouldn't trigger new screenshot generation)

As such I've disabled the automated screenshot generator, in favour of manually triggering as required. So that means running visual tests is a semi-manual process involving:
1. Generate the base screenshots by triggering `Run Workflow` from the [Test-Visual Generate](https://github.com/IDEMSInternational/open-app-builder/actions/workflows/test-visual-generate.yml) action page

2. Wait for action run to complete

3. Assign the `test - visual` label on the PR to generate and compare pr branch screenshots

Note that this is a temporary workaround which should hopefully be replaced in the future with automated generation as part of the `test - visual` tagging process

It might also be worth reviewing the the base screenshots generated (323 screenshots for the debug templates), as many are blank/limited (might be worth either tidying some up or including better filters for what to include or not). This isn't critical by any means, more to speed up the overall processes a bit (probably takes 1-2s per screenshot) and make it easier to compare final outputs (fewer false-positives)

## Review Notes
See passing [Test-Visual Generate](https://github.com/IDEMSInternational/open-app-builder/actions/runs/13124368153/job/36617638367?pr=2771) action run (triggered manually).

Example passing [Test-Visual Compare](https://github.com/IDEMSInternational/open-app-builder/actions/runs/13124843690/job/36619081578?pr=2771) action run (triggered by PR)

## Dev Notes
I decided to try and update all dependencies to latest versions, however that immediately caused some issues with commonJS/esm module interop (the codebase was set as commonJS but a number of the dependencies have been updated to use ESM). As such the following additional steps were required:

- Replace `ts-node` as the main code runner for [TSX]() which has broadly better interoperability between esm and commonJS
- Update all code to be ESM by default
- Drop axios dependency and replace with native fetch method for HTTP requests

Additionally minor updates were made to both bypass the new language select screen (which otherwise blocks underlying templates) and to bypass `Leave site` beforeunload alert when trying to close each tab 

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/.github/workflows/test-visual-generate.yml
+++ b/.github/workflows/test-visual-generate.yml
@@ -2,10 +2,6 @@
 name: Test-Visual Generate Screenshots
 on:
   workflow_dispatch:
-  push:
-    branches:
-      - master
-      - fix/ci-screenshots
 concurrency:
   group: test-visual-generate-${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/test-visual-generate.yml
+++ b/.github/workflows/test-visual-generate.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - fix/ci-screenshots
 concurrency:
   group: test-visual-generate-${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/test-visual.yml
+++ b/.github/workflows/test-visual.yml
@@ -132,10 +132,10 @@ jobs:
             ${{ env.BIGGEST_DIFFS }}
 
             **Download Link**
-            https://nightly.link/IDEMSInternational/parenting-app-ui/actions/runs/${{github.run_id}}
+            https://nightly.link/IDEMSInternational/open-app-builder/actions/runs/${{github.run_id}}
 
             **Run Details**
-            https://github.com/IDEMSInternational/parenting-app-ui/actions/runs/${{github.run_id}}
+            https://github.com/IDEMSInternational/open-app-builder/actions/runs/${{github.run_id}}
 
       # Alt implementation to DL artifact using action instead of download script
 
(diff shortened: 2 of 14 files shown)


esmeetewinkel (approved):
Thanks for fixing this. I was able to download the assets from the action summary and works  as expected.

However, the direct download link that appears as a comment on the PR doesn't seem to work
image

> It might also be worth reviewing the the base screenshots generated (323 screenshots for the debug templates), as many are blank/limited (might be worth either tidying some up or including better filters for what to include or not). 

I can look into this at some point (but don't think it should hold up this PR). Does this currently use all templates from the `debug` deployment or a subset? If the latter - where is it specified?

chrismclarke:
@jfmcquade 
I'm going to go ahead and merge here to test the generated screenshots on main. None of the code changes should have any knock-on on main app codebase, but feel free to add any comments post-merge and I can address in a follow-up

esmeetewinkel:
Follow-up issue: https://github.com/IDEMSInternational/open-app-builder/issues/2778

PR #2772, opened 2025-02-04
feat: data items local

### LINKED ISSUE #2766: [FEATURE] Intermediate variables in data items

chrismclarke:
PR Checklist

**TODO**
- [x] Merge after #2775
- [x] Merge after #2780
- [x] Add templateCalcService specs
- [x] Calc service changes rebased to #2793, merge after
- [x] Screenshot tests (trigger on master and compare PR)
- [x] Check whether changes fixes bug noted in `example_auth` #2795

## Description
Add support for creating local variables within a data_items loop

## Author Notes
Now any `set_variable` (or unnamed) row types within a data_items loop will create a temporary variable, available only within that loop. This can be used to generate more complex expressions to include within templated child rows

As a side-effect, more complex data can be used within data_items loops via intermediate variables, e.g. [example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705) can now show app theme via nested `@item.contact_fields["rp-contact-field._app_theme"]`

## Review Notes
A new [comp_data_items_dev](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=1795216644#gid=1795216644) sheet has been created for the purpose of testing. Once in a good state could ideally be merged into main comp_data_items sheet (or renamed and linked from the comp_data_items sheet)

Should also check whether addresses issue in original issue

There is some scope for knock-on to data_items sheets more generally, so would be good to cross-check any advanced authoring examples. There are fairly comprehensive examples in `comp_data_items` and `feat_data_actions` - I've gone through each of these and they appear to be working as expected, but would be good to get a second pass. 

I've also triggered a new set of screenshots generated on `master` to compare with this PR and the only thing that stands out to me as related to `data_items` is `debug_data_items_condition` sheet which appears broken on master and maybe fixed a bit on this branch (didn't look at the sheet itself as assuming it's an old sheet)

## Dev Notes
This PR further decouples the `data-items` system from the rest of templating. It no longer processes item loops using the same `ItemProcessor` (but does use the processor to pipe operators like filter, sort etc.). This is because local context requires data-items to process in chunks by item (so that local context can be reset between evaluation), but itemProcessor outputs a single array of all processed items. So new methods have been added to the `data_items` service and utils to handle row generation and processing in a more linear way.

Handling support for `@calc(...)` statements within the data-items local variables was complicated, as up till now the new `AppDataEvaluator` hasn't had to deal with them. As a short term solution these expressions are being explicitly checked for and handled separately, but in the future it might be useful to consider how this should interoperate with the rest of the templated data system. Ideally methods to extract lists of local variables used should also be able to extract any calc references along with list of global functions used within the expression. All of this should ideally happen at the parser level, however that is beyond the scope of this PR

## Git Issues

Closes #2766

## Screenshots/Videos
[comp_data_items_dev](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=1795216644#gid=1795216644)

image

image

**Example - data_item loop now able to show deep nested contact field**
image


jfmcquade (approved):
Looks good, pending functional testing. The data-items logic always seems pretty complicated, but good to have the tests for clarity and safety

esmeetewinkel (commented):
Thanks, I can confirm setting local variables with computation as described in https://github.com/IDEMSInternational/open-app-builder/issues/2766 now works, as well as the examples in the debug sheets.

One inconvenient knock-on: 
(In the same use case as described in https://github.com/IDEMSInternational/open-app-builder/issues/2766) I have a list of module task cards / plh_module_list_item components, where a module is "unlocked" exactly when the previous module is completed. I had used a loop as below to achieve this ([sheet](https://docs.google.com/spreadsheets/d/1nvyc-d_noz-W7guO6x_33c6cCXMXpexGt5HYf5T5J50/edit?gid=14877797#gid=14877797)), but this no longer works on the PR branch. I'm guessing my authoring pattern is kind of hacky anyway, so appreciate it might not make sense to fix. Can you think of another way I might achieve this?
image

chrismclarke:
> One inconvenient knock-on: (In the same use case as described in #2766) I have a list of module task cards / plh_module_list_item components, where a module is "unlocked" exactly when the previous module is completed. I had used a loop as below to achieve this ([sheet](https://docs.google.com/spreadsheets/d/1nvyc-d_noz-W7guO6x_33c6cCXMXpexGt5HYf5T5J50/edit?gid=14877797#gid=14877797)), but this no longer works on the PR branch (instead, it now shows a module as unlocked iff it is completed itself). I'm guessing my authoring pattern is kind of hacky anyway, so appreciate it might not make sense to fix. Can you think of another way I might achieve this? 

Thanks for checking this. I think the case you described is something that ideally we would want a `data_query` row type that more conveniently maps the results of a query to specific local/field variables, however until then there is a workaround using the fact that `data_items` loops can emit a `data_changed` action whenever their contents updates, and any `data_items` loop that does not have inner template rows will emit the full item

So it's possible to author something along the lines of 

| type             | name              | value              | action_list                                                                                                                   | parameter_list                             |
| ---------------- | ----------------- | ------------------ | ----------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------ |
|                  | current_module_id |                    |                                                                                                                               |                                            |
|                  | next_module_id    |                    |                                                                                                                               |                                            |
| begin_data_items |                   | @data.module_tasks | data_changed \| set_local : current_module_id : @items[0]?.id;<br>data_changed \| set_local : next_module_id : @items[1]?.id; | <br>filter: !@item.completed;<br>limit: 2; |
| end_data_items   |                   |                    |                                                                                                                               |

This uses a local variable for current/next module ids, and a data_items loop to set those values from the first 2 items that are not completed. If you only care about marking the current module as unlocked then you can remove the `next_module` references and just limit to the first row. 

I've added a debug sheet to demo here:

[module_path_debug](https://docs.google.com/spreadsheets/d/1nvyc-d_noz-W7guO6x_33c6cCXMXpexGt5HYf5T5J50/edit?gid=1386183901#gid=1386183901)

Example screen recording - see the variables update whenever a module is marked as completed

https://github.com/user-attachments/assets/dbfa2a04-f991-44e4-8cb7-3eb6bb9edc06

@jfmcquade - while looking into this issue I did notice a bug where the `data_actions` could be repeatedly triggered (the action would update the parent template that would force re-render the rows and re-trigger the actions). I've fixed with [e0e25f7](https://github.com/IDEMSInternational/open-app-builder/pull/2772/commits/e0e25f7f30d6471952adc0e68d5958cd7877fa31) so marking for re-review

chrismclarke:
@jfmcquade - while looking into the issue in the previous message I noticed a bug where the data_actions could be repeatedly triggered (the action would update the parent template that would force re-render the rows and re-trigger the actions). I've fixed with [e0e25f7](https://github.com/IDEMSInternational/open-app-builder/pull/2772/commits/e0e25f7f30d6471952adc0e68d5958cd7877fa31) so marking for re-review

esmeetewinkel:
@chrismclarke thanks for the example, I wasn't aware of the syntax `@items[0]?.id`. 

This _almost_ works for my use case, only that there seems to be some sort of order-of-rendering issue: 
- in the `module_path_debug` template, I can see a little flicker before the "locked" status (which depends on a variable set in data actions) displays correctly. 

[Screenity video - Feb 20, 2025 (1).webm](https://github.com/user-attachments/assets/fe0730dd-e6d1-4e6f-964a-ee37a4b758a1)

- (I think this is equivalent) in the `module_path` template the "locked" status only shows correctly when navigating back to the page, or by toggling debug mode. on first render or on refresh the "locked" status of the current module is incorrect. 

[Screenity video - Feb 20, 2025.webm](https://github.com/user-attachments/assets/a77e1579-6044-4619-804a-85eff0680f94)

chrismclarke:
> @chrismclarke thanks for the example, I wasn't aware of the syntax `@items[0]?.id`.
> 
> This _almost_ works for my use case, only that there seems to be some sort of order-of-rendering issue:
> 
> * in the `module_path_debug` template, I can see a little flicker before the "locked" status (which depends on a variable set in data actions) displays correctly.

> * (I think this is equivalent) in the `module_path` template the "locked" status only shows correctly when navigating back to the page, or by toggling debug mode. on first render or on refresh the "locked" status of the current module is incorrect.

That makes sense, as the data-items will only correctly update after being rendered for the first time, prompting a quick re-render after first evaluation. If moving to a `data_query` type it might be possible to more easily prevent rendering until data evaluated for the first time, although for now the simplest solution is to just use a condition to hide the content until the active module has been evaluated,e.g. 

| type             | name              | value              | action_list                                                    | parameter_list                             | condition                |
| ---------------- | ----------------- | ------------------ | -------------------------------------------------------------- | ------------------------------------------ | ------------------------ |
| begin_data_items |                   | @data.module_tasks |                                                                |                                            | @local.current_module_id |

HOWEVER - the one issue with doing this is that if all the modules are completed then there would be no current_module_id as evaluated as first non-completed, and no content would show.

So we need an additional to identify the case where all modules are completed, or at least to know the difference between current module being undefined due to all completed vs not yet evaluated. 

In this case I think it makes most sense to add a `completed_count`, which can evaluate from a second data_items loop and keep track of total number of completed modules. Then we can update the condition statement to show/hide content iff it has been computed, i.e. 

| type             | name            | value              | action_list                                                 | parameter_list           | condition                            |
| ---------------- | --------------- | ------------------ | ----------------------------------------------------------- | ------------------------ | ------------------------------------ |
|                  |                 |                    |                                                             |                          |                                      |
|                  | completed_count |                    |                                                             |                          |                                      |
| data_items       |                 | @data.module_tasks | data_changed \| set_local : completed_count : @items.length | filter: @item.completed; |                                      |
| begin_data_items |                 | @data.module_tasks |                                                             |                          | @local.completed_count !== undefined |

_Sidenote_ - as we're not including any inner items for the `data_items` loop, we don't actually need the begin/end statement. This might be a nicer way to implement item loops used for querying, as the syntax could later just be swapped out for `data_query` or whatever we settle on 
---
I've updated the demo to remove the unused `next_module_id` snippets, and the counter variable, and include a display condition. I've also added an intermediate `unlocked` variable (as that's the main purpose of this PR :p ), which simplifies some of the other conditions a bit

So hopefully that should be closer to what you need, but let me know if anything doesn't fully make sense or isn't working as you want it to.

jfmcquade (approved):
The updates in https://github.com/IDEMSInternational/open-app-builder/pull/2772/commits/e0e25f7f30d6471952adc0e68d5958cd7877fa31 make sense, and good to see the final authoring patterns in [module_path_debug](https://docs.google.com/spreadsheets/d/1nvyc-d_noz-W7guO6x_33c6cCXMXpexGt5HYf5T5J50/edit?gid=1386183901#gid=1386183901) looking pretty neat.

esmeetewinkel (approved):
Thanks @chrismclarke, that works! 

> as we're not including any inner items for the data_items loop, we don't actually need the begin/end statement. This might be a nicer way to implement item loops used for querying, as the syntax could later just be swapped out for data_query or whatever we settle on

I hadn't realised this was possible (so thanks for pointing out), makes sense.

> I've also added an intermediate unlocked variable (as that's the main purpose of this PR :p ), which simplifies some of the other conditions a bit. 

I also had that at first, but the non-debug template for some reason doesn't work when the condition is put in an intermediate variable. No clue why but no strict need to get to the bottom of right now... 

> Overall the authoring pattern feels pretty nice to me, so might be good to copy across some to the comp_data_items example

Done, see [example_data_actions](https://docs.google.com/spreadsheets/d/1kpRKj95Js24IYsYwT-3lbfFO-XAOuiDTycGZih0DWOU/edit?gid=345767231#gid=345767231)

PR #2773, opened 2025-02-04
style: new display group variant for MX button use case

jfmcquade:
PR Checklist

## Description

Adds a new variant to the display group to handle the use case referenced [here](https://github.com/ParentingForLifelongHealth/plh-facilitator-app-mx-content/issues/140#issuecomment-2616147702).

This variant, `box_secondary_alt`, applies a background colour, border and corner radius to the group, in line with the existing `box_*` variants. Additional styling for the use case should be handled through other means, see example below.

Corresponding debug content PR: https://github.com/IDEMSInternational/app-debug-content/pull/142

## Notes

I wasn't sure how best to handle this use case. While defining the styling itself is relatively trivial, there is a decision about how much styling to include in a variant vs relying on additional authored styling, to achieve a balance between reusability and flexibility. I've opted in favour of making the styling very flexible to authors: any content could be authored inside the styled display group in the usual way. Another approach to handling this request, at the other end of the scale, would be to author a new variant of `button` component which took an additional parameter to define the top text (e.g. `title`), and applied all the necessary styling. This would be easier to reuse in other contexts, but more difficult to adapt if, for example, we wanted the same visual component to contain an image and no top title, for example.

Adding a `box_secondary_alt` does not represent a step towards our planned theme-system improvements (#1468), but is a welcome departure from cases such as the display group "styles" below the new one in the [demo sheet](https://docs.google.com/spreadsheets/d/1IlM2fbVYzVdwPBa9UEaKZhGGlLFhRHCo--T6uC-EQiE/edit?gid=7532536#gid=7532536): e.g. `tool_1`, `tool_2` etc., which have deployment specific names and inflexible styling.

## Git Issues

Referenced in https://github.com/ParentingForLifelongHealth/plh-facilitator-app-mx-content/issues/140

## Screenshots/Videos

Design:

406990546-fa347aae-0305-46bb-9031-760cadd75dbe

[comp_display_group](https://docs.google.com/spreadsheets/d/1IlM2fbVYzVdwPBa9UEaKZhGGlLFhRHCo--T6uC-EQiE/edit?gid=7532536#gid=7532536):

Screenshot 2025-02-05 at 11 50 46

`default` theme:

Screenshot 2025-02-05 at 11 52 28

`plh_facilitator_mx` theme:

Screenshot 2025-02-05 at 11 52 50

### DIFF
--- a/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
+++ b/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
@@ -60,7 +60,6 @@
   &[data-variant~="box_primary"],
   &[data-variant~="box_secondary"],
   &[data-variant~="box_secondary_alt"],
-  &[data-variant~="solid_secondary"],
   &[data-variant~="box_white"] {
     margin-top: var(--regular-margin);
     padding: var(--regular-padding);
--- a/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
+++ b/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
@@ -89,3 +89,13 @@
     --border-color: var(--border-color-default, var(--ion-color-primary));
   }
 }
+
+.display-group-wrapper {
+  &[data-variant~="solid_secondary"] {
+    padding: var(--regular-padding);
+    background-color: var(--ion-color-secondary);
+    border-radius: var(--ion-border-radius-secondary);
+    border: var(--border-standard);
+    border-color: var(--ion-color-secondary-200);
+  }
+}
--- a/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
+++ b/src/app/shared/components/template/components/layout/display-group/display-group.component.scss
@@ -59,6 +59,8 @@
   &[data-variant~="box_gray"],
   &[data-variant~="box_primary"],
   &[data-variant~="box_secondary"],
+  &[data-variant~="box_secondary_alt"],
+  &[data-variant~="solid_secondary"],
   &[data-variant~="box_white"] {
     margin-top: var(--regular-margin);
     padding: var(--regular-padding);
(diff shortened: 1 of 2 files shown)


ChrisMarsh82 (approved):
Looks good for MX use case

esmeetewinkel (approved):

src/app/shared/components/template/components/layout/display-group/display-group.component.scss
```diff
@@ -89,3 +89,13 @@
     --border-color: var(--border-color-default, var(--ion-color-primary));
   }
 }
+
+.display-group-wrapper {
+  &[data-variant~="solid_secondary"] {
+    padding: var(--regular-padding);
```
chrismclarke, line 95: nit(blocking): I'm not sure it makes for just one variant to include padding by default whilst others do not (given that variants should also be able to stack/combine). I think likely better to either include as a deployment override, or add an additional variant that applies padding more consistently

src/app/shared/components/template/components/layout/display-group/display-group.component.scss
```diff
@@ -89,3 +89,13 @@
...
+    background-color: var(--ion-color-secondary);
```
chrismclarke, line 96: nit(non-blocking): when using `box_secondary` the background is `var(--ion-color-secondary-200)` but when using `solid_secondary` we have `var(--ion-color-secondary)`. Even if foregoing the possible confusion of what is a box vs what is solid, there's not really any indication of why one variant uses lighter colours and other darker (unless that is the distinction?)

I think it might be better to try and define either within the existing box semantics, e.g. `box_secondary_alt` if we're only changing a few of the colour weights, or again just overriding at the deployment level

chrismclarke (changes requested):

src/app/shared/components/template/components/layout/display-group/display-group.component.scss
jfmcquade: Good point. I thought I needed to handle separately to avoid the MX theme level overrides (the `!important` flags, for example, can make it difficult to go back and change styling at the component level). But I've opted for `box_secondary_alt` within the existing box semantics, and it works well with more minimal new style rules. Also addresses your other inline comment. 

See [new commits](https://github.com/IDEMSInternational/open-app-builder/pull/2773/files/5f653b14e36b45da7b2cd9147ebea304abd2ee9c..1cfd2470288a05ce04f8d60493f35cebf8b3ed30).

jfmcquade:
Thanks @chrismclarke, I've acted on your suggestions.

@ChrisMarsh82 Note that the variant is now called `box_secondary_alt` – I've updated the PR description and [comp_display_group](https://docs.google.com/spreadsheets/d/1IlM2fbVYzVdwPBa9UEaKZhGGlLFhRHCo--T6uC-EQiE/edit?gid=7532536#gid=7532536) template accordingly.

chrismclarke (approved):
Thanks for the changes @jfmcquade , the balance of display-group styles to deployment styles feels a lot better here, looks good to me

PR #2775, opened 2025-02-04
Chore: ng test ci and mock updates

chrismclarke:
PR Checklist

## Description

- Add `ng test` to ci test workflow to ensure all ng tests are passing before merge
- Change mock db to use dexie with `fake-indexeddb` instead of stubbing methods
- Alter various mock test suites to extend base instead of implementing partial

## Review Notes
Can see code changes in files ending `.ts` (but not `.spec.ts`) - I've never found a good way to filter these out but there's not too many changed files to quickly look through and check

The spec test code updates are non-breaking (full set of spec tests included in PR testing)

## Dev Notes
I've slowly been migrating most mock services to extend original services instead of partially implementing methods as I feel like in the long term this will be less prone to breakage and better at catching knock-on issues when core methods changed on underlying services (tests will break instead of just potentially showing type errors depending on how implemented)

This shouldn't result in many core code changes at all, just a few minor bits of refactoring in places to make it easier to override some service methods/variables when called from mocks

The CI tests have been separated into separate steps and `ng test` included so that the entire set of spec tests will be evaluated on any PR. If we wanted to be more optimal it would be possible to setup the src, shared, and scripts workspace tests to only [execute when files changed](https://www.meziantou.net/executing-github-actions-jobs-or-steps-only-when-specific-files-change.htm) within corresponding folders, however this isn't a high priority for now (could save a few minutes on each PR test run depending on what files change)

## Git Issues

Closes #

## Screenshots/Videos

**Example Passing Action - with `Test - src` step**
https://github.com/IDEMSInternational/open-app-builder/actions/runs/13144318569/job/36678739564?pr=2775

image

### DIFF
--- a/src/app/shared/components/template/components/data-items/data-items.service.spec.ts
+++ b/src/app/shared/components/template/components/data-items/data-items.service.spec.ts
@@ -9,6 +9,8 @@ import { firstValueFrom } from "rxjs";
 import { Injector } from "@angular/core";
 import { TemplateVariablesService } from "../../services/template-variables.service";
 import { TemplateTranslateService } from "../../services/template-translate.service";
+import { TemplateCalcService } from "../../services/template-calc.service";
+import { MockTemplateCalcService } from "../../services/template-calc.service.mock.spec";
 
 /***************************************************************************************
  * Test Setup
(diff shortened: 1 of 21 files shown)


src/app/shared/components/template/components/data-items/data-items.service.spec.ts
```diff
@@ -150,6 +153,8 @@ describe("DataItemsService", () => {
     console.log({ evaluated });
     expect(evaluated.args).toEqual(["my_local_var", 3]);
   });
+
+  it("...", async () => {});
```
jfmcquade, line 157: can this be removed?

jfmcquade (approved):
LGTM – really nice to finally have the spec tests integrated into the CI workflow

PR #2780, opened 2025-02-05
hotfix: data-items local variables

### LINKED ISSUE #2777: [BUG] ERROR: @local.filter not found
**Describe the bug**
The filtering radio buttons on the [activity_library](https://docs.google.com/spreadsheets/d/1V8iYNvVVc0whiZ3p--ulL_93DiOEH_SKwwwSqLX6OE4/edit?gid=0#gid=0) template in the `plh_kids_kw` deployment no longer work - they now throw errors `@local.filter not found`.

Going back through merged PRs, I found that this bug probably arose on PR #2767 (since the app is not broken on 02f04d9 and broken on eb720e9)

**App version**
v0.17.1

**Deployment Sheet(s)**
[activity_library](https://docs.google.com/spreadsheets/d/1V8iYNvVVc0whiZ3p--ulL_93DiOEH_SKwwwSqLX6OE4/edit?gid=0#gid=0) 
Image

**Screenshots and videos**
Current behaviour:
Image

Desired behaviour (e.g. on https://plh-kids-kw-uat1.web.app/template/activity_library):
Image

chrismclarke:
PR Checklist

## Description
Fix regression introduced by #2767 where data_items no longer passed outer local variables to inner items

## Review Notes
Minimal reproduction of issue available at [comp_data_items_debug_2](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=1510685976#gid=1510685976)

Should be broken on master but fixed in branch. Should also confirm original issue resolved

## Dev Notes
I should have been a bit more careful when adding the additional variable that is meant to stay in sync with the templateRowMap, not realising that the rowMap was updated in multiple places not just the main processor. As part of this fix I checked the codebase for anywhere the `templateRowMap` was updated and tried to ensure the related `templateRowMapValues` was also.

A better solution might have been to only allow the `templateRowMap` to be set/updated from a method (not directly), and just using that method to ensure both values updated, however that would have been a larger refactor so not included at this moment. Alternatively the values could be created through some sort of getter/function, however it would require a level of memoizing or using computed signal in service to prevent repeated recalculation as it is accessed by every row processor. 

I tried adding tests to capture the behaviour, however the code change is within the `hackProcessRows` method which is stubbed for testing so not easy to do. However the behaviour is captured within [comp_data_items](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892) Example 3 (accessing completed_counter inside of items), so now that visual tests are back working the component sheet should pick up any future regression if visually tested. 

I have also added some minor test updates to the `template-action.service.spec` that check the general case of the templateRowMapValues being updated (wouldn't have caught this specific issue but good to have for future ref)

## Git Issues

Closes #2777 

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/components/data-items/data-items.service.ts
+++ b/src/app/shared/components/template/components/data-items/data-items.service.ts
@@ -165,6 +165,11 @@ export class DataItemsService {
     } as any);
     // HACK - still want to be able to use localContext from parent rows so copy to child processor
     processor.templateRowMap = JSON.parse(JSON.stringify(templateRowMap));
+    const templateRowMapValues = Object.fromEntries(
+      Object.entries(templateRowMap).map(([key, { value }]) => [key, value])
+    );
+    processor.templateRowMapValues = templateRowMapValues;
+
     await processor.processContainerTemplateRows();
     return processor.renderedRows();
   }
--- a/src/app/shared/components/template/services/instance/template-action.service.spec.ts
+++ b/src/app/shared/components/template/services/instance/template-action.service.spec.ts
@@ -40,6 +40,10 @@ class MockTemplateRowService implements Partial<TemplateRowService> {
 
 class MockContainer implements Partial<TemplateContainerComponent> {
   templateRowService = new MockTemplateRowService() as any as TemplateRowService;
+
+  get templateRowMap() {
+    return this.templateRowService.templateRowMap;
+  }
 }
 
 /**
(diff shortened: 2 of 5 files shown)


esmeetewinkel (approved):
Functional test passed on highlighted `debug` deployment sheet, as well as on `plh_kids_kw` deployment where the issue was first discovered.

jfmcquade (approved):
Looks good, the changes and explanation make sense

PR #2781, opened 2025-02-06
feat(map): support multiple layers sharing the same scale

jfmcquade:
PR Checklist

## Description

Adds a new layer property, `scale_id`, used to identify layers which should share the same scale.

If no scale ID is provided, the layer functions as before, and any associated scale is rendered for that layer when it is visible. If multiple layers share a `scale_id`, then only one scale associated to these layers will ever be shown, and filters applied using that scale slider will apply to all such layers with that scale ID. Authors should ensure that each layer with a given scale ID has the same configuration for its scale (e.g. `scale_max`, `scale_min`, `scale_increment`, `scale_title` etc.)

## Dev Notes

The implementation is quite hacky. Currently, different scales are shown for different layers with the same `scale_id` (up to a maximum of 1) depending on which layers are visible, but these scales are assumed to be identical and kept in sync. 

Instead, it would be better to handle the scale configuration at a different level, only specifying a single scale that relates to multiple layers. This could be at the `layer_group` level, however in that case layers that are in different layer groups could not be identified as belonging to the same slider group. Alternatively, a single slider could be defined for each `property_to_plot` referenced by layers, although this would require such properties to be uniquely named across different data sources where necessary.

## Git Issues

Closes #

## Screenshots/Videos

Demo of [comp_map](https://docs.google.com/spreadsheets/d/1KgLTrFhqZdGhcSk6Zt0KQoy2mYf7NCqhCJz_zS0KEYw/edit?gid=569531329#gid=569531329).

See updates to [comp_map_layer_group_2](https://docs.google.com/spreadsheets/d/1KgLTrFhqZdGhcSk6Zt0KQoy2mYf7NCqhCJz_zS0KEYw/edit?gid=1396735088#gid=1396735088)

Screenshot 2025-02-07 at 12 06 03

https://github.com/user-attachments/assets/9504b54e-f737-452e-b4f3-636f7557dc38

### DIFF
--- a/src/app/shared/components/template/components/map/map.component.html
+++ b/src/app/shared/components/template/components/map/map.component.html
@@ -12,11 +12,11 @@
               <div class="scale-container">
                 <ngx-slider
                   [class.disabled]="!mapLayer.get('scaleSlider')"
-                  [value]="mapLayer.get('scaleSlider') ? mapLayer.get('scaleMin') : null"
-                  [highValue]="mapLayer.get('scaleSlider') ? mapLayer.get('scaleMax') : null"
+                  [(value)]="sliderValues[mapLayer.get('scaleId') || mapLayer.get('id')].min"
+                  [(highValue)]="sliderValues[mapLayer.get('scaleId') || mapLayer.get('id')].max"
                   [options]="getSliderOptions(mapLayer)"
                   (userChangeEnd)="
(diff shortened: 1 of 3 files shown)


Ateamate (commented):
This is currently working well. For the infrastructure data, the 'roads data' takes some time to respond to the filter applied but eventually works. This is probably due to how heavy the 'roads data' is; otherwise, it looks good!

PR #2782, opened 2025-02-07
Chore/git lfs opt out

chrismclarke:
PR Checklist

## Description
Attempt to reduce git-lfs usage by removing from main repo. This takes 2 main approaches:

1) Update main repo to replace current git-lfs assets with standard assets.

2) Update reusable workflows to avoid calling git-lfs on the main repo

## Review Notes
It would be good to track post-merge whether usage continues to increase or not. As of Feb 6th I can see 750GB usage. Billing should reset around Feb 12th, so that would be an estimated 30GB per day in use currently.

The line diff is huge although I think part of that is due to confusion when trying to compare git-lfs file references with the replaced binary assets. Even though most of the assets display the message 
```
Unable to render rich display
Invalid image source.
```
clicking on the 3-dot menu and `view file` does still show the correct asset as being available, so I again assume this is part of some git confusion

## Author Notes
It's assumed these changes should be non-breaking as content-repos provide their own assets, and any assets previously hardcoded into the main repo via git-lfs are still present as binary files.  

That being said, it would be good to check content repo actions still trigger as expected post-merge.

## Dev Notes
I followed the instructions at the link below to remove all active git-lfs references and replace with binary assets. I'm hoping that in doing so any 
https://www.ryangittings.co.uk/blog/removing-git-lfs-netlify/

This process does not rewrite any git history, although I'm still not entirely sure how easy it would be to restore the repo to a previous point in time with assets that had been managed by git-lfs. As such it might be advisable to take a backup of the current repo as synced locally (should be about 1GB in local `.git` folder including all checked out assets

I also separately searched the main repo for any references to git-lfs and examined to see what could be causing potential issue. A number of legacy/deprecated actions have been removed (for better clarity), and I've also tried to ensure that any triggered workflows cannot call git-lfs for the main content repo (but still allows checkout of child content repos using git-lfs)

## Git Issues

Closes #

## Screenshots/Videos


esmeetewinkel (approved):
Thanks! 

> it would be good to check content repo actions still trigger as expected post-merge.

I'll keep an eye on this. 

> it might be advisable to take a backup of the current repo as synced locally (should be about 1GB in local .git folder including all checked out assets

@ChrisMarsh82 Could you take a backup?

ChrisMarsh82 (approved):

chrismclarke:
I've taken a local backup with full git-lfs history and uploaded to weTransfer - available for download for the next 7 days in case anyone else wants a copy
https://we.tl/t-twyimNOn55

As of now bandwidth usage stands at 854GB (and steadily increasing), so it would be good to monitor changes over the next couple days
image

I'd likely recommend removing all the additional data packs for the next billing period, and if we see quotas starting to exceed free it might be worth contacting github to try and diagnose where the usage is coming from as there's also the chance it could be due to external factors out of our control

PR #2786, opened 2025-02-07
Feat/table component

### LINKED ISSUE #2519: [FEATURE] Data table component
**What?**
Add a component to make it easier to display tabular data

**Why?**
Quite often when working on debug sheets I want the ability to show all the data for a given data_list in a table. I expect this could be useful in other contexts also

**How?**
Will need to first identify the core functionality desirable for basic (and more advanced tables), and likely pick a 3rd party library that provides reasonable functionality out of the box. I expect by default we would want sort and filter functionality (all should have), we may also want to consider pagination/virtualisation options for larger data.

Whatever option is selected, there would be more discussion to be had as to how best to expose functionality to authors in a sensible way

**Other details**
I've previously worked a bit with ag-grid and datatables libraries. Both are quite feature-rich (although possibly a little over-engineered in places). Another alternative could be tanstack table

Examples:
https://tanstack.com/table/latest/docs/framework/angular/examples/filters
https://www.ag-grid.com/example-inventory/
https://datatables.net/examples/basic_init/multi_col_sort.html

https://npmtrends.com/@tanstack/table-core-vs-ag-grid-community-vs-datatables.net
image

jfmcquade:
PR Checklist

## Description

A basic table component using `@angular/material`'s [table](https://material.angular.io/components/table/overview) library.

Allows for sorting on columns.

Params
- `display_columns`: comma-separated list of column names to display. If not defined, all columns are shown by default 
- `show_search`: boolean to optionally display the search box at the top of the table. Default `false`.
- `sort_default`: string indicating name of column to sort on by default, and sort order (`asc`/`desc`)

## Notes

This is component is currently in use on the `conflict_forecast` deployment (which builds from the `deployment/conflict_forecast` branch). See web [riots_list_view_page](https://docs.google.com/spreadsheets/d/11_VvFXc4ycfEBndYHjUjJu8PIUrPxviP2nYPBitfdGg/edit?gid=216480713#gid=216480713) template, web preview [here](https://conflict-forecast.web.app/template/riots_list_view_page).

## Follow-ups

Future improvements tracked in #2848

## Git Issues

Closes #2519

## Screenshots/Videos

Demo of [comp_table](https://docs.google.com/spreadsheets/d/1utCg51oPxeT3rI1s74GmoNvdsS60aXOvCiVsKUTSMLY/edit?gid=603128459#gid=603128459):

https://github.com/user-attachments/assets/65be3ec6-f892-4d05-81cd-58e8b87fd11d


src/app/shared/components/template/components/table/table.component.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+@Component({
+  selector: "plh-table",
+  templateUrl: "./table.component.html",
+  styleUrls: ["./table.component.scss"],
+})
+export class TmplTableComponent extends TemplateBaseComponent implements AfterViewInit {
+  params = computed(() => this.getParams(this.parameterList()));
+  dataRows = computed(() => this.getDataRowsFromValue(this.value()));
```
chrismclarke, line 29: nit(blocking): the `getDataRows` from value function is async, so this computed will not behave as expected (no such thing as async computed). Instead it would be better to use an effect to respond to value changes and set the dataRows as a signal. I think it still kinda works in the frontend due to the additional cdr calls, but I'd consider that to be more of a bug than expected behaviour

src/app/shared/components/template/template.module.ts
```diff
@@ -26,14 +28,16 @@ import { DEMO_COMPONENTS } from "packages/components/demo";
   imports: [
     CommonModule,
     FormsModule,
-    ReactiveFormsModule,
     IonicModule,
-    SharedPipesModule,
-    NouisliderModule,
     LottieModule,
+    MatSortModule,
+    MatTableModule,
```
chrismclarke, line 34: nit(non-blocking): I can't remember how well we've implemented optimisation for this module, but might be more viable to include the imports in the table component itself and turn into a standalone component so that the modules are easier to remove from build when the component is not needed

chrismclarke (changes requested):
Thanks @jfmcquade 
Overall I think component code is set out nicely and feature works well. 

Longer term I'm still unsure whether we want to be importing the entire angular-material library just for one component, although at least with tree-shaking it only adds about 64kb to bundle size (which is still less than some components like pdf, lottie or odk viewer), and hopefully we can find a sensible way to include in optimisation system to remove from deployments that don't include the component

image

Longer term I think we'd come across similar bundle sizes if implementing AgGrid, so I think the only reasonable alternative would be Tanstack Table, although that would mean implementing quite a bit more UI code so I think maybe keep as-is for now.

Regarding my own use case to render data-items in a table, I think likely I'd be best creating another component (likely `data_table` component) that handles data subscriptions and finds a way to just pass the final row data to this component (may include minor refactor).

For the demo case you presented I think it might be good to include a few additional authoring options to implement pagination, as I can see the large data list is quite painful to render otherwise. I'll make a follow-up PR to address this and the other blocking issues I've raised

chrismclarke:
@jfmcquade 
See my proposed changes in https://github.com/IDEMSInternational/open-app-builder/pull/2838

Once happy may also want to tag Esmee for functional testing

jfmcquade:
Thanks for the updates, @chrismclarke. I've merged your changes and update the PR description.

I'll also experiment with deployment build optimisation(#2520) on this feature branch and see if it works for the table component as it expected.

@esmeetewinkel this should now be ready for functional testing.

chrismclarke:
> Thanks for the updates, @chrismclarke. I've merged your changes and update the PR description.
> 
> I'll also experiment with deployment build optimisation(#2520) on this feature branch and see if it works for the table component as it expected.
> 
> @esmeetewinkel this should now be ready for functional testing.

Thanks @jfmcquade 
I just quickly checked optimisation on debug build (setting optimisation config setting and then manually deleting the table component entry from the `reports/summary.json` which is read in for optimisation

Can confirm that the component is commented out as expected (no difference being a standalone component as the optimisation happens further down the page)

image

and also looks like it then omits all angular material/cdk modules from the build too

image

So I think that should be fine

chrismclarke (approved):

esmeetewinkel (approved):
Functional test passed. Comments below, most can be follow ups (happy for @jfmcquade to decide):

1. The `sort` function behaves a bit funny when it tries to sort `id_1` and `id_10`:

   [Screenity video - Mar 17, 2025.webm](https://github.com/user-attachments/assets/e601fdea-9324-44d0-8f76-b869a47037bb)

2. The icon to indicate sorting only appears on hover, and on click when used in device mode. Maybe on device-mode it would make sense to make the sort feature visible in some other way?
   image

3. The parameter `display_columns` is used in the `comp_table` sheet but not documented in the PR description. 

4. On initial read I was surprised that `display_columns` expects a comma separated list, whereas `sort_default` expects a space separated list, but I see now that `sort_default` is only about the _first_ column to sort by. In future I could imagine supporting something like `sort_columns: ward desc, county` which would be consistent with both proposed syntaxes.

chrismclarke:
> Functional test passed. Comments below, most can be follow ups (happy for @jfmcquade to decide):
> 
> 1. The `sort` function behaves a bit funny when it tries to sort `id_1` and `id_10`:

This is a bit of a classic human vs computer issue. As both IDs are stored as strings the sort is behaving as expected as when comparing strings something like `id_100` will come before `id_2` as sort order is determined by the first non-identical character. If wanting sort to behave the same as human interpretation you need to pad leading zeroes, e.g. `id_002` < `id_100`

I think more what is missing here is the ability to resort by initial `row_index` as extracted from the sheet, which is stored numerically. Could be worth a small follow-up to populate as an initial `#` column or similar

> 2. The icon to indicate sorting only appears on hover, and on click when used in device mode. Maybe on device-mode it would make sense to make the sort feature visible in some other way?

This appears to be a limitation with the package we use to implement the table sorting (https://material.angular.io/components/sort/examples). Looking at the styling the package uses, changing this behaviour would be pretty fiddly unfortunately, so I think unlikely to be able to fix unless we opt for a different solution (has been proposed, but would be low priority).

> 3. The parameter `display_columns` is used in the [comp_table](https://docs.google.com/spreadsheets/d/1utCg51oPxeT3rI1s74GmoNvdsS60aXOvCiVsKUTSMLY/edit?gid=603128459#gid=603128459) sheet but not documented in the PR description. Also, it is set there to `display_columns: ward, county` but I'm also seeing a column `subcounty` (so maybe this parameter is not supported?)

The change came from the changes I proposed in https://github.com/IDEMSInternational/open-app-builder/pull/2838, but also the original `column_list` changed name to `display_columns` (will update this PR)

The `subcounty` column only appears in the first example which doesn't use the `display_columns` option, but is omitted from the second example so I think is working as expected. It's a little extra confusing because there's 2 separate data lists used for different examples too

image

> 4. On initial read I was surprised that `display_columns` expects a comma separated list, whereas `sort_default` expects a space separated list, but I see now that `sort_default` is only about the _first_ column to sort by. In future I could imagine supporting something like `sort_columns: ward desc, county` which would be consistent with both proposed syntaxes.

Agreed, although would require a follow-up as by default the sorting package we're using only supports sort by a single column

So yeah, overall if planning a follow-up a few potential improvements could include
- Include row_index column (denoted by `#`)
- Consider providing authoring option to display all sortable column icons by default (possibly light colour)
- Add documentation
- Add support for custom column headers
- Add support for custom column value templating
- Add support for sort by multiple columns

Although I'd be hesitant to prioritise any of these until we know we have definite use-cases within deployments

jfmcquade:
I've added TODOs to #2848 for tracking. Merging this PR now

PR #2788, opened 2025-02-10
feat: Google Sign In button component

### LINKED ISSUE #2774: [FEATURE] Google sign in button
**What?**
Button variant that follows [Google's branding guidelines](https://developers.google.com/identity/branding-guidelines) to display when using the google auth action.

**Why?**
Rather than having a simple button saying "Login", Laurie [requested](https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/issues/150) to have a button with the Google icon and text "Login with Google". I believe we can only do this if we follow the branding guidelines.

**How?**
I tried achieving something like this this using CSS 
```
--background: #F2F2F2;
color: #1F1F1F;
font-family: 'Roboto';
```

in the `style_list`, however, this doesn't seem to get applied to the desired level of nesting: that applies to the `<plh-template-component>` level rather than the `<ion-button>`

Image

jfmcquade:
PR Checklist

## Description

Adds a new component: `google_sign_in_button`. This component has no exposed parameters, it's `value` will displayed as the test on the button.

On click, the button automatically triggers the equivalent of the `auth: sign_in_google` action. Additional actions can be authored as usual, e.g. `click | emit: force_reprocess`, and these will be triggered after the sign in logic completes, see example below.

This applies styling according to Google's [strict guidelines](https://developers.google.com/identity/branding-guidelines) for such a button – including rendering the Google logo inline.

Google's docs offer multiple styles (light/neutral/dark themes, pill/rectangular shape). Whilst we could expose options to customise these to authors, for now I've opted to just support the light/pill style. As the Google branded styling doesn't match our theming anyway, I don't think it's crucial to expose these options, and the "light" themed button is the most flexible for different applications (the border means it looks good on all background colours).

## Git Issues

Closes #2774

## Screenshots/Videos

[comp_google_sign_in_button](https://docs.google.com/spreadsheets/d/1L_AjZYIqTd6ubSYlmyrbGcrpgYaSVybzm7aLt_VFAHw/edit?gid=3937162#gid=3937162):

Screenshot 2025-02-11 at 14 44 44

https://github.com/user-attachments/assets/0d0c7de1-9dde-45a3-a47e-63b394575605


esmeetewinkel (changes requested):
Thanks! Looking good in RTL (broken record...)

It seems that Google's styles don't properly adapt to Arabic / RTL, see https://developers.google.com/identity/branding-guidelines?hl=ar
image and neither does the button when displayed in the app. 
image

One other comment is that the button doesn't amend well to long text strings (they just run off the page rather than wrapping and the button height increasing). However, I think we can accept this, since Google only allows very particular text strings to appear on this button anyway and trying a few languages I couldn't find any where the strings were so long that they would need wrapping on a screen size of 360px.

src/app/shared/components/template/components/button/google-sign-in/button-google-sign-in.component.scss
```diff
@@ -0,0 +1,116 @@
...
+  vertical-align: middle;
+  white-space: nowrap;
+  width: auto;
+  max-width: 400px;
+  min-width: min-content;
+}
+
+.gsi-material-button .gsi-material-button-icon {
+  height: 20px;
+  margin-right: 12px;
```
esmeetewinkel, line 44:
```suggestion
  margin-inline-end: 12px;
```

esmeetewinkel: @jfmcquade I think this addresses the issue I highlighted above
image

esmeetewinkel (approved):
Functional test passed

src/app/shared/components/template/components/button/button.component.ts
```diff
@@ -13,6 +13,7 @@ interface IButtonParams {
     | "card-portrait"
     | "flexible"
     | "full"
+    | "google_sign_in"
```
chrismclarke, line 16: nit(blocking): Given that the google_sign_in button has very specific behaviour I think it would be best if not included in default button component as a variant, but instead be it's own standalone component (e.g. `google_sign_in_button` )

chrismclarke (changes requested):
Thanks @jfmcquade

I'm not sure I can quite see in the code how the custom text is included in the component, however I think actually we don't want to include custom text. The strict guidelines you referred to are quite explicit

> To encourage users to click the button, we recommend the call-to-action text "Sign in with Google", "Sign up with Google", or "Continue with Google". It should be clear to the user that they are signing into your app or signing up for your app with their Google credentials, not signing up or registering for a Google account on your app.

So I think we should stick to a single  "Sign in with Google" and find a sensible way to mark within translations (not sure how we currently mark hardcoded strings for translation)

chrismclarke:
> In terms of the displayed text on the button, @esmeetewinkel and I discussed this morning and thought it would make sense to keep this as a directly author-able value.
> 
> For one thing, I don't think we do have a good system for handling translations of hardcoded strings: we have strings like "Next", "Previous" and "Done" in the tour service, see
> 
> https://github.com/IDEMSInternational/open-app-builder/blob/f8d523b7f33de454bd22ca2c1c077f9a291c879e/src/app/feature/tour/tour.service.ts#L55
> 
> , which I believe are just manually added to the translation inputs to ensure a translation exists; then there's https://github.com/IDEMSInternational/open-app-builder/blob/master/packages/data-models/app-config/globals.ts, which contains global hardcoded strings but is not integrated into the translation system (and hasn't been updated since 2022).

I can see a limited number of strings in `packages\data-models\app-config\globals.ts` which are renamed to as `APP_STRINGS` in `packages\data-models\appConfig.ts`. It would be good to create a follow-up issue to move all strings to this file (or a similar one) so that at some later stage it can be more easily integrated into translation workflows. 

> In any case, we thought that the authoring flexibility is desirable: it may be that we wish to have a longer or shorter text string displayed and these may vary by language (indeed, the direct translation of one string may be longer in another language). Potentially we could expose a variety of hardcoded messages via a param (e.g. `text: call_to_action_1` or `text: call_to_action_short`), but I think it's more straightforward to keep the text as directly authorable via the component row value.

I would say as a compromise perhaps we can set default value as "Sign in with Google" (imported from global app_strings) and allow users to override with a custom value - but should also mention in example/debug template the recommendations for text

chrismclarke (approved):
Thanks for the updates @jfmcquade , looks good to me

jfmcquade:
> I can see a limited number of strings in `packages\data-models\app-config\globals.ts` which are renamed to as `APP_STRINGS` in `packages\data-models\appConfig.ts`. It would be good to create a follow-up issue to decide on how we want to manage in the future for integrated into translation workflows, but I think for now having the link to google text recommendations on the comp example you've added seems reasonable to me

Thanks, I've made a follow-up issue here: #2797

PR #2791, opened 2025-02-11
Feat: auth user data

### LINKED ISSUE #2776: [FEATURE] Google auth display name and image
**What?**
Ability to display name and profile image of a user that is signed in with a Google account. 

There's also been a request to flag accounts that belong to someone under 16 (but @jfmcquade said that we don't have access to that information).

**Why?**
Requested in https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/issues/149. 

**How?**
Ideally this data would not sync to the postgres server, since this is Personally Identifiable Information (PII). This relates to https://github.com/IDEMSInternational/open-app-builder/issues/2516.

jfmcquade:
PR Checklist

## Update 2025-02-14
See notes from merged review PR #2796 

---

## Description

When signing in with Firebase (currently the only supported auth provider), as well as storing the `_auth_user_id` in a field, now some additional fields are updated: 
- `_auth_user_display_name`
- `_auth_user_family_name`
- `_auth_user_given_name`
- `_auth_user_profile_image_url`

These fields are not synced to our database and are only stored on device, to avoid any issues with storing PII data.

## Author notes

- The values of these fields come straight from Google, rather than being set by the user when logging in to our app specifically
- The family and given name fields are fetched from additional request to Google's API. If there is an issue with this step, then these fields will not be populated. Therefore I think it makes sense to use the `_auth_user_display_name` as a fallback for how to address the user, in the case that the first name cannot be fetched (see demo sheet below for an example)
- Since the request to Google can only be made at the point of the user signing in and not when re-opening the app for a user that has already signed in, the `_auth_user_family_name` and `_auth_user_given_name` fields will not be populated for users who already have the app installed and have signed in before the changes in the current PR were included. In this case, the user would need to re-authenticate with Google in order for us to get their names. I don't know if any users belong to this category.

## Dev notes

Please double-check my implementation of excluding the fields from the server sync, to confirm that the changes are sufficient to do so.

Rather than defining an `EXCLUDED_FIELDS` const in `packages/data-models/fields.ts`, I considered changing the `PROTECTED_FIELDS` enum to a type/interface that stores metadata alongside the value of each field, e.g.
```ts
AUTH_USER_DISPLAY_NAME: {
   value: "auth_user_display_name",
   exclude_from_server_sync: true
}
```
However, I opted to keep the simple enum as it is.

Also, I had hoped to be able to define the type of the user object returned by a given auth provider's API entirely within the file that defines that particular auth provider. I wasn't able to find a way to do this, so compromised by including the type in `src/app/shared/services/auth/types.ts`, which would need to be updated for any new auth provider added.

## Git Issues

Closes #2776

## Screenshots/Videos

[example_google_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705)

Screenshot 2025-02-12 at 12 36 42

Demo:

https://github.com/user-attachments/assets/11bef9a9-00e4-42e0-919c-3dc32bcd201e


chrismclarke (approved):
Looks good to me,

@esmeetewinkel, just to note that the PR description here will be slightly outdated due to changes introduced in review PR #2796 (would be good to also check author notes added there), but otherwise should be ready for functional testing

esmeetewinkel (commented):
Looking great, and thanks for adding separate given and family name fields. 

When I log in I can see the name fields populated correctly, but the picture doesn't behave as I expected: My Google account doesn't have an image, and so I normally see a purple square with an "E". However, the app displays a broken image icon, even though the image hyperlink does work when I paste it in my browser directly. 

image
image

@jfmcquade is this something that would be straightforward to address?

esmeetewinkel (approved):
@jfmcquade don't worry about my picture issue above - it is showing now (I must have done some sort of refresh)

jfmcquade:
Thanks both. I've updated the PR description to incorporate points from #2796. Notably, the names of the app fields have been updated.

chrismclarke:
> @jfmcquade don't worry about my picture issue above - it is showing now (I must have done some sort of refresh)

I've noticed myself that google can quite heavily rate-limit access to the image, so if you've refreshed a page a couple times it just stops allowing the image to be retrieved

@jfmcquade - that makes me think, at some point it will probably worth adding a way to persist an image retrieved from the web locally so that it does not have to be fetch every time it is displayed (and require active internet connection). Possibly worth a discussion/follow-up issue

PR #2793, opened 2025-02-11
refactor!: template-calc service

chrismclarke:
PR Checklist

## Breaking Changes
The `window.calc` accessor has been marked as deprecated, e.g. `@calc(window.calc.pick_random([1,2,3]))`
Any methods available via `window.calc.xxxx` are already available directly, e.g. `@calc(pick_random)`, so use that way instead.

I can't see any examples of this being used in [example_calc](https://docs.google.com/spreadsheets/d/15TccctjzepSPPv9xrLp46fhYldNN0wrho1zA5voTYcg/edit?gid=1595229908#gid=1595229908), so assume this is mostly a legacy behaviour anyways

The only method that should still be accessed via the window is the `date_fns` utility library, e.g. `@calc(window.date_fns.add_days(...))`

## Description
Changes merged from #2772 for easier review, a handful of updates to the `template-calc.service` which handles authored `@calc(...)` statements, specifically

## Author Notes
Beyond the marked deprecation, there shouldn't be any breaking changes or functional testing required.

## Review Notes
The code doesn't introduce changes to how calc statements are evaluated (for follow-up), so more just to confirm test cases behave as expected

## Dev Notes
The changes introduced here should be mostly non-functional (testing, mock and type-safety improvements). The additional method included is designed for future integration in #2772, so in itself shouldn't be breaking. 

Whilst diving into the service I did notice a handful of inconsistencies and clunky areas within the code, however have avoided larger refactor as would be best done through RFC procedures first (beyond marking `window.calc` as deprecated. Goals of a larger refactor would include:

- Improve ways to define deployment-specific calc functions
- Customisable list of 3rd party libs to include in calc (currently just date_fns included for all)
- Improved parser methods to extract globals required by calc statements and only including those when generating calc context

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/services/template-calc.service.ts
+++ b/src/app/shared/components/template/services/template-calc.service.ts
@@ -1,4 +1,4 @@
-import { IFunctionHashmap, IConstantHashmap } from "src/app/shared/utils";
+import { IFunctionHashmap, IConstantHashmap, evaluateJSExpression } from "src/app/shared/utils";
 import { Injectable } from "@angular/core";
 import { Device, DeviceInfo } from "@capacitor/device";
 import * as date_fns from "date-fns";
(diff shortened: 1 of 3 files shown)


src/app/shared/components/template/services/template-calc.service.ts
```diff
@@ -41,10 +62,23 @@ export class TemplateCalcService extends AsyncServiceBase {
...
+   * The expression is evaluated as JS, with additional access to global constants, function and
+   * evaluation context variables
+   * */
+  public evaluate(expression: string, evalContext: FlowTypes.TemplateRowEvalContext = {}) {
+    const calcContext = this.getCalcContext();
+    const calcExpression = expression.replace(/@/gi, "this.");
+    const { thisCtxt, globalFunctions, globalConstants } = calcContext;
+    const mergedContext = { ...thisCtxt, ...evalContext };
+    return evaluateJSExpression(calcExpression, mergedContext, globalFunctions, globalConstants);
+  }
```
chrismclarke, line 80: Note - this method has been adapted from the template-variable service which currently handles evaluation. It will be used as part of the refactor in #2772

esmeetewinkel (approved):
Noted about the breaking changes. I normally use the examples in the `debug` deployment as a starting point for writing `@calc()` statements, so it's plausible I never realised `@calc(window.calc. ...)` even existed as a valid syntax.

PR #2795, opened 2025-02-12
Feat!: auth restore data

chrismclarke:
PR Checklist

## Breaking Changes
As part of the work on this PR we need a way to create and expose to author data that has been created as part of internal services (in this case the list of auth profiles available for restoration). 

For convenience, I've kept the same `@data` convention for these internal lists, but used the prefix `app_` to mark as internal, so that the auth service populates an `app_auth_profiles` list that authors can read via `@data.app_auth_profiles`. 

This has the breaking knock-on that means any other lists that an author creates named starting `app_` will not be processed correctly (would be assumed internal and not attempt to load initial data from sheets)

## Description
- Add support for data created by internal app services to be exposed to to authors `@data` references (using prefixed `app_` flow names)
- Provide `app_auth_profiles` table for tracking restorable auth profiles
- Update [example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705) to include profile restore example

## Review Notes
See example in [example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705)

Note - should remove `app_auth_profiles` data_items loop condition to display (set to `false` to avoid breaking work in #2791)

</body>
</html>

## Author Notes
As an alternative to the `app_` naming convention we could use a different prefix to access auth restore data, e.g. `@app.auth_profiles` or `@auth.restore_profiles`. Each of these would take a bit of extra refactoring to manage data retrieval from the templating system but could be worthwhile if we expect extending in similar ways in the future.

Additionally the `app_auth_profiles` could be renamed to anything else (keen to keep the `auth_` prefix there to identify the service that creates the data.

As a minor bit of tidying I've renamed the `example_google_auth` sheet to `example_auth` and updated examples
https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705

Due to a quirk with how contact fields are stored only the `_server_sync_lateste` field currently works to retrieve from contact fields, because all the rest are named `@item.contact_fields.rp-contact-field.some_field` which does not parse in the current templating system (the last `"rp-contact-field.some_field"` is actually a single string not nested json). This could be resolved with proposed changes to item parsing in #2772

It might be worth creating a follow-up issue to consider removing these prefixes when storing on the server (and possibly locally as well as I'm not sure what purpose they really serve currently - originally designed to support multiple apps in same localstorage however this is only used when running on localhost as otherwise all apps have their own per-domain localstorage, and we don't actually use currently)

It would also likely be good to create a follow-up issue that allows a user to fully remove their profile from the server (or at least mark for deletion). This will both help in debugging and also GDPR (currently require manual request for data deletion)

## Dev Notes
Would be good to have additional feedback on the notes outlined in author notes above

## Git Issues

Closes #

## Screenshots/Videos

https://github.com/user-attachments/assets/94db4838-b401-4f6b-8ac5-b5faf1ebf06f

### DIFF
--- a/src/app/shared/services/auth/auth.service.ts
+++ b/src/app/shared/services/auth/auth.service.ts
@@ -12,6 +12,7 @@ import { toObservable } from "@angular/core/rxjs-interop";
 import { ServerService } from "../server/server.service";
 import { HttpClient } from "@angular/common/http";
 import type { IServerUser } from "../server/server.types";
+import { DynamicDataService } from "../dynamic-data/dynamic-data.service";
 
 @Injectable({
   providedIn: "root",
@@ -30,7 +31,8 @@ export class AuthService extends AsyncServiceBase {
     private injector: Injector,
     private templateService: TemplateService,
     private serverService: ServerService,
-    private http: HttpClient
+    private http: HttpClient,
+    private dynamicDataService: DynamicDataService
   ) {
     super("Auth");
     this.provider = getAuthProvider(this.config.provider);
@@ -51,6 +53,15 @@ export class AuthService extends AsyncServiceBase {
       },
       { allowSignalWrites: true }
     );
+    // expose restore profile data to authoring via `app_auth_profiles` internal collection
+    effect(async () => {
+      const profiles = this.restoreProfiles();
+      if (profiles.length > 0) {
+        const collectionData = profiles.map((p) => ({ ...p, id: p.app_user_id }));
+        await this.dynamicDataService.ready();
+        await this.dynamicDataService.setInternalCollection("auth_profiles", collectionData);
+      }
+    });
   }
 
   private get config() {
--- a/src/app/shared/services/auth/auth.service.ts
+++ b/src/app/shared/services/auth/auth.service.ts
@@ -60,6 +60,7 @@ export class AuthService extends AsyncServiceBase {
         const collectionData = profiles.map((p) => ({ ...p, id: p.app_user_id }));
         await this.dynamicDataService.ready();
         await this.dynamicDataService.setInternalCollection("auth_profiles", collectionData);
+        console.log("[Auth] Restore Profiles", profiles);
       }
     });
   }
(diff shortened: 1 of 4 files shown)


esmeetewinkel (approved):
Functional test passed.

jfmcquade (approved):
Looks good, and is working functionally.

Just for clarity, I would highlight that there are two differences between standard dynamic data and the new data accessible via the `@data.app_<field_name>` syntax:
- The new internal app data does not correspond to, and is independent of, any static base-level authored data list
- The new internal app data does not persist in indexedDB browser storage, and only exists in app data, so only relates to the current app session

In theory there could be another category of data introduced in the future that doesn't depend on an authored data list but does persist between app launches.

Regarding the points in the Author Notes, I'm happy with the `@data.app_*` syntax and so I think the other points makes sense as follow-ups. One query – would `@data._app_*` be a better syntax to identify this as a protected name?

chrismclarke:
> Looks good, and is working functionally.
> 
> Just for clarity, I would highlight that there are two differences between standard dynamic data and the new data accessible via the `@data.app_<field_name>` syntax:
> 
> * The new internal app data does not correspond to, and is independent of, any static base-level authored data list
> * The new internal app data does not persist in indexedDB browser storage, and only exists in app data, so only relates to the current app session

Correct, it is generated by the app, and from an authoring perspective read-only

> In theory there could be another category of data introduced in the future that doesn't depend on an authored data list but does persist between app launches.

Anything that is not authored is generated in the app, so I still think probably just safer to regenerate the data instead of persisting. Although I guess the one exception might be something like data retrieved from an external URL, although in that case I think still would likely include some sort of authored data_list that at least specifies where to retrieve the data from and what list to persist it to

> Regarding the points in the Author Notes, I'm happy with the `@data.app_*` syntax and so I think the other points makes sense as follow-ups. One query – would `@data._app_*` be a better syntax to identify this as a protected name?

I think using an underscore before the name makes a lot more sense (seems consistent with generated fields like `_auth_user_id`). I've updated the PR to drop the `app` and just use an underscore prefix, so 
```
@data._auth_profiles
```

PR #2796, opened 2025-02-12
Review/auth user data

chrismclarke:
PR Checklist

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/services/auth/providers/firebase.auth.ts
+++ b/src/app/shared/services/auth/providers/firebase.auth.ts
@@ -3,48 +3,43 @@ import { FirebaseAuthentication, User } from "@capacitor-firebase/authentication
 import { getAuth } from "firebase/auth";
 import { FirebaseService } from "../../firebase/firebase.service";
 import { AuthProviderBase } from "./base.auth";
+import { IAuthUser } from "../types";
 
 export type FirebaseAuthUser = User;
 
 @Injectable({
   providedIn: "root",
 })
-export class FirebaseAuthProvider extends AuthProviderBase<User> {
+export class FirebaseAuthProvider extends AuthProviderBase {
   public override async initialise(injector: Injector) {
     const firebaseService = injector.get(FirebaseService);
     // TODO - is service required here?
     if (!firebaseService.app) {
       throw new Error("[Firebase Auth] app not configured");
     }
-    this.addAuthListeners();
     // use firebase authStateReady to ensure any previously logged in user is available
     await getAuth().authStateReady();
   }
 
   public async signInWithGoogle() {
-    const result = await FirebaseAuthentication.signInWithGoogle();
-    const accessToken = result?.credential?.accessToken;
-    await this.getGoogleUserInfo(accessToken);
+    const { user, additionalUserInfo } = await FirebaseAuthentication.signInWithGoogle();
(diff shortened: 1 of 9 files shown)


src/app/shared/services/auth/providers/firebase.auth.ts
```diff
@@ -3,48 +3,72 @@ import { FirebaseAuthentication, User } from "@capacitor-firebase/authentication
...
   public async signInWithGoogle() {
-    const result = await FirebaseAuthentication.signInWithGoogle();
-    const accessToken = result?.credential?.accessToken;
-    await this.getGoogleUserInfo(accessToken);
+    const { user, additionalUserInfo } = await FirebaseAuthentication.signInWithGoogle();
+    if (user) {
+      // NOTE - additionalUserInfo is only returned on first signIn so persist to localStorage
+      // for access on automated sign-in following restart
+      const { profile } = additionalUserInfo;
+      if (profile) {
```
jfmcquade, line 32: I was wondering if it would be possible for a user to sign in successfully, but for the `profile` in `additionalUserInfo` that is returned by `signInWithGoogle()` to be undefined (or `null`). In that case, a user would be prompted to sign in again with Google on each app launch. It's not clear to me from [the Firebase docs](https://firebase.google.com/docs/reference/js/auth.additionaluserinfo) if that's possible, but I can imagine that a user could have an account without having associated additional user info (although I don't know if that is possible in actuality with a Google account). 

Instead off this check, possibly we could set the auth metadata field like:
```ts
localStorage.setItem(AUTH_METADATA_FIELD, JSON.stringify(profile || {}));
```
to track that an attempt was made to retrieve this data. However that wouldn't handle the case of trying again to retrieve this profile data if it available but for whatever reason a given attempt to retrieve it on initial sign-in was unsuccessful.

Either way, I'm think I'm happy with this as it is

jfmcquade (approved):
Thanks @chrismclarke, I like these improvements.

Good to have a standard profile info spec to work to ([openID](https://openid.net/specs/openid-connect-basic-1_0.html#StandardClaims))

> As we now need to store more information from when a user first signs in this PR will force all existing authenticated users to re-login when first launched. This will happen following app init and not part of any authored workflow.

This makes sense. See inline comment for a minor query.

> I noticed the manual method that had been added to retrieve userInfo directly from google API endpoints, however on closer inspection of the default auth method this same info was available from additionalUserInfo.profile in the signIn response. I attempted using this method to retrieve refresh userInfo for an already-signed-in user, however this wasn't possible due to the different tokens made available to signed in users vs sign in response (access token vs id token). I've retained the original method in case useful in the future (added to the draft supabase provider)

Nice, makes sense to keep it there provisionally in case it is required.

src/app/shared/services/auth/providers/firebase.auth.ts
chrismclarke: Good catch, the [api docs](https://firebase.google.com/docs/reference/js/auth.additionaluserinfo) says can be `null` (guessing if using a non-google provider with firebase auth that does not provide federated profile data), so good not to introduce potential infinite login loop (although at least would get an error logged about trying to stringify undefined)

Issue #2763, opened 2025-01-30
[FEATURE] Kuwait: map view styling


esmeetewinkel:
**What?**
Currently the map view `plh_progress_path` has a blue border styling of a module task card that seems related to task being in progress. I'd like to use a blue border instead for the unlocked modules that are not highlighted. 

Highlighted module images also appear slightly bigger than others - I think it would make sense for that to apply to all unlocked modules. In that case we have consistency between map and list homescreen:
- locked: greyed out, smaller image, lock icon
- unlocked: larger image
   - not highlighted: blue border
   - highlighted: yellow border

PR #2805, opened 2025-02-21
feat: plh_module_list_item highlighted param and circle variant

### LINKED ISSUE #2756: [FEATURE] KUWAIT: yellow plh_module_list_item
**What?**
For the PLH Kids KW deployment, we need a new variant of the `plh_module_list_item` component (available only in the `plh_kids_kw` theme) that is yellow/mustard colour instead of blue. 

**Why?**
This should be the equivalent in the module list view of the yellow/mustard border available in the `plh_progress_path` view.

**Design (doesn't include this specific case)**
Image

jfmcquade:
PR Checklist

## Description

Updates the `plh_module_list_item` component:

- Adds a new param, `highlighted`
  - If true, "highlighted" styling is applied (by default, the theme's secondary colour is applied as a highlight)
- Adds a new variant, `circle`
  - This variant is styled to very closely resemble the styling of the `circle` variant of the `task_card` component
  - This variant responds as expected to the same `highlighted`/`locked` parameters
  - The fact that this is now a variant of the same component should open up a neater authoring pattern for the map/list view on the `plh_kids_kw` deployment homepage
- Various other styling improvements
  - Brings the styling more inline with the original designs for the component
  - Tidies the style/component code, removing the need for theme-level override styling

## Follow-up

If the new circle variant is sufficient, and the task card's circle variant is no longer needed, the latter should be removed from the code, is it was always a somewhat messy workaround for a specific use-case.

## Git Issues

Closes #2756 
Related issues: #2763, https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/issues/145

## Screenshots/Videos

[comp_plh_module_list_item](https://docs.google.com/spreadsheets/d/1Tq8Ib_ZcJsABW9U9XbQxgnmtv_uayXk28DXoa94j1aE/edit?gid=569531329#gid=569531329) template:

Screenshot 2025-02-21 at 18 22 01

| default variant | circle variant |
|---|---|
| Screenshot 2025-02-25 at 12 48 09 | Screenshot 2025-02-25 at 12 48 16 |

[comp_plh_progress_path](https://docs.google.com/spreadsheets/d/130xy0FbCqgVMg8astiR5FBzuUMfVBHtg9k6wmweE1BA/edit?gid=603128459#gid=603128459)

Screenshot 2025-02-21 at 18 26 37

Screenshot 2025-02-25 at 12 48 34


esmeetewinkel (commented):
Minor comment: The comparison between the two styles would be slightly more consistent to me if the blue and yellow border styles were the same size also for the `circle` variant. In that case, across both variants: 
- locked: smaller, greyed out, lock appears
- unlocked (independent of whether or not it's highlighted): larger

This also seems desirable from a user perspective, where the initial state of the app they will see is `circle` variant with 1 unlocked unhighlighted module and the other modules  locked unhighlighted.

image

jfmcquade:
@esmeetewinkel yeh that makes sense.

I think you're suggesting two changes if I've understood correctly:
1. Consistent border width across the two variants
2. The size of the `circle` variant buttons should be the same when the instance is not locked – i.e. the unhighlighted unlocked circle button should be made larger to match the size of the highlighted unlocked circle button

Is that right? In which case:

1. This was bugging me too, but I chose the heavier border for the circle variant because of the way it has to appear over the background image in our current use case. I found that with a less heavy border, the colour was lost against the busy background, so I kept the heavier border to match the designs. I also tried giving the default variant a heavier border to match, but this didn't look right. So as the `circle` variant currently has the single use case of appearing over a background image, I think the inconsistency is acceptable/desirable at the moment

2. Yeh this makes sense, I can quickly make the update, just let me know if I've misunderstood the request

jfmcquade:
@esmeetewinkel I've updated with https://github.com/IDEMSInternational/open-app-builder/pull/2805/commits/bec9f03ccbb628a7f9583df7b0424869eac2de98 and updated the screenshots in the PR description acordingly

esmeetewinkel (approved):
Functional test passed.

PR #2807, opened 2025-02-22
fix: data-items action value

chrismclarke:
PR Checklist

## Description
Fix issue where actions triggered from within a `data-items` loop would not correctly parse the value of the triggering row `this.value` variable

## Review Notes
Should fix case in [debug_data_items_actions_2](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit?gid=1415551165#gid=1415551165)

## Author Notes
This also now added console logs anytime a `set_item` or `set_data` action is called

## Dev Notes
The issue occurs because due to the way data_items keep their own local variable references, and avoid notifying parent template of inner value changes (updated in #2706). The problem with this is that the parent template is in charge of parsing any `this.value` instances with the row value. 

There were two possible fixes/workarounds, this first would be to move similar logic to the `data-items` action interceptor to handle replacement, the second would be to pass the current value of the triggering row with the action so that it can be replaced from there. As we already have a `triggered_by` variable attached to each action it made the most sense to simply use this for replacement instead of looking at the parent `templateRowMap`

There may be some minor conflicts between here and draft PR #2753. It would probably be good to try and revisit/update #2753 post-merge to make sure the relevant changes are also copied across

## Git Issues

Closes #

## Screenshots/Videos

**Example - Using radio_button_grid inside data_items loop to dynamically set item to match value**

https://github.com/user-attachments/assets/bc98efc7-6ad8-48db-b7b4-ea71df66f6de

**Example - Updated console logs**
image


jfmcquade (approved):
Looks good, nice straightforward solution. And great to add debug logs for dynamic data updates

esmeetewinkel (approved):
Functional test passed. And I'm so happy to see the dynamic data changes logged in the console :)

PR #2808, opened 2025-02-22
Refactor: value changed actions

chrismclarke:
PR Checklist

## TODO
- [x] Builds on top of #2807 , should merge after

## Description
- Remove individual component `changed` action triggers
- Add `changed` action trigger to base component, and automatically trigger on any value change
- Update actions to not trigger when values unchanged (e.g. clicking same value in a combo_box)

## Review Notes
If testing [debug_data_items_actions_2](https://docs.google.com/spreadsheets/d/1HvJOcbYGtrkPxwkPp8lvNW63ngsvworzz_jEUo5Swe0/edit?gid=1415551165#gid=1415551165), should note before actions triggering every value click (not just different), and the template becoming broken if the same value is clicked multiple times in a row. After events only trigger if value clicked different to current value.

It might also be useful to pick one or two components from those changed in this PR and just check the `changed` action triggers as expected - a few good examples are in [example_changed_action](https://docs.google.com/spreadsheets/d/1tpno5SmIjX1VZ7OaECRCZXZhWDPrQloWRcVMhPLqb00/edit), possibly some other sheets too that I'm less aware of

## Author Notes
This change shouldn't cause authoring issues, although just to be aware that previously setting the same value on a component would trigger `changed` actions whereas now it doesn't (although assuming actions are designed to behave the same on each trigger then I can't imagine cases where this would be undesirable)

The one possible minor exception is the `plh_parent_point_box` which manually triggered `click` actions before `change`. Now the default behaviour is to trigger `change` before `click`, so if any authoring is order-dependent it may cause issue.

As an added benefit, now all components that are capable of setting values will trigger a `changed` action when the value is changed - previously this had to be included by developers on a per-component basis

## Dev Notes
I've removed all manual calls to trigger `changed` action from components to call from the base component instead. I've manually checked all updated components to make sure they include `extends TemplateBaseComponent` and so inherit the change

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade (approved):
Nice, the code changes definitely make sense. Good to have the "changed" action as a fundamental property of components as opposed to handled on a case-by-case basis. 

I've double-checked your changes to components: I can see from the changes that the "changed" action is was only being triggered when components were also calling `this.setValue()`. Conversely, I've checked all instances of components calling `this.setValue()` to see whether there might be any cases that we don't want to fire "changed" – I'm not sure what case this might by in theory, and can't find any examples on practice.

esmeetewinkel (commented):
This seems to work well for the `radio_button_grid`. On the `combo_box` component I found that:
- actions do trigger again when the same value is selected
- in that case, the console log shows that the item is set to "undefined" (this was already the case prior to this PR)

Screenshot in `plh_kids_kw` deployment, template `article_consequences_practise`

**After (running locally on PR branch)**
image

**Before (on https://plh-kids-kw1.web.app/template/article_consequences_practise)**
image

chrismclarke:
> Whoops, I had missed this PR since it's draft. @chrismclarke I assume it should be marked as ready-for-review now that the PR it depends on is merged?

Yep that will be it, think it was waiting on another PR that got delayed a little and forgot to unmark the draft post-merge. Thanks for picking up on it

>This seems to work well for the radio_button_grid. I can't notice any difference on the toggle-bar nor number-selector (but neither allows for selecting the same value twice in a row). On the combo_box component I found that:

> - actions do trigger again when the same value is selected
> - in that case (and there is a set_item action), the console log shows that the item is set to "undefined" (this was already the case prior to this PR)

Thanks for checking. I think I vaguely remember setting this behaviour explicitly due to an issue where some deployments want to essentially be able to unselect a previously selected option, so good to know it's still working the same as before.

PR #2809, opened 2025-02-24
server: increase payload size limit

jfmcquade:
PR Checklist

## Description

increases the size limit on body payloads posted to API server. This is required to resolve an issue where attempting to sync user data would result in a `413 (Payload Too Large)` response to the POST request to the server.

Solution taken from [this answer](https://stackoverflow.com/a/59978098).

The value of 32mb was chosen to match the `client_max_body_size` as specified in `packages/server/docker/config/nginx/templates/server.conf.template`, but could likely be much lower.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "api",
-  "version": "1.6.0",
+  "version": "1.6.1",
   "scripts": {
     "prebuild": "rimraf dist",
     "build": "nest build",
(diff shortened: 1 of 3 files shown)


chrismclarke (approved):
Thanks for updating, oh yeah I can see the default is 100kb so seems reasonable to bump it. 

I can see why matching the server limit makes sense, although I'm going to reduce to `2mb` as I think if we start reaching payloads that sort of size would really want to come up with a better mechanism to sync

I'll deploy to the server to check whether it is working as expected

chrismclarke:
Server updated, can confirm larger payloads accepted when testing locally

**Before**
image

**After**
image

PR #2811, opened 2025-02-24
fix: auth restore profiles order

jfmcquade:
PR Checklist

## Description

Fixes a minor issue where the list of auth profiles available to be restored were not sorted correctly by the `updatedAt` property.

I'm assuming that the original logic was an error as I don't see the significance of the comparison, although I may well be missing something. It's also worth noting that the author can sort the list themselves, see the example of this in the [settings_login](https://docs.google.com/spreadsheets/d/1hy3nSqXZaLjVFRCQgayZjzSMVCX26P3HNuR_KtKjE3c/edit?gid=56143870#gid=56143870) template on the `plh_kids_kw` deployment:

Screenshot 2025-02-24 at 17 40 45

## Git Issues

Closes #


chrismclarke (approved):
Thanks, no that will be a typo for sure, makes sense to merge

PR #2816, opened 2025-02-25
fix: combo box visibility bug

jfmcquade:
PR Checklist

## Description

Fixes a bug where the combo box component would not always be visible on initial render of parent template.

## Testing

As well as testing on the `plh_kids_kw` templates referenced in the issue, it would be good to have a quick check of [comp_combo_box](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=569531329#gid=569531329) debug template to ensure all the functionality is still working as expected. I have checked myself and haven't found any issues.

## Dev notes

The bug was due to issues with reactivity and change detection on some level – likely the value of the `variant` property as read in the html template not updating reliably (if `variant` is undefined, the component would not render any content). This is fixed by refactoring the component to make all param values reactively read from a `params()` signal (in line with recent updates to other components, e.g. button). Further improvements could probably be made to the component's logic, but as the handling of the `modal` variant is quite complex, via the `combo-box-modal` sub-component, I'm wary of changing it unnecessarily.

## Git Issues

Closes https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/issues/152

## Screenshots/Videos

Testing the template referenced in the issue, `article_oneononetime_explore_5` on the `plh_kids_kw` deployment, all the combo boxes now consistently render, whereas before there was almost always an issue with at least one when loading the page.

Screenshot 2025-02-25 at 12 08 43


chrismclarke (approved):
Thanks @jfmcquade - I had wondered originally whether the combo_box was just set-up to hide when data not available but makes more sense that it's a reactivity issue. Definitely happy to have the code refactored to use signal patterns, looks good to me

esmeetewinkel (approved):
Functional test passed on `debug` and `plh_kids_kw` deployments.

PR #2822, opened 2025-02-26
chore: tidy server sync code

### LINKED ISSUE #2819: [FEATURE] Allow authors to disable server syncing at a template level
**What?**
The author should be able to configure a template such that while that template is open, the syncing to the server is paused.

**Why?**
As part of allowing user profile restore from the same device, we need to ensure that the user's data is not overwritten before being imported back to the app. One way to accomplish this is by disabling the server sync whilst the template that handles logging in and restoring data is active.

**How?**
This could be achieved by exposing an app config property to disable/enable server syncing, which could then be overridden at the template metadata level.

**Other details**

jfmcquade:
PR Checklist

## Description

- Moves `SERVER_SYNC_FREQUENCY_MS` property from app config (runtime) to deployment config (compile time)
  - In theory this is a breaking change, but I don't believe any deployments actually override this property's default value
- Tidies logic in `db-sync` and `server` services and adds explanatory comments

## Git Issues

Related to #2819 – tidies up the related code but doesn't close the issue. To be discussed whether we want to expose a way to pause server syncing, e.g. via a template action.


chrismclarke (changes requested):
I'm not quite sure if this PR is pushing in the right direction to solve the underlying issue. I feel like putting the onus on authors to enable/disable syncing for specific cases is a bit of a minefield, and one we might be better off trying to avoid entirely. Equally it seems a bit strange that a single template would have the power to put a block entirely on server syncing - I could imagine maybe a general action like `server_sync: disable`, but not using something that depends on rendered UI. 

Given that this also doesn't fully address the initial issue, I think it might be better to look for alternate approaches. E.g. if needing to avoid syncing during profile restore action then likely best if the user service that handles restoration also handles temporarily suspending the sync service as required.

At the same time I think it also highlights how probably this server sync frequency doesn't make sense as a runtime/changing property. I've added a PR targeting this branch to make it a deployment config option instead, although appreciate that no longer closes this linked issue

#2827

chrismclarke (approved):
Thanks @jfmcquade 
I'm hopeful that improvements made to profile sync/restore more generally as discussed on the call should remove the need for more explicit author-driven sync changes, and so will move any ongoing discussion related to #2819 to that PR when open

Issue #2652, opened 2024-12-21
[FEATURE] Reset all data action


esmeetewinkel:
**What?**
An action to clear all dynamic data / reset all data lists to their original values. 

**Why?**
I often include a "Reset App" button (with action `click | reset_app`) somewhere in the app which can be helpful for testing. I noticed that this does reset fields to their original state (and removes fields that didn't have a value initially), but the dynamic data doesn't get cleared. 

**How?**
We already have authoring patterns `click | reset_data | _list_id: feat_data_actions_list` and `click | reset_app` so I think this could be `click | reset_data`.

PR #2824, opened 2025-02-27
Feat: user import dynamic data merge strategy

### LINKED ISSUE #2810: [BUG] User import action does not update progress negatively
**Describe the bug**
Observed on the `plh_kids_kw` deployment, e.g. in [this comment](https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/issues/53#issuecomment-2665952446).
At a high level, the bug as observed is that when signing in with Google and choosing to load previously saved progress, the content does not seem to update in the case that the data being loaded is "behind" the current progress – in this case, when the user has completed more modules in the current session than they have in the data that is being loaded.

More technically, the issue is that when performing the `user: import` action to import dynamic data from a remote database user entry, the changes being applied from the imported data are just applied as updates to the apps dynamic data. This means that if updates have been made in the current session to a row that has no updates in the imported data, the current values will persist and will not be altered by the updates. This explains the behaviour as observed: when the user has currently made more progress through the modules than the data they import, the additional progress persists after import.

**App version**
code: 0.17.1
`plh_kids_kw` content: 1.2.26

**To Reproduce**
See this video demonstrating the behavior for the [plh-kids-kw-uat1.web.app](plh-kids-kw-uat1.web.app) web preview. Progress has been made in the incognito on the left hand side. Progress is reset in the window on the right hand side, where the user is signed in, and a sync is performed. The window on the left then signs in and chooses to load saved progress (from the recently updated data from the other window). The additional progress persists.

https://drive.google.com/file/d/1gbddE8xTMLBX7mjjrH3d9zfmrzb8_UGE/view?usp=sharing

**How**
It is not completely clear whether this should be seen as a feature or a bug. It may be desirable to perform this kind of "merge" of the data rather than replacing any previously saved data. When loading previously saved data in the `plh_kids_kw`, a warning is shown that explicitly states that the current data will be lost. If that warning wasn't there, perhaps the current behaviour would be what the user expects?

However, it seems like for this use case, we probably do want to reset the user's data before applying the imported data. This could be achieved by resetting all local dynamic data before applying the updates from the imported data (we don't currently have a method to do so but could add).

Assuming its desirable, we could expose an option to the `user: import` action to toggle this behaviour, e.g. `click | user: import: a_user_id | replace_data: true`. Or, moving to our preferred action syntax, `click | user: import | user_id: a_user_id, replace_data: true`.

In fact, we may want to support other merge strategies in future, so could potentially anticipate that with the following syntax: `click | user: import | user_id: a_user_id, merge_strategy: replace`, or similar.

jfmcquade:
PR Checklist

## Description

When restoring user profile info via the `user: import` template action, the user's dynamic data is now reset before the new data is imported. Previously, the imported data would be merged as updates.

The new underlying logic to reset all dynamic data is exposed in these additional ways:
- Via calling the `reset_data` template action without specifying a `_list_id` to reset
- On the `/user` debug page

## Git Issues

Closes #2810

## Screenshots/Videos

### DIFF
--- a/src/app/feature/user/pages/user-debug/user-debug.page.ts
+++ b/src/app/feature/user/pages/user-debug/user-debug.page.ts
@@ -1,4 +1,5 @@
 import { Component, Injector, OnInit, signal } from "@angular/core";
+import { FlowTypes } from "packages/data-models";
 import { TemplateActionService } from "src/app/shared/components/template/services/instance/template-action.service";
 import { TemplateFieldService } from "src/app/shared/components/template/services/template-field.service";
 import { AuthService } from "src/app/shared/services/auth/auth.service";
@@ -82,6 +83,7 @@ export class UserDebugPage implements OnInit {
   /** Prepare table data to display for provided dynamic data entry */
   public setDynamicEntryView(entry?: IDynamicDataEntry) {
     if (entry) {
+      console.log("[USER] entry", entry);
       const rows = Object.values(entry.data);
       this.dynamicDataTable = { headers: Object.keys(rows[0]), rows };
     } else {
@@ -90,6 +92,17 @@ export class UserDebugPage implements OnInit {
   }
   public dynamicEntryCompareFn(a: IDynamicDataEntry, b: IDynamicDataEntry) {
     return a.id === b.id;
+  }
+
+  public async resetSelectedDynamicData() {
+    const { flow_type, flow_name } = this.dynamicDataSelected;
+    await this.dynamicDataService.resetFlow(flow_type as FlowTypes.FlowType, flow_name);
+    location.reload();
+  }
+
+  public async resetAllDynamicData() {
+    await this.dynamicDataService.resetAll();
+    location.reload();
   }
 
   /** Retrieve localStorage entries prefixed by field service prefix */
--- a/src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
+++ b/src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
@@ -155,6 +155,11 @@ export class PersistedMemoryAdapter {
     this.persistStateToDB();
   }
 
+  public async deleteAll() {
+    this.state = {};
+    await this.persistStateToDB();
+  }
+
   /** Trigger persist handler. Requests will be debounced and notified when complete */
   public async persistStateToDB() {
     this.statePersist$.next("pending");
@@ -198,7 +203,7 @@ export class PersistedMemoryAdapter {
     const db = await this.mapDBToObject({});
     const idsToDelete = [];
     for (const flow_type of Object.keys(db)) {
-      const { deleted } = compareObjectKeys(db[flow_type], this.state[flow_type]);
+      const { deleted } = compareObjectKeys(db[flow_type], this.state[flow_type] || {});
       for (const flow_name of deleted) {
         const rowHash = db[flow_type][flow_name];
         for (const row_id of Object.keys(rowHash)) {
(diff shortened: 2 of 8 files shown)


esmeetewinkel:
@jfmcquade Does this also close https://github.com/IDEMSInternational/open-app-builder/issues/2652?

esmeetewinkel (changes requested):
I can't consistently get this to work (in video below I get no data, but I also think I've seen merged data...). I'm testing from the PR preview branch in a regular and private browser window.


https://github.com/user-attachments/assets/d6552646-8016-461c-bf38-8465bdef0a62

jfmcquade:
> I can't consistently get this to work (in video below I get no data, but I also think I've seen merged data...). I'm testing from the PR preview branch in a regular and private browser window.

@esmeetewinkel Apologies for the delay, I think I've addressed the source of the issue you highlighted (with https://github.com/IDEMSInternational/open-app-builder/pull/2824/commits/e31acd126d71b14172dcd735b72aa91f5c45c540) – I now consistently get the data coming through on import as expected.

@chrismclarke tagging you as I think this is now ready for review

esmeetewinkel (approved):
Functional test passed

src/app/feature/user/pages/user-debug/user-debug.page.ts
```diff
@@ -141,6 +142,17 @@ export class UserDebugPage implements OnInit {
...
 
+  public async resetSelectedDynamicData() {
+    const selected = this.dynamicDataSelected();
+    await this.dynamicDataService.resetFlow("data_list", selected);
+    location.reload();
+  }
+
+  public async resetAllDynamicData() {
+    await this.dynamicDataService.resetAll();
+    location.reload();
```
chrismclarke, line 153: nit(non-blocking): resetting all data also probably doesn't need full reload. As this is a debug page I think we can expect the user to manually refresh if something doesn't appear quite right after reset (but I'd expect main use case would be author staying on page and observing a single table which should still be fine)

src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
```diff
@@ -155,13 +155,37 @@ export class PersistedMemoryAdapter {
...
+          return null;
+        }
+        // Since each collection relates to a single flow, just take the first
+        const doc = docs[0].toMutableJSON();
+        const { flow_type, flow_name } = doc as any;
+        return { flow_type, flow_name };
+      })
+    );
+
+    return allResults.filter((result) => result !== null);
```
chrismclarke, line 187: nit(blocking): This seems like quite a specialised/niche method to be exposing publically from what is meant to be a fairly generalised service. In these types of situation I'd usually prefer to write a `hack` - style method in the service that just retrieves the private underlying db as required, e.g.  `dynamicDataService['writeCache']['db']`

For this specific case, I'm not even sure if it's required - struggling to follow fully but looks like you're trying to get a full list of tables used by dynamic data service so that you can then reset them one-by-one (excluding any that are internal)... I'm guessing it might be simpler to just wipe everything and and reload the app/page so that any internals can be re-initialised as required

chrismclarke (changes requested):
I've added a handful of comments in-line, but attempted my best to address in #2874

I haven't had time to fully test the proposed changes, but would still appreciate if you could take a look and see whether makes sense to merge into this PR for final testing

chrismclarke (approved):
@esmeetewinkel 

This PR was updated from #2874, so if you get a chance could you re-confirm functional tests still passing? Have tagged for re-review

The main change from an authoring perspective is that you should now see the page fully reload after a user is imported to ensure the data is updated correctly

PR #2834, opened 2025-03-05
chore(ios): update ios build files for firebase auth

jfmcquade:
PR Checklist

## Description

In order to enable firebase auth on ios, additional configuration is required. These changes were applied by following the instructions in [these docs](https://github.com/capawesome-team/capacitor-firebase/blob/main/packages/authentication/docs/setup-google.md#ios) for the capacitor plugin we are using. Additionally, I ran the following to update the CocoaPods accordingly:

```sh
cd ios/App
pod repo update
pod update
```

These changes were made specifically for the `plh_kids_kw` deployment. There are two obstacles to generalising these changes:
1. The CocoaPods plugin changes (see changes to `Podfile` and `Podfile.lock`) are only required for deployments that make use of the Firebase auth feature.
    - For now, I think the best strategy is probably just to include these changes for all deployments as we don't have a good way of removing these on a deployment-level basis
2. The "URL Scheme" that must be included in the `Info.plist` file comes from Firebase. Specifically, it is the value of `REVERSED_CLIENT_ID` in the `GoogleService-Info.plist` file that must be downloaded from the Firebase console and manually added to the iOS build files before building the app (step 7 [here](https://open-app-builder.com/developers/building-and-deploying-to-ios/#building-for-the-first-time)).
    - The value included in this PR relates specifically to the `plh_kids_kw` deployment (this isn't considered sensitive)
    - This means that it isn't easy to incorporate into our existing `ios configure` workflow. Currently all the values required for populating the iOS build files as part of this workflow come directly from the deployment config. We don't rely on the `GoogleService-Info.plist` being populated first.
    - One option would be to require this property in the deployment config for apps with Firebase auth enabled and use this value to populate the relevant field as part of the `ios configure` workflow. E.g. the config could have a new property such as one of the following: `config.ios.firebase_auth_url`; `config.firebase.auth_url`; `config.auth.firebase_url`
        - However, this property isn't particularly easy for an author to understand, and would require navigating the Firebase console to get the required value

## Git Issues

Closes #

## Screenshots/Videos

Successful login on iOS emulator after making the changes:

Screenshot 2025-03-05 at 11 10 48


chrismclarke:
> These changes were made specifically for the `plh_kids_kw` deployment. There are two obstacles to generalising these changes:
> 
> 1. The CocoaPods plugin changes (see changes to `Podfile` and `Podfile.lock`) are only required for deployments that make use of the Firebase auth feature.
>    
>    * For now, I think the best strategy is probably just to include these changes for all deployments as we don't have a good way of removing these on a deployment-level basis

I'm inclined to agree, it's very much the pathway the original react-native/expo ecosystem took of just including everything by default and allowing users to manage their own opt out processes (although over time did manage to implement a system where everything becomes opt-in). I doubt this will have a big impact on bundle sizes and assuming it doesn't require any extra permissions added so hopefully won't be a big issue. The one thing worth checking if there are any knock-ons when updating an app that specifically doesn't require this (e.g. are there any extra checks that happen on the app store because they see new SDKs - android can be quite finicky for these types of things, e.g. if it see an ad-mob SDK expects certain ads declaration even if not using ads)

> 2. The "URL Scheme" that must be included in the `Info.plist` file comes from Firebase. Specifically, it is the value of `REVERSED_CLIENT_ID` in the `GoogleService-Info.plist` file that must be downloaded from the Firebase console and manually added to the iOS build files before building the app (step 7 [here](https://open-app-builder.com/developers/building-and-deploying-to-ios/#building-for-the-first-time)).
>    
>    * The value included in this PR relates specifically to the `plh_kids_kw` deployment (this isn't considered sensitive)
>    * This means that it isn't easy to incorporate into our existing `ios configure` workflow. Currently all the values required for populating the iOS build files as part of this workflow come directly from the deployment config. We don't rely on the `GoogleService-Info.plist` being populated first.
>    * One option would be to require this property in the deployment config for apps with Firebase auth enabled and use this value to populate the relevant field as part of the `ios configure` workflow. E.g. the config could have a new property such as one of the following: `config.ios.firebase_auth_url`; `config.firebase.auth_url`; `config.auth.firebase_url`
>      
>      * However, this property isn't particularly easy for an author to understand, and would require navigating the Firebase console to get the required value
> 

If possible I think it would be better to update the code to read the value from a template, just like how things like `<string>$(MARKETING_VERSION)</string>` are parsed.
Naively I'd assume if you define the variable in the same `buildSettings` part of the `ios\App\App.xcodeproj\project.template.pbxproj` file then we can use an environment variable to populate the build settings which in turn should update the value in the plist file.

As for extracting the variable, agreed it doesn't make sense for authors to have to specify. Assuming that eventually all these builds will happen as part of a github action anyways which would likely have some sort of `env.GOOGLE_SERVICES_PLIST` variable, better to focus on reading that value from the plist instead. You could add a devDep [plist.js](https://www.npmjs.com/package/plist) and include a step in the ios configure task that parses the plist file and populates the `REVERSED_CLIENT_ID` as part of the envReplace. E.g.

```ts
const { REVERSED_CLIENT_ID } =  await parsePlist(pathToGoogleServicesPlist)
 await envReplace.replaceFiles({
     // ...
      REVERSED_CLIENT_ID: parsedPlist.,
  });
  ```

PR #2835, opened 2025-03-06
fix: firebase auth issues on ios

jfmcquade:
PR Checklist

## Description

Fixes an issue where the Auth service with Firebase as the provider would not initialise correctly on ios. This manifested as the issues logged in [this mattermost thread](https://chat.idems.international/all/pl/1z7w5ue55jb38xiarrgx7yx3bh).

## Testing

@esmeetewinkel to confirm login flows are working:
- [iOS appetize](https://appetize.io/app/ios/international.idems.plh-kids-kw?device=iphone14pro&osVersion=16.2)
- [Google Play Internal App Sharing link](https://play.google.com/apps/test/international.idems.plh_kids_kw/1002033) (building locally and uploading to appetize is not currently working for me)

## Dev notes

These issues proved difficult to debug, in part because initially I could only replicate them in appetize and not via my local emulator. This is explained by the fact that the issue is related specifically to serving the app from a `capacitor://localhost` domain, and I was testing locally by serving from my own IP address via `yarn ng serve --configuration=external`. See error log below.
```
Cross-origin redirection to http://developers.google.com/ denied by Cross-Origin Resource Sharing policy: Origin capacitor://localhost is not allowed by Access-Control-Allow-Origin. Status code: 301
```

Also, as the login functionality was still working via the dedicated google sign in button component, it wasn't immediately obvious that the auth service wasn't successfully initialising. The issue was with the `FirebaseAuthentication.getCurrentUser()` method specifically, run as part of the service init, which wasn't resolving and so blocked the full init whilst `signInWithGoogle()` could still be called successfully. The issue seems to be in how Firebase Auth stores its persisting data. This PR configures it to use indexedDB when on native platforms, in accordance with [this answer](https://stackoverflow.com/a/74904270) and [these docs](https://firebase.google.com/docs/auth/web/custom-dependencies#platform-specific_considerations). Using this configuration on web does not seem to work, so the platform check is required.

## Git Issues

Closes #


esmeetewinkel (approved):
Functional test passed on iOS Appetize, Android native device and web preview

image

For future reference, I pasted the two screen videos of the issues this addresses in the mattermost chat referenced in the issue (the two screen videos that sit in the chat so far were addressed by authoring changes, I believe)

chrismclarke (approved):
Super interesting issue, thanks for the deep-dive on it @jfmcquade 

Looking at the links you've shared and a quick google leads me to agree with your assessment - the `getAuth()` method assumes it is in a browser environment uses methods incompatible with capacitor. 

I don't actually think it's an issue with indexeddb vs localstorage, but more how `getAuth` includes the `browserPopupRedirectResolver` which will not work when called. Setting the `indexedDB` persistence also enforces the `cordovaPopupRedirectResolver` instead
https://github.com/firebase/firebase-js-sdk/issues/5019
https://firebase.google.com/docs/reference/js/auth.dependencies.md

There is part of me wondering whether all of this has been fixed in more recent updates for firebase and/or capacitor-firebase, as essentially the automatic behaviour used to check for URLs starting `ionic://` which broke when capacitor updated to `capacitor://` URLs way back in Capacitor 3 (discussed a bit in this thread
```
https://github.com/firebase/firebase-js-sdk/issues/5019
``` 
However I still think the fix you have provided makes a lot of sense and also reflects the guidance in the current capacitor-firebase docs

https://github.com/capawesome-team/capacitor-firebase/blob/main/packages/authentication/docs/firebase-js-sdk.md

I'm assuming post-merge any existing android users will be forced to log back in (could be worth checking), although I don't think that should have any real impact except perhaps if being prompted to restore profile which would depend a bit on deployment. It might still be good to quickly test following merge though @esmeetewinkel @jfmcquade

jfmcquade:
> I'm assuming post-merge any existing android users will be forced to log back in (could be worth checking), although I don't think that should have any real impact except perhaps if being prompted to restore profile which would depend a bit on deployment. It might still be good to quickly test following merge though @esmeetewinkel @jfmcquade

I think this is true (I've been prompted to log back in when relaunching the app under some circumstances). @esmeetewinkel and I to test on planned release of `plh_kids_kw` deployment this week. I'm merging this PR now.

esmeetewinkel:
> I'm assuming post-merge any existing android users will be forced to log back in (could be worth checking), although I don't think that should have any real impact except perhaps if being prompted to restore profile which would depend a bit on deployment. It might still be good to quickly test following merge though @esmeetewinkel @jfmcquade

I just updated my native android app (`plh_kids_kw`) to content v1.2.36 / code v0.18.0 which includes this PR, and my login persisted normally. I was not prompted to login again.

Issue #2438, opened 2024-09-23
[BUG] Splash screen inconsistently applied


chrismclarke:
**Describe the bug**
As part of API changes to Android 12, splash screens running on Android 12 or higher (API 31+)  devices display differently to those running on lower versions
https://developer.android.com/develop/ui/views/launch/splash-screen#how

The new behaviour by default displays the app logo on a plain background. This is likely inconsistent if a custom splash screen has been specified 

**To Reproduce**
_Describe the steps to reproduce this bug._

**Screenshots and videos**
(TODO - check examples of existing apps comparing Android 12 and Android 11)

**Additional context**
The API 12 changes come both with new functionality but also limitations. Splash icons can now be [Animated Vector Drawable](https://developer.android.com/develop/ui/views/animations/drawable-animation) format which supports a form of animated SVG, however the size of the icon is tightly constrained (fits within a 192dp diameter circle within a 288dp square).

These diagrams are however not compatible on Android 11 and below, so that gives a couple options to ensure consistency:

1. Icon-based splash
Remove support for custom splash screen images and request content teams instead use the app logo (ideally provided as SVG). We could provide an additional authoring option to control background color on the splash screen as a means to enhance the logo. This svg can be converted into a standard vector asset and displayed in all android versions.

2. Animated Icon with Fallback
Alternatively we could allow support for content teams to pass their own custom XML that defines animated vector drawings to show on android 12+ and use a separate fallback image defined as previous. We could also consider wrapping a couple common animation styles and exposing as a utility to automatically generate animation from the initial SVG (e.g. fade in, scale in etc.). 

3. Custom splash activity
A third option would be to display a mostly blank initial splash, and then launch an activity which shows a custom splash screen after on all devices. If launched from within the app itself then this could be templated in a standard way (but would likely be slow to launch), or could be implemented at the android level with whatever functionality we can expose (e.g. lottie animation files, images etc.). Assuming the initial splash background matches the new activity background this could still be fairly seamless, just a little bit slower depending on implementation (initial splash dismisses after signal from main activity that it is ready to render, usually around 300-500ms or so, but can be configured for longer).

**Resources**
https://shapeshifter.design/
https://itnext.io/a-comprehensive-guide-to-android-12s-splash-screen-api-644609c811fa

[https://developer.android.com/develop/ui/views/launch/splash-screen#how](Android 12 splashscreen docs)
[...]

PR #2837, opened 2025-03-07
feat: reusable-ios-build action

chrismclarke:
PR Checklist

## Description
- Add `reusable-ios-build` action to handle build of ios app

- Improve reusable build action to support custom post-build scripts and artifact upload. This enables triggering things like ios configuration scripts following the app build, so that follow-up runner doesn't have to manage deployment-specific code.

## Author Notes
Once this PR is merged similar actions as the one in the debug repo PR can be added to other content repos to generate ios simulator builds. As the action depends on the latest version of the main code repo, the content repo variable for `APP_CODE_BRANCH` will also need to be set to one that includes these merged changes

The action does not (currently) upload automatically to testflight, nor can they generated artifacts be uploaded to the app store (requires follow-up work to handle code signing). So for now the main use-case is manually triggering the build action and then manually uploading the generated artifact to appetize for testing on ios simulator device

The content repo will also need and additional `GOOGLE_SERVICES_PLIST` secret if using the android-equivalent `GOOGLE_SERVICES_JSON` to enable firebase features. This file can be generated from the firebase console for the associated project.

## Review Notes
The action is designed to be triggered by a content repo. A testing PR has been added to the debug repo used to trigger this action: https://github.com/IDEMSInternational/app-debug-content/pull/154

See output artifacts from workflow run. 
[Example Deployment Repo Workflow Run](https://github.com/IDEMSInternational/app-debug-content/actions/runs/13745252358)

This should be a simulator build, able to run locally on a mac or uploaded directly to testflight.

## Dev Notes
There's quite a lot going on in the action, and a whole lot more that will still require further work/follow-up. For now this really just seeks out to lay the groundwork for triggering ios builds, and is likely not yet in a state where it will drastically reduce the amount of ongoing manual work. 

It should however at least be a decent proof-of-concept, and lead to a clearer understanding on the next steps required (the number of [debug workflow runs](https://github.com/IDEMSInternational/app-debug-content/actions/workflows/debug-ios-build.yml) on the content repo should be a good indicator of the learning involved!).

**General action overview**
1. Trigger the reusable web build to set and build code for deployment. Use an additional post-build script to sync the generated www folder to the ios app via the `npx cap sync` command. This is done as part of the build action to avoid having to reconfigure deployment and populate ios config in step 2. Upload the full ios folder as an artifact for use in next step

2. Checkout the builder repo and replace the default ios folder with the artifact generated in the build step (will contain public www contents as well as compiled capacitor.config.json). Install node_modules again as native plugins store podfiles in node_modules folders, and are required for build

This step could possibly be optimised if there were a cleaner way to store just the node_modules required for ios build in an artifact, so that we don't need to run a `yarn install` at all in this step.

3. Generate an unsigned build for simulator and upload.

It would also be possible to generate an unsigned `.xcarchive` that could be imported locally into xcode, signed, and uploaded, although I think it would probably still be better to focus on integrating signing processes into CI instead. So for now the main purpose is generating builds capable of uploading to appetize.

**Github Actions - Choice of Tools**

When it comes to building IOS via CI, there are generally 3 approaches
1. Use raw commands to trigger various `xcrun` or `xcodebuild` commands directly on the runner, with additional steps to [provision apple certificates](https://docs.github.com/en/actions/use-cases-and-examples/deploying/installing-an-apple-certificate-on-macos-runners-for-xcode-development)
2. Use a utility like [fastlane](https://fastlane.tools/) as a wrapper around commands and certificate provisioning
3. Use a 3rd party action, the most popular being [sparkfabrik/ios-build-action](https://github.com/sparkfabrik/ios-build-action) and [yukiarr/ios-build-action](https://github.com/yukiarrr/ios-build-action). On github marketplace these are called `build-ios-action` and `ios-build-action` respectively which is super confusing

I initially went for option 3. however found both actions were just thin wrappers around fastlane scripts (option 2), and included a mix of both a bunch of functionality that we don't want (e.g. direct upload to browserstack, enforce appstore login even when not uploading build), as well as lacking some functionality we may want (e.g. generate simulator/xcarchive builds, upload to appetize, generate screenshots etc.). I also noted the sparkfabrik repo started as a fork of the yukiarr repo, however both are loosely maintained/active. 

Next I went for option 1. however quickly found there to just be a huge amount of boilerplate code required to achieve basic builds and very difficult to deubg/test locally (no access to xcode cli without a mac to test examples). Here's an [example gist](https://gist.github.com/huynguyencong/004e98e4d9e7671f93fec280ddb7fc18) that should give an idea

So I settled on option 2. There is a follow-up choice of how to manage fastlane (namely whether to define scripts in a `fastfile` or execute inline. For now I've opted to execute inline although we may want to setup specific tracks in a fastfile in the future

**Certificate Management**
Building for IOS is typically more involved than android, as both certificates and provisioning profiles are required that need to be provided from a registered apple developer account

There's a few different approaches to this, as outlined in the fastlane [codesigning docs](https://docs.fastlane.tools/codesigning/getting-started/), although the only viable for CI are:

1. Using fastlane's [match](https://docs.fastlane.tools/actions/match/) tool which stores all certificates in external storage (e.g. private git repo or google cloud bucket) for import into workflows
2. Provide files directly via github secrets as base64 encoded files

I've experimented with both options, although limited in what can be done without access to various certs and keys that require developer account admin access as well as a local mac device. For now I've skipped the signing to process to generate only unsiged builds, which can still be exported as `.app` files to run on simulator or `.xcarchive` which can later be signed and exported as `.ipa` via xcode locally. 
I'd recommend a follow-up discussion to decide best way to handle signing builds before following up with action code to automate deploying to testflight. 

For now I think the best option would likely be using [fastlane match](https://docs.fastlane.tools/actions/match/) (1), authenticating via app-store-connect API. The benefit of this over (2) is the amount of secrets that would need encoding into the github action is fewer (1-2 vs 4-5). There would then be a follow-up choice to decide if fastlane match would store credentials in a private git repo or a google cloud storage bucket, here I would be leaning towards GCS as it has slightly better options for user-management (uses service accounts which are reasonably easy to set-up whereas github relies on personal access tokens which have much broader scope)

**Debugging workflow**
To speed up development workflow for triggered actions, I noted it would be possible to just run the build action once and then reference the generated artifacts in future builds via the run-id (instead of rebuilding each time). Local modification to the workflow looks like below (should note that build artifacts expire after 24h, so needs at least one new run per day while testing)
```yml
 #   build_web:
  #     uses: ./.github/workflows/reusable-app-build.yml
  #     with:
  #       branch: ${{ inputs.branch }}
  #       post-build-cmd: npx cap sync ios
  #       additional-artifact: ios
  #     secrets: inherit

  debug:
    runs-on: macos-latest
    # needs: build_web
    steps:
      - name: Download IOS artifact
        uses: actions/download-artifact@v4
        with:
          name: ios
          #   NOTE - need to explicitly include token to download from a different workflow run
          # https://github.com/actions/download-artifact/issues/320
          github-token: ${{ github.token }}
          run-id: 13727815102
```
In order to trigger build, typically new commits need to be pushed on the debug repo. Empty commits can be added via command line
```
git commit --allow-empty -m "chore: Trigger Build"
```

There is added complication that any changes made to builder repo files will only be available to the action after they have been merged into master, due to the way the `APP_CODE_BRANCH` variable specifies which builder branch to checkout. As such if making changes to builder repo files (e.g. creating fastlane files in the ios folder), then this should temporarily be updated to target the current pr branch

**Misc Notes**
I've registered an ios app in the debug-app firebase project and added the google services plist to the debug content repo `GOOGLE_SERVICES_PLIST` secret for use in testing

## TODO - Follow-up
- [ ] Remove debug action from debug repo and replace with named action.
- [ ] Revert debug app `APP_CODE_BRANCH` variable in debug repo (currently targeting this branch, should target `master`) 
- [ ] Fully automate all ios config workflow to setup deployment build as required (migrate manual methods, handle splash/icon and anything else as required)
- [ ] Create additional workflow to upload output to testflight directly from action (likely refactoring ios_build workflow to accept custom input args that could be used to generate either simulator or production build)
- [ ] Organise IDEMS org certificates and include in test action (migrating any existing certs). If exporting credentials from local machine see [example commands](https://www.andrewhoog.com/post/how-to-build-an-ios-app-with-github-actions-2023) from blog post, although might be better to create new credentials for use throughout
- [ ] Setup fastlane locally to remove the amount of boilerplate required for actions (e.g. `fastlane init` files in ios folder)
- [ ] Replace testing credentials with updates on debug repo
- [ ] Extend actions to also submit to testflight using fastlane with [app-store-connect-api](https://docs.fastlane.tools/app-store-connect-api/)
- [ ] Migrate changes to standalone actions repo
- [ ] Refactor android actions to use post-build and additional artifacts to improve efficiency
- [ ] Consider migrating to fastlane match (with google storage bucket)
- [ ] Documentation

## Git Issues

Closes #

## Screenshots/Videos
**Example - content repo action artifacts. The `simulator_app` artifact contains app for upload to appetize**
image

**Example - Debug App built and (manually) uploaded to appetize**
image

### DIFF
--- a/.github/workflows/reusable-app-build.yml
+++ b/.github/workflows/reusable-app-build.yml
@@ -52,19 +52,21 @@ on:
           description: Optional folder name to upload as additional artifact (e.g. ios/android folders)
           type: string
           default: ""
-  
+        workflow-runner:
+          description: Optionally specify github runner to use for build action
+          type: string
+          default: ubuntu-latest
       outputs:
         GIT_SHA:
           description: "Git SHA of build head"
           value: ${{ jobs.build.outputs.GIT_SHA }}
-  
 jobs:
     build:
       outputs:
         GIT_SHA: ${{ steps.populate.outputs.GIT_SHA }}
       env:
         ARTIFACT: test
-      runs-on: ubuntu-latest
+      runs-on: ${{inputs.workflow-runner}}
 
       steps:
         - name: Check out app code
--- a/.github/workflows/reusable-ios-build.yml
+++ b/.github/workflows/reusable-ios-build.yml
@@ -116,14 +116,11 @@ jobs:
       # https://blog.richardszalay.com/2016/08/22/ios-deploy-pipeline-3-setup-and-build-test-phase/
       # https://stackoverflow.com/a/66278353
       - name: Build IOS
-        env:
-          # Supported configurations in IOS project are Debug and Release
-          configuration: Debug
         working-directory: ios/App
         # Call fastlane build_ios_app track
         # https://docs.fastlane.tools/actions/build_ios_app/
         run: |
-          bundle exec fastlane build_ios_app --configuration "${{env.configuration}}" --scheme "App" --skip_codesigning true --skip_package_ipa true --build_path "./build"
+          bundle exec fastlane build_ios_app --configuration "Release" --scheme "App" --skip_codesigning true --skip_package_ipa true --skip_archive true --build_path "./build" --output_directory "./build"
 
       #############################################################################
       #         Upload
@@ -133,7 +130,7 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: ios_build
-          path: /ios/App/build
+          path: ios
           # you can also archive the entire directory
           retention-days: 30
 
--- a/.github/workflows/reusable-ios-build.yml
+++ b/.github/workflows/reusable-ios-build.yml
@@ -47,16 +47,16 @@ jobs:
   #         Build Deployment Web
   #############################################################################
 
-  # build_web:
-  #   uses: ./.github/workflows/reusable-app-build.yml
-  #   with:
-  #     branch: ${{ inputs.branch }}
-  #     post-build-cmd: yarn workflow ios configure && npx cap sync ios
+  build_web:
+    uses: ./.github/workflows/reusable-app-build.yml
+    with:
+      branch: ${{ inputs.branch }}
+      post-build-cmd: yarn workflow ios configure && npx cap sync ios
 
-  #     additional-artifact: ios
-  #     # Ensure build also run on macos to allow npx cap sync to configure ios pods
-  #     workflow-runner: macos-latest
-  #   secrets: inherit
+      additional-artifact: ios
+      # Ensure build also run on macos to allow npx cap sync to configure ios pods
+      workflow-runner: macos-latest
+    secrets: inherit
 
   #############################################################################
   #         Build IOS
@@ -64,7 +64,7 @@ jobs:
 
   build_ios:
     runs-on: macos-latest
-    # needs: build_web
+    needs: build_web
     steps:
       # Checkout builder repo and install node_modules so that they can be used in ios build
       # (e.g. capacitor plugins store podfiles in node_modules)
@@ -77,11 +77,10 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: ios
-          # TODO - revert after debugging
-          #   NOTE - need to explicitly include token to download from a different workflow run
+          # Debug - use fixed build artifact (can skip build_web). Needs explicit token set
           # https://github.com/actions/download-artifact/issues/320
-          github-token: ${{ github.token }}
-          run-id: 13732579493
+          # github-token: ${{ github.token }}
+          # run-id: 13732579493
 
       # Remove default ios folder and replace with artifact generated
       - name: Setup IOS folder
--- a/.github/workflows/reusable-ios-build.yml
+++ b/.github/workflows/reusable-ios-build.yml
@@ -80,7 +80,7 @@ jobs:
           # Use build artifact generated either as part of this run or from previous run (if debug run id set)
           # https://github.com/actions/download-artifact/issues/320
           github-token: ${{ github.token }}
-          run-id: 13744145866
+          run-id: 13743953801
 
       # Remove default ios folder and replace with artifact generated
       - name: Setup IOS folder
--- a/.github/workflows/reusable-ios-build.yml
+++ b/.github/workflows/reusable-ios-build.yml
@@ -80,32 +80,32 @@ jobs:
           tar -xf artifact.tar --directory ios
           ls ios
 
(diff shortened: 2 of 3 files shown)


esmeetewinkel (approved):
Looking good. I was able to upload the output `simulator_app.zip` from the debug workflow run to appetize and the app was working as expected. 

image

jfmcquade (approved):
Thanks @chrismclarke. Quite a lot going on so I appreciate you documenting the process and various options.  

I haven't picked through the workflow code, but functionally I can confirm that I was able to download and run the build artefact in a simulator on my Mac (via `xcrun simctl install booted` command):
Screenshot 2025-03-10 at 11 27 24

It seems sensible to work on this incrementally, so I'd be happy to merge this as-is, tracking the future TODOs.

In terms of TODOs that I'm aware I'm responsible for:
- Handle ios asset generation (#2241)
  - #2364 makes some progress towards this, but needs updating
  - May intersect with #2438, but probably best to handle independently
- #2844

chrismclarke:
> In terms of TODOs that I'm aware I'm responsible for:
> 
> * Handle ios asset generation ([[FEATURE] Handle iOS app icon and splash generation #2241](https://github.com/IDEMSInternational/open-app-builder/issues/2241))
>   
>   * There's an open PR for me to revisit that makes some progress towards this, but needs updating: [Chore: Migrating from using the cordova-res package to using the capacitor-assets package for asset generation #2364](https://github.com/IDEMSInternational/open-app-builder/pull/2364)
>   * May intersect with updating the Android splash screen customisation ([[BUG] Splash screen inconsistently applied #2438](https://github.com/IDEMSInternational/open-app-builder/issues/2438)), but probably best to handle independently
> * [[FEATURE] Incorporate firebase auth configuration into ios configure workflow #2844](https://github.com/IDEMSInternational/open-app-builder/issues/2844)

Thanks for tracking these, yeah I can appreciate auto-submit to testflight is pretty much useless until we get these resolved, but I'd still say no major urgency as will take some time to sort out cert management strategy and such (have put on the agenda for next call). I'll make a note to re-visit the items tracked here when next working on the feature, and tidy-up longer term issues as required

PR #2838, opened 2025-03-10
Review/table component

chrismclarke:
PR Checklist

## Description

Targets branch of #2786 

- Add virtualised scroller to improve rendering performance for large data lists
- Add data searchbox and filter functionality

## Dev Notes
To solve the issue of poor rendering performance on large lists I initially tried to implement a mat-table paginator which functioned well to limit the number of rendered results, however it had a few shortcomings, notably
1) It required import of various material core and component themes to display generated UI in any reasonable way. This would require further work to integrate into theming in general and get particularly confusing for our different app themes. It would also require significant refactor when migrating from angular 17 -> 19 due to various theme breaking changes between versions
2) It included hardcoded text that would be tricky to incorporate into translation system
3) It would require exposing pagination options to authoring which could get confusing

So instead I opted for including a virtualised scroller instead, so that all rows can be displayed but only those visible in the viewport rendered. Virtual scrollers are often fiddly, usually relying on fixed item heights to calculate scroll bounds. I opted to use an experimental scroller from angular cdk which has support for dynamic item heights. It appears to work well for the current demo case, so should possibly wait and see whether we can just keep this setting for all lists or make it an opt-in authoring feature

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
   "private": true,
   "dependencies": {
     "@angular/animations": "^17.2.2",
+    "@angular/cdk-experimental": "^19.2.2",
     "@angular/common": "~17.2.2",
     "@angular/core": "~17.2.2",
     "@angular/elements": "^17.2.2",
--- a/src/app/shared/components/template/components/table/table.component.html
+++ b/src/app/shared/components/template/components/table/table.component.html
@@ -1,11 +1,18 @@
 <div class="table-container">
   <table mat-table [dataSource]="dataSource" matSort>
-    <ng-container *ngFor="let column of columnsToDisplay" [matColumnDef]="column">
-      <th mat-header-cell *matHeaderCellDef mat-sort-header>{{ column }}</th>
-      <td mat-cell *matCellDef="let element">{{ element[column] }}</td>
-    </ng-container>
-
-    <tr mat-header-row *matHeaderRowDef="columnsToDisplay; sticky: true"></tr>
-    <tr mat-row *matRowDef="let row; columns: columnsToDisplay"></tr>
+    <cdk-virtual-scroll-viewport autosize style="height: 100%">
+      @for (column of columnsToDisplay(); track column) {
+        <ng-container [matColumnDef]="column">
+          <th mat-header-cell *matHeaderCellDef mat-sort-header>{{ column }}</th>
+          <td mat-cell *matCellDef="let element">{{ element[column] }}</td>
+        </ng-container>
+      }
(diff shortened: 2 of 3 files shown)


packages/shared/src/utils/object-utils.ts
```diff
@@ -137,13 +137,20 @@ export function filterObjectByKeys<T extends Record<string, any>, K extends keyo
...
+  const loopArr = maxDepth ? arr.slice(0, Math.min(arr.length, maxDepth)) : arr;
+
   const keyHashmap: Record<string, boolean> = {};
-  for (const el of arr) {
+  for (const el of loopArr) {
     for (const key of Object.keys(el)) {
       keyHashmap[key] = true;
     }
   }
+  console.log({ keyHashmap });
```
jfmcquade, line 154: remove debug log

jfmcquade (approved):
Thanks, @chrismclarke, this seems to be working really well.

> Hardcoded sizing (80vh max-height) is proposed to be removed in favour of defining inline authoring styles and allowing table to default scroll within full page

The motivation for avoiding default scrolling was to allow for the sticky header functionality, but I think you're right that any sizing is best handled at the authoring level, which achieves the same thing.
 
> So instead I opted for including a virtualised scroller instead, so that all rows can be displayed but only those visible in the viewport rendered. Virtual scrollers are often fiddly, usually relying on fixed item heights to calculate scroll bounds. I opted to use an experimental scroller from angular cdk which has support for dynamic item heights. It appears to work well for the current demo case, so I think I feel reasonably comfortable making this the default behaviour for now, and could possibly investigate author-driven opt-out or alternate pagination if there is good reason to do so in the future.

Cool!

> As for whether angular-material is good to include in the repo or not, I'm probably leaning towards a no given the amount of effort it took when trying to use any of the functionality that is dependent on theming (and so prone to break in the future); it also gives another component library to keep updated with angular (when we already struggle to keep ionic updated). That being said, I think the table component is working quite nice as-is for now so would be inclined to keep until there is strong need for more table functionality (at which point I'd probably recommend refactoring to use the more primative [cdk/table](https://material.angular.io/cdk/table) and recreating UI features we need such as sort icons, or start fully from scratch with our own UI base on headless tanstack-table). I would say more generally best to avoid angular-material components where possible moving forwards, but open to using tools from the angular-cdk packages which typically underpin most of the components.

Yeh makes sense, I was looking for the library that seemed simplest to get something basic working. Agree that starting with angular-cdk packages could be a good option for this or similar components, I wasn't aware of it before.
 
I've extracted your suggested follow-up tasks to this issue: #2848

PR #2839, opened 2025-03-10
feat: user debug page dynamic updates

### LINKED ISSUE #2825: Show default values in `user` tables
Following PR #2815, the dynamic data tables in the `user` page show only the changed values (rather than the full rows of data where at least one value has been changed):

image

It might still be preferable to show the full list of values (default + changed). I'm guessing the most useful would probably be to only include columns that have changed values (to not show the entire data list), but then use some sort of highlighting/colour to show both the default value where unchanged. Maybe something like:

image

_Originally posted by @chrismclarke in https://github.com/IDEMSInternational/open-app-builder/issues/2815#issuecomment-2689004842_

### LINKED ISSUE #2826: `user` sidebar should show live updates
Dynamic data in the `user` sidebar doesn't update dynamically with the template, it requires a refresh to show the latest. For improved debugging it would be great to see live updates. 

_Originally posted in https://github.com/IDEMSInternational/open-app-builder/issues/2806#issuecomment-2689033392_

chrismclarke:
PR Checklist

## Description
Update user debug page to:
- Keep dynamic data datable live-updated with data changes
- Display full row data with visual distinction between default and overwritten values 

## Review Notes
Open a template that users dynamic data with user debug page sidebar, e.g.
```
http://localhost:4200/template/feat_data_actions(sidebar:user)
```

## Author Notes
The dynamic data table will only show the list of tables where data has been set, hence at the start of the demo video only a single table is available. 

When updating values the user debug table should reflect the updated values, showing in darker text the values that have been pulled from dynamic data vs default values. In cases where the dynamic data is updated to the same value as the default data, the text still appears darker as the update still exists in the database (would be quite a bit of added effort to always cross-check updated data against original, and more simple to just track all state changes instead)

## Dev Notes
The debug page code has been refactored to make more use of signals to handle live data updating. Given the amount of code related just to the dynamic data table it might be better split into a separate component, but kept as-is for now in case useful to see the git diffs during review

I decided to add this PR independent to the work on the authored table component as that would require quite a bit of additional follow-up to accept the input data rows from the dynamic data subscription as well as custom row styling

## Git Issues
Closes #2825
Closes #2826

## Screenshots/Videos

**Example - feat_data_actions template with user sidebar updating**

https://github.com/user-attachments/assets/a5494984-29c0-415e-8997-ac54c3d2b88b


jfmcquade (approved):
Really nice, I suspect @esmeetewinkel and @ChrisMarsh82 will be very happy with the feature. 

I might suggest moving the order of tables around on the page, as I suspect the Dynamic Data table is going to be more commonly useful than Contact Fields or Protected Fields, and currently I have to scroll to view it.

esmeetewinkel (approved):
Functional test passed for amending existing data rows. A few comments (non of which essential to be included in this PR):

1. While testing I found that rows that are dynamically created don't show on the `/user` page. See for instance the debug template `feat_data_actions_edit`
   image
2. When the `(sidebar:user)` is active, it could be nice to be able set the screen width of the LHS of the screen to a reasonable phone dimension, e.g. similar to `(sidebar:feedback)`: 
   image
3. Pop-ups and stacked templates appear in front of the sidebar. I believe that's a limitation we've been aware of from the start - but it's not great in the use case where I wanted to have a look at how the dynamic data changes as I'm navigating through a module, which is a stacked template...

   https://github.com/user-attachments/assets/052ea282-4de3-46db-8443-91f470b060c0

4. I noticed that, when navigating, `(sidebar:user)` shows up in multiple places in the hyperlink, not only with the current template but also with the `nav_parent` query parameter. E.g. http://localhost:4200/template/module_landing_selfcare(sidebar:user)?nav_parent=home_screen(sidebar:user)&nav_parent_triggered_by=unlocked_selfcare. Seems to work fine though, just not what I expected!

chrismclarke:
Thanks for the feedback @esmeetewinkel 

> Functional test passed for amending existing data rows. A few comments (non of which essential to be included in this PR):
> 
> 1. While testing I found that rows that are dynamically created don't show on the `/user` page. See for instance the debug template `feat_data_actions_edit`

Should now be fixed

> 2. When the `(sidebar:user)` is active, it could be nice to be able set the screen width of the LHS of the screen to a reasonable phone dimension, e.g. similar to `(sidebar:feedback)`:

> 3. Pop-ups and stacked templates appear in front of the sidebar. I believe that's a limitation we've been aware of from the start - but it's not great in the use case where I wanted to have a look at how the dynamic data changes as I'm navigating through a module, which is a stacked template...

I've attempted to address this issue in part by enabling the debug page to keep in sync when multiple browser tabs/windows are active. That way you can simply use multiple windows to resize as needed and won't come across issues with full-screen content. 

See attached demo:

https://github.com/user-attachments/assets/9c3d3f66-f35c-4b2d-b9ba-a458182427ab

Should note that this only appears to content on the debug page. If we want all dynamic data to stay in live-sync more generally across multiple tabs (i.e. without need to reload tabs) that would need to be a follow-up issue and PR

> 4. I noticed that, when navigating, `(sidebar:user)` shows up in multiple places in the hyperlink, not only with the current template but also with the `nav_parent` query parameter. E.g. http://localhost:4200/template/module_landing_selfcare(sidebar:user)?nav_parent=home_screen(sidebar:user)&nav_parent_triggered_by=unlocked_selfcare. Seems to work fine though, just not what I expected!

Yeah the navigation system in general is not very friendly for urls. The double reference is needed to track keeping the sidebar open when returning back to the previous page. I'd say this is something we're unlikely to want to dive into too much at this point, but will want to keep in mind in the future if we doing a larger overhaul on the navigation system

chrismclarke:
@jfmcquade 
> I might suggest moving the order of tables around on the page, as I suspect the Dynamic Data table is going to be more commonly useful than Contact Fields or Protected Fields, and currently I have to scroll to view it.

Good idea, updated in 414821edb7d0893c8f4df7ff7102273325a4a2c2 - you can mostly ignore the diffs, git is trying to be too clever and storing partial changes instead of just the full cut/paste I did

I've also pushed some changes to support multi-window support on the debug page in 
[01724b2](https://github.com/IDEMSInternational/open-app-builder/pull/2839/commits/01724b29f0f54116613ba1ef106acddd2b098c57)

I've requested re-review, although the only code to be aware of is the change in the commit mentioned (see demo in notes in previous comment)

chrismclarke:
> Thanks for the feedback @esmeetewinkel
> 
> > Functional test passed for amending existing data rows. A few comments (non of which essential to be included in this PR):
> > 
> > 1. While testing I found that rows that are dynamically created don't show on the `/user` page. See for instance the debug template `feat_data_actions_edit`
> 
> Should now be fixed
> 
> > 2. When the `(sidebar:user)` is active, it could be nice to be able set the screen width of the LHS of the screen to a reasonable phone dimension, e.g. similar to `(sidebar:feedback)`:
> 
> > 3. Pop-ups and stacked templates appear in front of the sidebar. I believe that's a limitation we've been aware of from the start - but it's not great in the use case where I wanted to have a look at how the dynamic data changes as I'm navigating through a module, which is a stacked template...
> 
> I've attempted to address this issue in part by enabling the debug page to keep in sync when multiple browser tabs/windows are active. That way you can simply use multiple windows to resize as needed and won't come across issues with full-screen content.
> 
> See attached demo:
> 
>  Screenity.video.-.Mar.13.2025.mp4 
> Should note that this only appears to content on the debug page. If we want all dynamic data to stay in live-sync more generally across multiple tabs (i.e. without need to reload tabs) that would need to be a follow-up issue and PR
> 
> > 4. I noticed that, when navigating, `(sidebar:user)` shows up in multiple places in the hyperlink, not only with the current template but also with the `nav_parent` query parameter. E.g. http://localhost:4200/template/module_landing_selfcare(sidebar:user)?nav_parent=home_screen(sidebar:user)&nav_parent_triggered_by=unlocked_selfcare. Seems to work fine though, just not what I expected!
> 

Yeah the navigation system in general is not very friendly for urls. The double reference is needed to track keeping the sidebar open when returning back to the previous page. I'd say this is something we're unlikely to want to dive into too much at this point, but will want to keep in mind in the future if we doing a larger overhaul on the navigation system

I've requested re-review so that you can test the updates and decide whether any of the outstanding issues are worth follow-up issue or not.

esmeetewinkel (approved):
Functional test passed. 

image

(nit) I noticed that the `row_index` and `_user_created` columns don't show their default values. It might make sense to add this, but also happy to merge as is.

jfmcquade (approved):
Looks good.

Interesting to see the changes in https://github.com/IDEMSInternational/open-app-builder/pull/2839/commits/01724b29f0f54116613ba1ef106acddd2b098c57. I don't see a need for keeping dynamic data in sync across multiple tabs/windows in general (it may even be desirable not to), but nice to have it implemented for the user debug page.

chrismclarke:
> Looks good.
> 
> Interesting to see the changes in [01724b2](https://github.com/IDEMSInternational/open-app-builder/commit/01724b29f0f54116613ba1ef106acddd2b098c57). I don't see a need for keeping dynamic data in sync across multiple tabs/windows in general (it may even be desirable not to), but nice to have it implemented for the user debug page.

Yeah I was interested to see while looking into the RXDB code that they already had a system of broadcast events in place to handle sync across tabs. I think that given multiple tabs/browser windows still access the same indexeddb where rxdb stores data it probably would make sense to keep in sync more generally, given that they will always be in sync after page refresh anyways (so would just be slightly more consistent behaviour) - but definitely wouldn't see this as a high priority issue to work on given the changes it would need to how we manage the in-memory system that writes to the cache system (would now need a 2-way relationship where the in-memory can also be updated from the indexeddb cache)

PR #2842, opened 2025-03-11
fix: ios visual bugs on elements with drop-shadow

jfmcquade:
PR Checklist

## Description

- Fixes an issue with the `plh_module_list_item`, where the component would not display properly on iOS (see https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/issues/174).
- Fixes related undocumented issues for some other components:
  - `task_card`
  - `button` (`card-portrait` variant)

## Notes

The underlying issue is with the CSS `filter: drop-shadow` on iOS. Similar issues are noted in various places (e.g. [1](https://stackoverflow.com/questions/65801487/drop-shadow-not-working-properly-on-ios-safari), [2](https://www.reddit.com/r/css/comments/bweqdc/safari_bug_css_dropshadow_after_elementhover_issue/)), although not well documented. 

There are two possible workarounds for this issue:
1. Refactor the styling to replace `filter: drop-shadow` with `box-shadow:`
    - The latter does not seem to cause issues on iOS
    - This change does entail some subtle visual changes, since the shadow applied by the `drop-shadow` filter appears to be softer and more blurred than that applied by the `box-shadow` property
2. Make use of the workaround we introduced for similar issue in #2288 – apply the CSS mixin designed for troublesome iOS elements. 

Testing the components that make use of the `drop-shadow` filter, `plh_module_list_item`, `task_card`, and `button` with `card-portrait` variant all exhibited related bugs on iOS. `parent_point_box` and `combo-box` also use a `drop-shadow` filter, and whilst I wasn't able to observe any bugs on these components, I've applied the workaround in anticipation of such bugs.

### `plh_module_list_item`

In the case of the `plh_module_list_item` component for which the issue was noticed, it made most sense to apply option 1., since the other variants of this component already made use of `box-shadow`, so this change improves consistency as well as resolving the issue. See screenshots section for example of the changes.

### `task_card`

For this component, on iOS the shadow around the status badge doesn't appear when overlapping with the main card element. I've resolved this with workaround 1. There is a slight visual difference (the shadow around the card is slightly less blurred), but I don't think this is a problem.

| before | after |
|---|---|
| Screenshot 2025-03-11 at 12 19 19 | Screenshot 2025-03-11 at 12 22 30 |

### `button` (`card-portrait` variant)

I applied workaround 1. As well as fixing this flickering issue, this styling brings the component in line with the shadow for the updated task-card component for a closer visual match, as well the other button variants, which already use `box-shadow`.

Before:

https://github.com/user-attachments/assets/824ebafb-8226-402d-a6e7-c88f47f9ea92

After:

Screenshot 2025-03-11 at 15 28 51

### `parent_point_box`

I couldn't detect any issues, but applied workaround 1. 

### `combo-box`

This is a case where the difference between a `drop-shadow` filter and `box-shadow` is more pronounced, because of the way they interact with transparency. Since the button element consists of just an outline, using a `drop-shadow` filter means the element gets a shadow on the inside as well as the outside, unlike a `box-shadow` which is just applied to the outside. Whilst this change isn't necessarily undesirable, I've opted to implement workaround 2. to preserve the existing look.

Here's the difference that workaround 1. would apply:

| current | potential change (not applied) |
|---|---|
| Screenshot 2025-03-11 at 12 24 15 | Screenshot 2025-03-11 at 12 23 56 |

## Git Issues

Closes https://github.com/ParentingForLifelongHealth/plh-kids-app-kw-content/issues/174

## Screenshots/Videos

 component in situ within [comp_plh_progress_path](https://docs.google.com/spreadsheets/d/130xy0FbCqgVMg8astiR5FBzuUMfVBHtg9k6wmweE1BA/edit?gid=603128459#gid=603128459):

| before | after |
|---|---|
| Screenshot 2025-03-11 at 12 01 44 | Screenshot 2025-03-11 at 12 00 15 |

### DIFF
--- a/packages/components/plh/parent-point-box/parent-point-box.component.scss
+++ b/packages/components/plh/parent-point-box/parent-point-box.component.scss
@@ -17,7 +17,7 @@ $item-border: var(--points-item-border, none);
     max-width: 100%; // Don't allow overlap
     height: var(--points-item-height);
     background: $item-background;
-    filter: drop-shadow(var(--ion-default-box-shadow));
+    box-shadow: var(--ion-default-box-shadow);
     border: 1px solid $item-border;
     border-radius: var(--ion-border-radius-standard);
     position: relative;
--- a/packages/components/plh/plh-kids-kw/components/module-list-item/module-list-item.component.scss
+++ b/packages/components/plh/plh-kids-kw/components/module-list-item/module-list-item.component.scss
@@ -90,12 +90,12 @@ $highlighted-color: var(--module-list-item-color-highlighted, var(--ion-color-se
   --circle-width: 84px;
   &[data-locked~="true"] {
     --circle-width: 76px;
+    box-shadow: -2px 2px 5px rgba(0, 0, 0, 0.2);
   }
   width: var(--circle-width);
   height: var(--circle-width);
   border-radius: 50%;
   border-width: 4px;
-  filter: drop-shadow(-2px 2px 5px rgba(0, 0, 0, 0.2));
   overflow: hidden;
   background-color: white;
   position: relative;
--- a/src/app/shared/components/template/components/button/button.component.scss
+++ b/src/app/shared/components/template/components/button/button.component.scss
@@ -86,7 +86,6 @@ ion-button[data-variant~="card"] {
   border-radius: var(--ion-border-radius-standard);
   min-height: var(--buttons-short-height);
   box-shadow: var(--ion-default-box-shadow);
-  // filter: drop-shadow(var(--ion-default-box-shadow));
   display: flex;
   flex-direction: column;
   align-items: center;
--- a/src/app/shared/components/template/components/button/button.component.scss
+++ b/src/app/shared/components/template/components/button/button.component.scss
@@ -85,7 +85,8 @@ ion-button[data-variant~="card"] {
   border: 1px solid var(--ion-color-gray-light);
   border-radius: var(--ion-border-radius-standard);
   min-height: var(--buttons-short-height);
-  filter: drop-shadow(var(--ion-default-box-shadow));
+  box-shadow: var(--ion-default-box-shadow);
+  // filter: drop-shadow(var(--ion-default-box-shadow));
   display: flex;
   flex-direction: column;
   align-items: center;
--- a/src/app/shared/components/template/components/task-card/task-card.component.scss
+++ b/src/app/shared/components/template/components/task-card/task-card.component.scss
@@ -6,7 +6,7 @@
   border-radius: var(--ion-border-radius-secondary);
   padding: var(--regular-padding);
   max-width: 400px;
-  filter: drop-shadow(var(--ion-default-box-shadow));
+  box-shadow: var(--ion-default-box-shadow);
 
   // Variants should be set using "variant" param, but support legacy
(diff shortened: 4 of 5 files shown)


chrismclarke (approved):
Thanks @jfmcquade , really nice breakdown of the issue and set of solutions.

Given the sources referenced I agree that `drop-shadow` should likely be avoided where possible given the inconsistencies on ios safari, preferring to use `box-shadow` instead. I also found a pretty good running thread on the issue and workarounds from the mdn browser-compat repo
```
https://github.com/mdn/browser-compat-data/issues/17726
```

Looking at the following css-tricks article, it seems the main difference is around handling of transparent content (which you also noted), but nice that we already have the previous workaround you implemented to force gpu-acceleration for those occasions where drop-shadow is definitely required
https://css-tricks.com/breaking-css-box-shadow-vs-drop-shadow/

I'm wondering whether it might be worth considering adding CSS linting to help avoid similar issues regressing in the future, by supplying a lint warning for the specific drop-shadow filter function? Looks like it should be possible with stylelint
https://stylelint.io/user-guide/rules/function-disallowed-list

But in the short term I think it makes a lot of sense to simply manually migrate all instances. Can confirm the only instance of drop-shadow now in use is the one you have manually fixed
image

esmeetewinkel (approved):
Looks good from description and screenshots/videos. If you want me to test anywhere (not sure if the issue was present on Appetize?)  feel free to add a link. 

One change I observed in your `task-card` screenshot that to me is more noticeable than the more-or-less blurred shadow, is that circle with tickmark icon in the top-right corner now also has a shadow in front of the card, rather than only "behind" the card. Therefore it now looks more like the circle is floating in front of the card which might suggest it is clickable itself. But I think that difference is still acceptable. 
image

chrismclarke:
> Thanks, @chrismclarke. From the [thread you linked](https://github.com/mdn/browser-compat-data/issues/17726#issuecomment-1373163624), it seems our SCSS mixin is likely doing something very similar to tailwind's `transform-gpu` class.
> 
> > I'm wondering whether it might be worth considering adding CSS linting to help avoid similar issues regressing in the future, by supplying a lint warning for the specific drop-shadow filter function? Looks like it should be possible with stylelint https://stylelint.io/user-guide/rules/function-disallowed-list
> 
> I like this idea. I just had a look into it, I was able to set up stylelint with a custom rule. This works when running manually, e.g. `yarn stylelint src/app/shared/components/template/components/combo-box/combo-box.component.scss`, however I struggled to get this working as real-time in-editor linting, or integrate it with eslint. I have also noticed in the past that eslint is responsible for a lot of the high memory usage from VSCode on my machine. So I think that additionally linting all SCSS files in real-time is probably not worth it at the moment for this one gain. I've made a quick issue to track this here (#2843), although don't see it being a priority.

Did you try installing the stylelint extension for vscode? As far as I'm aware it's independent to eslint so would likely need its own extension to show UI markup in vscode. 
https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint

Definitely can appreciate vscode memory hog issues, I loose track myself although at least the footprint would only increase if dev chooses to install the extension. I think the larger benefit would just be including the stylint test as part of our CI lint/test just to make sure the rule doesn't accidentally work its way into production, but no reason for any of this to hold up this PR

Issue #2845, opened 2025-03-13
[FEATURE] plh_kids_tz: styling of module buttons on homepage


jfmcquade:
**What?**
The new design for the homepage of the `plh_kids_tz` deployment includes two views on a list of modules. Additional functionality is needed to handle the proposed button styling:

![Image](https://github.com/user-attachments/assets/9bd328c3-544f-4f8c-a974-c33ea1ff35c0) 

![Image](https://github.com/user-attachments/assets/61bcc826-1530-4062-b104-7554695f0e3c)

**Why?**
See https://github.com/ParentingForLifelongHealth/plh-kids-app-tz-content/issues/161

**How?**
It seems sensible to reuse the `plh_module_list_item` component introduced for the `plh_kids_kw` deployment. This component has recently been updated, see #2805. This means that it can now handle both the list/map view styling via the `default`/`circle` variants.

### Initial styling changes
- [ ] Move `plh_module_list_item` component from `plh_kids_kw` scope to parent `plh` scope
- [ ] Styling tweaks to both variants
  - e.g. border width; shadow/glow handling
  - Some of these styles should be scoped as overrides of the `plh_kids_tz` theme, although it may make sense to make some the defaults of the component, and move some existing styles to the `plh_kids_kw` theme.
- [ ] Apply theme colours from `plh_kids_tz` theme for default and highlighted states 

### Possible follow-up/stretch goal

chrismclarke:
> NB: the yellow/pink/green styling signifies modules that are unlocked and default/highlighted/completed respectively.

I'm guessing this has already gone through design phase, but just to check whether the warning visual in the pink version is the intended indicator for highlighted task. To me that feels a little bit confusing (both you are meant to click this and you are not)

I might also suggest making a bit more use of border radius to distinguish between highlighted/non. I.e. I could imagine completed, locked and unlocked all having similar, slightly smaller border radius, whereas highlighted could have slightly larger and/or a box-shadow/glow around it to draw emphasis

jfmcquade:
> I'm guessing this has already gone through design phase, but just to check whether the warning visual in the pink version is the intended indicator for highlighted task. To me that feels a little bit confusing (both you are meant to click this and you are not)

I agree that it's confusing. It might be better if it's a round badge rather than triangular, which wouldn't have the association of a stop/warning symbol. The icon itself could be populated by an author supplied icon via a param exposed to authoring.

[...]

PR #2851, opened 2025-03-17
feat: plh_module_list_item 'completed' state

jfmcquade:
PR Checklist

## Description

Adds a new param for the `plh_module_list_item` component: `completed`. This takes a boolean value and, if true, applies custom styling to represent a "completed" state.

By default, the completed state makes colours the component instance green, taking the green colour from the theme-level green (all themes include a green and red colour). This can be overridden for a given theme either by specifying the theme's `green` colour globally, or by specifying a theme-level override for the component's `module-list-item-color-completed` variable.

This means that the component now has three possible "states": default, completed, highlighted and locked. These are independent to each other, and take precedence in the order listed (i.e. the locked state will take precedence over the highlighted state, etc.). For the immediate use case, I believe that the only two states that could co-exist by design are completed and highlighted. For more detail, see #2845.

## Git Issues

Addresses a task from #2845

## Screenshots/Videos

[comp_plh_module_list_item](https://docs.google.com/spreadsheets/d/1Tq8Ib_ZcJsABW9U9XbQxgnmtv_uayXk28DXoa94j1aE/edit?gid=569531329#gid=569531329):

https://github.com/user-attachments/assets/fdbcd1ad-f8fb-4e77-90a8-123355b0072e

Screenshot 2025-03-17 at 15 27 22


esmeetewinkel (approved):
Functional test passed on `plh_kids_tz` deployment

image

PR #2854, opened 2025-03-18
chore: rename some app config properties; add deprecation migration system

jfmcquade:
PR Checklist

## Description

- Renames some app config properties for consistency. After merge, the new names should be used going forward.
  - `APP_FOOTER_DEFAULTS.templateName` -> `APP_FOOTER_DEFAULTS.template`
  - `APP_HEADER_DEFAULTS.colour` -> `APP_HEADER_DEFAULTS.background`
- Introduces a way to handle renaming app config properties such as these in such a way as to continue to support deprecated names

There should be no functional changes. 

## Testing

Test all templates within [feature_app_layout](https://docs.google.com/spreadsheets/d/1Z46gTluTPeffqpiAZMapYo0imm59GU03j0sW9UChvPM/edit?gid=3937162#gid=3937162), e.g. [feat_app_layout_custom_footer](https://docs.google.com/spreadsheets/d/1Z46gTluTPeffqpiAZMapYo0imm59GU03j0sW9UChvPM/edit?gid=569531329#gid=569531329), to ensure header and footer overrides are still functioning as expected.

## Git Issues

Closes #

## Screenshots/Videos

Highlighting of deprecated properties in deployment config (`debug` deployment):

Screenshot 2025-03-18 at 15 14 25

### DIFF
--- a/packages/data-models/appConfig.ts
+++ b/packages/data-models/appConfig.ts
@@ -233,7 +233,12 @@ const APP_CONFIG = {
 };
 
 /**
- * Deprecation migrations to support renamed propertiesfor backwards compatibility
+ * List of deprecated app_config properties
+ * These will be replaced at runtime if applied as config overrides
+ *
+ * Mappings should be maintained until methods exist in the parser to ensure authored
+ * configurations are valid. This will require updates to check for template-level
+ * config overrides (deployment and skin configs are otherwise type-checked)
  */
 const DEPRECATION_RULES = {
   APP_FOOTER_DEFAULTS: {
@@ -250,9 +255,7 @@ const DEPRECATION_RULES = {
   },
 };
 
-/**
- * Applies deprecation migrations to the app config, converting deprecated properties to their new equivalents
- */
+/** Convert any deprecated properties to their new equivalents */
 export function applyAppConfigDeprecations(config: RecursivePartial<IAppConfig>) {
   Object.entries(DEPRECATION_RULES).forEach(([configSection, rules]) => {
     const sourceSection = config[configSection];
--- a/packages/data-models/appConfig.ts
+++ b/packages/data-models/appConfig.ts
@@ -84,7 +84,7 @@ export type IHeaderFooterBackgroundOptions = "primary" | "secondary" | "none";
 /** The "compact" variant reduces the header height and removes the title */
 export type IHeaderVariantOptions = "default" | "compact";
 
-interface IAppConfigHeaderBase {
+interface IAppConfigHeader {
   back_button: {
     hidden?: boolean;
   };
@@ -97,10 +97,6 @@ interface IAppConfigHeaderBase {
   template: string | null;
   title: string;
   variant: IHeaderVariantOptions;
-}
-
-/** Include deprecated properties for backwards compatibility */
-interface IAppConfigHeader extends IAppConfigHeaderBase {
   /** @deprecated, use "background" instead */
   colour?: IHeaderFooterBackgroundOptions;
 }
@@ -125,13 +121,9 @@ const activeRoute = (location: Location) => {
   return path;
 };
 
-interface IAppConfigFooterBase {
+interface IAppConfigFooter {
   background: IHeaderFooterBackgroundOptions;
   template: string | null;
-}
-
-/** Include deprecated properties for backwards compatibility */
-interface IAppConfigFooter extends IAppConfigFooterBase {
   /** @deprecated, use "background" instead */
   templateName?: string | null;
 }
--- a/packages/data-models/appConfig.ts
+++ b/packages/data-models/appConfig.ts
@@ -97,7 +97,7 @@ interface IAppConfigHeader {
   template: string | null;
   title: string;
   variant: IHeaderVariantOptions;
-  /** @deprecated, use "background" instead */
+  /** @deprecated use "background" instead */
   colour?: IHeaderFooterBackgroundOptions;
 }
 
@@ -124,7 +124,7 @@ const activeRoute = (location: Location) => {
 interface IAppConfigFooter {
   background: IHeaderFooterBackgroundOptions;
   template: string | null;
-  /** @deprecated, use "template" instead */
+  /** @deprecated use "template" instead */
   templateName?: string | null;
 }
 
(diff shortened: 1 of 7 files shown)


chrismclarke (changes requested):
Thanks for tidying these up

The name changes look good to me, I also like how you have retained the original properties and marked as deprecated to provide backwards compatibility.

The migration method is also really nice to have, although I think it would be better if we can move it to the scripts so that it is run just once when deployment is compiled instead of at runtime.

The deployment compilation system is a little bit messy/confusing, but I think could likely try to add the `applyAppConfigDeprecations` method to `packages\scripts\src\commands\deployment\compile.ts` just before cleaning the config via
```ts
// remove empty placeholders populated by override config
  converted.app_config = cleanEmptyObject(converted.app_config);
```

packages/data-models/appConfig.ts
```diff
@@ -119,9 +125,20 @@ const activeRoute = (location: Location) => {
   return path;
 };
 
-const APP_FOOTER_DEFAULTS = {
-  templateName: null as string | null,
-  background: "primary" as IHeaderFooterBackgroundOptions,
+interface IAppConfigFooterBase {
```
chrismclarke, line 128: nit(non-blocking): I'm not sure if we really need to keep two type definitions for base and extending with deprecated. I think you can probably just define a single config interface that includes the marked deprecations - I think probably a bit easier to mentally parse if everything is in one place

packages/data-models/appConfig.ts
```diff
@@ -223,6 +240,44 @@ const APP_CONFIG = {
   TASKS,
 };
 
+/**
+ * Deprecation migrations to support renamed propertiesfor backwards compatibility
+ */
+const DEPRECATION_RULES = {
```
chrismclarke, line 246: nit(non-blocking): really nice to have the set of rules and clear authoring messages. I think there might be some minor issue if we need to change entirely either a root property name (e.g. `APP_FOOTER_DEFAULTS -> FOOTER`) or if we had more deeply nested structures. 

I'm guessing the solution would be flattening the entire config and deprecations so that you could define
```ts
{
  "APP_FOOTER_DEFAULTS.templateName":{
    newProperty: "APP_FOOTER_DEFAULTS.template"  
  }
}
```
however I'm also happy to just deal with this when the need arises

jfmcquade:
Thanks for the review @chrismclarke 
 
> The migration method is also really nice to have, although I think it would be better if we can move it to the scripts so that it is run just once when deployment is compiled instead of at runtime.

The motivation for applying the migrations at runtime is that they then apply to any config overrides applied at runtime – i.e. overrides from the skin and template levels as well as the deployment level. If we move the logic to the deployment compilation step, then this will be a breaking change that will require updating any skins/templates that override these properties, e.g. the template-level app config overrides in the [feature_app_layout](https://docs.google.com/spreadsheets/d/1Z46gTluTPeffqpiAZMapYo0imm59GU03j0sW9UChvPM/edit?gid=3937162#gid=3937162) debug examples, and any cases in production deployments. 

I'm not sure which I'd prefer: it would be nice to not introduce breaking changes (which would be the case every time we introduced any more such migrations in the future), but it is also suboptimal to run these migrations every time the app config is updated at runtime. Let me know what you think, I'm happy to move the logic if we think that's better on balance.

chrismclarke:
> The motivation for applying the migrations at runtime is that they then apply to any config overrides applied at runtime – i.e. overrides from the skin and template levels as well as the deployment level. If we move the logic to the deployment compilation step, then this will be a breaking change that will require updating any skins/templates that override these properties, e.g. the template-level app config overrides in the [feature_app_layout](https://docs.google.com/spreadsheets/d/1Z46gTluTPeffqpiAZMapYo0imm59GU03j0sW9UChvPM/edit?gid=3937162#gid=3937162) debug examples, and any cases in production deployments.
> 
> I'm not sure which I'd prefer: it would be nice to not introduce breaking changes (which would be the case every time we introduced any more such migrations in the future), but it is also suboptimal to run these migrations every time the app config is updated at runtime. Let me know what you think, I'm happy to move the logic if we think that's better on balance.

Ah okay, that makes a lot of sense, forgot about the template-level override system. I've added an extra comment to the code to make more explicit in e2005b4

If we wanted to provide a more balanced solution right now, likely it would be to only check for deprecation mappings when applying a template-level override (we track the override `source` level so that would be fine to do), and rely on type-checking the build to manage deployment and skin level. However I don't think any of these operations are too costly to do at runtime so probably not worth the semi-optimisation work.

I think instead it would be better to create a follow-up issue that can pick up potential config errors from template overrides, so that we can at least eventually remove deprecations and be confident that build will fail instead of invalid config making its way into runtime. At that point we can also hopefully move all overriding to parser-level.

chrismclarke (approved):

packages/data-models/appConfig.ts
```diff
@@ -119,9 +121,16 @@ const activeRoute = (location: Location) => {
   return path;
 };
 
-const APP_FOOTER_DEFAULTS = {
-  templateName: null as string | null,
-  background: "primary" as IHeaderFooterBackgroundOptions,
+interface IAppConfigFooter {
+  background: IHeaderFooterBackgroundOptions;
+  template: string | null;
+  /** @deprecated, use "background" instead */
```
esmeetewinkel, line 127:
```suggestion
  /** "templateName" is @deprecated, use "template" instead */
```

packages/data-models/appConfig.ts
```diff
@@ -88,21 +88,23 @@ interface IAppConfigHeader {
...
   collapse: boolean;
-  colour: IHeaderFooterBackgroundOptions;
   hidden?: boolean;
   menu_button: {
     hidden?: boolean;
   };
   template: string | null;
   title: string;
   variant: IHeaderVariantOptions;
+  /** @deprecated, use "background" instead */
```
esmeetewinkel, line 100:
```suggestion
  /** "colour" is @deprecated, use "background" instead */
```

esmeetewinkel (approved):
Functional test passed, deprecated properties are still working. 

Can confirm seeing console log warnings when using deprecated properties
image

I can also see the deprecated properties being crossed out in the config, although I don't find the definition reference too insightful: why does it tell me about `background` when I'm looking at something to do with the deprecation `template`/`templateName`? Also it's not super clear which line (i.e. property) the comment relates to.

image

PR #2859, opened 2025-03-20
feat(campaigns): provide config for opt-out

### LINKED ISSUE #2265: Notifications as opt-in/out feature

chrismclarke:
PR Checklist

## Description
Provide deployment config option `campaigns.enabled` to allow _opt-out_ of campaigns system

Due to the way various other services interact with the campaigns system, it's not currently possible to fully remove the service, and so the console will still show the system as initialised (albeit resolving much more quickly than otherwise would). Opting out should at least prevent user permission prompts for notifications.

## Review Notes
If running a regular deployment that makes use of campaigns and notifications should check things still function as expected (e.g. notification prompt, campaign/notification debug pages)

To test behaviour opted out should pick a deployment that doesn't make use of campaigns and update app config accordingly:
```
config.campaigns.enabled = false
```
Expected that if running in web preview should not prompt for notification permission (not applicable when running in localhost as permissions never requested)

## Author Notes
The main purpose of this feature is to provide a short-term bypass to avoid notification permission prompts for deployments not making use of notifications (as currently the only means of interacting with notifications in the app is via campaigns).

In the future we would likely want to auto-detect whether the feature is required and provide optimisation to auto opt-in/out accordingly. 

## Dev Notes
This originally started as a local-notifications opt out, however I realised this could lead to lots of errors if campaigns still try to schedule. As such the focus here is on making the campaigns system more modular for easier opt-in and out, with various knock-on effects

**Local Notification Interaction Service** 
This was previously created as a global service, but only serves the purpose of storing local notification history in a database for use in debugging. As such it's been renamed to the `LocalNotificationPersistAdapter` and accessed as a child class of the local notification service instead of a global service. This makes it easier to decouple other dependencies

**App Initialisation**
Given that the only way to interact with notification is through the campaigns system, hardcoded calls to init notifications within the app.module have been removed. Similarly hardcoded calls to init the campaign service have been removed, and placed within the campaign module instead with deployment config check. The purpose of refactoring init into the module is for easier future optimisation where the module import can be commented out if not in use. 

A knock-on effect of this is that the module constructor code is not async, so the init method will fire but not block anything else happening in the app. This isn't much of an issue as any methods that call the service typically have their own logic to check it has been initialised (e.g. evaluating `@campaign.some_value` from template variable service), however there will be a small warning coming from the `template-processor` service to say that it has injected a service that is not yet ready

The benefit of this is saving time it typically takes to load the campaign service and avoiding prompt to allow notifications when not in use

**Service Dependencies**
It's not currently possible to consider removing the campaign service entirely from the app due to inter-service dependencies
image

So would likely need future work to decouple, namely

- [ ] Template variable service - Add custom registry to allow services to register their own variable namespaces instead of relying on service as source of truth (like custom action registry). This would let the campaign service register it's own processor for `@campaign` statements

- [ ] Template process service - Create `campaign` action namespace in campaign action registry to handle re-evaluation instead. This could either be triggered by authors in any start-up templates that require, or just refactor existing code to trigger standalone action (and allow to silently fail if campaign service not registered)

- [ ] Task service - Likely same as above, triggering the template action service to call campaign re-evaluation instead of calling service directly 

**Further Improvements**
Looking more broadly at the campaigns system, a few follow-up tasks to improve management would likely include:

- [ ] Much of the campaign system code would benefit from use of signals, e.g. for debug page variables or improved traceability of computed effects 

- [ ] Parser methods to identify whether campaign system in use (either from campaign flows or `@campaign` references), and ability auto comment-out module if not required (like components). Would also require include either decoupling or optimising all other services that call the campaign service (e.g. task service, template-process service, template-variable service)

## Git Issues

Closes #2265

## Screenshots/Videos
**Example - warning now present but can be safely ignored**
image

**Before**
Campaigns system initialising on load - typically the last blocking service to load
image

**After**
Campaigns system still marked as initialised, although resolves near-instantly
image


esmeetewinkel:
I believe this closes https://github.com/IDEMSInternational/open-app-builder/issues/2265

packages/data-models/deployment.model.ts
```diff
@@ -45,6 +45,15 @@ export interface IDeploymentRuntimeConfig {
     /** prevent user accessing app pages without being logged in. Specified template will be shown until logged in */
     enforceLoginTemplate?: string;
   };
+  campaigns: {
+    /**
+     * Specify whether campaigns notification used by deployment. Default `true`
+     *
+     * NOTE - if disabled functionality will still enable if explicitly called from templates (e.g. `@campaign` authoring)
```
jfmcquade, line 52: Is this still true? It looks like the campaign service won't initialise properly in this case

jfmcquade (approved):
Looks good. I've added one query on a comment inline.

It could be good to make clear in the PR description that this doesn't affect the permissions that native apps will register with their respective app store platforms (and possibly include as a future TODO – I thought we might have an open issue for this in general but I can't find one)

packages/data-models/deployment.model.ts
chrismclarke: Good catch, thanks. No it is no longer true - I realised after the fact that the task and template-process force init on startup anyways, so no use in attempting to defer until needed. Tidied up with a8fade8cc2b03fa7249750f8f55519d31aef0667

chrismclarke:
@esmeetewinkel 
I've added a web preview to this PR just to test default state of campaigns/notifications still working as expected. I'm not really sure what's the best way to go about testing the opt-out functionality (would require a pr preview for a deployment that is opted out of the config and targeting this build branch) - so might be easier to test on deployment repo after if there are any deployments that don't make use of campaigns and also target the master branch for releases. 

Alternatively, we could create a new builder release post-merge for deployment repo to target.

esmeetewinkel (commented):
## With notifications

On `debug` deployment PR preview link (has notifications):
- `Campaigns (1053ms)`

image

## Without notifications
On `plh_kids_tz` deployment [PR](https://github.com/ParentingForLifelongHealth/plh-kids-app-tz-content/pull/178) preview link (notifications disabled in config, running off this PR branch):

- `Campaigns (6ms)`
- Still received prompt for notification permission (contradicting statement below)

> Expected that if running in web preview should not prompt for notification permission (not applicable when running in localhost as permissions never requested)

image

chrismclarke:
> ## Without notifications
> On `plh_kids_tz` deployment [PR](https://github.com/ParentingForLifelongHealth/plh-kids-app-tz-content/pull/178) preview link (notifications disabled in config, running off this PR branch):
> 
> * `Campaigns (6ms)`
> * Still received prompt for notification permission (contradicting statement below)
> 
> > Expected that if running in web preview should not prompt for notification permission (not applicable when running in localhost as permissions never requested)
> 

Thanks for checking this. Very strange indeed, yes I can also see the notifications are still being initialised even though explicitly opting out. I can't fully figure out the reason why (could be a hard-to-trace side-effect somewhere else, or some sort of JS quirk I've missed), but in any case I think I've managed to fix with a small workaround for now and expect should also not be an issue in the future if addressing the items raised for follow-up (as that addresses a number of potential side-effect issues).

I've re-run the same preview action on the test pr - you should be able to try again in a new private browsing window and see the notification prompt no longer appearing.

PR #2862, opened 2025-03-21
feat: include firebase google auth requirements in ios configure workflow

jfmcquade:
PR Checklist

## Description

In order to handle Google auth, iOS apps are required to declare a URL that will be used to reopen the app as part of the Google login flow. This is achieved by including the URL it in a list of "custom URL schemes" declared in the iOS manifest file, `Info.plist`.

This is available through the Firebase-specific `GoogleService-Info.plist` file, but I believe this `REVERSED_CLIENT_ID` URL would be required when integrating Google sign in via any auth provider.

This PR also includes the updated cocoapods configuration from #2834.

## Testing

There isn't too much to test here beyond my testing of the ios builds. I've uploaded a debug build to appetize [here](https://appetize.io/app/ios/international.idems.debug-app?device=iphone14pro&osVersion=16.2&toolbar=true), built from this branch, for which the login flow is working (navigate to template `example_auth` to test).

## Git Issues

Closes #2834

### Deployment without auth enabled

Deployments without an auth provider (here I used the debug deployment and commented out the relevant lines in the config) populate this property with an empty string on `yarn workflow ios configure`:

Screenshot 2025-03-21 at 16 29 14

In this case, the app still builds and runs successfully:

Screenshot 2025-03-21 at 16 33 21

### Deployment with auth enabled

Here I used the debug deployment, and finished configuring Google Auth from the firebase console. I've updated the `GOOGLE_SERVICES_PLIST ` secret on the [debug content repo](https://github.com/IDEMSInternational/app-debug-content/settings/secrets/actions). 

When running the ios configure workflow, the new var is populated:

Screenshot 2025-03-24 at 11 36 16

And the login flow functions as expected:

Screenshot 2025-03-24 at 11 38 02

### `plh_kids_kw` deployment

The login flow functions as expected when built for iOS simulator from this branch:

Screenshot 2025-03-24 at 11 19 47


esmeetewinkel (approved):
Functional test passed.

documentation/docs/developers/building-and-deploying-to-ios.md
```diff
@@ -33,17 +33,17 @@ If the deployment has not previously been built for iOS, follow these steps
...
 4. Run `yarn build`
 5. Run `npx cap sync ios`
 6. Run `npx cap open ios` to open the project in XCode.
-7. Open the Firebase console and fetch the relevant `GoogleService-Info.plist` file for the deployment's registered app. Copy the file to `ios/App/App/GoogleService-Info.plist`.
-8. In order to sign the app, in XCode navigate to `App > Targets > App Signing & capabilities`. Set the value for the development team, using "Add account" if you have not previously configured XCode with a development team. Ensure that the option "Automatically manage signing" is checked.
-9. Now build the app to a target simulator to test functionality.
-10. When ready to upload the app to App Store Connect, set the target as "Any iOS device (arm64)", and go to `Product > Archive`.
+7. In order to sign the app, in XCode navigate to `App > Targets > App Signing & capabilities`. Set the value for the development team, using "Add account" if you have not previously configured XCode with a development team. Ensure that the option "Automatically manage signing" is checked.
```
chrismclarke, line 42: Thanks for updating this docs, still handy to have as we slowly migrate more over to CI

ios/App/App.xcodeproj/project.template.pbxproj
```diff
@@ -359,6 +359,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = ${VERSION_CODE};
 				DEVELOPMENT_TEAM = CHLBG6683N;
+				GOOGLE_REVERSED_CLIENT_ID = "${GOOGLE_REVERSED_CLIENT_ID}";
```
chrismclarke, line 362: +1 - good to see this pattern working, will make it much easier to avoid need for hardcoded changes.

I'm also guessing if we need to opt out in the future we could just do a regex search for the templated string and remove the line when templating (as part of the env-parser script), but not a priority for now

packages/@idemsInternational/env-replace/src/index.ts
```diff
@@ -195,8 +195,9 @@ class EnvReplaceClass {
   private logSummary() {
     const tableSummary = [];
     for (const [fileName, replacements] of Object.entries(this.summary)) {
-      for (const [varName, value] of Object.entries(replacements)) {
-        tableSummary.push({ fileName, varName, value });
+      for (let [varName, value] of Object.entries(replacements)) {
+        const truncatedValue = value.length > 40 ? value.slice(0, 37) + "..." : value;
```
chrismclarke, line 199: +1 - nice idea, yeah console tables don't wrap or flow nicely at all. 

nit(non-blocking): the `let` above can go back to `const` (guessing you had originally updated the value directly and then decided to use intermediate variable which I agree is a little bit cleaner)

packages/scripts/src/tasks/providers/ios.ts
```diff
@@ -56,6 +76,43 @@ function convertToValidIOSAppId(appId: string) {
...
+/**
+ * Get deployment-specific custom URL schemes, used for deep linking.
+ * Currently just REVERSED_CLIENT_ID, required for Google Auth integration
+ * TODO: make this step optional, allowing for deployments without Google Auth integration, for example
+ */
+function getCustomUrlSchemes(authProvider: IiOSBuildOptions["authProvider"]) {
+  let GOOGLE_REVERSED_CLIENT_ID = "";
+  if (authProvider) {
+    // TODO: handle other auth providers
+    if (authProvider === "firebase") {
```
chrismclarke, line 88: nit(non-blocking): I'm not sure if this method really needs generalising right now as even if using supabase I'm pretty sure we'd need google services/plist if wanting to offer google-based sign-in, although if looking ahead I'd say might be good to throw an error if any other authProvider detected just so we know we need to review the code

chrismclarke (approved):
Really like how all these pieces have come together for this PR, feels like a tidy way to go about supporting the various configuration updates required.

See a few minor (non-blocking) comments inline, but overall I'd say good to go

PR #2836, opened 2025-03-06
fix: android build lfs and debug signing

chrismclarke:
PR Checklist

- [ ] PR title descriptive (can be used in release notes)

## Description

- Fix issue where content repos using git-lfs would not correctly pull the lfs assets needed to generate android build (splash/icons). E.g. debug content repo

- Add signing to debug builds to ensure same SHA-1 and SHA-256 generated as production builds (required for testing firebase auth on appetize)

## Review Notes
A new `Android - Debug` action has been created in draft on the debug content repo for use in testing changes to the reusable action. 
https://github.com/IDEMSInternational/app-debug-content/pull/153

See workflow runs at:
https://github.com/IDEMSInternational/app-debug-content/actions/workflows/android-debug.yml

For review see
[example 1 - failed run prior to updates](https://github.com/IDEMSInternational/app-debug-content/actions/runs/13705945603) - this is due to the android assets not checking out via git lfs as required

[example 2 - passed run using default signing](https://github.com/IDEMSInternational/app-debug-content/actions/runs/13706504225?pr=153) - fixed android assets checkout, prior to signing credential update

[example 3 - passed run with custom signing](https://github.com/IDEMSInternational/app-debug-content/actions/runs/13709024742) - this includes update to sign debug apk with same signing key as release

Despite using the environment keystore I don't actually see the fingerprints registered in the firebase console, although I'm assuming that is because the keystone has been changed and debug app not tested in this way. Although as a confirmation a different app known to be working with auth in production could also be tested

## Dev Notes
The sha fingerprint of any of the generated apks can be analyzed via command line keytool, e.g.
```sh
keytool -printcert -jarfile app-release.apk
```
For run 2 this output default fingerprints as created by default android debug keystore. If testing login feature via appetize these would need to be manually added to the firebase console for every build
```
Signature:

Owner: C=US, O=Android, CN=Android Debug
Issuer: C=US, O=Android, CN=Android Debug
Serial number: 1
Valid from: Thu Mar 06 11:32:46 PST 2025 until: Sat Feb 27 11:32:46 PST 2055
Certificate fingerprints:
         SHA1: 53:2E:D0:B5:5F:D7:xxxxxxxx
         SHA256: 4A:3B:58:51:AB:2C:xxxxxxxxxxxx
Signature algorithm name: SHA256withRSA
Subject Public Key Algorithm: 2048-bit RSA key
Version: 1
```
[...]

PR #2869, opened 2025-03-25
Feat: apple auth

jfmcquade:
PR Checklist

## Description

Enables sign in via Apple using Firebase auth.

To enable the feature for given deployment, additional configuration is required in both the associated project in the Firebase console and the associated app in the Apple Developer console. See Firebase docs, with links to relevant Apple docs, [here](https://firebase.google.com/docs/auth/ios/apple).

## Testing

- [iOS appetize link](https://appetize.io/app/b_yswli5qmzbdnltj7dmf6pdmhlq?device=iphone14pro&osVersion=16.2&toolbar=true)
- Web funcitonality can be tested by checking out this feature branch, syncing debug content, and running the app locally
  - The functionality is not currently working on web preview for me via the web preview for this PR: I get an "unauthorised domain" error, despite having added the URL to the whitelisted domains in the Authentication section of the [Firebase console for the debug project](https://console.firebase.google.com/u/1/project/debug-e8934/authentication/settings). This may be due to the fact that the hosting for the debug deployment is set to use a different Firebase project (`plh-teens-app1`) vs the one for which auth is configured (`debug`).
- Android functionality should be testable after merging https://github.com/IDEMSInternational/open-app-builder/pull/2836
  - Will need to upload a signed build to appetize and configure the associated SHA keys for the android app in the `debug` firebase project

## Follow-ups

- [ ] Create dedicated Apple Sign In button component, equivalent to #2788
  - #2871
- [ ] Enable authors to tailor content based on device platform
  - #2870
- [ ] Enable Apple Sign In for iOS apps of the specific desired deployments
  - chiefly `plh_kids_kw`, but also `plh_facilitator_mx`
- [ ] Test and debug feature on Android
  - #2876

## Git Issues

Closes #

## Screenshots/Videos

Demos use [example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705) template

Screenshot 2025-03-25 at 15 20 34

### Web

https://github.com/user-attachments/assets/5138fd8a-d743-4094-9492-ac3ff11a7568

#### emulator
Initially I wasn't signed in with my Apple ID at the level of the emulated device settings, and got this message:

https://github.com/user-attachments/assets/aaec3230-f699-43ca-acfd-e08ccee08db7

I've edited out the process of signing in at the device level, but returning to the app after doing so, I was able to sign in via the template action:

https://github.com/user-attachments/assets/3f37bc9d-7c92-4a83-9556-217d151a6196

#### appetize

Sign in success on appetize (again, I needed to sign in at the device level first):

Screenshot 2025-03-25 at 15 28 32

### DIFF
--- a/src/app/shared/services/auth/providers/firebase.auth.ts
+++ b/src/app/shared/services/auth/providers/firebase.auth.ts
@@ -1,5 +1,9 @@
 import { Injectable, Injector } from "@angular/core";
-import { FirebaseAuthentication, User } from "@capacitor-firebase/authentication";
+import {
+  AdditionalUserInfo,
+  FirebaseAuthentication,
+  User,
+} from "@capacitor-firebase/authentication";
 import { getAuth, initializeAuth, indexedDBLocalPersistence, Auth } from "firebase/auth";
 import { FirebaseService } from "../../firebase/firebase.service";
 import { AuthProviderBase } from "./base.auth";
@@ -28,15 +32,22 @@ export class FirebaseAuthProvider extends AuthProviderBase {
     await this.handleAutomatedLogin();
   }
 
+  public async signInWithApple() {
+    console.log("signInWithApple");
+    const result = await FirebaseAuthentication.signInWithApple();
+    const { user, additionalUserInfo } = result;
+    console.log("signInWithApple result", result);
+    if (user) {
+      // Note: Apple allows for anonymous sign-in so profile info may be minimal
+      this.saveUserInfo(user, additionalUserInfo);
+    }
+    return this.authUser();
+  }
+
   public async signInWithGoogle() {
     const { user, additionalUserInfo } = await FirebaseAuthentication.signInWithGoogle();
     if (user) {
-      // NOTE - additionalUserInfo is only returned on first signIn so persist to localStorage
-      // for access on automated sign-in following restart. Use fallback empty object if null
-      const { profile = {} } = additionalUserInfo;
-      localStorage.setItem(AUTH_METADATA_FIELD, JSON.stringify(profile));
-
-      this.setAuthUser(user, profile);
+      this.saveUserInfo(user, additionalUserInfo);
     }
     return this.authUser();
   }
@@ -67,8 +78,18 @@ export class FirebaseAuthProvider extends AuthProviderBase {
     const authUser: IAuthUser = {
       ...profile,
       uid: user.uid,
+      name: user.displayName,
     };
     this.authUser.set(authUser);
+  }
+
+  private saveUserInfo(user: User, additionalUserInfo: AdditionalUserInfo) {
+    // NOTE - additionalUserInfo is only returned on first signIn so persist to localStorage
+    // for access on automated sign-in following restart. Use fallback empty object if null
+    const { profile = {} } = additionalUserInfo;
+    localStorage.setItem(AUTH_METADATA_FIELD, JSON.stringify(profile));
+
+    this.setAuthUser(user, profile);
   }
 
   /**
--- a/src/app/shared/services/auth/providers/firebase.auth.ts
+++ b/src/app/shared/services/auth/providers/firebase.auth.ts
@@ -33,13 +33,12 @@ export class FirebaseAuthProvider extends AuthProviderBase {
   }
 
   public async signInWithApple() {
-    console.log("signInWithApple");
-    const result = await FirebaseAuthentication.signInWithApple();
-    const { user, additionalUserInfo } = result;
-    console.log("signInWithApple result", result);
+    const { user, additionalUserInfo } = await FirebaseAuthentication.signInWithApple();
     if (user) {
       // Note: Apple allows for anonymous sign-in so profile info may be minimal
-      this.saveUserInfo(user, additionalUserInfo);
+      const { profile = {} } = additionalUserInfo;
+      this.setAuthUser(user, profile);
+      this.saveUserInfo(user, profile);
     }
     return this.authUser();
   }
@@ -47,7 +46,9 @@ export class FirebaseAuthProvider extends AuthProviderBase {
   public async signInWithGoogle() {
     const { user, additionalUserInfo } = await FirebaseAuthentication.signInWithGoogle();
     if (user) {
-      this.saveUserInfo(user, additionalUserInfo);
+      const { profile = {} } = additionalUserInfo;
+      this.setAuthUser(user, profile);
+      this.saveUserInfo(user, profile);
     }
     return this.authUser();
   }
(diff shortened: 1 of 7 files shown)


src/app/shared/services/auth/providers/firebase.auth.ts
```diff
@@ -28,15 +32,22 @@ export class FirebaseAuthProvider extends AuthProviderBase {
...
+  public async signInWithApple() {
+    console.log("signInWithApple");
+    const result = await FirebaseAuthentication.signInWithApple();
+    const { user, additionalUserInfo } = result;
+    console.log("signInWithApple result", result);
+    if (user) {
+      // Note: Apple allows for anonymous sign-in so profile info may be minimal
+      this.saveUserInfo(user, additionalUserInfo);
+    }
+    return this.authUser();
```
chrismclarke, line 44: nit(non-blocking): a couple console logs could be cleaned up

src/app/shared/services/auth/providers/firebase.auth.ts
```diff
@@ -67,10 +78,20 @@ export class FirebaseAuthProvider extends AuthProviderBase {
...
     this.authUser.set(authUser);
   }
 
+  private saveUserInfo(user: User, additionalUserInfo: AdditionalUserInfo) {
+    // NOTE - additionalUserInfo is only returned on first signIn so persist to localStorage
+    // for access on automated sign-in following restart. Use fallback empty object if null
+    const { profile = {} } = additionalUserInfo;
+    localStorage.setItem(AUTH_METADATA_FIELD, JSON.stringify(profile));
+
+    this.setAuthUser(user, profile);
```
chrismclarke, line 92: nit(blocking): I can see how `setAuthUser` is required by both however I think better to duplicate the code in the signIn methods instead of moving it into the common function for better separation of concerns (the method would otherwise need to be called `saveUserInfoAndSetAuthUser`

Also would be good to preserve the operation order in case of knock-ons (previously was set then save, here it is save then set, which could throw error before user set if issues saving profile)

chrismclarke (changes requested):
Thanks for adding, nice to see the demos too. I can see how the apple flow is a little less user friendly given you need to input your apple id password to authorise, but guessing on some other devices this might include fingerprint scan/face id if setup so maybe isn't as big an issue. I'd still personally recommend including google sign-in option on apple devices if it makes sense to do so.

Saying that - do we have a way that authors can detect platform? If not we might want to quickly at a protected field for `@fields._is_ios` or `@fields._platform` or whatever binding makes most sense. I thought I remembered looking into something similar before related to theming but I can't actually remember now.

Otherwise see a few minor inline comments

esmeetewinkel (commented):
I can't test on Appetize since it refuses to let me create an Apple account (I'll see if there's another apple-deviceless way to create one) 
image

I've also tried running the app locally on this PR branch with synced `debug` content, however, when I click the Sign in with Apple button I'm getting an error: `ERROR FirebaseError: Firebase: Error (auth/operation-not-allowed).`
image


______________________

Update: I actually do have an Apple account because I use it to login to App Store Connect. However, even with that I'm not managing to login on Appetize

image

src/app/shared/services/auth/providers/firebase.auth.ts
jfmcquade: Yeh that makes sense about duplicating the code, my instincts said the same thing.

I believe I did originally preserve the order (save then set), but for the reason you've suggested I think it makes sense to do set then save.

See updates in https://github.com/IDEMSInternational/open-app-builder/pull/2869/commits/8bdb4c98f6386fb4e0e669a2eab48cbe66b9d3d1

jfmcquade:
> I've also tried running the app locally on this PR branch with synced `debug` content, however, when I click the Sign in with Apple button I'm getting an error: `ERROR FirebaseError: Firebase: Error (auth/operation-not-allowed).` image

Ah thanks for flagging, I missed a step. I had updated the `debug` deployment's firebase config to point to the dedicated `debug` project (for which authentication is now enabled). I've pushed that to the content repo here: https://github.com/IDEMSInternational/app-debug-content/pull/158

After running `yarn workflow deployment set debug` to pull the latest version, you should find the feature works as in my screen recording.

> Update: I actually do have an Apple account because I use it to login to App Store Connect. However, even with that I'm not managing to login on Appetize (I changed my pw just now, will try again later)
> 
> image

I had a similar issue, I was getting this error for one Apple account so tried a second one that I have and that worked. I think to the extent that this is an issue it lies with Apple and there overly stringent login criteria (I often have to do 3FA or "unlock" my account when logging in on a new device, although in this case the "username or password is incorrect" message seems plain wrong to me.


Overall, I suggest we merge this PR and plough on towards Apple sign in support on iOS, with the aim of releasing a version of the `plh_kids_kw` iOS app that includes the feature ASAP. I'll add a follow-up issue for testing and debugging the feature on Android.

esmeetewinkel (approved):
I guess I (or someone with access to an iPhone) can test further when the feature has been added, so approving here.

I'm still not able to login on Appetize - did you update that link since I last reviewed? Right now pressing the Sign in with Apple button doesn't seem to do anything at all. 

Also tried again running locally on latest debug content. Still getting the same error in a regular browser tab, in a private browser tab I get a different error (but that might just be related to private browsing).

image

image

jfmcquade:
> I'm still not able to login on Appetize - did you update that link since I last reviewed? Right now pressing the Sign in with Apple button doesn't seem to do anything at all.

Ah I think the link I included was possibly going to the latest build, which has since been usurped by a build for a different PR. Try [this link](https://appetize.io/app/b_yswli5qmzbdnltj7dmf6pdmhlq?device=iphone14pro&osVersion=16.2&toolbar=true) (I've also updated the link in the description), which is working for me. 

One issue is that if you haven't previously logged in on the device, the "Sign in with Apple" button will send you to do so, but then will not receive the updated information automatically when returning the app. If you reload the template (e.g. by navigating back within the app and selecting the template again) then it works. And if you've already logged in on the device before clicking the button, it works seamlessly. I don't think this is a major issue, as it's very unlikely that user won't have already logged in at the device level before accessing our app (they will need to have done so to download it from the app store).

> Also tried again running locally on latest debug content. Still getting the same error in a regular browser tab, in a private browser tab I get a different error (but that might just be related to private browsing).

Interesting, I suspect the issue is that the app config isn't recompiling correctly to include the updated decrypted firebase config values. Open `.idems_app/deployments/activeDeployment.json` and check the `firebase` property. The `projectId` should be `"debug-e8934"`. If it's showing `"plh-teens-app1"` then try deleting the `.idems_app/deployments/activeDeployment.json` file and running `yarn workflow deployment set debug` again to force it to recompile. Then check the `firebase` property in the newly compiled `.idems_app/deployments/activeDeployment.json`.

If it is already showing `"debug-e8934"` and you're still getting the same error then something else is the problem.

chrismclarke (approved):

esmeetewinkel:
> Try [this link](https://appetize.io/app/b_yswli5qmzbdnltj7dmf6pdmhlq?device=iphone14pro&osVersion=16.2&toolbar=true) (I've also updated the link in the description), which is working for me.

Ah thanks, on the new link I'm back to the error that my password or username is incorrect. 

> Interesting, I suspect the issue is that the app config isn't recompiling correctly to include the updated decrypted firebase config values. Open .idems_app/deployments/activeDeployment.json and check the firebase property. The projectId should be "debug-e8934". If it's showing "plh-teens-app1" then try deleting the .idems_app/deployments/activeDeployment.json file and running yarn workflow deployment set debug again to force it to recompile. Then check the firebase property in the newly compiled .idems_app/deployments/activeDeployment.json again.

You're right, however even when I force it to recompile it still shows "plh-teens-app1"

image

PR #2874, opened 2025-03-27
Review: User Import Action (#2824)

chrismclarke:
PR Checklist

## Description
Alt PR to #2824

Main changes
- Simplify user reset logic. Now includes reload after user import to simplify some of the more destructive operations
- Add debug page prompt to reset
- Add reset action spec tests

## Review Notes
Should confirm same test cases passing as in original pr

## Screenshots/Videos

### DIFF
--- a/packages/shared/src/utils/object-utils.ts
+++ b/packages/shared/src/utils/object-utils.ts
@@ -151,6 +151,5 @@ export function uniqueObjectArrayKeys(arr: Record<string, any>[], maxDepth?: num
       keyHashmap[key] = true;
     }
   }
-  console.log({ keyHashmap });
   return Object.keys(keyHashmap);
 }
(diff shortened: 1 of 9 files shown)


jfmcquade (approved):
Generally looks good, I've added some tidying with https://github.com/IDEMSInternational/open-app-builder/pull/2874/commits/5bc7a4112a99311adaa845e9e320301d47f02ec7

I like the warning pop-up on the user debug page when resetting all dynamic data.

I have found one issue. Resetting the selected data list works when accessing the user debug page through the sidebar or navigating there after making changes to a data list, but it doesn't work when accessed in another tab without reloading after the data has updated. See screen recording below:

https://github.com/user-attachments/assets/7b72e700-7252-41e9-ad12-83e7f7f121e0

Multi-window support for the user debug page was added in https://github.com/IDEMSInternational/open-app-builder/pull/2839/commits/01724b29f0f54116613ba1ef106acddd2b098c57. I've had a brief attempt at debugging but haven't managed to resolve the issue. As this multi-window feature doesn't have general support, I don't think it's crucial to resolve before merging this PR (and its parent PR), but I'm not sure how we could make it clear that the behaviour tested here isn't working as expected.

PR #2877, opened 2025-03-27
Feat: apple sign in button

### LINKED ISSUE #2871: [FEATURE] dedicated Apple Sign In button component
**What?**
Create dedicated Apple Sign In button component, equivalent to #2788 

See Apple's [docs](https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple#Displaying-buttons) for guidance

jfmcquade:
PR Checklist

- [ ] Builds on #2869, which should be merged first

## Description

Adds a new component, `apple_sign_in_button`, equivalent to the Google version introduced in #2788.

The component is styled according to [Apple's guidelines](https://developer.apple.com/documentation/sign_in_with_apple/displaying-sign-in-with-apple-buttons-on-the-web) (they offer a web API to pull a styled button from directly, however this wouldn't be compatible with our offline-first configuration). 

## Git Issues

Closes #2871

## Screenshots/Videos

[comp_apple_sign_in_button](https://docs.google.com/spreadsheets/d/1P6EhHmElOJOAoMH4h_vingVL82jyCfSTYPXj8eY2Iek/edit?gid=569531329#gid=569531329)

Screenshot 2025-03-27 at 16 15 40

Screenshot 2025-03-27 at 16 16 30


esmeetewinkel (approved):
Functional test passed. I would have maybe gone with a pill shaped button to look better together with the Google button, and it seems to be allowed by Apple (see [here](https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple#Button-size-and-corner-radius) and [here](https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple#Creating-a-custom-Sign-in-with-Apple-button)). Not sure if that would increase our chances for Apple to reject it, though. I'll make sure to display the apple sign in button first on ios :)

image  image

jfmcquade:
> ... I would have maybe gone with a pill shaped button to look better together with the Google button, and it seems to be allowed by Apple (see [here](https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple#Button-size-and-corner-radius) and [here](https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple#Creating-a-custom-Sign-in-with-Apple-button)).

I was thinking that as it's styled differently anyway (including font etc.) it might look more jarring if the buttons were similar but not identical. But I think you're right that actually the pill style would look better, should be easy to implement

esmeetewinkel (approved):

src/app/shared/components/template/components/button-apple-sign-in/button-apple-sign-in.component.scss
```diff
@@ -0,0 +1,33 @@
...
+  padding: 10px 30px 14px;
+  border-radius: 20px;
+  font-size: 16px;
+  font-weight: 600;
+  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif;
+  border: none;
+  cursor: pointer;
+  transition: background-color 0.2s ease-in-out;
+  height: 40px;
+  min-width: 140pt;
```
chrismclarke, line 19: nit(non-blocking): the use of `pt` sizing seems a little strange as it's not used anywhere else. I can see on the link you shared for `40px` height the min-width is set at `210px` which might be a better value to use (and would correspond to the standard 1.33 px per pt which aligns with standard dpi values)

chrismclarke (approved):
One minor (non-blocking) comment inline but otherwise looks good to me

src/app/shared/components/template/components/button-apple-sign-in/button-apple-sign-in.component.scss
jfmcquade: Good point (/`pt`...).

Apple's guidelines are spread across a few different pages, but I've opted to set the minimum width to `140px`, which seems to be the required minimum width according to [this section](https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple#Button-size-and-corner-radius), and I think looks appropriate for when the button text string is shorter

Screenshot 2025-03-27 at 16 46 29

PR #2879, opened 2025-03-28
Feat: Combo box search filter

### LINKED ISSUE #2219: [FEATURE] String search feature for combo box
**What?**
Add a "string search" feature to the combo_box component, to more easily find an entry in a long list. 

This should search through the "text" strings, not the "name"s. 

**Why?**
Use case in FacilitatorApp Mexico: users need to select their state from a list of 32 states.
image

**How?**
I'm imagining something like the string search feature in the template list

image

markharmon:
PR Checklist

- [x] Merge #2890 first, and update if necessary

## Description

When the default variant of the combo_box component has > 8 selectable items a search bar will be shown that can be used to filter the displayed items.

This adds a 'sub' component 'src/app/shared/components/template/components/combo-box/combo-box-search/*' which is opened via an Ionic modal from the existing combo-box component.

It is intended that the existing combo-box-modal component will be removed and replaced by the dropdown variant in the near future.

[comp_combo_box](https://docs.google.com/spreadsheets/d/168Z8Ne3VsF5XQCOTW_kkylUnsgjt6TohmNe3DvBHZ9A/edit?gid=839796397#gid=839796397).

## Git Issues

Closes #2219 

## Screenshots/Videos

image

image

image


src/app/shared/components/template/components/combo-box/combo-box-search/combo-box-search.component.ts
```diff
@@ -0,0 +1,60 @@
...
+  styleUrls: ["./combo-box-search.component.scss"],
+})
+export class ComboBoxSearchComponent implements OnInit {
+  public answerOptions = input.required<IAnswerListItem[]>();
+  @Input() row: FlowTypes.TemplateRow;
+  @Input() selectedValue: string;
+
+  textTitle: string | null;
+  searchTerm = signal<string>("");
+  showSearch = computed(() => this.answerOptions().length > 8);
```
jfmcquade, line 21: nit(non-blocking): This is duplicated from the parent component, so could either have the value passed in instead of recomputing, or be removed here entirely as it will always be true if this subcomponent is being called

src/app/shared/components/template/components/combo-box/combo-box-search/combo-box-search.component.ts
```diff
@@ -0,0 +1,60 @@
...
+  );
+
+  constructor(private modalController: ModalController) {}
+
+  ngOnInit() {
+    this.getParams();
+  }
+
+  getParams() {
+    this.textTitle = getStringParamFromTemplateRow(this.row, "text", null);
```
jfmcquade, line 40: nit(non-blocking): seeing as this is the only use of the `row` input, it might be better to expose a `title` property on this sub-component, and get the value from the row from within the parent component and pass that through. That would involve adding a new property to `IComboBoxParams` and the object returned from the combo-box component's `getParams` method (`modal_title` would be a better name for the param rather than `text`, but I can see `text` is already in use)

jfmcquade (changes requested):
Thanks @markharmon, generally looks good. I've added quite a few comments inline, mostly minor nit-picky things.

I think integrating much further with the theming system would better as a follow-up so as not to hold up this PR, as there's likely some explanation needed of how the themes work.

Also as a convention, good to link to any relevant debug sheet(s) in the PR description for ease of reviewing, in this case [comp_combo_box](https://docs.google.com/spreadsheets/d/168Z8Ne3VsF5XQCOTW_kkylUnsgjt6TohmNe3DvBHZ9A/edit?gid=839796397#gid=839796397).

@chrismclarke I'll tag you for review in case you want to double check my suggestions, otherwise feel free to remove yourself and I can re-add you when Mark's addressed some of minor points.

src/app/shared/components/template/components/combo-box/combo-box.component.scss
```diff
@@ -70,3 +70,48 @@ $background-with-value: var(
     }
   }
 }
+
+// Search variant
+.open-search-combobox {
```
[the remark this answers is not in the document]
markharmon: Yeah I did duplicate them intentionally, it's a bit of a mix of  Modal variant ones and dropdown variant.  I thought it would be easier to remove the Model variant ones without having to unpick and wanted to make clear that there was a custom component pretending to be a Ionic one

src/app/shared/components/template/components/combo-box/combo-box-search/combo-box-search.component.scss
```diff
@@ -0,0 +1,40 @@
...
+}
+
+ion-list, ion-item {
+  --background: var(--ion-background-color, #fff);
+  background-color: var(--background);
+}
+
+ion-item.selected
+{
+  --background: rgba(var(--ion-color-primary-rgb, 56, 128, 255), 0.08);
```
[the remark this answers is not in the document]
markharmon: I implemented your better still suggestion

src/app/shared/components/template/components/combo-box/combo-box-search/combo-box-search.component.ts
markharmon: Yep, this is gone now too. It was in use before I reworked it but I forgot about it

src/app/shared/components/template/components/combo-box/combo-box-search/combo-box-search.component.scss
```diff
@@ -0,0 +1,40 @@
...
+  --background-focused: var(--ion-color-primary, #3880ff);
```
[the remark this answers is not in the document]
markharmon: I removed them

markharmon:
Thanks for reviewing @jfmcquade I've addressed some of the issues and will take a look at the remaining ones sometime tomorrow

src/app/shared/components/template/components/combo-box/combo-box-search/combo-box-search.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+@Component({
+  selector: "combo-box-search",
+  templateUrl: "./combo-box-search.component.html",
+  styleUrls: ["./combo-box-search.component.scss"],
+})
+export class ComboBoxSearchComponent implements OnInit {
+  public answerOptions = input.required<IAnswerListItem[]>();
+  @Input() row: FlowTypes.TemplateRow;
+  @Input() selectedValue: string;
```
chrismclarke, line 17: nit(non-blocking): the `@input()` decorator notation is meant to be roughly equivalent with the newer `input` notation used for `answerOptions`. Angular introduced around 18 months back and has been promoting them more and more since (so a lot of legacy tutorials and ai code generators can be a bit out of date when referencing, and our own codebase is somewhat inconsistent)

It doesn't make a huge difference which one you use (signals are better for things like computed side-effects, but if the component is only receiving a static input one-time makes little difference).

So I would recommend here just making the inputs consistent, either all decorator or all signal. Given you're using some computed side-effects probably signals better

src/app/shared/components/template/components/combo-box/combo-box-search/combo-box-search.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  textTitle: string | null;
+  searchTerm = signal<string>("");
```
chrismclarke, line 20: nit(non-blocking): if defining a default value for a signal typescript should be smart enough to infer the type (so no need to explicitly state `string`). Where type definitions come in particularly handy is if the default value is undefined but you expect subsequent values to be a different type, e.g.

```ts
initialData = signal<string | undefined>(undefined)
```

src/app/shared/components/template/components/combo-box/combo-box-search/combo-box-search.component.ts
```diff
@@ -0,0 +1,55 @@
...
+  selectedItem = computed(() => {
+    if (!this.selectedValue) return null;
+    return this.answerOptions().find((item) => item.name === this.selectedValue);
+  });
```
chrismclarke, line 24: nit(non-blocking): it looks like this variable is no longer in use (guessing no longer required as you now pass the item value on click with the modal dismiss). 

As a general tip I'd recommend marking any variables or methods used by the HTML template as `public` and all others `private`. That way typescript will be able to detect if any variables marked as private are not referenced anywhere else in the code. Unfortunately it can't do this for public variables/methods as public variables could in theory be consumed elsewhere beyond the template.

This again is something we are pretty inconsistent with in the code (I typically only remember myself when manually noticing additional variables), so not a major issue, more just a general tip

src/app/shared/components/template/components/combo-box/combo-box.component.html
```diff
@@ -13,7 +13,21 @@
...
-  <div class="combo-box-container dropdown">
+  <button *ngIf="showSearch()"
+    class="open-search-combobox margin-t-regular"
+    (click)="openSearch()"
+    [ngClass]="{
+      disabled: disabled(),
+      'placeholder-style': (!_row.value && params().placeholder) || params().prioritisePlaceholder,
+      'with-value': _row.value,
+      'no-value': !_row.value
+    }"
```
chrismclarke, line 24: nit(non-blocking): It might be a little bit cleaner to calculate these dynamic classes within the `ts` file and pass as a variable. 

Given that the value depends on both the row and params you could use a computed signal, using the existing `params()` signal as well as the `value` signal which is also made available by the `TemplateBaseComponent` that most other template components extend (less obvious, but does exist - there's also a `rowSignal()` signal for the entire row). e.g.

```ts
public searchButtonClass = computed(()=>{
    const value = this.value();
    const params = this.params();
   return{
       disabled: disabled(),
      'placeholder-style': (!value && params.placeholder) || params.prioritisePlaceholder,
      'with-value': value ? true : undefined,
      'no-value': value ? undefined : true
   }
})
```

chrismclarke (approved):
Overall looking pretty good to me! Appreciate there's quite a lot to get used to both in terms of angular functionality and keeping consistent with our existing components which themselves can be pretty inconsistent, but I think the main logic I would expect to see is all there in ways that look familiar. 

I've added a handful of non-blocking comments in-line but happy to approve and leave it to you @jfmcquade to iron out any fine-details.

One final thought I might have would be whether some of the styling might be easier if integrating the search bar into the existing `combo-box-modal`, although also don't mind the separation of concerns and could always look to do a mini refactor at some point in the future if either component needs to get extended further

markharmon:
Thanks for the comments @chrismclarke I'll get them implemented

jfmcquade:
NB should probably merge after #2890

src/app/shared/components/template/components/combo-box/combo-box.component.ts
```diff
@@ -54,6 +55,7 @@ export class TmplComboBoxComponent
     }
     return getAnswerListParamFromTemplateRow(this.rowSignal(), "answer_list", []);
   });
+  showSearch = computed(() => this.answerOptions().length > 8);
```
jfmcquade, line 58: could be explicitly `public` for clarity

src/app/shared/components/template/components/combo-box/combo-box-search/combo-box-search.component.ts
```diff
@@ -0,0 +1,38 @@
...
+    )
+  );
+
+  constructor(private modalController: ModalController) {}
+
+  select(item: IAnswerListItem){
+    this.closeModal({ answer: item });
+  }
+
+  closeModal(value) {
```
jfmcquade, line 29: this could be made `private`, and the other methods explicitly `public`

jfmcquade (approved):
Looks good. I've added some comments to suggest making some properties/methods explicitly `public`/`private`, but not blocking so I think it should be good to merge either way

src/app/shared/components/template/components/combo-box/combo-box-search/combo-box-search.component.ts
```diff
@@ -10,7 +11,7 @@ import { ModalController } from "@ionic/angular";
 export class ComboBoxSearchComponent {
   public answerOptions = input.required<IAnswerListItem[]>();
   public title = input<string>();
-  public selectedValue = input<string>();
+  public selectedValue = input<FlowTypes.TemplateRow>();
```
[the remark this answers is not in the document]
markharmon: AAAh yeah, that's much better

PR #2887, opened 2025-04-10
fix: local vars defined in data-items only support string type

jfmcquade:
PR Checklist

## Description

Follow-up to #2772: fixes an issue where local vars defined within a `data-items` loop could only be strings.

# Further discussion

The associated bug was quite difficult to debug, and initially I thought there was an issue with the `@calc` functionality, which is complicated by the fact that there are at least three different methods in the code for evaluating a javascript expression. I think there is an issue for authors of not knowing what functionality is available where, when it comes to working with variables and javascript, which is partly a feature of having built out variable support in different contexts in a patchwork way (notably from inside/outside data-items loops). I think it could be good to include something along the lines of "doing variable things with variables across authoring contexts" as an explicit goal for future refactoring, and I think the place for that would likely be as part of the planned "dynamic templating" work (see [RFC](https://docs.google.com/document/d/1a8-lyF1LaNTTgx8lVWO0H91Twrr5ZEv_KjriddAk96s/edit?tab=t.0#heading=h.nd960lmg5uuf)).

## Git Issues

Closes #

## Screenshots/Videos

[debug_data_items_context](https://docs.google.com/spreadsheets/d/1p5j0rR7hUW8oyoj3Dz8RpUflxiLBliV0uz_tAt6NYfA/edit?gid=1422144003#gid=1422144003)

Screenshot 2025-04-11 at 12 46 43

| On `master` | On PR branch |
|----|----|
| Screenshot 2025-04-11 at 12 31 55 | Screenshot 2025-04-11 at 12 32 56 |


chrismclarke (approved):
Thanks for identifying this one, makes sense that the previous implementation would only ever assign intermediate variables if they were string values and that the updated code should fix.

I've added a quick test for the expected behaviour with [3eb143b](https://github.com/IDEMSInternational/open-app-builder/pull/2887/commits/3eb143b59f2e3ed7e2d7c863af9498729ef5fd4a), but otherwise I'd say should be good to go

chrismclarke:
> The associated bug was quite difficult to debug, and initially I thought there was an issue with the `@calc` functionality, which is complicated by the fact that there are at least three different methods in the code for evaluating a javascript expression. I think there is an issue for authors of not knowing what functionality is available where, when it comes to working with variables and javascript, which is partly a feature of having built out variable support in different contexts in a patchwork way (notably from inside/outside data-items loops). I think it could be good to include something along the lines of "doing variable things with variables across authoring contexts" as an explicit goal for future refactoring, and I think the place for that would likely be as part of the planned "dynamic templating" work (see [RFC](https://docs.google.com/document/d/1a8-lyF1LaNTTgx8lVWO0H91Twrr5ZEv_KjriddAk96s/edit?tab=t.0#heading=h.nd960lmg5uuf)).

Agreed, I have a deep backlog TODO item to revisit calc system and tidy-up, but don't really expect the work to be prioritised anytime soon - a lot depends on more general refactor to the templating system and unifying the multiple rendererers used across templates, items and data-items (leaning more heavily into the current data-items implementation which servers as a partial alternative rendering system, as mentioned in the RFC). 

I also had a call with Chris M a couple months back where we identified some limitations of the current calc system and ways to address, essentially either offering a new data_list type or inline row type to allow calculations authored in JS over multiple lines. Not sure if these suggestions ever made it into a working doc or not (screenshot below from the meeting notes I shared as a word doc following the call)

image

esmeetewinkel (approved):
Functional test passed on `debug` deployment. 

One authoring [question](https://docs.google.com/spreadsheets/d/1p5j0rR7hUW8oyoj3Dz8RpUflxiLBliV0uz_tAt6NYfA/edit?disco=AAABiDmFOpc) added to debug sheet.
image

I also confirmed this doesn't break the workaround offered on the `plh_kids_tz` template where this issue first appeared (sheet [here](https://docs.google.com/spreadsheets/d/1BoC8Zide9G0WNwqkVA2VYAEsE47kSUrBeFUJ8TJI7f0/edit?disco=AAABhmsjA3w)), but I've not gone back to check whether the original syntax now works (not wanting to break it now that it works...)

PR #2890, opened 2025-04-14
chore(deps): update ionic to v8

### LINKED ISSUE #2510: [FEATURE] Upgrade ionic packages to v8
**What?**
Upgrade all ionic packages to their latest version (currently v8). As noted in [this comment](https://github.com/IDEMSInternational/open-app-builder/pull/2374#issuecomment-2307539886), at least some of the `ionic-native` packages can be removed as they are unused.

**Why?**
As well as the inherent motivations for upgrading ionic itself, upgrading to v8 would unblock #2428 as v7 is not necessarily compatible with Angular 18. 

**How?**
See upgrade docs v7 -> v8 [here](https://ionicframework.com/docs/updating/8-0). `ionic/angular` is the primary package, but be sure to consider all ionic packages that are used.

jfmcquade:
PR Checklist

## Description

Updates the core ionic library to the latest version, 8.x (the core dependency is the [@ionic/angular](https://www.npmjs.com/package/@ionic/angular) package).

The included changes were made by following the steps in the [Required Changes](https://ionicframework.com/docs/updating/8-0#required-changes) section of the migration docs, as well as working through the list of breaking changes specified in the [changelog](https://github.com/ionic-team/ionic-framework/blob/main/BREAKING.md#version-8x).

## Author notes

The changes to be aware of are minor:
- Minor changes to styling of some elements on debug pages, e.g. the feedback sidebar, campaign and notification debug pages
- `simple_checkbox` component
    - Clicking the text label now selects/deselects the checkbox
- `timer` component
  - No expected changes
- `toggle_bar` component
  - No expected changes

### ionic-native packages

The [ionic-native](https://www.npmjs.com/package/@ionic-native/core/v/4.18.0) libraries are no longer actively maintained, having been usurped by Capacitor APIs. The codebase makes use of `ionic-native/status-bar`, `ionic-native/media`, `ionic-native/device`, and `ionic-native/http`. `ionic-native/status-bar` was only referenced in a spec test and so could be removed cleanly (https://github.com/IDEMSInternational/open-app-builder/pull/2890/commits/614042302490ff7f1b92b288b0771068ba70b775). The other libraries are used in the audio recording service and push notification service. I'm not aware whether either of these services are actively in use – if not, we could remove them and their associated dependencies, or otherwise consider refactoring them to use modern Capacitor libraries instead.

## Git Issues

Closes #2510


chrismclarke:
(forgot that the current test-visual needs new screenshots generated in advance of comparison - just created new screenshots via https://github.com/IDEMSInternational/open-app-builder/actions/runs/14447572622 and will re-run compare for clearer diffs)

chrismclarke:
@jfmcquade 
Looking through visual diffs the only ones that stand out to me are a handful of demos that include checkbox. Appears spacing changed quite a bit (not sure whether an ionic 8 change or diff elsehwere). E.g.

|               | before | after |
|---------------|--------|-------|
| debug_dg_form |   debug_dg_form before     |  debug_dg_form after     |

src/app/shared/components/template/components/simple-checkbox/simple-checkbox.component.html
```diff
@@ -1,14 +1,10 @@
...
->
+<div class="checkbox-wrapper margin-t-large" [class]="style" [class]="position">
   <ion-checkbox
     #checkbox
     [checked]="_row.value"
     (ionChange)="handleChange(checkbox.checked)"
     color="primary"
-  ></ion-checkbox>
-  <label *ngIf="label_text" [innerHTML]="label_text | markdown">{{ label_text }}</label>
+    [labelPlacement]="reverse ? 'start' : 'end'"
```
chrismclarke, line 7: question: is the `reverse` currently used by authors to define whether text is to the left/right of the checkbox? If so have you had a chance to check whether these updates also work as expected with RTL language variants? (I guess 'by expected' I mean either consistent with previous implementation, or improved if previous version had issue)

chrismclarke (approved):
Code changes look goo to me. I've added one minor comment noticed from the screenshot diffs and a question about RTL for checkbox labels but otherwise nice to see not a huge number of breaking changes to deal with.

As for deprecated ionic native deps:

`ionic-native/status-bar` - yeah just remove. Could be something authors want to use in the future to set specific status bar color, but whenever implementing we can just use Capacitor

`ionic-native/media` - as it's only used by the audio-recording service which itself is only a demo/poc service I would suggest just to remove both the service and dep and associated `IonicNativeAudioPlayer` component. 

`ionic-native/http` - this appears to be used by the pushNotificationService which itself is depcreated/unused and also I can't think of any real reason why it's included there (not used by methods directly). Native http calls is something that actually has very little benefit (unless streaming a large blob file directly to disk, which would require additional plugin implementation anyway) - as otherwise streaming on the native layer still needs to pass the full response over the Capacitor bridge back into the webview context, so likely better performance if just keeping requests within the webview context (browsers manage fetch pretty well).

`ionic-native/device` - again just used referenced push notification service and both unused and without purpose in the code (assuming would have been for native device check, which can be done in capacitor anyways and seems to be missing from code). 

So I think safe to say just get rid of all the ionic-native deps unless there's anything else you think I might have misssed

src/app/shared/components/template/components/simple-checkbox/simple-checkbox.component.html
jfmcquade: Thanks for flagging. It looks like the behaviour is unchanged: under the previous implementation (available on the [debug web preview](https://idems-debug.web.app/template/comp_simple_checkbox)), the label would appear on the right hand side of the checkbox by default for LTR languages and on the left hand side when the `reverse` param is true. For RTL languages, the order is reversed. This behaviour is preserved in the new implementation.

I think the "reverse" param isn't very well-named, it would be clearer to expose a parameter like "label_placement" with possible values `"start" | "end"` (and possibly `"fixed" | "stacked"` to match the [ionic options](https://ionicframework.com/docs/api/checkbox#label-placement)), but I'm happy to avoid breaking changes in this PR for now.

RTL demo:

Live web preview:

Screenshot 2025-04-15 at 11 41 20

Pr branch:

Screenshot 2025-04-15 at 11 41 13

jfmcquade:
> So I think safe to say just get rid of all the ionic-native deps unless there's anything else you think I might have misssed

Thanks for looking into it, yeh that was my understanding too. I've removed in [these commits](https://github.com/IDEMSInternational/open-app-builder/pull/2890/files/f366be89aa46be1c56d9bffe32027fc256c981d2..c054bf0ec43b4c4ca86d922be0c759ba91ddf2e1).

esmeetewinkel (approved):
Functional test passed. 

> Minor changes to styling of some elements on debug pages, e.g. the feedback sidebar, campaign and notification debug pages

Changes on feedback and campaign pages look acceptable to me, on the notification page I can't distinguish a difference. 

image

image

image

> `simple_checkbox` component

Changed behaviour works okay, and I actually prefer the reduced height (on the PR branch).

Issues I observed while testing that are not related to this PR:
- the `timer` component is broken (already on `master`), I'll track as a separate issue.
- toggle bar iOS variant needs a refresh to show.

PR #2892, opened 2025-04-15
Feat: Add clipboard copy action

### LINKED ISSUE #2886: [FEATURE] Copy action
**What?**
Ability to copy to clipboard from an action (click) 
Would need to pass in the variable to copy 

**Why?**
needed for https://github.com/ParentingForLifelongHealth/plh-facilitator-app-ph-content/issues/27

**How?**
click | copy | text: @local.share_message

**Other details**
This will be used instead of share as certain platforms do not allow editing the message. Therefore, users will copy the message instead. 

We need to think how this would work if the message contains an image as well i.e. click | copy | text: @local.share_message, file: @local.asset

markharmon:
PR Checklist

## Description

Adds the 'copy' action which has parameter 'text'

`click | copy | text: @local.text_value`

## Git Issues

Closes #2886

## Screenshots/Videos

Template: [feature_action_copy](https://docs.google.com/spreadsheets/d/1rFMgk8YGFMG_bN8WKY_4Yg1KWXaQn4FtohtWbsrIZe4/edit?gid=2063929093#gid=2063929093)


markharmon:
I think if we extend support to something like `click | paste` then we could register an additional action 'paste' in the ClipboardService.

markharmon:
I think this could possibly be extended in future to support:

`click | copy | text: @local.text_value, image:@local.image_value `

This function would likely be android specific though using @capacitor/clipboard label option:

image

jfmcquade (approved):
Looks good, nice clean logic. 

As discussed, it's not clear how passing the other data types to the [Capacitor Clipboard API](https://capacitorjs.com/docs/apis/clipboard#writeoptions) is supposed to work – it's implicit that multiple data types could be passed at once but I'm not sure how this is handled in practice. It may be that the behaviour is platform specific, as is the case for the share functionality introduced in #2768. I'd be happy to get this PR merged first and address any further functionality in a follow-up if required.

chrismclarke (approved):
Code all looks nice and clean to me. 

There might be an authoring decision about whether better to namespace as `clipboard copy` or `clipboard_copy` instead of just `copy`. I can't really think of other actions that would want to use copy/paste keywords beyond the clipboard, just more whether we want to make more effort to namespace actions or not (and the likelihood of doing anything else with the clipboard API... guessing paste isn't really an action to trigger from a button rather something a user would just do themselves)

I'll tag @esmeetewinkel to review also in case any strong feelings either way, but otherwise I'd say good to go

esmeetewinkel (approved):
Functional test passed. Happy with proposed syntax

PR #2899, opened 2025-04-16
Feat: shared data

chrismclarke:
PR Checklist

**TODOs**
- [x] Merge after #2900
- [x] Merge after #2901
- [x] Merge after #2902

## Review Notes
**Firebase Setup**
If using a new firebase project will need to enable firestore, set firestore rules and create query indexes. See screenshots below.
These should ideally be added to standalone documentation in the future

**General Setup**
New deployment config requires compiling
```sh
yarn workflow deployment set debug
```

Debug page available at `/shared-data`. Functionality will only be available if a user is signed in to firebase

Data updates can be viewed from firebase console project, within the `firestore` tab

**NOTE** - the current debug repo is setup to use a project named `debug-e8934.firebaseapp.com`. Will need access to firebase project to view console (message in mattermost about changing)

**Testing**
1. Create a new shared collection 
2. Set key-value pairs for the shared data, and see updates in firebase console
3. Invite additional users by auth_id to the project and confirm they also receive the same data updates (should sync in realtime)

## Author Notes
This PR doesn't expose authoring for the shared data components (will be part of future follow-up PR). For now it is designed to be called directly in a custom component

The collection `delete` functionality is currently limited to only delete collections where no additional users have been added. This is due to the fact that if multiple users have data from a collection synced, and then one user deletes the collection, there is currently no way for other users to know that the collection has been deleted (the app queries for updates to collections where the user is a member, so deleted collections will just not send any update data - essentially keeping the local collection in its current state). Future PR will be required to implement a tidier system of keeping reference to deleted collections on the server so that users can sync deletes.

Deployments will need to be using firebase as auth provider in order to use the shared data feature. 

In cases where multiple apps use the same firebase project, any data stored in firestore will be available across all apps. Ideally non-related projects should have their own firebase projects, however if this is a technical limitation then would recommend using careful naming conventions on projects to avoid overlap. It is also possible to setup multiple firestore databases within a project, however only the initial default database receives free resource quotas. (Could discuss as follow-up if needed)

All shared data currently sits within a top-level collection named `shared_data`, with random ids assigned to generated collections. Follow-up PR could be used to allow changing top-level names for easier data separation.

## Dev Notes
When using with a new project for the first time an error will appear in the console as firestore needs to build indexes used for collection queries. Follow the link in the console to setup (example screenshot below)

**Follow-up Tasks**
- [ ] Add support for using different top-level collection names than `shared_data` - e.g. `shared_parent_groups` (keep shared prefix for easier custom rules). Will require service-level changes and rules update
- [ ] PLH-specific component to interact with shared data for parent groups
- [ ] Add Tests
- [ ] Documentation
- [ ] (see additional tasks in feature Readme)

## Git Issues

Closes #

### Demo
Demo showing a signed in user create a collection, share with second user (incognito window), and update shared data viewing updates in realtime

Demo also shows issue of removing users (as mentioned in author notes), where removed users will still keep a snapshot of data at the time they were removed until cache cleared

https://github.com/user-attachments/assets/a21de452-5a2c-4ee0-a20b-f92f35f9739a

### Firebase Setup
**Initial setup will request reqion to store firestore data. This cannot be changed after creation, and different regions have different pricing structures so recommend being intentional when setting**
image

**Firestore will also require setting rules. Recommend using production rules as defined in feature readme (TODO - general documentation)**
image

**If cached data exists will require one-time console setup of query index to allow searching for shared collections newer than cached. So will need to create a collection and reload the page before following link to create**
image


jfmcquade (commented):
Thanks @chrismclarke. A few initial comments from testing:

I've enabled Firestore on the [debug project](https://console.firebase.google.com/u/1/project/debug-e8934/firestore/databases/-default-/data). Creating and updating shared collections seems to work. Deleting throws a series of errors but does seem to work as expected:

Screenshot 2025-04-23 at 13 17 20

Some queries on points in the PR description:

> General Setup
> New deployment config requires compiling

Are there still required updates to the deployment config? Deployment config updates aren't included in the code repo, and I can't see any associated updates on the [debug content repo](https://github.com/IDEMSInternational/app-debug-content)

> **When using for first time console will show error and link to setup query index - follow link to generate**

I didn't get a link to generate indexes, but the feature seems to be basically working without having configured indexes

chrismclarke:
> Thanks @chrismclarke. A few initial comments from testing:
> 
> I've enabled Firestore on the [debug project](https://console.firebase.google.com/u/1/project/debug-e8934/firestore/databases/-default-/data). Creating and updating shared collections seems to work. Deleting throws a series of errors but does seem to work as expected:

Oh that makes sense, it's because there are two separate data subscriptions (one for full list and one for selected), but when deleting from the full list it is still trying to render the selected. Should be fixed with  f1c470eccf3fb69836477d90a5f2f2876f97e739
 
> Are there still required updates to the deployment config? Deployment config updates aren't included in the code repo, and I can't see any associated updates on the [debug content repo](https://github.com/IDEMSInternational/app-debug-content)

I've set the default config to have shared data enabled with firebase as a provider, so authors don't need to author explicitly, but do still need to compile the deployment to apply the new defaults. This shouldn't be an issue if the deployment does not have firebase enabled as the shared data will only attempt to interact with the underlying firebase db when it encounters a shared-data component (no background processing involved)

> > **When using for first time console will show error and link to setup query index - follow link to generate**
> 
> I didn't get a link to generate indexes, but the feature seems to be basically working without having configured indexes in the firebase console

This will only happen if there is cached data, as the query made to firebase searches for shared data collections newer than most recent cache. So you should see after page reload (will make pr description clearer)

chrismclarke:
@jfmcquade 
I've also just created a new collection myself and invited you to it if you want to test any of the data sharing functionality

image

jfmcquade:
> @jfmcquade I've also just created a new collection myself and invited you to it if you want to test any of the data sharing functionality
> 
> image

Nice, I'm able to see the collection (after adding an index on the db via the link).

But I'm not able to update with new data, despite my auth_id appearing as an admin

Screenshot 2025-04-23 at 17 18 00

chrismclarke:
@jfmcquade - just to check-in. Now that you've had a chance to test with #2910, what do you think are the current priorities/blocking issues to address ahead of merging this?

src/app/feature/shared-data/shared-data.actions.ts
```diff
@@ -0,0 +1,35 @@
+import type { IActionHandler } from "src/app/shared/components/template/services/instance/template-action.registry";
+import { SharedDataService } from "./shared-data.service";
+
+interface INavStackActionParams {
```
jfmcquade, line 4: nit(non-blocking): update/remove

jfmcquade (approved):
> @jfmcquade - just to check-in. Now that you've had a chance to test with #2910, what do you think are the current priorities/blocking issues to address ahead of merging this?

I don't think I'm aware of any blocking issues. I think it probably makes sense to get this merged and deal with any issues as they arise?

I did find that I wasn't able to use the exposed `SharedDataService.query$()` method for my desired purpose – see [line 247 of the proposed plh-parent-group service](https://github.com/IDEMSInternational/open-app-builder/pull/2910/files#diff-4bd37b5e366ab5768cb5fe6be9ba0b6722de528c4e76788dec8fa39f7a2eefaeR247). But I think this is an issue with me trying to get a snapshot from the result: taking `firstValueFrom()` gave a stale result from the cache, and I couldn't figure out how to wait for the latest value to come through and take that. So I ended up effectively bypassing the cache and just calling the methods that query the live data directly. But I think that's more of an issue with how I'm consuming the query results, and casting them to static values: the querying and cache works when used as designed, i.e. subscribing to the observables, as on the debug page.

chrismclarke:
> I don't think I'm aware of any blocking issues. I think it probably makes sense to get this merged and deal with any issues as they arise?
> 
> I did find that I wasn't able to use the exposed `SharedDataService.query$()` method for my desired purpose – see [line 247 of the proposed plh-parent-group service](https://github.com/IDEMSInternational/open-app-builder/pull/2910/files#diff-4bd37b5e366ab5768cb5fe6be9ba0b6722de528c4e76788dec8fa39f7a2eefaeR247). But I think this is an issue with me trying to get a snapshot from the result: taking `firstValueFrom()` gave a stale result from the cache, and I couldn't figure out how to wait for the latest value to come through and take that. So I ended up effectively bypassing the cache and just calling the methods that query the live data directly. But I think that's more of an issue with how I'm consuming the query results, and casting them to static values: the querying and cache works when used as designed, i.e. subscribing to the observables, as on the debug page.

Yeah that makes sense, and I think your approach is the right one. I've just added a comment to the follow-up PR, but think essentially whilst it would be nice to better track the source of the data (i.e. `server` | `cache`), doing so would take a fair bit of refactoring which I wouldn't be keen to do at this point. Likely a better follow-up will probably be to move more of the 1-time data request methods into the service which have option to specify query source

chrismclarke:
@esmeetewinkel 
I'm going to go ahead an merge this without functional test as user/author-enabling updates will be applied in #2910, but just to note that there is an extra debug page at `/shared-data` that could be further extended in the future if we want to generalise authoring shared data (currently relying on a specific plh component to handle case for user groups given time constraints)

PR #2902, opened 2025-04-17
Fix: dynamic data delete

chrismclarke:
PR Checklist

**TODO**
- [x] Merge after #2901 and rebase (currently targets pr branch for clearer diff)

## Description
Part of #2899 tidying
Fix issue where dynamic data would not delete correctly

## Dev Notes
The issue is related to the way the state persist methods would compare what is currently stored in the indexeddb vs what is in cache, relying on looking at keys in hashmaps such as `{data_list: {list_1 : {row_1: ...}}}`. 

There main problem was that the comparison only looked one level deep, so would detect if an entire list was added/removed, but not individual rows. So to fix I've refactored to generate a flatter hashmap, where every entried is keyed by a combination of flow_type, list_id and row_id, e.g. `{data_list__list_1__row_1:{...}}`

I don't think this has much impact for current apps as resetting flows to original state still work as expected (would only have been noticed if deleting rows from data-items on lists that do not have initial data) 

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
+++ b/src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
@@ -17,7 +17,8 @@ addRxPlugin(RxDBUpdatePlugin);
 import { debounceTime, filter, firstValueFrom, Subject } from "rxjs";
 
 import { FlowTypes } from "data-models";
-import { compareObjectKeys, deepMergeObjects } from "../../../utils";
+import { deepMergeObjects } from "../../../utils";
+import { arrayToHashmap, diffObjects } from "packages/shared/src/utils/object-utils";
 
 /**
  * All persisted docs are stored in the same format with a standard set of meta fields and doc data
--- a/src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
+++ b/src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
@@ -140,7 +140,22 @@ export class PersistedMemoryAdapter {
     this.persistStateToDB();
   }
 
-  public delete(flow_type: FlowTypes.FlowType, flow_name: string, ids?: string[]) {
+  /** Set data for a given entry. Will overwrite any existing data */
+  public set(update: IDataUpdate) {
+    const { flow_name, flow_type, id, data } = update;
+    if (!this.state[flow_type]) this.state[flow_type] = {};
+    if (!this.state[flow_type][flow_name]) this.state[flow_type][flow_name] = {};
+    // Remove any values marked as undefined
+    for (const [key, value] of Object.entries(data)) {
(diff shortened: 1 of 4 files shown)


src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
```diff
@@ -140,7 +140,22 @@ export class PersistedMemoryAdapter {
...
+    if (!this.state[flow_type]) this.state[flow_type] = {};
+    if (!this.state[flow_type][flow_name]) this.state[flow_type][flow_name] = {};
+    // Remove any values marked as undefined
+    for (const [key, value] of Object.entries(data)) {
+      if (value === undefined) {
+        delete data[key];
+      }
+    }
+    this.state[flow_type][flow_name][id] = data;
+    this.persistStateToDB();
```
jfmcquade, line 155: Should this be async/awaited so the `set()` method can also be awaited?

src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
```diff
@@ -149,11 +164,11 @@ export class PersistedMemoryAdapter {
         delete this.state[flow_type][flow_name][id];
       }
     }
-    // delete all
+    // delete all - mark as empty so it still persists to disk
     else {
-      delete this.state[flow_type][flow_name];
+      this.state[flow_type][flow_name] = {};
```
jfmcquade, line 169: Is it potentially messy to leave empty flows in the DB vs deleting them entirely? Doesn't seem like a major issue either way

jfmcquade (approved):
Looks good I think. See couple of queries inline

src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
chrismclarke: not on this occasion, as it's part of a synchronous `set` method which is designed to update the in-memory value, and let the persistence just happen as a side-effect (trigger but don't wait to complete). I can't actually remember why I wanted the sync set/update methods, but better to keep non-async as persistence always has a bit of debounce/delay that isn't super important to whatever prompted the update

src/app/shared/services/dynamic-data/adapters/persistedMemory.ts
chrismclarke: This is just the in-memory state, so when persisting it iterates over row entries within the hashmap (e.g. `this.state['data_list']['example_flow']['row_id_1']`, so it being empty will also prompt any existing rows to be deleted when persisting.

The previous issue with deleting entirely is that when it compared rows there was no flow to compare against (everything was happening one level less deep than required), and I found adding the empty placeholder worked to fix

PR #2905, opened 2025-04-22
Feat: Add date time picker

### LINKED ISSUE #2878: [FEATURE] DateTime picker
**What?**
Ability to author datetime picker

**Why?**
A date picker is required for https://github.com/ParentingForLifelongHealth/plh-facilitator-app-ph-content/issues/14

**How?**
This should have similar functionality as text_box i.e. placeholder, events

markharmon:
PR Checklist

## Description

Adds a date time picker component. There is scope to add more params such as minDate, maxDate, etc in future PRs.

Localisation including the translated month should be automatic based on the device https://ionicframework.com/docs/api/datetime#localization

When seeting a value in the spreadsheet it must be plain text and in this format "YYYY-MM-DDTHH:mm"

# DateTimePicker

## Example

| type             | name          | value | parameter_list |
| ---------------- | ------------  | ----- | -------------- |
| date_time_picker | date_time_1   | ----- | -------------- |	

[Google Sheet Demo](https://docs.google.com/spreadsheets/d/1Ff1QbrAKWpENp6QNeHVAw6_7KtkXz6zngRTSKqHMssw/edit?gid=569531329#gid=569531329)   
[Live Preview Demo](https://idems-debug.web.app/template/comp_date_time_picker)

## Parameters

| Parameter          | Value               | Description                          |
| ---------             | ------------------  | ------------------------------------ |
| disabled	            | false (default)      | Enables the button                   |
| disabled	            | true                | Disables the button and greys it out |
| type                   | date-time (default) | Allows date and time to be selected  |
| type                   | date                | Allows date only to be selected      |  
| type                   | time                | Allows time only to be selected      |  

## Events

changed

## Git Issues

Closes #2878

## Screenshots/Videos

image

image

image


src/app/shared/components/template/components/base.ts
```diff
@@ -23,7 +23,7 @@ export class TemplateBaseComponent implements ITemplateRowProps {
   // TODO - main row should just be an input.required and child code refactored to avoid set override
   // TODO - could also consider whether setting parent required (is it template row map or services?), possibly merge with row
   rowSignal = signal<FlowTypes.TemplateRow>(undefined, { equal: isEqual });
-  value = computed(() => this.rowSignal().value, { equal: isEqual });
+  value = computed(() => this.rowSignal()?.value, { equal: isEqual });
```
markharmon, line 26: I added this to solve a problem I had when this.rowSignal was undefined, I'm not sure if this will have any unforeseen consequences elsewhere?

src/app/shared/components/template/components/date-time-picker/date-time-picker.component.html
```diff
@@ -0,0 +1,28 @@
...
+  @if (dateTimeValue() === undefined) {
+    <ion-icon name="calendar-outline" size="large" slot="date-target"></ion-icon>
+    <ion-icon name="time-outline" size="large" slot="time-target"></ion-icon>
+  }
+</ion-datetime-button>
+
+<ion-modal [keepContentsMounted]="true">
+  <ng-template>
+    <ion-datetime
+      [id]="rowSignal()?.name"
```
markharmon, line 11: Not sure if this is a good idea? is name required? I could use a uuid instead

chrismclarke: I guess the other question is whether the `id` is even required? I can see you've also added a template variable for `#datetime` below which is how you reference in the buttons below so I'm guessing probably not required

src/app/shared/components/template/components/date-time-picker/date-time-picker.component.md
```diff
@@ -0,0 +1,27 @@
+# DateTimePicker
```
chrismclarke, line 1: +1 - thanks for adding the documentation, really nice that it also links through to the demos

src/app/shared/components/template/components/date-time-picker/date-time-picker.component.ts
```diff
@@ -0,0 +1,41 @@
...
+  type: string;
+}
+
+@Component({
+  selector: "plh-date-time-picker",
+  templateUrl: "./date-time-picker.component.html",
+  styleUrls: ["./date-time-picker.component.scss"],
+})
+export class TmplDateTimePickerComponent extends TemplateBaseComponent implements OnInit {
+  public dateTimeValue: WritableSignal<string | undefined>;
```
chrismclarke, line 18: nit(blocking): Oh yeah, I can see why this was tricky to implement and why the template base component was causing issue. So we need a variable that can both be initially set from the template value, but then also modified from the datetimepicker (with the ngmodel binding).

That leaves us with a single variable that is getting updated from two different places (ngmodel from HTML and row value from templating system). It might be a little bit cleaner to drop the `ngModel` use entirely from the template and just rely on the `ionChange` updating the the row value, e.g.
```html
<ion-datetime [value]="value()"  (ionChange)="handleChange(datetime.value)" >...
```
I'd at least hope that this should keep everything in sync without the need for ngmodels, but haven't tested myself. If working it would also remove the need for the `ngOnInit` call, which should help to simplify a bit more. But if it doesn't work then what you've done still seems pretty reasonable.

src/app/shared/components/template/components/date-time-picker/date-time-picker.component.ts
```diff
@@ -0,0 +1,41 @@
...
+  public ngOnInit(): void {
+    this.dateTimeValue = signal(this.value() || undefined);
+  }
+
+  public handleChange(value: string | string[]) {
+    this.setValue(value);
+  }
+
+  private getParams(): IDateTimePickerParams {
+    let parameterList = this.parameterList();
```
chrismclarke, line 34: nit(non-blocking): might be useful to cast the type definition you created above so we know what the expected type is, i.e.
```ts
const parameterList = this.parameterList() as IDateTimePickerParams
```

src/app/shared/components/template/components/date-time-picker/date-time-picker.component.ts
```diff
@@ -0,0 +1,41 @@
+import { Component, computed, OnInit, signal, WritableSignal } from "@angular/core";
+import { TemplateBaseComponent } from "../base";
+import { getStringParamFromTemplateRow, parseBoolean } from "src/app/shared/utils";
+
+interface IDateTimePickerParams {
+  /** TEMPLATE PARAMETER: "disabled". If true, date time picker is disabled and greyed out */
+  disabled: boolean;
+  /** TEMPLATE PARAMETER: "type". Determines the type of picker date & time ('datetime'), date only ('date') or time only ('time') picker*/
+  type: string;
```
chrismclarke, line 9: nit(non-blocking): If any of the parameters are optional for authors then probably best to update the type definition accordingly, e.g.
```ts
  /** TEMPLATE PARAMETER: "disabled". If true, date time picker is disabled and greyed out */
  disabled?: boolean;
  /** TEMPLATE PARAMETER: "type". Determines the type of picker date & time ('datetime'), date only ('date') or time only ('time') picker*/
  type?: string;
```

src/app/shared/components/template/components/date-time-picker/date-time-picker.component.ts
```diff
@@ -0,0 +1,41 @@
...
+  public handleChange(value: string | string[]) {
+    this.setValue(value);
+  }
+
+  private getParams(): IDateTimePickerParams {
+    let parameterList = this.parameterList();
+
+    return {
+      disabled: parseBoolean(parameterList.disabled),
+      type: getStringParamFromTemplateRow(this._row, "type", "date-time"),
```
chrismclarke, line 38: nit(non-blocking): a lot of these `getParamFromTemplate` style utility functions are a bit clunky/messy, and more designed for an earlier time where we struggled a lot with passing different data types between components. I think if creating a new component it might be better to define in simpler terms where possible (like you've done for the disabled parameter), 

i.e. just using parameter list with fallback value (using logical OR)
```ts
type: parameterList.type || 'date-time'
```

chrismclarke (changes requested):
Overall the component looks well layed out to me, really appreciate having the extra docs and examples. I haven't done any functional testing, but added a few (mostly minor/non-blocking) comments inline. 

See one specific (blocking) comment about some of the data binding that I'm not entirely sure myself about

src/app/shared/components/template/components/date-time-picker/date-time-picker.component.html
markharmon: Yes, it is required. This is how Ionic links the datetime to the datetime button

markharmon: I found that I needed a unique value, any datetime components with the same value will share the same properties. I thought this might possibly a nice feature (side effect) to have but on reflection I think a uuid is safer

src/app/shared/components/template/components/date-time-picker/date-time-picker.component.ts
markharmon: Yeah that's better, I thought to separate from using the actual value signal though as I anticipate there might be some pre-processing to do on the row value - maybe to allow for defaulting to todays date via the spreadsheet. We can adapt it later I guess and maybe that logic belongs earlier in the template process

src/app/shared/components/template/components/date-time-picker/date-time-picker.component.html
chrismclarke: Oh that is super weird, but in that case the safest would probably be to use `rowSignal()._nested_name` which ensures for unique row names even in cases where an author might have used the same name by authoring a template that is reused multiple times (e.g. an author might have an `activity` template flow that includes a row named `date_input`, but then use the `activity` template multiple times on a page, resulting in multiple `data_input` rows. The nested name includes full prefixes of any nested pathing, e.g. `activity_1.date_input`, `activity_2.date_input` etc.). 

Or if the signal has been giving issue should be safe to use the non reactive `row._nested_name` property given that the property should always be static (authors shouldn't be using dynamic variables while setting row names, and even if they did pretty sure our change detection would treat as a new component and destroy the old when creating the new).

markharmon: rowSignal()._nested_name does work, I think generating the unique id ourselves in the component is a bit more explicit though. We don't need anything meaningful here just unique. What do you think of the uuid solution I committed?

src/app/shared/components/template/components/date-time-picker/date-time-picker.component.ts
chrismclarke: If we do need some preprocessing then hopefully you can use a similar pattern and just a `computed` signal to map the value as required. Although I'm guessing most of the time if an author wants to pass a specific date into the component it will likely have been generated first (e.g. restoring a previously selected value) or maybe passing today's date - but yeah can address those cases as they come up. 

Authors should also have some flexibility in the meantime if wanting to do something more bespoke as the authoring system exposes the [date-fns package](https://date-fns.org/docs/Getting-Started) directly for use in calc statements, e.g. `value: @calc.date_fns.addDays(new Date(), 7)` (I think this works at least), so could also map to isoString as required for the input, `value: @calc.date_fns.addDays(new Date(), 7).toIsoString()`

I guess the biggest motivator for using a computed value could be if we want to support date objects being passed, and have something like
```ts
dateValue = computed(()=>{
  const value = this.value()
  if(value instanceof Date) return value.toIsoString()
  return value
})
```

chrismclarke (approved):
Thanks for all the changes @markharmon , I've added one minor comment with 4bd3e2281ab61b8b5c2dff65f7186141d530dd0d but otherwise looks good to me

ChrisMarsh82 (approved):

PR #2900, opened 2025-04-16
chore(deps): angular minor update

chrismclarke:
PR Checklist

- [ ] PR title descriptive (can be used in release notes)

## Description
Update angular v17.2.0 -> v17.3.0

## Dev Notes
I realised as part of review on #2879 that I proposed using [output signals](https://angular.dev/guide/components/outputs) but realised these were only introduced in version 17.3.0 whilst we were running v17.2.2.

I figured it would be easier to make the minor update that introduced the feature (which I've started using in some upcoming PRs) than push for major update right now (remember some previous spike PRs that were held up for various reasons)

## Git Issues

Closes #

## Screenshots/Videos

_If useful, provide screenshot or capture to highlight main changes_

PR #2907, opened 2025-04-22
fix: template-injection-context

### LINKED ISSUE #2906: [BUG] RuntimeError: NG0203: inputFunction() can only be used within an injection context
**Describe the bug**
ESM app home screen shows blank and error in console.

``error-handler.service.ts:69 ERROR RuntimeError: NG0203: inputFunction() can only be used within an injection context such as a constructor, a factory function, a field initializer, or a function used with `runInInjectionContext`. Find more at https://angular.io/errors/NG0203``

**App version**
Not working on 0.18.2+ (current `master`, e.g. commit 585188f)
Still working on commit 99d7b22
So I'm guessing the culprit is PR #2900 

**Deployment Sheet(s)**
[Early School Math](https://github.com/IDEMSInternational/early-school-math-app-content) deployment home screen

**Screenshots and videos**

Image

chrismclarke:
PR Checklist

## Description
Fix issue where startup templates would fail to load due to updated way angular checks templates before render (likely following #2900)

## Dev Notes
When running as a standalone template the signal input isn't actually used, but it appears the angular compiler has gotten more sensitive to the potential of using when not in injection context. So simple workaround is to just pass an injection context to the standalone templates (even if not used)

Now that the container is created in an injection context it's possible to remove the previous workaround for handling effect injection context

## Git Issues

Closes #2906

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/services/instance/template-process.service.ts
+++ b/src/app/shared/components/template/services/instance/template-process.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, Injector } from "@angular/core";
+import { Injectable, Injector, runInInjectionContext } from "@angular/core";
 import { FlowTypes } from "data-models";
 import { AppDataService } from "src/app/shared/services/data/app-data.service";
 import { getGlobalService } from "src/app/shared/services/global.service";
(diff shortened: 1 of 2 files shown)


esmeetewinkel (approved):
Functional test passed.

jfmcquade (approved):
Nice, thanks for fixing. I think the workaround introduced is a bit clearer than the one removed, so feels like an improvement overall

PR #2913, opened 2025-04-24
feat(Qol): Add google sheet folder URL to json

### LINKED ISSUE #2884: Content jsons should display spreadsheet folder
**Details**
Now that deployment configs no longer point to a single Google Drive folder, but to a list of folders (following #2081), it can be quite tricky to find the corresponding sheet in Google Drive to a given content json. 

The json file _does_ display the relative folder path to the file containing this sheet, but not which of the folders in the config this is relative to. Can we add this? Ideally it would be a hyperlink rather than the folder ID so that you can click and access the folder immediately. 

**Screenshots and videos**
Example json: 
Image

Example config:
Image

markharmon:
PR Checklist

## Description

Adds _sheetsFolderUrl property to the template JSON file so that it is easier to locate the template source sheet

```
  "_xlsxPath": "component_sheets/component_date_time_picker.xlsx",
  "_sheetsFolderUrl": "https://drive.google.com/drive/u/0/folders/1YcG0GDJME1Ds1UzwXcEmo97Ckb5tnUWc"
```

## Git Issues

Closes #2884


chrismclarke (commented):
question(non-blocking):
I can see how this solves the linked issue and the code changes look good to me. 

I have 2 potential questions however, linked more towards authoring so I'd be keen to get @esmeetewinkel 's opinion

1. This produces an output like below, where the `_sheetsFolderUrl` links to the top-level sheets folder, from where you can follow the `_xlsxPath` to locate the specific sheet.

```json
{
  "flow_type": "data_list",
  "flow_name": "debug_data_items_ctxt_data",
  "status": "released",
  "_xlsxPath": "debug_sheets/debug_data_items_context.xlsx",
  "_sheetsFolderUrl": "https://drive.google.com/drive/u/0/folders/1YcG0GDJME1Ds1UzwXcEmo97Ckb5tnUWc"
}
```

Would it however be more useful to try and get the URL to link directly to the sheet instead of the top-level folder? I think all of the metadata to do this should be passed through the google sheets API (just might take a bit of extra work to track correctly)

Alternatively/in addition It might also be possible to update the xlsxPath to include the top-level folder name in case of multiple content repos, e.g. `Dev: debug deployment/debug_shets/debug_data_items_context.xlsx`

2. Once merged, content repos will now include direct links to drive folders. Given repos are public, this could make it easier for outside parties to try to access the folders (already possible given the folder id is in config files, just now it would be fully formatted so would come up in more searches by things like bots). I don't think in itself this would be a huge issue, but might be worth highlighting the potential security issue and advising deployment repos to either only be shared with specific individuals, or if supporting public link to be readonly via public link

esmeetewinkel (approved):
Looks good, thanks. When moving between this PR branch and `master` (both ways) and running a content sync for the `plh_kids_teens_za` deployment, on the first (few) syncs I often get errors:

image
image

The errors stop showing after a few syncs. I'm guessing this is not related to this PR? Wanted to check just in case. 

On @chrismclarke's comments: 

1. I'm happy with the current, link to folder + relative path. About your alternatives:

> Would it however be more useful to try and get the URL to link directly to the sheet instead of the top-level folder? I think all of the metadata to do this should be passed through the google sheets API (just might take a bit of extra work to track correctly)

I often have a sheet or workbook with the same name (e.g. `home_screen` or `app_navigation`)  in multiple of the Drive folders. So it can actually be quite handy to start on the top level to figure out which one is being synced.

> Alternatively/in addition It might also be possible to update the xlsxPath to include the top-level folder name in case of multiple content repos, e.g. Dev: debug deployment/debug_sheets/debug_data_items_context.xlsx

The Drive folders don't always have unique names (e.g. might be a folder `sheets` within a deployment folder) so might have to go a few levels up to ensure uniqueness. In this case my preference would be to reference names that the user gives to each folder path in the deployment config. 

2. Thanks for pointing out the security issue, I hadn't thought about that. If you don't think it's too much of a problem, I'm happy to go ahead anyway. I've normally been giving minimal necessary access rights on the files, so I'll continue to do so and also advise others.

chrismclarke:
> Looks good, thanks. When moving between this PR branch and `master` (both ways) and running a content sync for the `plh_kids_teens_za` deployment, on the first (few) syncs I often get errors:

> The errors stop showing after a few syncs. I'm guessing this is not related to this PR? Wanted to check just in case.

Strange, I'd be surprised if related to this PR but will take a quick look to see if I can recreated

> 
> 1. I'm happy with the current, link to folder + relative path. About your alternatives:
> 
> > Would it however be more useful to try and get the URL to link directly to the sheet instead of the top-level folder? I think all of the metadata to do this should be passed through the google sheets API (just might take a bit of extra work to track correctly)
> 
> I often have a sheet or workbook with the same name (e.g. `home_screen` or `app_navigation`) in multiple of the Drive folders. So it can actually be quite handy to start on the top level to figure out which one is being synced.

> > Alternatively/in addition It might also be possible to update the xlsxPath to include the top-level folder name in case of multiple content repos, e.g. Dev: debug deployment/debug_sheets/debug_data_items_context.xlsx
> 
> The Drive folders don't always have unique names (e.g. might be a folder `sheets` within a deployment folder) so might have to go a few levels up to ensure uniqueness. In this case my preference would be to reference names that the user gives to each folder path in the deployment config.

Makes sense, happy to keep things as-is for now then. I'm guessing what's missing in this case would be a simpler way to include an alias for different google drive repositories in the config, e.g. 
```ts
config.google_drive.sheet_folder_ids: [{id:'abcd1234': label: 'ZA Sheets'}]`
```
or some similar option that could include in path `ZA Sheets/navigation/modules` or whatever it ends up being. 

But if we want anything like this then lets save for a follow-up.

> 2. Thanks for pointing out the security issue, I hadn't thought about that. If you don't think it's too much of a problem, I'm happy to go ahead anyway. I've normally been giving minimal necessary access rights on the files, so I'll continue to do so and also advise others.

Sounds good - yeah as I mentioned this PR itself doesn't really change the fact that the folder ids themselves are made public in the git repo, which in turn could be used to try accessing the drive url - but if a deployment is more sensitive then can easily limit access to the drive and/or make the content repo private

chrismclarke:
@esmeetewinkel 
Just to note that the errors you mention appear to be authoring issues, the error specifically occurs if the authoring encounters a `begin_` statement without an `end_`

E.g. in one of the issues I can see a `begin_display_group` without a corresponding `end_display_group`

[ZA module_navigation](https://docs.google.com/spreadsheets/d/1dazL3XDoB-wlmlam0hjUSqrJzy7atd6QUJVK2pJou1o/edit?gid=2073969664#gid=2073969664)
image

I've added a small commit to this PR to make the errors more interpretable, now reads as:

image

But otherwise I think this should be good to go

chrismclarke (approved):

PR #2920, opened 2025-04-30
fix: service init defer

chrismclarke:
PR Checklist

## Description

Fix issue where shared-data service would attempt to register even when not in use. E.g. shared-data
image

## Review Notes
Confirm that `shared-data` service no longer tries to initialise (which throws error on any deployment without firebase enabled)

## Dev Notes
Looking at the async service base, it appears that all services were set by default to call their init function after a random delay, with an optional override to call immediately. I think both of these options are somewhat part of legacy code (now pretty much all services explicitly await `ready` calls which in-turn triggers init if not already triggered), however I'm slightly wary of refactoring right now in case of unintended knock-ons

As such I've added a new option to `defer` init, which will explicitly prevent the delayed init being called. I've also removed the `callImmediately` option as no services in the existing codebase set that corresponding arg

## Git Issues

Closes #

## Screenshots/Videos
Example checking code for any services that register init function with the `callImmediately` variable set. This would be indicated by `this.registerInitFunction(..., true)` in previous syntax.
image

### DIFF
--- a/src/app/feature/shared-data/shared-data.service.ts
+++ b/src/app/feature/shared-data/shared-data.service.ts
@@ -49,7 +49,7 @@ export class SharedDataService extends AsyncServiceBase {
   ) {
     super("Shared Data");
     this.provider = getDataProvider(this.config.provider);
-    this.registerInitFunction(this.initialise);
+    this.registerInitFunction(this.initialise, "defer");
   }
 
   /**
@@ -265,6 +265,7 @@ export class SharedDataService extends AsyncServiceBase {
   }
 
   private async initialise() {
+    console.trace("[shared data] init");
     await this.provider.initialise(this.injector);
     await this.dynamicDataService.ready();
   }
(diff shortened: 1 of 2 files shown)


esmeetewinkel (approved):
Functional test passed, this resolves the issue seen.

jfmcquade (approved):
Looks good.

I was initially slightly confused about the use of "defer", as it suggests postponing something rather than never calling it at all. But I see now that deferring in this case means to defer until ready is called for the service, which could mean deferring indefinitely.

chrismclarke:
> Looks good.
> 
> I was initially slightly confused about the use of "defer", as it suggests postponing something rather than never calling it at all. But I see now that deferring in this case means to defer until ready is called for the service, which could mean deferring indefinitely.

Yeah, it's a bit confusing because I feel like the default behaviour should be to not do anything and depend on explicit ready calls, but wanting to avoid too many knock-ons instead need a way to essentially opt-out. Could definitely be a candidate for future optimisation, especially some of these auto-init services can still slow down or block bits of the app if there's a bunch of services all trying to call things on the capacitor native bridge (like checking/scheduling notifications and such)

PR #2929, opened 2025-05-06
[STYLING] TZ theme - Secondary button contrast

### LINKED ISSUE #2754: [FEATURE] PLH Kids TZ: primary button text on secondary bg
**What?**
In the `plh_kids_tz` theme, the text that appears on (round) buttons with secondary background colour is required to be a dark shade of the primary colour. I think `colour-primary-800` or `colour-primary-900` would work.

**Current look vs design**

Image 

Image  Image

**Other details**
Perhaps relevant, @jfmcquade commented [here](https://github.com/IDEMSInternational/open-app-builder/pull/2259#issuecomment-2034922270) that the button text colour is auto computed to be white or black (I actually expected the highest contrast on this yellow shade to be black).
image image

FaithDaka:
PR Checklist

## Description

Secondary button (coloured **yellow**) text colour refactored to darker shade, to increase accessibility.

## Git Issues

Closes #2754 

## Screenshots/Videos
Default:

Refactor:

### DIFF
--- a/src/theme/themes/plh_kids_tz/_overrides.scss
+++ b/src/theme/themes/plh_kids_tz/_overrides.scss
@@ -14,4 +14,11 @@ body[data-theme="plh_kids_tz"] {
       }
     }
   }
+
+  // Button
+  plh-button {
+    .navigation.standard {
+      color: var(--ion-color-primary-shade);
+    }
+  }
 }


jfmcquade (approved):
Looks good, should be good to merge following merge conflict resolution

esmeetewinkel:
@FaithDaka / @jfmcquade I think it looks a bit bad that this is not the same contrast colour as the text colour on white (or is that just my eyesight)? Especially on the home screen where the two appear very close to each other.

image

FaithDaka:
@esmeetewinkel I'll adjust the colour. 

But I've seen that it's easy to create inconsistency when the theme doesn't have a design system. Ideally the `button-text` colour should be a darker shade of that button colour rather than match the `body-text` colour. 

Like in this image, you'll see that the `Header` text in the large card is a different colour from the purple `button-text` at the bottom but they are both dark and a visual heirarchy has been created:



Looking into visual hierarchy of the home screen might have good results.

Issue #2935, opened 2025-05-08
[FEATURE] disabled state for text box component


jfmcquade:
**What?**
The text box component should have a "disabled" state, exposed via a `disabled` param, that makes the element greyed out and not interactive, whilst still displaying the current value.

**How?**
Should be trivial to implement via a new parameter and the existing `disabled` html attribute on `ion-input` (in line with existing components with disabled state, like button, combo box, date time picker)

PR #2938, opened 2025-05-08
Added disabled parameter to text-box and combo-box

ChrisMarsh82:
## Description

Added option for disabled parameter for text-box and combo-box.
With combo-box the disabled-text will replace placeholder text when marked as disabled.

## Git Issues

Closes https://github.com/IDEMSInternational/open-app-builder/issues/2935


ChrisMarsh82:
Not sure what has happened to this pull request but should be 4 files changes

PR #2941, opened 2025-05-09
feat: navigation_bar component style variants

### LINKED ISSUE #2856: [FEATURE] Navigation bar should interpret footer background colour
**What?**
The `navigation_bar` component should amend when the footer background colour is none so that
-  the text colour is a dark shade of the primary colour when the footer background colour is set to none. Currently, the colour of text shown in the `navigation_bar` is `var(--ion-color-primary-contrast)`.
- there is no line appearing underneath the text.

**Other details**
Use case in PLH Kids TZ: 

Image

Image

We had similar for PLH Kids KW but for that a stand-alone component was developed. I'm also happy to integrate that if easier.

jfmcquade:
PR Checklist

## Description

- Adds new variants for styling the `navigation_bar` component:
  - `text_primary_contrast` (default)
    - Sets the text colour to `primary-contrast`
  - `text_primary`
    - Sets the text colour to `primary`
  - This differs slightly from the suggested implementation in #2856 but resolves the fundamental issue. I wanted to keep the `navigation_bar` component independent from the footer settings, so the authoring changes required to resolve the underlying issue would be to update the footer template that instantiates the `navigation_bar` component to set `variant: text_primary` on that component.
- Minor refactoring of `navigation_bar` component code

## Notes

The syntax of these variants is a step towards the utility styles proposed in the discussion on #1971. Eventually, utility styles like this should live in the `style_list` column, but it seems acceptable to supply them as a list to the `variant` param for now. 

After going back and forth about how to handle the authoring of these particular style changes, this halfway house seems like an acceptable option. It would be nice if this styling could be authored in the `style_list`, but setting `color: var(--ion-color-primary);` does not have the expected result because of the nested component structure and the way styles are inherited. I also considered exposing equivalent variants to apply styles to the background (e.g. `background_primary`, `background_none`), however this styling can be applied through a simple css rule (e.g. `background-color: var(--ion-color-primary);` – see example in component demo sheet), so have not included.

## Git Issues

Closes #2856

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/components/navigation-bar/navigation-bar.component.html
+++ b/src/app/shared/components/template/components/navigation-bar/navigation-bar.component.html
@@ -1,14 +1,17 @@
-<div class="navigation-bar-wrapper">
-  <a
-    *ngFor="let button of buttonList"
-    class="button-wrapper"
-    routerLinkActive="active-link"
-    [routerLink]="'template/' + button.target_template"
-    [attr.data-name]="button.name"
-  >
-    <div class="image-wrapper">
-      <img [src]="button.image | plhAsset" />
-    </div>
-    <p *ngIf="button.text">{{ button.text }}</p>
-  </a>
+<div class="navigation-bar-wrapper" [attr.data-variant]="params().variant">
+  @for (button of params().buttonList; track button.name) {
+    <a
+      class="button-wrapper"
+      routerLinkActive="active-link"
+      [routerLink]="'template/' + button.target_template"
+      [attr.data-name]="button.name"
+    >
+      <div class="image-wrapper">
+        <img [src]="button.image | plhAsset" />
+      </div>
+      @if (button.text) {
+        <p>{{ button.text }}</p>
+      }
+    </a>
+  }
 </div>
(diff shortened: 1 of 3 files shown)


jfmcquade:
> From an authoring perspective I'm not sure we really want authors to have to be passing ionic colour variables. Longer term I expect some utility that maps named variables like `primary`, `secondary` could then go through some sort of intermediate process to assign,

Yes I think that's definitely preferable, we certainly haven't encouraged authors to be aware of ionic colour variables up to now

> but short term I'm wondering if it might just make more sense to have `primary` (default using primary colour background and contrast colour text) and `inverted` (swapping primary and background) as variants?

That's a good idea, I think I prefer it to the currently proposed implementation. I feel like `primary` and `inverted` don't necessarily apply well generally in a predictable way (for example, for many components we might expect `primary` to apply the primary colour to the foreground elements rather than the background). But they're probably preferable and more author friendly in this case than the ones I've included that gesture at wider applicability. There is one caveat that there would then be no way of authoring a footer with a dark, non-primary background and light text, but I don't think we have a use case for that.

Let us know what you think @esmeetewinkel.

I'm aware that we often get into the weeds about naming conventions in cases such as this, which is definitely important to get right but also we're often talking about interim solutions anyway. I think it would be good to start prioritising some work towards addressing the issues in #1971, and that's probably something for me to push on.

esmeetewinkel (approved):
Functional test passed, LGTM.

I don't understand how you're proposing that the naming convention using `primary` and `inverted` on the `navigation_bar` component would interact with the footer config setting `config.app_config.APP_FOOTER_DEFAULTS.background`. Or not at all? Either way, happy to leave as is for now.

Issue #2943, opened 2025-05-13
[BUG] NG0955: The provided track expression resulted in duplicated keys for a given collection.


esmeetewinkel:
**Describe the bug**

I've recently started seeing the console warning below. I'm wondering if this indicates something has changed about the order in which certain processes run (cause I'm guessing that beforehand `@item.id` would have been evaluated first, in which case these keys would have been different). 

The error links to https://angular.dev/errors/NG0955. Could this be due to the recent Angular updates? 

```
NG0955: The provided track expression resulted in duplicated keys for a given collection. Adjust the tracking expression such that it uniquely identifies all the items in the collection. Duplicated keys were: 
key "dg_all_modules.loop_modules.module_@item.id" at index "0" and "1", 
key "dg_all_modules.loop_modules.module_@item.id" at index "1" and "2", 
key "dg_all_modules.loop_modules.module_@item.id" at index "2" and "3", 
key "dg_all_modules.loop_modules.module_@item.id" at index "3" and "4", 
key "dg_all_modules.loop_modules.module_@item.id" at index "4" and "5", 
key "dg_all_modules.loop_modules.module_@item.id" at index "5" and "6", 
key "dg_all_modules.loop_modules.module_@item.id" at index "6" and "7", 
key "dg_all_modules.loop_modules.module_@item.id" at index "7" and "8", 
key "dg_all_modules.loop_modules.module_@item.id" at index "8" and "9", 
key "dg_all_modules.loop_modules.module_@item.id" at index "9" and "10", 
key "dg_all_modules.loop_modules.module_@item.id" at index "10" and "11", 
key "dg_all_modules.loop_modules.module_@item.id" at index "11" and "12", 
key "dg_all_modules.loop_modules.module_@item.id" at index "12" and "13", 
key "dg_all_modules.loop_modules.module_@item.id" at index "13" and "14", 
key "dg_all_modules.loop_modules.module_@item.id" at index "14" and "15", 
key "dg_all_modules.loop_modules.module_@item.id" at index "15" and "16", 
key "dg_all_modules.loop_modules.module_@item.id" at index "16" and "17", 
key "dg_all_modules.loop_modules.module_@item.id" at index "17" and "18", 
key "dg_all_modules.loop_modules.module_@item.id" at index "18" and "19", 
key "dg_all_modules.loop_modules.module_@item.id" at index "19" and "20", 
key "dg_all_modules.loop_modules.module_@item.id" at index "20" and "21", 
[...]

PR #2948, opened 2025-05-15
Fix: nested name dynamic

chrismclarke:
PR Checklist

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/components/data-items/data-items.service.spec.ts
+++ b/src/app/shared/components/template/components/data-items/data-items.service.spec.ts
@@ -255,6 +255,14 @@ describe("DataItemsService", () => {
     expect(translateDataListRowsSpy).toHaveBeenCalledTimes(1);
   });
 
+  // TODO - unique id created as part of row processor which is stubbed during tests
+  // Should refactor when replacing hack method with alt. row processor class
+  xit("creates unique _nested_name for row", async () => {
+    const obs = service.getItemsObservable(MOCK_DATA_ITEMS_ROW(), {});
+    const data = await firstValueFrom(obs);
+    expect(data[0]._nested_name).toEqual(`button.id_1`);
+  });
+
   // TODO - requires improvement to mocked dynamic data service
   // it("provides live update when data changes", async () => {
   //   const obs = service.getItemsObservable(MOCK_DATA_ITEMS_ROW, {});
--- a/src/app/shared/components/template/components/data-items/data-items.service.ts
+++ b/src/app/shared/components/template/components/data-items/data-items.service.ts
@@ -236,6 +236,12 @@ export class DataItemsService {
     processor.templateRowMapValues = templateRowMapValues;
 
     await processor.processContainerTemplateRows();
-    return processor.renderedRows();
+    const processedRows = processor.renderedRows();
+
+    // Ensure the _nested_name includes item id for unique id in for loop track
+    return processedRows.map((row) => {
+      row._nested_name = `${row._nested_name}.${row._evalContext.item._id}`;
+      return row;
+    });
   }
 }
--- a/src/app/shared/components/template/services/template-variables.service.spec.ts
+++ b/src/app/shared/components/template/services/template-variables.service.spec.ts
@@ -229,4 +229,27 @@ describe("TemplateVariablesService", () => {
     );
     expect(res).toEqual({ number: 2 });
   });
+
+  it("Evaluates _nested_name metadata fields", async () => {
+    const res = await service.evaluatePLHData(
+      { _nested_name: `button_{@item.id}` },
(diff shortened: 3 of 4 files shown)


jfmcquade (commented):
Looks good.

I've added https://github.com/IDEMSInternational/open-app-builder/pull/2948/commits/f8c359deef021e7116f7c6dfb98a63b660e46bb8 to resolve resolve an issue with the navigation bar component that was causing a console warning of the type described in https://github.com/IDEMSInternational/open-app-builder/issues/2943.

I'm also still getting a single console warning when initialising the [comp_data_items](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892) template. I'm not quite sure which example is responsible, and it may be that this is highlighting a legitimate authoring issue in the row naming.
Screenshot 2025-05-19 at 14 57 46

esmeetewinkel (approved):
1. I can confirm this resolves the error seen in #2943.
2. I'm still seeing a very similar error on the template [activity_library](https://docs.google.com/spreadsheets/d/1ejf7JQSM2-kvclr1oZW6jH_a0K_aEZuGQCYOD7RcO8U/edit?gid=0#gid=0) in the `plh_kids_teens_za` deployment, only that the keys are empty in this case. I think this is the same issue @jfmcquade was seeing. I can't see a problem in the template authoring - all rows in my template have explicitly defined names. Please let me know in case I should log that issue separately. 

   image

3. Not sure it matters, but it seems that row 25 in [menu_sub_tasks_base](https://docs.google.com/spreadsheets/d/1a_caZOGQfuJ6UEXGwdh78q4WBWXAUO0Uz3w391DaDeo/edit?gid=251359046#gid=251359046&range=A25) (`plh_facilitator_ph` deployment) now gets processed quite differently - in particular its name `set_variable_1` is now `undefined_1_@item.id`
   image

chrismclarke:
> I'm also still getting a single console warning when initialising the [comp_data_items](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=845703892#gid=845703892) template. I'm not quite sure which example is responsible, and it may be that this is highlighting a legitimate authoring issue in the row naming

Ah, yes - that's from the edge case where we don't have any child rows of the data_items (but the items are still processed as a means of triggering the data_changed action just with the items themselves, as opposed to rendered child rows). That means the rendered rows are original data_list and so don't have a nested name column. 

I've added a fix with 3d6299a8a097a5db3b24d28061a8f79c0c5f75f0 to avoid rendering anything in this case. It should fix the warning

chrismclarke:
> 2. I'm still seeing a very similar error on the template [activity_library](https://docs.google.com/spreadsheets/d/1ejf7JQSM2-kvclr1oZW6jH_a0K_aEZuGQCYOD7RcO8U/edit?gid=0#gid=0) in the `plh_kids_teens_za` deployment, only that the keys are empty in this case. I think this is the same issue @jfmcquade was seeing. I can't see a problem in the template authoring - all rows in my template have explicitly defined names. Please let me know in case I should log that issue separately.

Should be fixed now

> 3. Not sure it matters, but it seems that row 25 in [menu_sub_tasks_base](https://docs.google.com/spreadsheets/d/1a_caZOGQfuJ6UEXGwdh78q4WBWXAUO0Uz3w391DaDeo/edit?gid=251359046#gid=251359046&range=A25) (`plh_facilitator_ph` deployment) now gets processed quite differently - in particular its name `set_variable_1` is now `undefined_1_@item.id`

I think in general there will be somewhat large diffs on all content repos given the new way that any data_items rows explicitly include the `@item.id` reference in name (for unique keys), which in turn tracks a dynamic reference in the metadata.

I can't think of a quick/sensible way to resolve this for now (would require larger scale update to how we track variables in templates), so I think probably still a better tradeoff for now (more data in sheets as opposed to more code to workaround/add default names during runtime)

Example before/after diffs created for debug repo here:
https://github.com/IDEMSInternational/app-debug-content/pull/177

**HOWEVER** - the fact that undefined values are now creeping in doesn't quite feel right... I'm guessing there might be an edge case when dealing with `set_variable` statements inside items and sub-items lists. Would you be able to confirm before/after on that deployment to see if working as expected? I don't see any other undefined values coming up in the debug repo so guessing it might be a debug case we don't currently have in these sheets (searching for `undefined` in the diff pr shows no results)

jfmcquade (approved):
Thanks for the updates, @chrismclarke. I've looked over the changes since my last review and all looks good.

PR #2949, opened 2025-05-15
fix: parse percentage values from sheets

### LINKED ISSUE #2947: [BUG] Percentage values in sheets are parsed to decimals
**Describe the bug**
When a cell in a sheet contains a percentage value, e.g. `95%`, then the value is parsed to a percentage, e.g. `0.95`. This is the case for values authored in both templates and data lists.

**Context**

Displaying percentage values in a table is required for the `conflict_forecast` deployment.

**App version**
0.19.1

**Debug Sheet(s)**

[debug_parse_percentage](https://docs.google.com/spreadsheets/d/1_x5-AVV5eVcXicjpqcbKXgVZ3US-cWvqB8SdoT5DhVE/edit?gid=1422144003#gid=1422144003):

Image

Image

jfmcquade:
PR Checklist

## Description

When a sheet cell contains a percentage value, e.g. `50%`, then this value is preserved through parsing and remains a string. Previously, the string would be converted to a number, e.g. `0,5`, due to the fact that spreadsheet software stores the raw value as a number with percentage formatting applied.

## Git Issues

Closes #2947

## Screenshots/Videos

[debug_parse_percentage](https://docs.google.com/spreadsheets/d/1_x5-AVV5eVcXicjpqcbKXgVZ3US-cWvqB8SdoT5DhVE/edit?gid=1422144003#gid=1422144003):

Screenshot 2025-05-15 at 12 44 29

Screenshot 2025-05-15 at 12 43 44

### DIFF
--- a/packages/scripts/src/commands/app-data/convert/processors/xlsxWorkbook.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/xlsxWorkbook.ts
@@ -7,7 +7,7 @@ import BaseProcessor from "./base";
 import { existsSync } from "fs-extra";
 import { IContentsEntry, parseAppDataCollectionString } from "../utils";
 
-const cacheVersion = 20250429.0;
+const cacheVersion = 20250515.0;
 const sheetsFolderBaseUrl = "https://drive.google.com/drive/u/0/folders";
 
 export class XLSXWorkbookProcessor extends BaseProcessor<IContentsEntry> {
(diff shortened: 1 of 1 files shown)


packages/scripts/src/commands/app-data/convert/processors/xlsxWorkbook.ts
```diff
@@ -61,6 +61,18 @@ export class XLSXWorkbookProcessor extends BaseProcessor<IContentsEntry> {
           worksheet[cellId].v = html;
         }
       });
+      // If authored value was a percentage, override converted decimal value to preserve percentage representation
+      Object.keys(worksheet).forEach((cellId) => {
```
chrismclarke, line 65: nit(blocking): The codeblock above already loops over all the cells so would probably be a little bit tidier to include the logic there instead of looping over for a second time.

chrismclarke (changes requested):
Makes sense why we would want to preserve percentages as strings (easy enough to provide a decimal value if wanting to be numeric). See minor (blocking) comment inline, 

Adding a couple tests would be nice (as suggested in pr notes), ideally just one minimal test for each of the conversion cases (HTML replacement and percentage conversion)

jfmcquade:
Thanks @chrismclarke, I've tidied up the code and added tests

chrismclarke (approved):
Thanks @jfmcquade - looking much cleaner. 
Looks good to me

esmeetewinkel (approved):
LGTM

PR #2950, opened 2025-05-15
feat: add app env fields

### LINKED ISSUE #2927: [FEATURE] Identify localhost users in pg data
**What?**
Having a way to identify users that are running the app locally when inspecting postgres data. Most of the time we will want to filter out these users.

chrismclarke:
PR Checklist

## Description

Adds two new generated fields for `_app_environment` and `_app_hostname` for use in metadata tracking

## Author Notes
`_app_environment` will show up either as `development` when running locally, or `production` when running either as native app or web preview. Specifically this value represents the app build process, where a development build is used to run locally and a production build when deploying to web, web-preview or native app runtimes.

`_app_hostname` refers to the URL hostname if the app is being delivered on web. When running locally this will show as `localhost`. When running in web/web-preview it will display the partial URL (without http/https and trailing path), such as `plh-teens-app1--pr2885-feat-plh-progress-pa-9jd4wzx2.web.app` for a preview link or `idems-debug.web.app` for web version.
When running on native device this value will either also be `localhost` or simply an empty string depending,although can be distinguished between development localhost either by the app_environment field or the `_platform` field (which will show ios/android on native)

Both of these properties are deployment-independent, and will be automatically populated on app run

## Git Issues

Closes #2927

## Screenshots/Videos

### DIFF
--- a/packages/data-models/fields.ts
+++ b/packages/data-models/fields.ts
@@ -3,7 +3,11 @@
  * They are used to store store computed or exported variables and are not user overridable
  */
 enum PROTECTED_FIELDS {
+  /** Runtime environment, `production` or `development` */
+  APP_ENVIRONMENT = "app_environment",
   APP_FIRST_LAUNCH = "app_first_launch",
+  /** Location hostname of app, e.g. `localhost`, `example.web.app` */
+  APP_HOSTNAME = "app_hostname",
   APP_LANGUAGE = "app_language",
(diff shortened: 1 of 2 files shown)


jfmcquade (approved):
Looks good, simple implementation

esmeetewinkel (approved):
Functional test passed - confirmed locally and on PR web preview the two fields are as described.

image

image

PR #2952, opened 2025-05-20
Handle parent 'changed' actions when setting local variables

markharmon:
PR Checklist

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/services/instance/template-action.service.ts
+++ b/src/app/shared/components/template/services/instance/template-action.service.ts
@@ -390,10 +390,43 @@ export class TemplateActionService extends SyncServiceBase {
       }
     });
 
+    // Handle any actions that should be triggered by the changed value
+    this.handleParentChangedActions(key, value);
+
     // update parent reference in case actions force a re-intialisation
     // (not currently implemented)
     if (this.container.parent) {
       this.container.parent.children[this.container.name] = this.container;
+    }
+  }
+
+  /**
+   * If the row type is template, get parent variables
+   * that match by key and handle any 'changed' events.
+   * @param key variable name
+   * @param value variable value
+   */
+  private handleParentChangedActions(key: string, value: any) {
+    if (this.container?.row?.type === "template") {
+      const parentRows = this.container.row?.rows;
+      const parentVariableRows = parentRows?.filter(
+        (r) => r.type === "set_variable" && r.name === key
+      );
+
+      for (const row of parentVariableRows) {
+        if (!row.action_list) continue;
+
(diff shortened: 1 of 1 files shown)


markharmon:
This method is called after processing a set_local action but seems to redraw the child template in an invalid state

https://github.com/IDEMSInternational/open-app-builder/blob/f829dc8347c640a9007bf8b8d5504e7ac4d6dfca/src/app/shared/components/template/template-component.ts#L176

  /**
   * If the current template is generated as a child of data_items then updates to
   * parent variables will not propagate down. Force reprocessing to workaround
   * See issue https://github.com/IDEMSInternational/open-app-builder/issues/2636
   */
  private async hackForceReprocessNestedTemplate() {
    let row = this._row;

    if (this._row.type === "template") {
      const componentRef = this.componentRef as ComponentRef<TemplateContainerComponent>;
      if (componentRef.instance.parent) {
        // HACK - when a parent updates there may be cases when the updated variables need to be
        // reflected in the child template (e.g. set_nested_variable contains parent local variable)
        // As it is not currently easy to isolate this case, fully recreate any child template when
        // the parent template updates
        if (componentRef.instance.templatename()) {
          this.componentRef.destroy();
          this.renderTemplateComponent(this._row);
          // TODO - test better ways to manage from container, confirming test cases from
          // https://github.com/IDEMSInternational/open-app-builder/issues/2636
        }
      }
    }
  }
}

markharmon:
Hey @chrismclarke do you mind taking a look at this draft, what I'm trying to do is make the action on line 5 here work [sheet](https://docs.google.com/spreadsheets/d/1lD3elJjIf2c3MMyaxAXdeIqZng_i58TjmWCt9jKQYiA/edit?gid=2017326959#gid=2017326959)

It seems to work well if the problematic method hackForceReprocessNestedTemplate mentioned in the comment above is excluded

But this breaks this type of behaviour [sheet](https://docs.google.com/spreadsheets/d/14XXwZ7R_7qIBo5xKYuxnPVQbQidI--PPJ7TVvZxwy1k/edit?gid=1990453574#gid=1990453574) which the hackForceReprocessNestedTemplate is probably resolving.

I think the correct solution is probably to find an alternative solution to hackForceReprocessNestedTemplate but would be good to get your opinion on my approach so far or any ideas you might have. I Hacked the hack in my last commit to try to illustrate the issue.

Thanks

markharmon:
Thanks @chrismclarke I think we'll start by discussing this as a group in the meeting on Thursday

PR #2968, opened 2025-05-28
chore: performance optimisation for set_data action

jfmcquade:
PR Checklist

## Description

Optimises `set_data` action logic, which should make it more performant when applying large numbers of updates at once.

## Git Issues

Closes #

## Screenshots/Videos

@ChrisMarsh82 should understand the following example:

On the `plh_facilitator_ph` deployment, when switching between parent groups on the homepage, the changes in this PR bring a significant performance gain – a drop in processing the changes from ~4 seconds to ~1 second (still quite slow)

| master | PR branch |
|---|---|
| Screenshot 2025-05-28 at 19 07 37 | Screenshot 2025-05-28 at 19 06 20 |


ChrisMarsh82 (approved):
This makes a big different on the home screen when changing between groups. 
There are still issues when looping through pages but this is a big improvement

src/app/shared/services/dynamic-data/actions/set_data.action.spec.ts
```diff
@@ -208,4 +212,93 @@ describe("set_data Action", () => {
...
+  /*************************************************************
+   *  mergeUpdatesById Tests
+   ************************************************************/
+  describe("mergeUpdatesById", () => {
+    it("merges updates for the same ID correctly", () => {
+      const updates = [
+        { id: "123", user: { name: "Jasper" } },
+        { id: "123", user: { age: 9 } },
+        { id: "123", status: "active" },
+        { id: "123", status: "inactive" },
```
chrismclarke, line 225: Question - How exactly would we end up in a case like this from authoring? I thought a `set_data` action already should include all the updates to apply to a specific id e.g. 
```
click | set_data | 
_list_id: feat_data_actions_list,
completed:true
number: 20
```
In this case yes there would be batch update for all items in the data_list, but each item should already be collated for the `completed: true, number:20` partial update? I'm confused how we end up with successive updates to the same ID.

In the case of multiple actions each would be processed in turn so also wouldn't be batched together

jfmcquade:
> > @ChrisMarsh82 should understand the following example:
> > On the `plh_facilitator_ph` deployment, when switching between parent groups on the homepage, the changes in this PR bring a significant performance gain – a drop in processing the changes from ~4 seconds to ~1 second (still quite slow)
> 
> Any way you can help me understand too please? 😅 I have the `plh_facilitator_ph` running locally but don't see any way to change between parent groups. Would be keen to test the performance improvements and see if I can make any further recommendations

OK, you're allowed...

Here's a screen recording showing what's required from a fresh install: setting your user role to "System Tester" and then adding at least 2 parent groups. After that, you can switch between parent groups via the dropdown at the top of the home screen:

https://github.com/user-attachments/assets/103cd09a-401a-4635-a09c-7daa732da96a

I achieved the logs that I've screenshotted in the PR description by bookending the default exported function from `src/app/shared/services/dynamic-data/actions/set_data.action.ts` like this:

Screenshot 2025-06-03 at 15 21 43

src/app/shared/services/dynamic-data/actions/set_data.action.spec.ts
jfmcquade: Ah yes I think you're right. I got caught up in trying to optimise generally without thinking about what is actually possible through authoring. In which case, the performance gains I noticed, if genuine, must all be coming from the batching optimisation.

chrismclarke:
> 
> OK, you're allowed...
> 
> Here's a screen recording showing what's required from a fresh install: setting your user role to "System Tester" and then adding at least 2 parent groups. After that, you can switch between parent groups via the dropdown at the top of the home screen. The logic triggered is authored on row `14` of this [menu_task_flow](https://docs.google.com/spreadsheets/d/1JiKjLEZucBCeFuErhDnIjfTB3dSCrrxg9h0-YHlA7ns/edit?gid=0#gid=0) template. Essentially, it's setting a value for a single property for each row in a 200+ row data list.
> 

Thanks @jfmcquade, always nice to be in with the in-crowd. Also wow, yeah that's one slow update! Even with the fix in this branch my laptop struggles so would be good to optimise a bit better for sure. A couple suggestions

1. Lets try to get https://github.com/IDEMSInternational/open-app-builder/pull/2948 merged to fix issue with console errors and some of the data-items refs that are causing issue

2. I'm guessing this issue was a bit motivation for asking for a `filter` operator on the `set_data` action. That would help to streamline quite a bit, but as that's currently held up a bit it might be worth adding an automated filter to the action so that it first calculates an optimistic update (compare server data with rows for processing), and filters out all ops that would be unchanged. I think can probably use that instead of the custom update combination methods

3. Use a bulk processing method on the db itself. I can see `bulkUpsert` already exists on the service, so I think likely could just use that instead of the custom batching (or exposing a similar `bulkUpdate` if we there is good reason to provide separate update/upsert bulk ops for this use case (I can't really think of one myself, given that the set action already targets specific ids so shouldn't be any risk of accidentally inserting unwanted data)

Let me know if you're happy to work on any of these optimisations as part of this PR, or if you would prefer I create issues for standalone dev

chrismclarke (changes requested):
> 
> OK, you're allowed...
> 
> Here's a screen recording showing what's required from a fresh install: setting your user role to "System Tester" and then adding at least 2 parent groups. After that, you can switch between parent groups via the dropdown at the top of the home screen. The logic triggered is authored on row `14` of this [menu_task_flow](https://docs.google.com/spreadsheets/d/1JiKjLEZucBCeFuErhDnIjfTB3dSCrrxg9h0-YHlA7ns/edit?gid=0#gid=0) template. Essentially, it's setting a value for a single property for each row in a 200+ row data list.
> 

Thanks @jfmcquade, always nice to be in with the in-crowd. Also wow, yeah that's one slow update! Even with the fix in this branch my laptop struggles so would be good to optimise a bit better for sure. A couple suggestions

1. Lets try to get https://github.com/IDEMSInternational/open-app-builder/pull/2948 merged to fix issue with console errors and some of the data-items refs that are causing issue

2. I'm guessing this issue was a bit motivation for asking for a `filter` operator on the `set_data` action. Although I feel like this shouldn't be strictly necessary as the `parseUpdateData` method should be setup to filter out any updates that don't change data. It might be worth investigating a little bit deeper to see what part of the timing takes longest (e.g. pre-processing updates vs db writes), and maybe confirming things like whether the `isItemChanged` is functioning as expected

Taking a quick look at the data going through I think the issue might be because `completed` is being set to `undefined` but original data is `false`. I'd hope that by making that authoring change most updates will get dropped

image

4. Use a bulk processing method on the db itself. I can see `bulkUpsert` already exists on the service, so I think likely could just use that instead of the `hackProcessUpdatesInBatches` method (or exposing a similar `bulkUpdate` if we there is good reason to provide separate update/upsert bulk ops for this use case (I can't really think of one myself, given that the set action already targets specific ids so shouldn't be any risk of accidentally inserting unwanted data)

Let me know if you're happy to work on any of these optimisations as part of this PR, or if you would prefer I create issues for standalone dev

chrismclarke (changes requested):
Alt PR suggested: https://github.com/IDEMSInternational/open-app-builder/pull/2982

PR #2555, opened 2024-11-25
Chore: component base tidying

chrismclarke:
PR Checklist

- [ ] PR title descriptive (can be used in release notes)

**TODO**
- [ ] Builds on top of #2531, update post-merge

## Description

Following on from #2531, some general housekeeping on components to remove manual specification of inherited base properties

## Dev Notes
I'm guessing a lot of these duplicate specifications are due to copy-paste from older templates. There's still a handful of components that don't extend the base template definition (changed a couple as I noticed them), but I haven't extensively reviewed.

## Git Issues

Closes #

## Screenshots/Videos

_If useful, provide screenshot or capture to highlight main changes_

PR #2977, opened 2025-06-03
fix: completely hide data-items component in case of no child rows to render

jfmcquade:
PR Checklist

## Description

Completely hides the `data_items` component in the case where there are no child rows to render – either because there are no child template rows or because there are no item rows returned for the specified data list. Previously, a row representing the data-items component would still render with no content, including default row margins, resulting in unexpected gaps between content.

## Git Issues

Follow-up to issue identified in https://github.com/ParentingForLifelongHealth/plh-facilitator-app-ph-content/issues/87#issuecomment-2930878961

## Screenshots/Videos

[comp_data_items_debug_3](https://docs.google.com/spreadsheets/d/1khGTrDekzqDQyW1r2x0ADqknanrlkmhOz1RRzhP6-7c/edit?gid=820434161#gid=820434161):

Screenshot 2025-06-04 at 15 26 41

| `master` | PR branch |
|---|---|
| Screenshot 2025-06-04 at 15 24 45 Screenshot 2025-06-04 at 15 24 54| Screenshot 2025-06-04 at 15 23 51 Screenshot 2025-06-04 at 15 24 03 |

### DIFF
--- a/src/app/shared/components/template/components/data-items/data-items.component.ts
+++ b/src/app/shared/components/template/components/data-items/data-items.component.ts
@@ -47,11 +47,7 @@ export class TmplDataItemsComponent extends TemplateBaseComponent {
     // If there are no child rows to render, hide the parent container (otherwise it will be visible as a blank row with default margins)
     effect(() => {
       const shouldHide = !this.rowSignal().rows?.length || !this.itemRows()?.length;
-      if (shouldHide) {
-        this.parent.elRef.nativeElement.setAttribute("data-hidden", "true");
-      } else {
-        this.parent.elRef.nativeElement.setAttribute("data-hidden", "false");
-      }
+      this._row.hidden = shouldHide;
     });
   }
 


chrismclarke (changes requested):
Unfortunately this does not work for me when navigating to the debug page from the template list page (but surprisingly does when reloading directly)

If I had to guess I'd imagine maybe there's a change-detection issue going on here when setting the row property directly - none of the template-component code is setup to use signals so the HostBinding update is dependent on ngZone picking up the changes as required. 

The general pattern also seems a little bit odd where the parent renders the child, but then the child tells the parent a property should be different forcing the parent to re-render...

image

Annoyingly the templating code exposes a `parent` component property however this refers to the `plh-template-container` component which sits 2 levels up, and not the direct `plh-template-component` parent.  So I'm guessing as a workaround we would either want to:

1) Use the data items element ref to navigate the dom tree up one level to the `plh-template-component` and just modify styles directly on the HTML component (ignoring all angular change detection)

2) Add bindings to the `template-component` to make component itself accessible from children (alongside the current `parent` reference), and add methods there that can be triggered from the template base component to toggle visibility

If wanting to go the route of (2), we would likely need to tidy up the inconsistencies in how components sometimes do and sometimes don't correctly extend the TemplateBaseComponent - this was started in https://github.com/IDEMSInternational/open-app-builder/pull/2555 but I'm not sure why it was abandoned

I'm also somewhat reluctant to go down path (1) and modify the dom directly as would be somewhat prone to breaking if we were to refactor hierarchy in the future (no type-safety). Open to suggestion/discussion

chrismclarke:
@jfmcquade - I've made an attempt at implement option 2 with #2978, quite a bit more code but I think it should behave a little more consistently - let me know what you think

Issue #2979, opened 2025-06-05
[BUG] Firestore issues on iOS


jfmcquade:
**Describe the bug**
When running on iOS, the shared parent group functionality does not work, throwing a firebase permissions error when attempting to connect to firestore. The shared parent group functionality (#2910) which is built upon the shared data functionality (#2899), which in turn is built upon firebase firestore.

The firebase related services seem to initialise correctly, but when attempting an operation that queries firestore (e.g. the `PlhParentGroupService.handlePull()` method), the following error is thrown:

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/26744cb5-9183-4e48-aa36-4e849a070356" />

This behaviour is consistent whether logged in via Apple or Google. The app running on web platforms does not have this issue, it is specific to iOS.

**Possible causes**

One possible cause is a discrepancy between the firebase libraries used. We're using [@capacitor-firebase/authentication](https://www.npmjs.com/package/@capacitor-firebase/authentication) for authentication, and [@firebase/firestore](https://www.npmjs.com/package/@firebase/firestore) for firestore. The former uses the native SDKs from Firebase, Google, etc., uses native Firebase SDKs on iOS/Android, whereas the latter uses the web SDK. So potentially the two may not always share authentication state correctly. Therefore one approach to try is to migrate from using the `firebase/firestore` library to instead use the capacitor-specific [@capacitor-firebase/firestore](https://www.npmjs.com/package/@capacitor-firebase/firestore)

**App version**
Code: `0.19.4`
`plh_facilitator_ph` deployment content: `1.2.2`

**To Reproduce**
_Describe the steps to reproduce this bug._

**Deployment Sheet(s)**
_Specify the sheet name and include a link where the bug appears if appropriate._

**Debug Sheet(s)**
_If you have a debug sheet to show demonstrate the bug specify its name and provide a link._

**Screenshots and videos**
_If applicable, add screenshots or videos to help explain your problem._

**Additional context**
Additionally, the firestore functionality does not work at all when running in an iOS emulator, throwing the following error on init. This issue seems to be present only when running in an iOS simulator – it is not present when running the same app build on a connected iOS device through xcode.

<img width="1000" alt="Image" src="https://github.com/user-attachments/assets/66e0ef38-e4a1-4f2e-bbab-2c6461d073a3" />

One fix I attempted was to use the `experimentalForceLongPolling: true` option when initialising firestore, but this didn't resolve the issue:

[...]

PR #2980, opened 2025-06-05
Fix: Apple auth

jfmcquade:
PR Checklist

## Description

- Handles automated login on app init in the case that user logged in using Apple (previously assumed Google was used)
- Handles case where `firebase.crashlytics` is not defined in deployment config
- Includes latest iOS podfile updates (following #2959)

## Git Issues

Preliminary step towards resolving #2979

## Screenshots/Videos


chrismclarke (approved):
Changes here make sense to me, can see how it is always trying to sign in with google auth when app might be using apple.

I don't have access to an apple device for testing, but given the current implementation is broken anyway I think should be fine to just test once in production.

I've added one small tidying commit eb626db, but otherwise I think should be good to go. I'm going to go ahead and merge here just to tidy up the diff on #2981

PR #2981, opened 2025-06-05
fix: shared data on iOS

### LINKED ISSUE #2979: [BUG] Firestore issues on iOS

jfmcquade:
PR Checklist

- [x] Builds on #2980 (should be merged first)

## Description

Fixes an issue where shared data would not work as expected on iOS, due to issues with firestore authentication.

## Testing

Since this refactors the firestore functionality, the following should be tested (on web and Android platforms) to ensure still working as expected:
- The core shared data functionality should be tested via the `/shared-data` debug page introduced in #2899.
- The "shared parent group" functionality should be tested, see screen recording in PR description of #2910.

## Dev notes 

This PR migrates the firebase shared data provider (`src/app/feature/shared-data/providers/firebase.ts`) to use the [@capacitor-firebase/firestore](https://www.npmjs.com/package/@capacitor-firebase/firestore) plugin instead of [@firebase/firestore](https://www.npmjs.com/package/@firebase/firestore) used previously. I believe the cause of the issues was the way in which authentication is handled by these packages: the `@firebase/firestore` uses the web SDK on all platforms, whereas the capacitor specific plugin interops with the [@capacitor-firebase/authentication](https://www.npmjs.com/package/@capacitor-firebase/authentication) plugin, which uses the native SDKs on native platforms.

The code migration was aided by AI, and should be reviewed thoroughly.

## Git Issues

Closes #2979 (including the emulator-specific issue)
Closes https://github.com/ParentingForLifelongHealth/plh-facilitator-app-ph-content/issues/165

## Screenshots/Videos

Shared parent group data successfully "pulled" on `plh_facilitator_ph` app (previously would throw firestore permissions error):

Screenshot 2025-06-05 at 12 41 23

Screenshot 2025-06-05 at 12 25 04


jfmcquade:
Thanks @chrismclarke, it's good to have such a clear explanation of the underlying issue. 

However I'm not sure I agree with the suggestion that maintaining both a capacitor SDK and a web SDK is the best option. My understanding is that using the capacitor library should work on web platforms anyway – just as `@capacitor-firebase/authentication` works across both web and native platforms. The `@capacitor-firebase/firestore` package should be automatically using the native SDK or web SDK as appropriate to the current platform, in accordance with how capacitor works across platforms generally.

In the conversation above, the reason that gemini 2.5 pro gave for why we might want to support both packages is that we could have too much existing code based on the web SDK package that would make migration difficult. But I don't think that applies to us – we would only need to migrate the code in that one service provider (the changes included in this PR), which, whilst not trivial due to the different APIs of the two packages, isn't a huge change and seems simpler than maintaining support for two separate packages.

So I think we want to push ahead with a migration of the type included in this PR.

Seeing as we want to release the `plh_facilitator_app` today, with a final release at the end of next week, I think we have two options for today's release:
1. Build the app using this PR branch (at least for the iOS release), and test the functionality before the final release next week
2. Merge changes to add error handling to the parent group service, to avoid the firebase error from blocking, as in https://github.com/ParentingForLifelongHealth/plh-facilitator-app-ph-content/issues/165. And acknowledge that the shared parent group functionality will not currently work on iOS.

Seeing as we have some time to test next week, I would favour option 1. I will be on leave so will be unable to test on iOS, but I believe it should be possible via the github action and appetize. On reflection, we probably want to implement the error handling of option 2 either way, so I will make a separate PR for that.

jfmcquade:
> Do you know if anyone else is able to test things on IOS? I can access simulator but I know you said generally things just do not work there (do you know if it's a simulator-specific issue, or more around debug keys being added to firebase?).

The simulator-related issue I was seeing with firestore (a disparity between the behaviour on simulator vs physical device) does seem to be resolved by using the capacitor firestore library – it isn't present when running from this PR branch in its current state.

> I think it's going to be difficult to rush a deployment without being able to test. Although if these issues are also affecting Android I'm happy to try more debugging on Android native. In which case I would probably suggest prioritising web and android releases, and delaying ios until your back

Yeh I think you're right, we can't really rush it. If you have capacity, it would be good if you could proceed with debugging this PR on web and Android next week, and iOS via a simulator if possible. But I can handle the iOS testing when I'm back.
I'm actually not sure whether the issues with the shared data functionality were present on Android or not, @ChrisMarsh82 did you manage to test on Android?

chrismclarke:
@jfmcquade - you're right, there's full support for web in the capacitor-firebase version
image

chrismclarke:
Quick update on method A) - whilst I can modify methods to avoid the references to `Timestamp()` and `serverTimestamp()` on native, the default handling of date objects on android is to call the `toJson()` method on date objects when passing across the native bridge, which in turn converts dates to isoString

There wouldn't be huge harm in doing this, were it not for the fact that we have existing data stored as firestore timestamps, and so would break existing queries (where some of the `_created_at` is a string value and others a firestore Timestamp

So I will need to first check how reasonable it would be to do this. The other options of either forcing use of only web SDK (and syncing auth token) would still be valid, as would (potentially) be using the pr branch that generates correct timestamp formats on the native layer

ChrisMarsh82 (approved):
This doesn't break anything so should be good to merge then we can test on apps

ChrisMarsh82:
@jfmcquade just to note, there was a change that meant the group_id was sent as a number if it was a number, although previously this was always a string (it is a string in our database). This stopped any parents from being sent as it could not match them with the group_id. This change is not in this PR as its the same on master. I have solved the issue with a quick fix of adding a underscore to the beginning of group_id.

chrismclarke:
@jfmcquade @ChrisMarsh82 
Just to provide another quick update, I've now tested syncing auth from native to web layer so that the native auth credentials can be used when making firestore requests using web api and it's working - I also checked returning to my device after a weekend to see if both credentials updated as expected and they did.

I'm still not 100% confident in this approach as I'm not sure whether there would be any impact on existing signed in users whose credentials are not synced (would likely need to sign out and back in again after this PR). So I think I'd still be keen to raise a PR that provides a more unified solution that uses native-only version for both auth and firestore, but with same code as web (breaking change as would need to refactor timestamp management)

So as a quick summary of current state of things:

**Problem Diagnosis**
Firebase has separate SDKs that is used on native and web platforms, that do not communicate with each other. On web platforms this is not an issue as we use the web SDK for both, but on native we use the native SDK for auth, but the web SDK for firestore requests. This means that any requests to firestore endpoints that sit behind security rules requiring authenticated user fail, due to the fact the request coming from the web layer does not include any auth credentials (only exist in native layer).

Whilst it is possible to send firestore requests directly from the native layer (using capacitor-firebase), there are some incompatibilities between the two SDKs, notably the inability to use firebase's custom `Timestamp` generator for specifying things like document `_created` and `_updated_at` fields. This is a limitation with the capacitor-firebase SDK, with various open issues and PRs to address in the future

**State of Potential Solutions**
There were 3 potential solutions identified to resolve this:

**A) Refactor code to avoid using field values**

Initially I tried implementing separate adapters for web/native platforms, using standard JavaScript Date objects to record timestamps instead of the firebase Timestamp generator. This didn't work very well as the date objects were automatically converted to ISOString timestamps when sending to the server (due to the way data is serialised when passing between frontend JS and native layers). This had a knock-on of meaning data generated on web was incompatible with data generated on native device (queries for most recent docs would fail, as some of the data was stored as firestore Timestamps and some was just string values)

A better workaround would be to refactor all data to just use isoStrings instead of timestamps, however this would be a breaking change. I think this would still likely be the best approach and plan to create a follow-up PR to test (this thread has gotten quite long!)

**B) Target PR branch for dependency and try to use from there**

As mentioned, there is an open PR to add support for firestore Timestamps to the capacitor native SDK. In theory it would be possible to target our codebase at that branch, although as there are few guarantees the PR would be merged as-is or receive future updates, it is unlikely to be a very sustainable solution. Equally we could try to maintain our own fork, but would add a lot of technical overhead that would be better-off avoided. 

If the issue was less time-sensitive likely the best option would be to try prompt the library maintainer to merge the PR (providing additional support as useful) and wait for next release

**C) Use only web SDKs and add workaround to sync the auth between native auth layer and web auth layer**

In theory it is possible to use the same credential to sign in to auth on both web and native layers concurrently, so that requests made from either SDK should contain the correct auth info. This is implemented in this branch currently and appears to be working (hence will stop active development on this branch) - however it does add a bit of tech-debt to the code (any time we add new sign-in providers beyond apple and google we would need to remember to manually add code to sync auth across web-native sdks, because the implementation varies slightly by auth provider). 

So it should be possible to test on native device targetting this branch, however I would still like to open up an alternate PR to revisit option A

PR #2983, opened 2025-06-06
Chore: plh parent group error-handling

jfmcquade:
PR Checklist

## Description

Adds error handling for core methods of plh-parent-group service. This should prevent errors with this service from blocking other processes.

## Git Issues

A workaround for https://github.com/ParentingForLifelongHealth/plh-facilitator-app-ph-content/issues/165
Doesn't address the underlying issue (#2979), but prevents the uncaught exception from crashing the app.

## Screenshots/Videos

Demonstration of `plh_facilitator_ph` home screen functionality. Errors are still thrown, but do not interrupt other processes.

https://github.com/user-attachments/assets/2311389d-306c-486e-834f-f4f4dba6b49b


jfmcquade:
> I'm not sure I quite understand the point of logging and ignoring what seems like a fairly critical error for the functioning of the app. I'm guessing this is related to auth issues on native, in which case maybe would be better to opt-out the service/method on native and show a temporary user alert/message? This would likely take a lot less code and be much easier to revert once fixed than adding multiple try-catch blocks
> 
> Or is there another reason why it might be desirable for methods to fail and be ignored?

Hmm yeh, good point.

My thought was that there me be multiple reasons why firestore can't be reached at any given time. And without an API for authors to get a success/fail status back from an action they fire, we might need an alternate way of handling the exceptions.

One intermediate approach might be to consider changing the authoring pattern: if we change the order of actions fired when clicking a button so that the `parent_group` actions are fired last in the list, then at least their failure won't result in the other button logic not being run. @ChrisMarsh82, I'm thinking in particular of the `groups_go_to` button on [menu_override](https://docs.google.com/spreadsheets/d/1JiKjLEZucBCeFuErhDnIjfTB3dSCrrxg9h0-YHlA7ns/edit?gid=0#gid=0):

Screenshot 2025-06-06 at 16 50 23

But I think we probably need to think along the lines you suggest. Possibly something within the service that checks in the background if firestore is available, and if it isn't, disables the relevant methods in the background, showing a user-facing warning? It might initially be a bit difficult for this to integrate with the authoring system. Ultimately, it feels like we might need something resembling API response status codes for authors to respond to (@ChrisMarsh82, I know you have felt this need).

chrismclarke:
I think probably best to close here for now, and can reopen/revisit depending on whether the underlying issue is resolved by related PRs

PR #2991, opened 2025-06-11
fix!: shared data on iOS (alt implementation)

### LINKED ISSUE #2979: [BUG] Firestore issues on iOS

chrismclarke:
PR Checklist

## Breaking Changes
- Minor update to how auth is managed on native device will likely result in existing native users being forced to log in again
- Changes to shared_data timestamps may require migration on server (see author notes)

## Description

Replacement for https://github.com/IDEMSInternational/open-app-builder/pull/2981, but built on top to keep git history

This creates a unified firebase adaptor which correctly uses native sdks for both auth and firestore on native, and web sdks on web. In doing so some specific functionality previously only available on web has been dropped (namely creating firestore timestamps, now instead using iso strings)

## Review Notes
Should test shared-data system on web, android and ios. Would also be good to check existing data still synced as expected.

Shared data debug page available at http://localhost:4200/shared-data, or can use any of the deployments that make use in specific pages/features/components

E.g. shared-data debug page correctly shows both legacy and newer docs (for me as signed in user)
image

## Author Notes

Whilst testing locally it seems that the expected data incompatibility from timestamp changes actually isn't an issue. It appears (to me at least) that firestore is smart enough to retrieve legacy docs with timestamps when search for latest docs by a string value. 

E.g. - legacy doc stored metadata using firestore Timestamp
image

Now docs store using isoString
image

If any incompatibilities are detected any legacy server docs could be manually migrated (updated_at will update whenever doc updates, but the created_at would need to be manually converted)

## Dev Notes
The firebase web SDK methods have been fully re-written to work with the capacitor-firebase package. I gave a couple passes with ai-support but still ended up rewriting quite a bit of the code so I think it should hopefully be in a reasonable state. Changes in `src/app/feature/shared-data/providers/firebase.ts`

I also made some updates to the firebase auth implementation to remove previous workaround that partially worked to handle some cross-platform compatibility issues (persisting native auth to indexeddb, which should no longer be necessary), as well as adding a unified event listener to handle detecting when automatic login complete (web has standalone method that was used, but native wasn't waiting correctly). Changes in `src/app/shared/services/auth/providers/firebase.auth.ts`

Testing potential incompatibility is a little awkward due to the way docs are cached locally - however given that the legacy local cache converts timestamps to isoString anyways I think should be reasonably safe to just keep all endpoints as-is. Additionally I think the shared-data feature wasn't being heavily used in production anyways. That being said, it would probably be a sensible follow-up to add support for automatically clearing cached documents when a user signs out (follow-up git issue?)

Moving forwards it would be important to avoid using the JS sdks and only use capacitor-firebase instead. I had tried to add an eslint rule to throw an error via the `no-restricted-imports` plugin, however for some reason either my eslint isn't working properly locally or there might be upgrade issues. So I'll hold off resolving for now and could also be a future update

## Git Issues

Closes #2981
Closes #2979 (including the emulator-specific issue)
Closes https://github.com/ParentingForLifelongHealth/plh-facilitator-app-ph-content/issues/165

## Screenshots/Videos


jfmcquade:
I've tested the auth functionality on both iOS and Android emulators, for both Google sign in and Apple (iOS only). Everything seems to be working as it should, and logging in persists over closing and relaunching the app.

Appetize links for reference:
- [iOS](https://appetize.io/apps/ios/international.idems.debug-app/builds/b_k2esooozbtl44443qlaag7uxie)
- [Android](https://appetize.io/apps/android/international.idems.debug_app/builds/b_il3ca4sg4onrgazng7yrewfoha)

@ChrisMarsh82 has confirmed that the shared parent group functionality is working as expected for the `plh_facilitator_ph` deployment, including accessing pre-existing shared data.

So I think we should be good to merge.

Dev note: I had to add my SHA keys to the Android app in the firebase console to enable Google sign-in on Android. Before doing so, I was getting an unhelpful error:
```
ERROR Error: 10: 
    at returnResult (<anonymous>:947:32)
    at win.androidBridge.onmessage (<anonymous>:922:21)
```
I viewed the relevant keys by running `keytool -keystore ~/.android/debug.keystore -list -v -alias androiddebugkey -storepass android -keypass android`. This should now configure Android builds of the `debug` deployment that I build locally to make use of Google sign-in through firebase, but I believe that they are unique dev credentials and so if wanting to test, you would need to add your own SHA keys to the app in Firebase.

jfmcquade:
> If building via CI it should also be possible to re-use the same creds that are used in production (seen notes in #2836), although don't think any of these were ever added to the debug repo so makes sense to just add your own local dev keys (I did the same with mine)

Ah yeh, makes sense. I had tried building via CI, and assumed that the SHA keys already in firebase were the relevant production ones, but the login feature was still failing. I have now added the production SHA keys (from Google Play Console) to firebase, so testing auth on debug builds produced through the github action should now be possible, although I haven't tested again.

PR #2992, opened 2025-06-12
chore: lint rules fix and update

chrismclarke:
PR Checklist

## Description

Following #2991, adds eslint rule to prevent users accidentally importing the firestore web SDK 

## Review Notes
PR tests should fail until after #2991 merged and firestore web SDK removed.
If running locally should see error in vscode

## Dev Notes
I couldn't figure out why I wasn't getting vscode linting popups for the longest time. It turns out to be due to some awkward incompatibilities between the version of eslint that angular uses automatically, version installed in the project, and version expected by the vscode eslint extension. 

Given that these can often change without core dev work (e.g. local devs use their own eslint extension, and angular isn't transparent about version of typescript-eslint in use), I've opted simply to patch the current version of eslint to avoid throwing the incompatibility errors. Linting should still work fine without, and hopefully the patch reference will serve as a reminder to review when upgrading eslint.

## Git Issues

Closes #

## Screenshots/Videos
Example vscode highlighting for import error and recommendation
image

### DIFF
--- a/android/app/capacitor.build.gradle
+++ b/android/app/capacitor.build.gradle
@@ -12,6 +12,7 @@ dependencies {
     implementation project(':capacitor-community-file-opener')
     implementation project(':capacitor-firebase-authentication')
     implementation project(':capacitor-firebase-crashlytics')
+    implementation project(':capacitor-firebase-firestore')
     implementation project(':capacitor-firebase-performance')
     implementation project(':capacitor-app')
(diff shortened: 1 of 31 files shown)


jfmcquade (approved):
Looks good, the changes to the eslint config make sense. Maintaining the patch could be a bit fiddly, but I can see how it might be time better spent than tracking down the cause of the referenced error.

PR #2997, opened 2025-06-18
feat: translate text component value to specified language

### LINKED ISSUE #2996: [FEATURE] Set text in text component to display in arbitrary language
**What?**
A text component should be able to translate its source text to a named language, specified by language code, which may not be the app language.

The actual value of the text component should update, so that the value can be use elsewhere, e.g. copied or shared outside the app via an action.

**Why?**
@ChrisMarsh82 Detailed a request for the facilitator app where a facilitator may want to share a message in one of a number of languages, which may not match the app language that the facilitator themselves is using.

**How?**
Likely a param on the text component, e.g. `language_code`, that dynamically translates the text value when changed.

**Other details**
In theory this logic might be better suited to sitting outside of a specific component, but it's not clear where that would be, and the current use case is urgent

jfmcquade:
PR Checklist

## Description

Exposes a new parameter on the text component, `language_code`. If provided, and if a valid language code for which translations exist for the current app and string in question, then the value of the text component will be translated to this language, which may not match the current app language.

## Dev notes

Because of the requirement of the issue that the actual value of the component must change, not just the value rendered in the DOM, there are some complications with the implementation that involve storing the previous language in order to translate from a non-default language back to the default. Additionally, the fact that the default language doesn't have a translations file associated with it results in a reverse object lookup being used.

## Git Issues

Closes #2996 

## Screenshots/Videos

[feature_translate_text](https://docs.google.com/spreadsheets/d/1gg953sra_4FnUFKcAkoZQKxM0NHHNIc2qA9-z26Azdw/edit?gid=569531329#gid=569531329)

Screenshot 2025-06-18 at 17 13 17

Screenshot 2025-06-18 at 17 13 05

### DIFF
--- a/src/app/shared/components/template/components/text/text.component.html
+++ b/src/app/shared/components/template/components/text/text.component.html
@@ -1,14 +1,12 @@
-@if (this._row.value) {
+@if (this.value()) {
   <div
     class="large standard normal"
-    [class]="params.style"
+    [class]="params().style"
     [ngStyle]="!hasTextValue ? { display: 'none' } : { display: 'block' }"
     [innerHTML]="
-      params.type === 'numbered'
-        ? (this._row.value | number)
-        : (this._row.value?.toString() | markdown)
+      params().type === 'numbered' ? (this.value() | number) : (this.value()?.toString() | markdown)
     "
     [style]="_row.style_list | styleList"
-    [style.textAlign]="params.textAlign"
+    [style.textAlign]="params().textAlign"
   ></div>
 }
--- a/src/app/shared/components/template/components/text/text.component.ts
+++ b/src/app/shared/components/template/components/text/text.component.ts
@@ -1,11 +1,18 @@
-import { Component, OnInit } from "@angular/core";
+import { Component, computed, effect } from "@angular/core";
 import { TemplateBaseComponent } from "../base";
 import { getStringParamFromTemplateRow } from "../../../../utils";
+import { FlowTypes } from "packages/data-models";
+import { TemplateTranslateService } from "../../services/template-translate.service";
 
 interface ITextParams {
-  style: string | null;
-  textAlign: string | null;
-  type: string;
+  style?: string | null;
+  textAlign?: string | null;
+  type?: string;
+  /**
+   * TEMPLATE PARAMETER: "language_code". The target language code to translate the value to.
+   * Used for displaying text in a different language to the global app language.
+   * */
+  languageCode?: string;
 }
 
 @Component({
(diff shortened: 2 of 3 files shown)


chrismclarke (changes requested):
I'm wondering if this is the best approach to resolve the issue. I think adding a new feature but only for the text component might be a little bit confusing as authors might expect similar functionality to be available more generally - e.g. if developing an app that included both English and translated text side-by-side either for educational or debugging purposes.

Do you think it might be possible instead to use intermediate variables with a parameter that enforces the language? E.g. 

|type|name|value|parameter_list|
|---|---|---|---|
|set_variable| source_text | hello | |
|set_variable| text_in_spanish | @local.source_text| language: es|
|text| |@local.source_text| | |
|text| |@local.text_in_spanish| | |

There might be slight challenges in making sure the intermediate variable evaluation bypasses translation, but at least then could be used in any component, parameter list etc.

jfmcquade:
> I'm wondering if this is the best approach to resolve the issue. I think adding a new feature but only for the text component might be a little bit confusing as authors might expect similar functionality to be available more generally - e.g. if developing an app that included both English and translated text side-by-side either for educational or debugging purposes.
> 
> Do you think it might be possible instead to use intermediate variables with a parameter that enforces the language? E.g.
> 
> type	name	value	parameter_list
> set_variable	source_text	hello	
> set_variable	text_in_spanish	@local.source_text	language: es_sp
> text		@local.source_text	
> text		@local.text_in_spanish	
> There might be slight challenges in making sure the intermediate variable evaluation bypasses translation, but at least then could be used in any component, parameter list etc.

That would definitely be a nicer way to do it, agreed that it doesn't really make sense to limit the functionality to a single component's value. But I am a bit unsure of how to implement it – it seems like the code for the targeted language would need to be attached to the value as metadata in order to ensure that it doesn't get translated automatically to the app language and also can be translated back to its original language, but that feels like a significant change to how local variable data is handled and I'm a bit reluctant to touch that code as it's used so extensively. I will investigate though and see how it could be achieved.

Another option would be to move the logic that I've currently put in the text component into the general base component, so that any component can specify the language that it's value should be translated to. We don't currently have any parameters that are common to all components, but I don't see why we shouldn't. However this would still be limited by the fact that it would only allow for direct translation of component _values_, it wouldn't allow for those values to be directly passed to parameter lists etc. without an intermediate component. So allowing variables themselves to have a specified language would be much more powerful. 

As always, I'm aware of the need to balance the immediate use case with a more robust approach. I think that the requested functionality is pretty niche so doesn't necessarily warrant a lot of effort to generalise it, but I can see how the current implementation is not a desirable workaround really.

ChrisMarsh82 (approved):
This looks good to me. Would fit the purpose. I agree that having it on set_variable would be better and give greater flexibility but that could be a follow up issue if it is much bigger

PR #3006, opened 2025-06-19
feat: translate local var from default language to specified language when setting

### LINKED ISSUE #2996: [FEATURE] Set text in text component to display in arbitrary language

jfmcquade:
PR Checklist

## Description

Alternative to #2997.

Exposes an optional parameter, `language`, on a `set_variable` (aka `set_local`) row. This takes a language code (e.g. `gb_en`, `tz_sw`, etc.), which specifies a language to which the value should be translated, if available. This language may be different to the app language, but a translation for the string must exist, configured in the usual way.

Note that this only works for translating instances of strings that are originally in the app's default language into a different language. The reverse is not currently possible (see comments below for more details. This is acceptable for the immediate foreseeable use cases.

## Git Issues

Closes #2997, closes #2996

## Screenshots/Videos

[feature_translate_text](https://docs.google.com/spreadsheets/d/1gg953sra_4FnUFKcAkoZQKxM0NHHNIc2qA9-z26Azdw/edit?gid=1259420210#gid=1259420210)

Screenshot 2025-06-19 at 15 10 15

Screenshot 2025-06-19 at 15 11 54
Screenshot 2025-06-19 at 15 12 07


chrismclarke:
@jfmcquade - just to quickly check in as I think I overlooked the latest updates on this one. Is there still anything you think required to address the questions raised by Chris M. or is it ready for full review?

jfmcquade:
@chrismclarke I'd appreciate your input as I'm struggling to see how to achieve the desired functionality.

When it comes to having the app language set to a non-default language and the specified var being displayed in the default app language, there are two complications:
1. (minor) Translating from a non-default language to the default language is non-trivial since we only store translation strings for non-default languages. If we want to supply a base string in a non-default language and have it translated to the default language, this could be achieved by exposing two parameters when setting a variable, `source_language` and `target_language` and then using a reverse look-up using the source language's translation strings to translate to the default language (I originally used this approach in the old PR for this issue, [here](https://github.com/IDEMSInternational/open-app-builder/pull/2997/files#diff-edb52794ce4b3ee3d6779ea8ba378e3ff6719a908ad5fd666f586c07d9ce0fd7). However, I think it's a reasonable restriction for now to require that variables that are intended for targeted translation in this way should always be supplied in the default language, so resolving this doesn't seem necessary.
2. (major) A more difficult challenge is in ensuring that the variable does not get translated back to the app language (as predicted in https://github.com/IDEMSInternational/open-app-builder/pull/2997#pullrequestreview-2939956999), since a corresponding translation does exist. I can imagine this working via some `do_not_translate` metadata property on the variable, however adding metadata to the existing variable system in such a way that it is available wherever variables are translated is proving difficult. (We already have the `exclude_from_translation` property on rows, but this is to flag data within the translation generation system, i.e. crowdin, as opposed to being used at runtime, and is a property of rows rather than variables)

I can see two possible ways to move forward and resolve issue 2.:
a. Implement a variable metadata system. The core `translateValue` method takes a single `value` argument, and it would be difficult to refactor all the variable code to pass through an independent metadata property. So one way this could be achieved as a workaround by including the metadata in the string itself: e.g. the string value being assigned to a variable that is marked as being in the app's default language (via the new `language` param) gets a prefix attached to it when being set, e.g. `{NOT FOR TRANSLATION}How are you feeling today?", which is then read and removed by the `TemplateTranslateService.translateValue()` method. 
b. Return to the approach of #2997 where the translation is handled at the component level rather than the variable level. It may be an improvement on that PR to create a new dedicated `translated_text` component as opposed to adding functionality to the existing `text` component, which would help to keep the distinction clear that the functionality is not available on all components.

jfmcquade:
@ChrisMarsh82 Following our discussion earlier in the week, I've updated the PR description and added comments to the code (https://github.com/IDEMSInternational/open-app-builder/pull/3006/commits/6a6232286fa543ab9fe03a48e99abb32e165b0a9) to make explicit the limitation that this currently only works for translating default language -> non-default language. If working as expected for that use case, I think should be good to merge

chrismclarke (approved):
Thanks for the work on this @jfmcquade - I think this is definitely the best compromise for now given the challenge of bypassing default translation behaviours

PR #3008, opened 2025-06-19
Fix(!): display group variable hoisting

### LINKED ISSUE #2942: Data items within display group
**Details**
I regularly replicate (and adapt) the example [here](https://docs.google.com/spreadsheets/d/1kpRKj95Js24IYsYwT-3lbfFO-XAOuiDTycGZih0DWOU/edit?gid=345767231#gid=345767231&range=G4) to make sure the components displayed on the screen await the data items loop evaluation. 

I found that this doesn't work when the computation is done within a display group. See debug sheet [debug_data_actions_dg](https://docs.google.com/spreadsheets/d/1kpRKj95Js24IYsYwT-3lbfFO-XAOuiDTycGZih0DWOU/edit?gid=1079947998#gid=1079947998). 

@chrismclarke is this the behaviour you would expect or is this a bug? 

**Screenshots and videos**

Image

Image

### LINKED ISSUE #2989: [BUG] Local variable set by radio group in display group doesn't update
**Describe the bug**
_Include a clear and concise description of what the bug is._

**App version**

**To Reproduce**

**Deployment Sheet(s)**

**Debug Sheet(s)**

**Screenshots and videos**

**Additional context**

chrismclarke:
PR Checklist

## Breaking Changes
Any variables declared as a direct child of a `begin_display_group` statement will now be treated as if it were declared at the top-level of a sheet. This is created at the parser level, so the extracted rows will be visible in synced sheets on update

Previously authors would need to refer to full nested name of variables created within display groups, e.g. `@local.display_group_1.var_1`, but now need to refer to directly as `@local.var_1`. The previous behaviour wasn't documented (and I'm not entirely sure even worked), so would assume knock-ons would only impact already broken templates

## Description
Add specific workaround to allow variables defined within a display group to be referenced outside, e.g.

| type | name | value |
|---|---|---|
|begin_display_group|dg_1|  |
|  |var_1| my text value|
|end_display_group| | |
|text| | @local.var_1|

## Review Notes
See example from linked issue [efm_chapters](https://docs.google.com/spreadsheets/d/1i_46RSdv8NzyqSIhjBXabUuW1ndiYRdjCgwgn5ZXsGk/edit?gid=568791997#gid=568791997) - video below

Would be good to confirm also fixes issue in related deployment

## Author Notes
**Use Cases**
The dev recommendation for variables is still that all variables simply be defined at the top of the sheet (outside of any nested row structures), however this workaround should at least fix the specific case to allow variables defined within a display_group

This PR doesn't add more general support for hoisting all variables within UI components (could be extended as follow-up), nor does it handle the case where the `display_group` sits within another nested structure (e.g. `display_group` within a `display_group`, or `display_group` within data_items). 

The latter case (within data_items) is mentioned in #2942, and although isn't fixed by this PR I've marked as closed as I think we should probably consider an authoring issue resolved by moving the variable declaration outside of the display_group. The logic to generally hoist any variables from within any component to the proper level (which could either be sheet top-level or top of nested `begin_data_items` row) gets quite complex, and so if we really want to generalise the behaviour would be better to create a follow-up issue.

**Not Supported**
| type | name | value |
|---|---|---|
|begin_data_items| | @data.item_list  |
|begin_display_group|dg_1|  |
|  |var_1| my text value|
|end_display_group| | |
|text| | @local.var_1|
|end_data_items||  |

**Recommended Alternative - manually hoist `var_1` outside of display group**
| type | name | value |
|---|---|---|
|begin_data_items| | @data.item_list  |
|  |var_1| my text value|
|begin_display_group|dg_1|  |
|end_display_group| | |
|text| | @local.var_1|
|end_data_items||  |

**QC**
I'm not sure if we want to document this anywhere else. Perhaps it might be worth either adding a specific example to the current `display_group` component sheet to demonstrate referring to variables within the display_group from outside, or creating a new `example_variable_hoisting` sheet that we could use to update over time if we plan to change behaviours

Given that any issues related to naming collision are now (once again) an authoring issue, we could also update either the parser or the report generator to flag any instances where the same variable name is used multiple times. This could either be a warning or an error generated - I'd probably recommend sticking to a warning for cases like debug/non-critical sheets, or cases where the variable isn't referred to elsewhere but just named for the sake of tidiness (so wouldn't have a knock on). The warning could appear either in console logs during parsing, as a markdown report in the repo, or just runtime warning (like the current warning in issue referenced). In any of these cases would recommend defining as a quick follow-up issue.

## Dev Notes
I've treated this more as a workaround as opposed to trying to impose full structural changes, partly to simplify the PR (resolve immediate issue quicker), partly as I'm still not entirely sure we've pinned down the exact behavioural changes (identifying instances when things should be hoisted and where all the potential boundary levels are). But I have at least added a test to catch regression issues in the future

## Git Issues

Closes #2989 
Closes #2942 (?)

## Screenshots/Videos

https://github.com/user-attachments/assets/0b262675-4678-456c-9000-1f3dc309e892


markharmon (approved):

chrismclarke:
> Thanks Chris, I don't have any further comments other than that we've added another hack method hackHoistDisplayGroupVariables which isn't ideal. I don't see an alternative to get this implemented right now though aside from the template rewrite/refactor stuff that we're working on separately.

Agreed, far from ideal but hopefully at least this time having an explicit test case will help avoid regressions when planning the upgraded system. I've also added a brief note in an appendix to the [Templating v2 RFC](https://docs.google.com/document/d/1Td5QJYpWq2vJ2LLZ10bQ3LbBsMWhwVKcBlzfsDF44u0/edit?tab=t.kivudux6qes2), as I expect capturing all the edge cases and nuances in the main proposal would be a bit of an uphill task

chrismclarke:
@esmeetewinkel - whenever you get a chance would you be able to give a quick functional test of this PR? Would be good to have included for some potential follow-up work on the templating system

esmeetewinkel (approved):
Thanks! Functional test passed. I can confirm this closes both https://github.com/IDEMSInternational/open-app-builder/issues/2989
and https://github.com/IDEMSInternational/open-app-builder/issues/2942.

I can't easily confirm whether this closes the issue on the specific deployment, as that has now been changed to use an authoring workaround. Happy for this to go in.

PR #3011, opened 2025-06-20
feat: add 'auth: delete_account' template action

### LINKED ISSUE #3010: [FEATURE] Expose option for user to delete their account (auth)
**What?**
Users should be able to delete their auth account. I don't believe that this needs to delete all their associated data (postgres), for the time being deleting the auth account in Firebase should be sufficient.

**Why?**
General good data practice, but also required by Apple for iOS apps. See review notes for recent review of rejected `plh_facilitator_ph` app:

> Guideline 5.1.1(v) - Data Collection and Storage
> Issue Description
> 
> The app supports account creation but does not include an option to initiate account deletion. Apps that support account creation must also offer account deletion to give users more control of the data they've shared while using an app.
> 
> Follow these requirements when updating an app to support account deletion:
> 
> - Only offering to temporarily deactivate or disable an account is insufficient.
> - If users need to visit a website to finish deleting their account, include a link directly to the website page where they can complete the process.
> - Apps may include confirmation steps to prevent users from accidentally deleting their account. However, only apps in highly-regulated industries may require users to use customer service resources, such as making a phone call or sending an email, to complete account deletion.
> 
> Next Steps
> 
> Update the app to support account deletion. If the app already supports account deletion, reply to App Review in App Store Connect and identify where to locate this feature.
> 
> If the app is unable to offer account deletion or needs to provide additional customer service flows to facilitate and confirm account deletion, either because the app operates in a highly-regulated industry or for some other reason, reply to App Review in App Store Connect and provide additional information or documentation. For questions regarding legal obligations, check with legal counsel.
> 
> Resources
> 
> Review frequently asked questions and learn more about the [account deletion requirements](https://developer.apple.com/support/offering-account-deletion-in-your-app).
> Support
> - Reply to this message in your preferred language if you need assistance. If you need additional support, use the [Contact Us module](https://developer.apple.com/contact/topic/#!/topic/select).
> - Consult with fellow developers and Apple engineers on the [Apple Developer Forums](https://developer.apple.com/forums/).
> - Provide feedback on this message and your review experience by [completing a short survey](https://essentials.applesurveys.com/jfe/form/SV_esVePfih7uqt4NM?campaignid=0001).

**How?**
It should be straightforward to expose an `auth: delete_user` action to templating. Presenting the "Are you sure you want to delete" would likely be best handled in authoring

**Other details**

jfmcquade:
PR Checklist

## Description

Adds a new template action: `auth: delete_account`. This action deletes the account of the currently signed in user from Firebase auth, and resets local data to ensure that the user does not appear to be signed in.

This action does not affect any data in the postgres database, it simply deletes the user's auth account in firebase. If other functionality is used that makes use of the user's auth ID, e.g. shared data in Firebase Firestore, then this is also unaffected: in the case of shared data, there will still exist documents that grant access permissions to a user with the old auth ID, but there will no longer exist a user with such an ID.

## Testing

I have confirmed that deleting an account from within the app causes the account to be deleted in the Firebase Auth console. Signing into the app with the same email address creates a new account with a different auth ID.

When testing this functionality on deployments other than the debug deployment, it may be worth using an additional Google/Apple account to the one usually used, in case you want to keep the data associated with that account for testing other features (e.g. shared data functionality, profile data restore).

## Git Issues

Closes #3010

### DIFF
--- a/src/app/shared/services/auth/auth.service.ts
+++ b/src/app/shared/services/auth/auth.service.ts
@@ -118,6 +118,7 @@ export class AuthService extends AsyncServiceBase {
           sign_in_google: async () => await this.provider.signInWithGoogle(),
           sign_in_apple: async () => await this.provider.signInWithApple(),
           sign_out: async () => await this.provider.signOut(),
+          delete_account: async () => await this.provider.deleteAccount(),
         };
         if (!(actionId in childActions)) {
           console.error(`[AUTH] - No action, "${actionId}"`);
--- a/src/app/shared/services/auth/providers/base.auth.ts
+++ b/src/app/shared/services/auth/providers/base.auth.ts
@@ -20,4 +20,9 @@ export class AuthProviderBase {
     this.authUser.set(undefined);
     return this.authUser();
   }
+
+  public async deleteAccount() {
+    throw new Error("Delete account not enabled");
+    return this.authUser();
+  }
 }
--- a/src/app/shared/services/auth/providers/firebase.auth.ts
+++ b/src/app/shared/services/auth/providers/firebase.auth.ts
@@ -64,6 +64,13 @@ export class FirebaseAuthProvider extends AuthProviderBase {
     return this.authUser();
   }
 
+  public async deleteAccount() {
+    await FirebaseAuthentication.deleteUser();
+    this.authUser.set(undefined);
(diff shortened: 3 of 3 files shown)


esmeetewinkel (approved):
Functional test passed

ChrisMarsh82 (approved):

PR #3013, opened 2025-06-21
Feat: notification triggered actions

chrismclarke:
PR Checklist

## TODO
- [ ] Builds on top of #3009, merge after
- [ ] Service code to trigger actions (ideally via current rendered template to support all possible actions, or headless if not easily viable)
- [ ] Unit tests
- [ ] Consider support for multiple actions via named buttons
- [ ] Consider whether better to author as `action` or `action_list` (assume `action` to allow future multiple `actions`)

## Description

Add support for 

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/deployment-features.module.ts
+++ b/src/app/deployment-features.module.ts
@@ -4,6 +4,7 @@ import { AnalyticsModule } from "./shared/services/analytics";
 import { NavStackModule } from "./feature/nav-stack/nav-stack.module";
 import { AuthModule } from "./shared/services/auth/auth.module";
 import { CampaignModule } from "./feature/campaign/campaign.module";
+import { NotificationsModule } from "./feature/notification/notification.module";
 import { provideSharedData } from "./feature/shared-data";
 import { PLH_FEATURE_MODULES } from "packages/components/plh";
 
@@ -25,6 +26,7 @@ import { PLH_FEATURE_MODULES } from "packages/components/plh";
     AnalyticsModule,
     CampaignModule,
     NavStackModule,
+    NotificationsModule,
     ...PLH_FEATURE_MODULES,
     /* */
   ],
--- a/src/app/feature/notification/notification.actions.ts
+++ b/src/app/feature/notification/notification.actions.ts
@@ -1,10 +1,15 @@
 import type { IActionHandler } from "src/app/shared/components/template/services/instance/template-action.registry";
 import { NotificationService } from "./notification.service";
 import { INotification } from "./notification.types";
+import { FlowTypes } from "packages/data-models";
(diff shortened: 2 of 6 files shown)


chrismclarke:
> I attempted tested this on native using [IDEMSInternational/app-debug-content#196](https://github.com/IDEMSInternational/app-debug-content/pull/196). However the ios simulator build action is failing with `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory` so I didn't test iOS.

That will be an issue with Node itself not assigning enough memory. Assuming your system has 8GB memory or more, you should be able to safely set a higher value

_Windows_
```ps1
setx NODE_OPTIONS "--max-old-space-size=4096"
```
Should apply from next vscode restart

> I was able to test on Android using the Release to Google Play action (content v1.7.4). Generally seems to work well, but I found that most of the time, when my app is closed and a notification is due to arrive, I only receive it once I open the app. See screen recording.

This is most likely just a timing issue - Android notification timings aren't precise (unless requesting specific permission, like used in alarms). So when scheduling I'd expect the notification to arrive in +/- around 30s. I think opening the app and see the notification is coincidental, I'd expect the notification to arrive eventually, but there might be something in opening the app that forces the system to re-check notifications.

But I don't believe either of these issues are particularly related to code changes in this PR, so still think would be good if @jfmcquade could review

chrismclarke:
> > > I attempted tested this on native using [IDEMSInternational/app-debug-content#196](https://github.com/IDEMSInternational/app-debug-content/pull/196). However the ios simulator build action is failing with `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory` so I didn't test iOS.
> > 
> > 
> > That will be an issue with Node itself not assigning enough memory. Assuming your system has 8GB memory or more, you should be able to safely set a higher value
> > _Windows_
> > ```powershell
> > setx NODE_OPTIONS "--max-old-space-size=4096"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Should apply from next vscode restart
> 
> Sorry if I wasn't clear, that error came up in the iOS simulator build GitHub action [here](https://github.com/IDEMSInternational/app-debug-content/actions/runs/16750135653/job/47419054211). I'm not sure how to set a higher value there.

Ah, thanks that makes more sense. I can manually specify the setting on the github action used, I've done so via (#3131) and re-run the action

chrismclarke:
Just to mention @esmeetewinkel - if you need to trigger another build for any reason I've changed the debug app repo target branch to a different one as part of #3136

PR #3017, opened 2025-06-26
Update Theme Generator to create two alternative colour palettes

### LINKED ISSUE #3016: [FEATURE] Update Theme generator code
**What?**
Update the theme generator code to generate additional sets of colour variables for two new colours, derived from the primary and secondary source colours respectively

**Why?**
To reduce the need for colour overrides across themes

**How?**
_Do you have an example template that demonstrates the desired behaviour or syntax? If so, specify its name and provide a link._

**Other details**

FaithDaka:
PR Checklist

## Description

- Updates the theme generator code to create two alternative/variant palettes to the primary and secondary colours. These two colours can still be overridden for pre-designed palettes. Darker colours get a 15% more saturation boost for vibrancy.
- Refactor of the `contrast-colour` code to use more recent math functions.
- `primary-variant` and `secondary-variant` palettes made visible on the debug `themes` page.

More on the discussion here: [link](https://www.figma.com/slides/HOnetqQMqKRYjWdw6wrqLw/OAB-Theming-Design?node-id=1-332&t=WQMKk1ubMCkXWKp2-1)

## Git Issues

Closes #3016


jfmcquade (changes requested):
Thanks @FaithDaka, this looks really good. Nice to have these changes independent in a way that means that no additional updates are needed to existing themes, whilst making the new colour vars available.

I've added some minor tweaks to the theme debug page with https://github.com/IDEMSInternational/open-app-builder/pull/3017/commits/65edeaa566e9819d6dd8e74e62dff25dae96cf50. I think that the swatches appearing next to each other produced a subtle optical illusion of gradients, so I think it's easier to identify the colours with a border around them:

Screenshot 2025-07-08 at 14 46 14

If there's anything else you think would be useful to include on the "theme editor" page, feel free to add it or make a suggestion. It would be nice eventually if that page could be used for testing out a colour palette, i.e. taking user supplied values of the source colours and generating the theme colours dynamically. Perhaps that could even be a first step towards reading in the source colour values from an authored spreadsheet?

I haven't tested out the new contrast colour generation logic, but I'm happy to trust it! Nice to have that simplified. 

Also nice to have the new "saturation boost" modifier to generate nicer colour palettes. Please do keep adding improvements like these when you see the opportunity, although would be good to also flag in the PR description to draw attention to all of the changes. This is a change that will affect all existing themes, but for the better.

One minor change to request: following our conversation, and consulting the [Material 2 docs](https://m2.material.io/design/color/the-color-system.html#color-theme-creation), I think `*-variant` would be a better name than `*-alt`. Please could you make the substitution with a find-and-replace (including the `$pa` and `$sa` vars)?

@esmeetewinkel I'm tagging you for review and to make you aware of the changes. I don't think much is required in terms of functional testing, but maybe good to have a look at the `/theme` page and see the new colours that will be available.

esmeetewinkel (approved):
Looks good, thanks! Curious to find out how well the automated computations of primary and secondary variant colours hold up. 

Only comment is that this branch conflicts with the renaming of the mapa/ph theme, @FaithDaka could you resolve these? 

image

FaithDaka:
Sorry for the delay @jfmcquade 
Thanks for the feedback. I've updated the naming of the variants and changed the hex colour codes to hsl values (makes it easier for the `generate-theme` to work with)

jfmcquade (approved):
Thanks @FaithDaka, looks good. 

> I've ... changed the hex colour codes to hsl values (makes it easier for the generate-theme to work with)

Makes sense. Presumably supplying a hex code still works? I can imagine some theme authors may want to supply a colour value in a hex code or some other format (e.g. to match brand colours exactly), although I appreciate it would be trivial for them to convert to hsl format

PR #3018, opened 2025-06-26
Spike: Reactive variables & Handling 'changed' actions in parent when setting local variables in child templates

### LINKED ISSUE #2894: [FEATURE] Ability to access local variables in nested templates
**What?**
Ability to access all variables within a nested template

**Why?**
Building up template components would require this. i.e. creating a validated text box that displays a label if incorrect. This would be a template used within many templates. In order to save values we would need to be able to access the variable within the template

**How?**
@local.[template_defined_name].[variable_name] 
(name column)

**Other details**
This would be good if there is no restrictions on levels i.e. @local.[first_template].[second_template].[variable_name] 
Although this is not essential as workarounds could be done for this and it would generally not be used.

markharmon:
PR Checklist

### Child variable changed events

Allows parent templates to react to changes in child template variables & first step towards a fully reactive template system.

[parent_template](https://docs.google.com/spreadsheets/d/1lD3elJjIf2c3MMyaxAXdeIqZng_i58TjmWCt9jKQYiA/edit?gid=2017326959#gid=2017326959)

image

[child_template](https://docs.google.com/spreadsheets/d/1lD3elJjIf2c3MMyaxAXdeIqZng_i58TjmWCt9jKQYiA/edit?gid=33001911#gid=33001911)

image

### Variable Store

Identified a problem due to child templates being forcefully redrawn and so 'resetting' child template variables. The variable store now reacts to parent variable changes reactively using the VariableStore

Problem code (now replaced):

```
  /**
   * If the current template is generated as a child of data_items then updates to
   * parent variables will not propagate down. Force reprocessing to workaround
   * See issue https://github.com/IDEMSInternational/open-app-builder/issues/2636
   */
  private async hackForceReprocessNestedTemplate() {
    if (this._row.type === "template") {
      const componentRef = this.componentRef as ComponentRef<TemplateContainerComponent>;
      if (componentRef.instance.parent) {
        // HACK - when a parent updates there may be cases when the updated variables need to be
        // reflected in the child template (e.g. set_nested_variable contains parent local variable)
        // As it is not currently easy to isolate this case, fully recreate any child template when
        // the parent template updates
        if (componentRef.instance.templatename()) {
          this.componentRef.destroy();
          this.renderTemplateComponent(this._row);
          // TODO - test better ways to manage from container, confirming test cases from
          // https://github.com/IDEMSInternational/open-app-builder/issues/2636
        }
      }
    }
  }
```

The new variable store implementation aims to make each template responsible for "re-rendering" itself by subscribing to it's dependencies.

### TemplateBaseComponent

The following changes to TemplateBaseComponent allow the template to react to it's own value and dependency value changes.

```
if (this.variableStore) {
      this.value = this.variableStore.asSignal(this._row._nested_name);
    } else {
      console.warn("Variable store not set for TemplateBaseComponent, row:", this._row);
    }
```
```
  private subscribeToParentVariables() {
    if (!this.parent || this.parent.row?.type !== "template") return;

    let parentVariable = this.parent.row.rows.find((parentRow) =>
      this._row._nested_name.endsWith(parentRow.name)
    );

    if (!parentVariable) {
      return;
    }

    if (!this.variableStore.has(parentVariable._nested_name)) return;
    const subscription = this.variableStore
      .watch(parentVariable._nested_name)
      .subscribe(async (value) => {
        await this.setValue(value);
      });

    this.subscriptions.push(subscription);
  }
```

### Examples

[Parent ➡️  Child](https://idems-debug.web.app/template/debug_data_items_local)
[Parent ⬅️  Child](https://idems-debug.web.app/template/debug_nested_var_emit_parent)

## Git Issues

Closes #2894

## Next steps for template rendering:

Introduce more stores, these stores will also be responsible for permanent strorage e,g, to Local Storage/DB

* FieldStore
* DataItemsStore

Methods to investigate replacing

* hackUpdateActionSelfReferenceValues
* hackProcessRows in data-items.service
* evaluatePLHData / evaluatePLHString

## Reference

https://www.figma.com/board/eCGDAgROVSwMpqYSGVJa0B/Reactive-Templates---Open-App-Builder?node-id=0-1&p=f&t=VA0gfriUILqa9NZi-0


markharmon:
I need to add code that watches variables for all dependencies of a row so that rows like these react when their own values change:

image
image

chrismclarke:
Thanks for adding @markharmon , a few initial thoughts:

**General Thoughts**
I think the delegation of more logic and data tracking to the base component makes a lot of sense, and it's also nice to have explicitly laid out in a PR all the parts of the code that want to be changed when updating local variable management (it's a lot!).  I think in particular it has a lot of potential for reducing inefficiencies when re-rendering components due to nested changes.

**Global vs Local Instance**
I see that you've created a global store provider for the reactive locals system. I think this could risk name collision and accidentally allowing for variables to spill from one template into another, particularly if we have two templates that use the same variable names which are either navigated between, loaded in a popup, or in split-screen routing. I think likely we need something that will either be instantiated at the same time top-level containers are, or support some sort of tracking/namespacing to handle either when the same template reused, or similar templates used in different contexts. 

**Parent-child Reference**
I'm working my head around the initial example, where the parent template sets the initial value on a child, but also adds a change action to update the parent action when changed (2-way data binding). I'm not entirely sure I see this pattern as desirable from an authoring perspective, the additional action doesn't really hold much benefit over just allowing direct reference, i.e.

|type|name|value|
|---|---|---|
||parent_variable|parent_value|
|begin_template|child_1|child_template_name|
||child_value|@local.parent_variable|
|end_template|||
|text|current_state_of_child_value|@local.child_1.child_value|

I'm not sure if the changes here would preclude or make possible the approach, and having the ability to use actions isn't a bad thing, just that I would want to make sure we prioritise getting the simpler case above working (if not already)

**Code complexity**
I'm slightly concerned that we are now adding another feature directly on top of a quite shaky foundation, without removing very much (nice to remove the `hackForceReprocessNestedTemplate`, but mostly everything else is additive). 

I think my biggest concern is the existence of both the variable store and the templateRowMap which have quite significant overlap. I can see from your comment on the data-items store that you are also thinking about this, so I guess I'm left wondering whether the scope of this PR would include that or not? If so I'd be cautious of how best to expand our current testing procedures to make it easier to evaluate a large-scale replacement, if not I'm not sure where we stand with trying to improve the templating system given we are adding without removing.

There's also a handful of the worst offenders for reprocessing, like the list of `reprocessActions` list in `template-action.service.ts` that take a very heavy hand to re-rendering which should ideally no longer include set_local or set_self actions following changes introduced, but are still currently present.

**Next Steps**
So I think if possible a couple next steps could also involve:

1. Identify whether any of the changes introduced here would be ready to come in standalone and open up separate PR. I.e. whether the changes proposed for handling changed actions are independent to the reactive variables

2. Setup test cases for yourself that involve variables with similar named variables. I believe there are some debug examples that include opening a popup of the same template currently registered (ability to open infinitely nested popups), although it might still be easier for you to create some of your own testing templates to cover a navigation and popup example.

3. Test removing the various forced_rerender methods to see whether the new subscriptions do adequately remove the need for it - this would be a major performance win if we could get it working

4. Consider instantiating at a more local level, potentially replacing the `templateRowMap` directly with a `templateLocalStore` or similar which extends your current code to also include a simple variable getter method to use as a drop-in replacement

PR #3030, opened 2025-07-07
chore(deps)!: upgrade capacitor to v7

jfmcquade:
PR Checklist

- [x] Merge after #3038
- [x] PR title descriptive (can be used in release notes)

## Description

Upgrades Capacitor core and plugin packages to v7.

This includes targeting Android API level 35 (closes #3021), and iOS v14.

## Breaking changes

Drops support for Android API level 22 (`v5.1`) so the new minimum requirement for Android devices is `v6.0`.

## Dev notes

I performed the upgrade by following the instructions in [Capacitor's official docs](https://capacitorjs.com/docs/updating/7-0). Some local dev updates were required, not committed to the repository, namely upgrading Android Studio (and some configuration of Android Studio to use JDK v21).

Note that Node v18 is no longer supported. We should explore supporting Node v22 for our repo generally.

The official capacitor libraries were all upgraded automatically via the `npx cap migrate` command:

```json
    "@capacitor/android": "^7.0.0",
    "@capacitor/app": "^7.0.0",
    "@capacitor/clipboard": "^7.0.0",
    "@capacitor/core": "^7.0.0",
    "@capacitor/device": "^7.0.0",
    "@capacitor/filesystem": "^7.0.0",
    "@capacitor/ios": "^7.0.0",
    "@capacitor/local-notifications": "^7.0.0",
    "@capacitor/push-notifications": "^7.0.0",
    "@capacitor/screen-orientation": "^7.0.0",
    "@capacitor/share": "^7.0.0",
    "@capacitor/splash-screen": "^7.0.0",
    "@capacitor/toast": "^7.0.0",
    
    "@capacitor/cli": "^7.0.0",
 ```

Additionally, we use various third-party capacitor plugins that I upgraded manually:
```json
    "@capacitor-community/file-opener": "^7.0.1",
    "@capacitor-firebase/authentication": "^7.2.0",
    "@capacitor-firebase/crashlytics": "^7.2.0",
    "@capacitor-firebase/firestore": "^7.2.0",
    "@capacitor-firebase/performance": "^7.2.0",
    ...
    "@capawesome/capacitor-app-update": "^7.0.1",
```

## Testing

I have confirmed that the app builds and runs successfully on bot Android and iOS platforms. In theory we should test all  Capacitor API features, i.e. those features that make use of the following packages:

```json
    "@capacitor-community/file-opener": "^7.0.1",
    "@capacitor-firebase/authentication": "^7.2.0",
[...]

PR #3034, opened 2025-07-09
Fix: ios auth automated login

### LINKED ISSUE #3033: [BUG] on iOS, auth state does not persist after force restart
**Describe the bug**
On iOS specifically, triggering a `force_refresh` of the app causes a logged in user to be logged out (or at least be registered as logged out via the persisting app fields, the user may still be logged according to the third-party firebase auth package). Web and android are unaffected, the log in persists as expected on these platforms.

This is a critical issue for the `plh_facilitator_ph` deployment, where auth is required and the app restarts after initial login.

**App version**
Code `v0.19.7`, `plh_facilitator_ph` deployment content `v1.3.6`.

**To Reproduce**
Demonstration of log in not persisting on `plh_facilitator_ph` deployment:

https://github.com/user-attachments/assets/4c3f4bd2-5486-4689-b45d-62c236e839b3

**Deployment Sheet(s)**
The "continue" button in the user journey recorded above is authored in D25 of [app_launch_ph](https://docs.google.com/spreadsheets/d/1bULj_CfbpLcnPpXSyIW3QnXWOT_pHI_CPgzZjH6pUh8/edit?gid=500532234#gid=500532234), which triggers an `emit: force_restart` action.

**Debug Sheet(s)**
UPDATE: User flow on `debug` deployment via [example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705) template. The issue presents on both iOS and Android

iOS:

https://github.com/user-attachments/assets/0505b4ca-8de1-45cf-8091-b2b03f891892

Android:

https://github.com/user-attachments/assets/e5dc53aa-07af-4242-a474-fe05c9e53bd9

**Additional context**
Technically, the issue seems to be that on iOS, the `"authStateChange"` event is not necessarily fired by the Firebase Authentication plugin when initialising with a user already logged in. This causes our Firebase auth provider to not finish initialising and updating the `authUser` (see [initialisation logic](https://github.com/IDEMSInternational/open-app-builder/blob/master/src/app/shared/services/auth/providers/firebase.auth.ts#L24-L37)), which in turn means the associated [effect in the auth service](https://github.com/IDEMSInternational/open-app-builder/blob/master/src/app/shared/services/auth/auth.service.ts#L40-L55) is not fired with the updated user credentials. Debug logging for the user flow demonstrated above shows that the Auth service does not finish initialising, and that the effect only runs with the value of `authUser` as `null`.

Image

jfmcquade:
PR Checklist

## Description

Fixes an issue where logging in on iOS (via Apple or Google auth) would not persist following a `force_restart` event.

## Testing

I've tested on iOS and the issue seems to be resolved. See corresponding appetize build [here](https://appetize.io/apps/ios/international.idems.plh-facilitator-ph/builds/b_vpgcf3fh6bh2fbfynxcokxcyju). Since the new logic will only run on iOS, it shouldn't be necessary to test web and android platforms.

## Dev notes 

From #3033:

> Technically, the issue seems to be that on iOS, the `"authStateChange"` event is not necessarily fired by the Firebase Authentication plugin when initialising with a user already logged in. This causes our Firebase auth provider to not finish initialising and updating the `authUser` (see [initialisation logic](https://github.com/IDEMSInternational/open-app-builder/blob/master/src/app/shared/services/auth/providers/firebase.auth.ts#L24-L37)), which in turn means the associated [effect in the auth service](https://github.com/IDEMSInternational/open-app-builder/blob/master/src/app/shared/services/auth/auth.service.ts#L40-L55) is not fired with the updated user credentials. Debug logging for the user flow demonstrated above shows that the Auth service does not finish initialising, and that the effect only runs with the value of `authUser` as `null`.

This is resolved by using an alternate method of waiting for firebase auth initialisation on iOS: polling the `FirebaseAuthentication.getCurrentUser()` method multiple times on init to try and ensure we have the up-to-date user details, as opposed to waiting for the `"authStateChange"` event to fire. I've restricted this alternate method to only be used on iOS, as the existing method seems to work well on other platforms and should be more robust.

## Git Issues

Closes #3033 

## Screenshots/Videos

See user journey logging in on `plh_facilitator_ph` deployment, and compare to equivalent and corresponding debug logs in description of #3033.

https://github.com/user-attachments/assets/5a6ffa3e-71c6-425d-b22b-d2286dab0f9d

Screenshot 2025-07-09 at 12 33 46

### DIFF
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/flowParser.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/flowParser.ts
@@ -4,7 +4,7 @@ import { IConverterPaths, IFlowHashmapByType, IParsedWorkbookData } from "../../
 import { arrayToHashmap, groupJsonByKey, IContentsEntry } from "../../utils";
 import BaseProcessor from "../base";
 
-const cacheVersion = 20250515.0;
+const cacheVersion = 20250619.1;
 
 export class FlowParserProcessor extends BaseProcessor<FlowTypes.FlowTypeWithData> {
   public parsers: { [flowType in FlowTypes.FlowType]: Parsers.DefaultParser } = {
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.spec.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.spec.ts
@@ -195,6 +195,37 @@ describe("Template Parser PostProcessor", () => {
   });
 });
 
+describe("Template Parser PostProcess", () => {
+  let parser: TemplateParser;
+  beforeEach(() => {
+    parser = new TemplateParser({} as any);
+  });
+  it("hoists variable references from display_groups", () => {
+    const rows: FlowTypes.TemplateRow[] = [
+      {
+        _nested_name: "dg",
+        name: "dg",
+        type: "display_group",
+        rows: [
+          {
+            _nested_name: "db.inner_var",
+            type: "set_variable",
+            name: "inner_var",
+            value: "inner_val",
+          },
+        ],
+      },
+    ];
+    const res = parser.postProcessFlow({ flow_name: "", flow_type: "template", rows });
+    // hoisting should move inner set_variable to top-level, resulting in 2 flat rows
+    // the hoisted row should also have its nested_name updated to reflect new position
+    expect(res.rows.length).toEqual(2);
+    expect(res.rows[0].type).toEqual("set_variable");
+    expect(res.rows[0]._nested_name).toEqual("inner_var");
+    expect(res.rows[1].type).toEqual("display_group");
+  });
+});
+
 describe("Template Parser [QC]", () => {
   // TODO - confirm what checks to include and add to code
   // it("Ensures answer_list parameters refer to list variables", () => {
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.ts
@@ -82,6 +82,11 @@ export class TemplateParser extends DefaultParser {
     return row;
   }
 
+  public override postProcessFlow(flow: FlowTypes.FlowTypeWithData): FlowTypes.FlowTypeWithData {
+    const hoisted = this.hackHoistDisplayGroupVariables(flow);
+    return hoisted;
+  }
+
   public override postProcessFlows(flows: FlowTypes.FlowTypeWithData[]) {
     const flowsWithOverrides = assignFlowOverrides(flows);
     return flowsWithOverrides;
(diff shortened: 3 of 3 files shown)


src/app/shared/services/auth/providers/firebase.auth.ts
```diff
@@ -30,12 +31,51 @@ export class FirebaseAuthProvider extends AuthProviderBase {
...
     await this.handleAutomatedLogin();
   }
 
+  /**
+   * Wait for Firebase Auth to fully initialize and restore any existing session.
+   * This handles the race condition where getCurrentUser() might return null
+   * before Firebase Auth has finished restoring the session from persistent storage.
+   */
+  private async waitForAuthInitialization(): Promise<void> {
+    const maxAttempts = 6;
```
chrismclarke, line 52: nit(blocking): this feels like a somewhat arbitrary number. In practical testing did it ever take more than 1 attempt to settle on a user? Are there any other indicators we could use to (e.g. if we know there is a previous user stored in localStorage use that as an indication to retry)

jfmcquade: Yes, it is arbitrary. From my testing setup, it seemed to return populated user info on the third polling attempt. Good suggestion of exploring other indicators.

jfmcquade:
Thanks for the review, @chrismclarke. I agree that this implementation is arbitrary and heavy-handed, good suggestions for investigating alternatives. 

> 2. Merge #3030 just to check if issues still persist on latest version of capacitor and capacitor-firebase

I'm currently in the process of testing #3030 on all three platforms. It's a bit time consuming, but I want to briefly test each of the upgraded plugins for peace of mind. I have found that it doesn't seem to resolve the issue behind the current PR, where a `force_restart` logs the user out, indeed it possibly additionally introduces the issue on Android (although I haven't tested exactly like-for-like to know whether the same behaviour is present on master.

iOS:


https://github.com/user-attachments/assets/000fe580-94ac-4a4c-b9bf-3f5095788143

Android:


https://github.com/user-attachments/assets/aa26bb7a-723e-4da2-a6fa-0109c5e893ad

> 1. Add a simple `await this.platform.isReady()` (importing the capacitor platform). There is a chance that due to the way this service is loaded as part of a feature module the service could be trying to initialise before capacitor fully stable, and could be the reason some events are not transmitted as expected

Good suggestion, I'll look into this after I've cleaned up my dev environment following the testing mentioned above.

> 3. Refactor the code a bit so that the ios workaround code can be called in parallel with the regular wait code (might lead to faster race conditions), example inline

That would definitely be an improvement if this implementation seems like the best option after investigating others.

> 4. Consider using the localstorage check for previous user when deciding whether to retry polling method. If there hadn't previously been a user then we probably shouldn't be waiting for it to auto-sign-in

Another good improvement. Possibly we could even be waiting explicitly for the package class to return a populated user object in the case that there is a previous user in localstorage (and not if not), which would remove the arbitrary number of polling retries. There may possibly be cases where there is a previous user entry in local storage but firebase doesn't consider the user to be logged in, but I'm not sure what that would be.

I've also suggested that in the immediate term, the issue may be worked around by removing the `emit: force_restart` action from the login flow of the app deployment in question (`plh_facilitator_ph`). This would also work around the issue of the slightly confusing (and untranslated) browser message that faces users when pressing this button (documented in https://github.com/ParentingForLifelongHealth/plh-kids-app-tz-content/issues/238).

chrismclarke:
@jfmcquade - I've opened up an alternate PR which I _think_ should offer a simpler solution to the issue https://github.com/IDEMSInternational/open-app-builder/pull/3038

Issue #3012, opened 2025-06-20
[FEATURE] Alternative syntax for authoring actions


jfmcquade:
**What?**
Offer an alternative syntax for authoring actions. This should allow for: increased complexity in action logic and reusable sets of actions that can be referenced and called in multiple places. A stretch goal/follow-up would be to allow actions to receive responses that can then be reacted to conditionally, akin to a standard web API (e.g. success/error status (code), possibly some data sent)

**Why?**
The current syntax for authoring actions tries to fit a lot into a single cell. By using our existing template structures (name, value, parameter_list etc.), we can make the action authoring syntax more explicit and potentially more powerful.

Authoring actions in units that can be referenced, either within a single template or in a separate flow, enables authoring complex sets of actions that can be referenced in multiple places without needing to copy and paste.

Would help with some of the functionality proposed in #3007.

**How?**
See some proposed possible syntaxes in [proposed_action_syntax](https://docs.google.com/spreadsheets/d/1ljjne138IC-7Aaw7fj4_OW9-4CWsDJdrxSfDf8mDhQw/edit?gid=0#gid=0)

**Other details**
_Add any extra details that you think may be useful._

PR #3040, opened 2025-07-12
feat: template value action list

chrismclarke:
PR Checklist

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/data-models/flowTypes.ts
+++ b/packages/data-models/flowTypes.ts
@@ -311,7 +311,8 @@ export namespace FlowTypes {
   export interface TemplateRow extends Row_with_translations {
     type: TemplateRowBaseType | TemplateRowCoreType | TemplateRowDeploymentType;
     name: string;
-    value?: boolean | number | string;
+    /** Value can be set from authored value (string, number, boolean), or transformed from action_list or _list types */
+    value?: any;
     action_list?: TemplateRowAction[];
     style_list?: string[];
     parameter_list?: { [param: string]: string };
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.ts
@@ -6,6 +6,7 @@ import {
   extractDynamicDependencies,
   parseAppDataCollectionString,
   parseAppDataListString,
+  parseAppDataActionString,
 } from "../../../utils";
 
 export class TemplateParser extends DefaultParser {
@@ -39,24 +40,15 @@ export class TemplateParser extends DefaultParser {
     row._nested_name = nestedPath ? `${nestedPath}.${row.name}` : row.name;
     // convert any variables (local/global) list or collection strings (e.g. 'my_list_1')
     // in similar way to how top-level properties get converted by default parser
-    if (row.value && typeof row.value === "string") {
-      if (row.name?.endsWith("_list") || row.name?.includes("_list_")) {
-        row.value = this.parseTemplateList(row.value);
-      }
-      if (row.name?.endsWith("_collection") || row.name?.includes("_collection_")) {
-        if (row.value && typeof row.value === "string") {
-          // TODO - verify if case used and whether it might be better to use a different
-          // column to store parsed object literal in value (would require type defs update)
-          row.value = parseAppDataCollectionString(row.value) as any;
-        }
-      }
-    }
+    row.value = this.transformRowValue(row.name, row.value);
+
     if (row.parameter_list) {
       row.parameter_list = this.parseParameterList(row.parameter_list as any);
     }
     if (row.action_list) {
       row.action_list = this.hackUpdateActionSelfReferences(row.action_list, row.name);
     }
+
     // extract dynamic fields for runtime evaluation
     const dynamicFields = extractDynamicFields(row);
     if (dynamicFields) {
@@ -90,6 +82,31 @@ export class TemplateParser extends DefaultParser {
   public override postProcessFlows(flows: FlowTypes.FlowTypeWithData[]) {
     const flowsWithOverrides = assignFlowOverrides(flows);
     return flowsWithOverrides;
+  }
+
+  /** Apply custom value transformations to rows with specific names, e.g. _list or _collection */
+  private transformRowValue(rowName: string, rowValue: any) {
+    if (rowName && rowValue && typeof rowValue === "string") {
+      // NOTE - required if passing an action_list from variable as only the `value`
+      // column is retained when interpreting data at runtime (workaround)
+      if (rowName.includes("action_list")) {
+        const entries = parseAppDataListString(rowValue);
+        return entries
+          .map((actionString) => parseAppDataActionString(actionString))
+          .filter((action) => action !== null);
+      }
+      if (rowName.endsWith("_list") || rowName?.includes("_list_")) {
+        return this.parseTemplateList(rowValue);
+      }
+      if (rowName.endsWith("_collection") || rowName.includes("_collection_")) {
+        if (rowValue && typeof rowValue === "string") {
+          // TODO - verify if case used and whether it might be better to use a different
+          // column to store parsed object literal in value (would require type defs update)
+          return parseAppDataCollectionString(rowValue) as any;
+        }
+      }
+    }
+    return rowValue;
   }
 
   /**
--- a/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.spec.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.spec.ts
@@ -103,6 +103,21 @@ describe("Template Parser PostProcessor", () => {
       { key_1a: "textValue", key_1b: "@local.value" },
       { key_2a: "", key_2b: 5 },
     ]);
+    // CASE 5 - Action_list
+    const case5 = parser.postProcessRow({
+      ...ROW_BASE,
+      name: "test_action_list",
+      value: "click | set_field : test_field : hello",
+    });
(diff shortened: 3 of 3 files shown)


packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.ts
```diff
@@ -92,6 +84,31 @@ export class TemplateParser extends DefaultParser {
...
+      if (rowName.endsWith("_list") || rowName?.includes("_list_")) {
+        return this.parseTemplateList(rowValue);
+      }
+      if (rowName.endsWith("_collection") || rowName.includes("_collection_")) {
+        if (rowValue && typeof rowValue === "string") {
+          // TODO - verify if case used and whether it might be better to use a different
+          // column to store parsed object literal in value (would require type defs update)
+          return parseAppDataCollectionString(rowValue) as any;
+        }
+      }
```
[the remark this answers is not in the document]
chrismclarke: tidied

packages/data-models/flowTypes.ts
```diff
@@ -311,7 +311,8 @@ export namespace FlowTypes {
   export interface TemplateRow extends Row_with_translations {
     type: TemplateRowBaseType | TemplateRowCoreType | TemplateRowDeploymentType;
     name: string;
-    value?: boolean | number | string;
+    /** Value can be set from authored value (string, number, boolean), or transformed from action_list or _list types */
+    value?: any;
```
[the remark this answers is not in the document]
chrismclarke: Don't think that would be of much benefit - so many overlapping types that handling at component level still likely to involve search for specific type

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.ts
```diff
@@ -92,6 +84,27 @@ export class TemplateParser extends DefaultParser {
     return flowsWithOverrides;
   }
 
+  /** Apply custom value transformations to rows with specific names, e.g. _list or _collection */
+  private transformRowValue(rowName: string, rowValue: any) {
+    if (rowName && rowValue && typeof rowValue === "string") {
+      // NOTE - required if passing an action_list from variable as only the `value`
+      // column is retained when interpreting data at runtime (workaround)
+      if (rowName.endsWith("_action_list") || rowName?.includes("_action_list_")) {
```
jfmcquade, line 92: I think we probably want to also allow for `rowName.startsWith("action_list_")`, e.g. `action_list_1` (or even allow for a var just called `action_list` too?)

jfmcquade (changes requested):
I can see why this has immediate utility for the notifications use case and could be generally helpful for defining an action list in a variable such that it could be reused or referenced by multiple components. However I think that adding another special case row name isn't ideal and is quite opaque. So I'm happy for these changes to be merged, but I think they should be considered a hack or interim solution.

I think longer term it would be good to introduce an action syntax that is more explicit, like the [suggestions](https://docs.google.com/spreadsheets/d/1ljjne138IC-7Aaw7fj4_OW9-4CWsDJdrxSfDf8mDhQw/edit?gid=0#gid=0) linked to in #3012.

UPDATE: The below can be ignored – I now see that all is required is that all that is required to resolve is to add the new `notification_received` and `notification_interacted` action triggers to the list of allowed triggers in `parseAppDataActionString()`, which is included in #3013. It could be useful to have a quick debug sheet that goes along with the current PR to demonstrate the new parsing of an action list that uses an existing trigger?

> I think there is also an issue with the current implementation. When parsing the [feat_notification_actions](https://docs.google.com/spreadsheets/d/1m_q57wKk7rWCzW5J2DQu8IH4DNYM_rILbDifrN-d_iQ/edit?gid=1107968849#gid=1107968849) on this branch, the `example_action_list` doesn't get parsed to the JSON representation of the action list suggested in the PR description, where `notification_received` etc is correctly taken to be the action trigger. Rather, the parser (through the `parseAppDataActionString()` util function) incorrectly infers that `notification_received | set_field: show_notification_bell: true;` represents an action defined in `action ID | params` format with no trigger defined, and adds `click` as the default trigger:
> 
> Screenshot 2025-07-14 at 16 07 44
> 
> This is partly an issue with our ambiguous action syntax, where some actions are defined in the form `trigger | action ID | params` and others are defined as `trigger | action ID : arg1 : arg2`. This ambiguity would be resolved if all actions used the former syntax, e.g. in this case: 
> ```
> notification_received | set_field | show_notification_bell: true;
> notification_interacted | set_field | show_notification_bell: false;
> notification_interacted | go_to | path: /user;
> ```
> 
> However as we don't want to drop support for the older action syntax at this stage, I think we need to make sure they are parsed as expected.

chrismclarke:
> I can see why this has immediate utility for the notifications use case and could be generally helpful for defining an action list in a variable such that it could be reused or referenced by multiple components. However I think that adding another special case row name isn't ideal and is quite opaque. So I'm happy for these changes to be merged, but I think they should be considered a hack or interim solution.

Agreed, although I think likely to be in place until fundamental issues address:

1. There isn't a clear system to create ad-hoc values of specific data types.

Anything typed into any spreadsheet cell will be passed as the type specified by excel, namely string, number or boolean. For a number of years I've heard discussion of a _Universal Data Parser_ that IDEMS was working on which I though was meant to make it easier to author more complex data types within a cell (e.g. array, json-object, array of json-objects etc.), however I haven't heard many updates for a while and am assuming the work has stalled out. So we need to rely on hacky workaround using either the column name (parameter_list, action_list) or variable name (my_var_list, my_action_list) to infer data types 

2. Variables don't always preserve data types across references. 

E.g. if I were to create an ad-hoc answer_list and try to store as a field, it will be serialised on storage but not deserialised on retrieval. This is something we could probably more proactively work on, using tools like [superjson](https://www.npmjs.com/package/superjson) to preserve data properties when serialising and handle deserialization correctly. 

3. Our template calc system prioritises string replacement ahead of JS evaluation. 

This has been an ongoing issue mentioned in various places and proposals to swap, but for now if we had a json object `my_var: { "a" : {"b" : true}}`  and try to reference `@local.my_var.a.b` calculation would break having replaced just the first `@local.my_var` with stringified json.

4. Action lists support both args and params, and given they are typically defined in nested arrays can be quite tricky to identify on inspection (assume a string with `|` operator most likely an action list, but also sometimes used within parameter lists

At some point we will likely need to revisit and unify action_list syntax for a fixed number of args, and consider any other ways to author to make it clear that the value contains an action list and not any other data structure (similar to the comments you also made above). 

So for now this PR aims to be a least-bad solution based on current templating constraints to minimally achieve what we need for the inline notification authoring. But yeah, far from ideal

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.ts
chrismclarke: I'm keen to keep it consistent with the logic for `list` and `collection` keywords which have the exact same logic
```
 if (rowName.endsWith("_list") || rowName?.includes("_list_"))
 ....
 if (rowName.endsWith("_collection") || rowName.includes("_collection_")) {
 ```
 
 I have no issue changing all to allow the keyword to also be a prefix as well as suffix, but that would be a potentially breaking change so better done as a follow-up.

jfmcquade (approved):
Thanks @chrismclarke, the updates look good and it's really nice to have the new [example_row_transform](https://docs.google.com/spreadsheets/d/1iKfq3-wan4-GYvg4-XL07YLDt5ZDGIfoCCEK78-HzrI/edit?gid=1745157248#gid=1745157248) template for documentation purposes.

I've added one minor fix with https://github.com/IDEMSInternational/open-app-builder/pull/3040/commits/ae391231a7d8dfc7b4a14fa9f961a8444b4028d7. Previously the date parsing was broken as the logic was only applying to string values. E.g. of broken state when parsing [debug_campaign_schedules](https://docs.google.com/spreadsheets/d/1Jf38D4pkoBO0QIUcw1IFAqlEokrUzhOESi3hNlW0vZM/edit?gid=998700467#gid=998700467):

Screenshot 2025-07-15 at 10 43 42

In comparing the parsed sheets on this PR branch vs master, I find that there's one unexpected change. When parsing [comp_display_group](https://docs.google.com/spreadsheets/d/1IlM2fbVYzVdwPBa9UEaKZhGGlLFhRHCo--T6uC-EQiE/edit?gid=7532536#gid=7532536), there are two cases of an empty `style_list` that was previously parsed to an empty array but is now parsed to an empty string:

Screenshot 2025-07-15 at 15 10 46

On investigation, it appears that both of these cases (F85 and F89 and of the [template](https://docs.google.com/spreadsheets/d/1IlM2fbVYzVdwPBa9UEaKZhGGlLFhRHCo--T6uC-EQiE/edit?gid=7532536#gid=7532536)) are instances of a style list cell containing nothing but a newline character. As this is an authoring error, I don't think we need to worry about a fix, but in theory we should be handling such cases to ensure that a `_list` field remains an array to avoid unexpected behaviour.

packages/scripts/src/commands/app-data/convert/processors/flowParser/parsers/template.parser.ts
jfmcquade: yeh happy to leave as it is for now, I don't think it's too important especially as we're planning to move away from relying on these special value names eventually

chrismclarke:
@esmeetewinkel - this should also now be ready for final review, please note the potential small breaking change to authoring

esmeetewinkel (approved):
PR description and sheet sync looks good, will merge here and test further on https://github.com/IDEMSInternational/open-app-builder/pull/3013

Noted about breaking change to authoring - I agree this shouldn't be much of an issue.

PR #3044, opened 2025-07-15
feat: parent group access code generator

chrismclarke:
PR Checklist

## Description
Add a new `plh_parent_group: generate_access_code` action to generate a 4 character alphanumeric `access_code` to associate with parent groups. This code is then used by cloud functions to link from rapidpro

## Dev Notes
In order to link from rapidpro, we needed a top-level document property that could be efficiently queried. As such this PR also extends the shared data service to allow assigning any custom top-level data to collections (subject to same security rule restrictions as previously)

This code still needs to be merged with the rest of the parent group data to view in the UI, however I think that can likely be handled as follow-up (code is visible in firestore which could be used to start testing rapidpro integration)

The code doesn't currently have any checks in place to ensure uniqueness - this would require a query to firestore for all docs matching the access_code, however a user would not have permission to perform that query as the code could belong to a different private document. So it would instead require a server-side function to check (and probably generate in the first place), which would be a little overkill for now. Given we only have a handful of users generating groups the chances of duplicates are low, and if there is the rapidpro flow will display an error message which the facilitator can use to prompt regenerating.

### Follow-up Tasks
@jfmcquade - I've still got more work to do on the rest of the rapidpro integration, so was wondering if you might be able to take over from here on the frontend? Specific next tasks:

- [ ] Expose the access code in the UI for facilitators to see/share with parents that want to join via rapidpro. Currently this sits at the root level of the firebase doc so would likely need to get merged with the local data_list data when querying and merging rest of parent list

- [ ] Update UI to merge/display data incoming from rapidpro webhook. With the shared group data this can be tested by sending a POST request to the cloud function (future will be triggered by rapidpro), e.g. request

```sh
curl --request POST \
  --url https://us-central1-debug-e8934.cloudfunctions.net/groupJoin \
  --header 'Authorization: Bearer example_access_token_1234567890' \
  --header 'Content-Type: application/json' \
  --header 'User-Agent: Insomnia/2023.5.6' \
  --data '{
	"access_code":"YYDG",
	"rapidpro_uuid":"1d3ea366-cfb9-4640-87e4-74e160ab7220",
	"rapidpro_fields":{
		"parent_name":"Chris from Rapidpro",
		"parent_gender":"M"
	}
}'
```

The data is populated as-is into `parentGroupData.parents` array

- [ ] Consider whether it might be more useful to move the `type: parent_group` to top level for easier querying in firebase, as well as `parent_group_id`, and just using the `data` entry to store the rest of group data (would be less nested). Although can appreciate might not be an ideal transition to make if already using data in production

## Review Notes
See action example in [debug_shared_parent_group](https://docs.google.com/spreadsheets/d/1fSMNZSWxBRoH6WQCrT-xdd5kOJjpM-wW5RQM59uSGgQ/edit?gid=99275306#gid=99275306)

## Git Issues

Closes #

## Screenshots/Videos
**Debug Sheet** - Now includes button to trigger access_code generation and placeholder to display current code (needs integration)
image

**Firestore** - entry now includes `access_code` following action button press and rapidpro user data shared from webhook
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "frontend",
-  "version": "0.20.0",
+  "version": "0.21.0",
   "author": "IDEMS International",
   "license": "See LICENSE",
   "homepage": "https://idems.international/",


src/app/feature/shared-data/shared-data.service.ts
```diff
@@ -138,6 +149,19 @@ export class SharedDataService extends AsyncServiceBase {
...
+   * This is used to extend typical shared data properties and should not be used to overwrite
+   * core data (admins, members, metadata etc.)
+   */
+  public setCustomCollectionMeta(id: string, key: string, value: any) {
+    if (key in DUMMY_SHARED_DATA_COLLECTION) {
+      console.error(`[Shared Data] cannot overwrite protected key:`, key, value);
+      return;
+    }
+    return this.provider.updateCollectionMetadata(id, { [key]: value });
+  }
```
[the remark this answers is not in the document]
chrismclarke: validation should happen in service that calls code

jfmcquade (approved):
Thanks @chrismclarke. These additions look straightforward and I can confirm that the randomly generated access code is correctly applied to the document in firestore. A couple of queries:
1. Is it expected behaviour that repeatedly triggering the action overwrites the previous access code that was applied? I'm not clear on the use case but I'm happy not to be for now, I'm imagining that this `generate_access_code` action would only be used occasionally in practice.
2. I'm not entirely clear on the "collection"/"document" distinction in general, but I think potentially we're using the wrong terminology, see my inline comment.

I'm happy to take over on the UI-facing changes that you've outlined, thanks. I think it probably makes sense to merge this PR and then address those as follow-ups, I can discuss the implementation with respect to the use case with @ChrisMarsh82

src/app/feature/shared-data/shared-data.service.ts
```diff
@@ -138,6 +149,19 @@ export class SharedDataService extends AsyncServiceBase {
     return this.provider.updateSharedData(id, key, value);
   }
 
+  /**
+   * Update top-level collection properties.
+   * This is used to extend typical shared data properties and should not be used to overwrite
+   * core data (admins, members, metadata etc.)
+   */
+  public setCustomCollectionMeta(id: string, key: string, value: any) {
```
jfmcquade, line 157: nit(non-blocking) I'm a bit confused about the terminology, but isn't the entity at this level a _document_ rather than a _collection_? It sits inside the `shared_data` collection and has a unique ID and has data directly on it, so my understanding is that we should be talking about a `DUMMY_SHARED_DATA_DOCUMENT` and `setCustomCollectionMeta()` method.

chrismclarke: You're right, it is a document not collection. It get's a bit extra confusing because once in the document we have `doc.data` which firestore refers to as all of the document data, and the `data` that we refer to as being shared data.

So to try disambiguate a little I've updated to `setCustomSharedMeta` to describe everything in a top-level document that is not the child `data` (e.g. members, admins or custom properties such as access_codes)

Updated with [5d6da05](https://github.com/IDEMSInternational/open-app-builder/pull/3044/commits/5d6da050d777720d93aa621893c635ace80f91f9)

chrismclarke:
> Thanks @chrismclarke. These additions look straightforward and I can confirm that the randomly generated access code is correctly applied to the document in firestore. A couple of queries:
> 
> 1. Is it expected behaviour that repeatedly triggering the action overwrites the previous access code that was applied? I'm not clear on the use case but I'm happy not to be for now, I'm imagining that this `generate_access_code` action would only be used occasionally in practice.

Yes it is intended to overwrite. This would cover a number of potential cases such as a user has put an access_code into the public domain but wants it to stop being valid after a period of time (can re-generate a new one), or if the code happens to be duplicate and they need a new one. Of course if neither of these are desirable authors can quite easily only show a generate button if an access code does not exist already. 

> 2. I'm not entirely clear on the "collection"/"document" distinction in general, but I think potentially we're using the wrong terminology, see my inline comment.

You're right, updated with [5d6da05](https://github.com/IDEMSInternational/open-app-builder/pull/3044/commits/5d6da050d777720d93aa621893c635ace80f91f9)

> I'm happy to take over on the UI-facing changes that you've outlined, thanks. I think it probably makes sense to merge this PR and then address those as follow-ups, I can discuss the implementation with respect to the use case with @ChrisMarsh82

Sounds good, yes I'm sure there will be a lot more fine-details which I'm less aware of, so keen not to be a bottleneck in the rest of the process. From my side I think all the logic should be in place now to pass data one-time from rapidpro to the app (pending per-deployment function deploys), so would be nice to start polishing up the UI a bit and addressing other issues as they come up

PR #3048, opened 2025-07-17
fix: firebase auth

### LINKED ISSUE #3049: [BUG] Firebase auth fails on first attempt
**Describe the bug**
When logging in (with Google) via firebase auth, the first attempt fails and throws errors from the firebase auth library. The next attempt does succeed, but subsequent attempts may fail, often relaunching multiple sign-in pop-ups on failure.

**App version**
`master`/`v0.21.0` (unreleased)

**To Reproduce**
See screen recording below

**Deployment Sheet(s)**

**Debug Sheet(s)**
[example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705) (using `google_sign_in_button` component)

**Screenshots and videos**

Demo of [example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705)

https://github.com/user-attachments/assets/68195652-0b67-4ca1-ad3c-25e8d94de3fc

**Additional context**
Quite possibly a regression issue introduced by #3030

jfmcquade:
PR Checklist

## Description

Fixes an issue (likely a regression issue introduced by #3030) whereby the auth sign in feature would not work on first attempt.

## Dev notes 

The multiple pop-up windows described in #3049 suggested that multiple attempts to authenticate were being made in parallel. This PR adds a check to ensure that only one auth attempt can be made at a time. I'm not entirely sure what causes these multiple attempts, but limiting to one open attempt seems sensible in any case.

Additionally, there was an issue with mismatched package versions: `@capacitor-firebase/authentication` required `firebase` version `^11.2.0`, but the version of that package was at `^10.9.0`. I had hoped that this change alone would resolve the issue, which it didn't, but I've included the update here that should have been part of #3030.

## Git Issues

Closes #3049

## Screenshots/Videos

[example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705)

https://github.com/user-attachments/assets/97b63d4e-5928-4da3-b6fb-7e4885793870

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "dompurify": "^3.2.4",
     "extract-math": "^1.2.3",
     "file-saver": "^2.0.5",
-    "firebase": "^10.9.0",
+    "firebase": "^11.2.0",
     "globalthis": "^1.0.3",
     "howler": "^2.2.4",
     "html5sortable": "^0.13.3",
--- a/src/app/shared/services/auth/providers/firebase.auth.ts
+++ b/src/app/shared/services/auth/providers/firebase.auth.ts
@@ -17,6 +17,8 @@ export type FirebaseAuthUser = User;
   providedIn: "root",
 })
 export class FirebaseAuthProvider extends AuthProviderBase {
+  private isAuthenticating = false;
+
   public override async initialise(injector: Injector) {
     const firebaseService = injector.get(FirebaseService);
     // TODO - is service required here?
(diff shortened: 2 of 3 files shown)


jfmcquade:
> Another interesting issue (why is it always firebase auth! 😢 )

I know, it's getting quite annoying! I don't look forward to resolving firebase auth issues because the package feels like a bit of a black box (although less so after #2991 and better understanding the interactions between the plugin and the native/web SDKs), plus it means testing on all three of our targeted platforms, which is always time consuming. But at least it feels like we're gradually wrangling our implementation into shape...

Thanks for the alt PR, the changes look good so I'll continue the discussion there.

Issue #3015, opened 2025-06-24
[FEATURE] Link to RapidPro Parent Text


ChrisMarsh82:
**What?**
We will need to bring parent data down from rapidPro (Malaysia deployment - could expand to more)

**Why?**
We are having validation issues with ParentText ids being entered incorrectly. 
New flow to be.
Facilitator creates a parent group within the App
Facilitator sends the unique id for the parent group to parents during the onboarding session.
Parent sends the parent group id to ParentText when prompted.
Parent name and ID gets pulled into the App under the parent group (this data might need to be expanded later)

- [x] From the App will need a way of ensuring Parent Group ID is unique in the app - users may not be logged in

- [x] Password protected end point within the app server to receive parent information from ParentText (Parent Group ID + Parent Name + Parent Text ID) Must provide a success message to validate parent Group ID 

- [ ] Set up flow in parent Text to allow users to enter a Parent Group ID

- [ ] Create webhook on ParentText to push to the app server endpoint

ChrisMarsh82:
@istride @chrismclarke This is the initial issue we discussed on Thursday. Please feel free to comment and create break out issues.

istride:
It's important to emphasise that ParentText will push the data back to the app, after a parent initiates a message flow using the group id. This implies that an API endpoint will need to be implemented, on the app side, to receive the data.

chrismclarke:
Following a quick call today, this is the general data flow landed on for the initial phase:

<img width="1140" height="439" alt="Image" src="https://github.com/user-attachments/assets/c95b9063-42c0-4b9b-ba26-d202dd10388c" />

This should hopefully complete a simple workflow where users can request access to a group from rapidpro flow via 4-digit access code, and be notified of response

Initial sync will include user rapidpro id, and whatever contact fields desirable to share with app

Future follow-up could include another function that fetches updates to users via the rapidpro contacts API (using uuid stored from groupJoin)

chrismclarke:
> For error responses that might use HTTP status code [400](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/400), it is worth considering whether [422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/422) would be more appropriate. For example, "bad payload" may well be a 400 if, say, the JSON payload is syntactically invalid. An error such as "invalid access code" is more likely to come from a well-formed request with incorrect information in it that can't be processed - a 422.

[...]

PR #3053, opened 2025-07-18
feat: parent group access code in local data

jfmcquade:
PR Checklist

## Description

Builds on #3044.

- Makes the access code available locally by saving it to the `rp_access_code` property on the parent group in the parent groups data list
    - This protected field will not be synced to the nested parent group data when performing a `plh_parent_group: push`
    - Originally I used `_access_code` as the name, but dynamic data fields with a "_" at the start of the key name are considered metadata and are handled differently from regular fields (one consequence was that the rendered value did not update automatically when the value changed in the source dynamic data)
- The `plh_parent_group: generate_access_code` action now ensures that the specified parent group is pushed to shared data. I.e. if the parent group is not already shared, it will be shared. Previously, the action logic would throw an error unless the specified parent group was already shared.

## Notes

Since the `plh_parent_group: generate_access_code` action now handles  pushing parent group data (in the case that the parent group data was not already pushed), additional parameters are required. Previously, just an ID for the targeted parent group was required, e.g. `plh_parent_group: generate_access_code | parent_group_id: my_id`. Now, since pushing involves merging data to/from multiple local data lists, the author must also supply values for a `parent_groups_data_list` param and `parent_data_list` param. This is in keeping with the other `plh_parent_group` actions. 

It may not be apparent to authors why these additional parameters are required and it would be possible to configure the feature to be passed these data list names once on initialisation to avoid passing them each time an action is called. However, keeping the actions completely standalone is desirable, and avoids requirements like providing these data list names for a specific feature in the deployment config (which we used for the task system but was not a good pattern). It would be possible to pass these values to the service once, e.g. via a `plh_parent_group: configure` action that is authored to run on start-up. However, the main issue is one of duplication of authored params, which would be resolved by implementing #3012: once we have actions that can be authored in one place per deployment that include default parameter values, the issue of duplication is avoided.

## Git Issues

Work towards #3015

## Screenshots/Videos

[debug_shared_parent_group](https://docs.google.com/spreadsheets/d/1fSMNZSWxBRoH6WQCrT-xdd5kOJjpM-wW5RQM59uSGgQ/edit?gid=99275306#gid=99275306)

access code now visible in app front-end UI and in firebase firestore console:

Screenshot 2025-07-22 at 11 23 12

Screenshot 2025-07-22 at 11 23 23

### DIFF
--- a/packages/components/plh/parent-group/plh-parent-group.service.ts
+++ b/packages/components/plh/parent-group/plh-parent-group.service.ts
@@ -28,7 +28,7 @@ interface IParent {
 }
 
 interface IParentGroup {
-  _access_code?: string;
+  rp_access_code?: string;
   archived: boolean;
   hidden: boolean;
   id: string;
@@ -179,7 +179,7 @@ export class PlhParentGroupService extends SyncServiceBase {
 
       // Update local parent group data to add access code
       await this.setLocalParentGroupProperty(parent_group_id, parent_groups_data_list, {
-        _access_code: code,
+        rp_access_code: code,
       });
     }
   }
@@ -227,7 +227,7 @@ export class PlhParentGroupService extends SyncServiceBase {
   private formatParentGroupDataForPush(parentGroup: IParentGroup): IParentGroup {
     // Remove any field whose key starts with "_"
     return Object.fromEntries(
-      Object.entries(parentGroup).filter(([key]) => !key.startsWith("_"))
+      Object.entries(parentGroup).filter(([key]) => !key.startsWith("rp_"))
     ) as IParentGroup;
   }
 
@@ -530,7 +530,7 @@ export class PlhParentGroupService extends SyncServiceBase {
       parentGroupData = this.hackTransformReadonlyParentGroupData(parentGroupData, sharedId);
     }
     // Add access code to parent group data. This protected field will not be pushed to shared data
-    parentGroupData._access_code = sharedParentGroupDoc.access_code;
+    parentGroupData.rp_access_code = sharedParentGroupDoc.access_code;
 
     if (completionTrackingDataList) {
       await this.hackUpdateCompletionTrackingDataList(


chrismclarke (commented):
question(blocking)
From the screenshot it appears the `access_code` is duplicated to the `parentGroupData` and pushed back up to the server. I'm not sure I fully understand in the PR where the logic for this is, and whether it's actually necessary - couldn't we just merge the access code when accessing data locally and not write back to the server? E.g. the data available is a merge of metadata + groupdata

packages/components/plh/parent-group/plh-parent-group.service.ts
```diff
@@ -160,6 +165,22 @@ export class PlhParentGroupService extends SyncServiceBase {
...
+        parentGroupsDataList: parent_groups_data_list,
+        parentsDataList: parents_data_list,
+      });
+      // 2. Update parent group data in local data to add access code
+      await this.setLocalParentGroupProperty(parent_group_id, parent_groups_data_list, {
+        access_code: code,
+      });
+      // 3. Push parent group data to shared data to add access code
+      await this.handlePush(parent_groups_data_list, parents_data_list, parent_group_id);
+
```
chrismclarke, line 182: nit(blocking):
This feels like a lot of server roundtrips going on... Is there any reason that the user needs to pull data first - I'm assuming if they've previously generated an access_code that would be up-to-date locally, or if using a new device would have synced all the data when loading parent group for the first time. Or if using 2 devices would need alternate mechanism to keep devices in sync (not dependant on generating access codes)

At the same time I'm not sure if we really need to duplicate the access code back to the db group data - particularly as this then becomes writeable from the group and could get out of sync. I'd probably be more in favour of automatically merging top-level doc fields into the group data as required, and prefixing with `_` to make more distinct as metadata e.g. `{...group_data, _access_code: doc.access_code}`. Then when writing back to the DB remove metadata fields, `const {_access_code,...rest} = group_data`

If there's other good reason for the round trip I'd also suggest either only writing the local access code after db `setCustomSharedMeta` operation successful, or if planning an optimistic update then having some fallback to undo if the method fails

chrismclarke (changes requested):
See one comment below related to server roundtrips, although ultimately I don't care too strongly given that this component is a plh-specific one and not really intended for general use.

jfmcquade:
Thanks @chrismclarke, that was helpful. Indeed the round trip was unnecessary. After discussing the use case with @ChrisMarsh82, we agreed that it would be desirable for the `plh_parent_group: generate_access_code` action to push the specified parent group to shared data in the case that it is not already shared.

I've updated this PR and the description to do as you suggested and avoid duplicating the access code, instead saving it to an `_access_code` field locally and ignoring this field when pushing.

chrismclarke (approved):
Looking much more straight-forward, thanks for the updates @jfmcquade

PR #3055, opened 2025-07-21
Feat: plh_parent_group feature handles parents added from RapidPro

jfmcquade:
PR Checklist

- [ ] Builds on #3053, which should be merged first

## Description

Updates logic of various `plh_parent_group` actions to handle data populated via the RapidPro API.

## Git Issues

Work towards #3015

## Screenshots/Videos

Demo TBC

### DIFF
--- a/packages/components/plh/parent-group/plh-parent-group.service.ts
+++ b/packages/components/plh/parent-group/plh-parent-group.service.ts
@@ -28,6 +28,7 @@ interface IParent {
 }
 
 interface IParentGroup {
+  access_code?: string;
   archived: boolean;
   hidden: boolean;
   id: string;
@@ -160,9 +161,20 @@ export class PlhParentGroupService extends SyncServiceBase {
         return;
       }
       const code = generateRandomCode(4);
+
+      // Update shared data to add access code as collection meta
       // TODO - validate code to check no conflict with other groups
       // will likely require backend function to generate and check as user will not have query permission
       await this.sharedDataService.setCustomSharedMeta(shared_id, "access_code", code);
+
+      // Update parent group data in local data to add access code
+      await this.setLocalParentGroupProperty(parent_group_id, parent_groups_data_list, {
+        access_code: code,
+      });
+
+      // Update parent group data in shared data to add access code
+      parentGroupData.access_code = code;
+      await this.pushParentGroupData(parentGroupData);
     }
   }
 
--- a/packages/components/plh/parent-group/plh-parent-group.service.ts
+++ b/packages/components/plh/parent-group/plh-parent-group.service.ts
@@ -147,7 +147,11 @@ export class PlhParentGroupService extends SyncServiceBase {
     await this.updateLocalParentGroupData(parentGroup, parentGroupsDataList, parentsDataList);
   }
 
-  public async generateAccessCode(parent_groups_data_list: string, parent_group_id: string) {
+  public async generateAccessCode(
+    parent_groups_data_list: string,
+    parents_data_list: string,
+    parent_group_id: string
+  ) {
     const parentGroupQuery = this.dynamicDataService.query$("data_list", parent_groups_data_list, {
       selector: { id: parent_group_id },
     });
@@ -167,14 +171,19 @@ export class PlhParentGroupService extends SyncServiceBase {
       // will likely require backend function to generate and check as user will not have query permission
       await this.sharedDataService.setCustomSharedMeta(shared_id, "access_code", code);
 
-      // Update parent group data in local data to add access code
+      // Now update local and shared parent group data to add access code:
+      // 1. Pull parent group data from shared data to ensure local data is up-to-date
+      await this.handlePull({
+        parentGroupId: parent_group_id,
+        parentGroupsDataList: parent_groups_data_list,
+        parentsDataList: parents_data_list,
+      });
+      // 2. Update parent group data in local data to add access code
       await this.setLocalParentGroupProperty(parent_group_id, parent_groups_data_list, {
         access_code: code,
       });
-
-      // Update parent group data in shared data to add access code
-      parentGroupData.access_code = code;
-      await this.pushParentGroupData(parentGroupData);
+      // 3. Push parent group data to shared data to add access code
+      await this.handlePush(parent_groups_data_list, parents_data_list, parent_group_id);
     }
   }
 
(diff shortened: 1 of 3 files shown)


jfmcquade:
Thanks, @chrismclarke. Appreciate you fixing the test integration, I was led astray by AI I think, although still surprised that the solution is so simple in the end.

I've fixed the tests and tidied up some of the code following the gemini review. I think it should be in a state to review now, although I appreciate that some of the logic is a bit hard to follow out of the context of the way data is handled in our use case. I'm going to try and add some additional parent group editing functionality to the debug deployment for testing purposes, as it's not currently possible to do the same data manipulations we do in the production deployment, and therefore not possible to test the feature fully on debug.

I will re-request a review from you when that's ready.

jfmcquade:
I've tidied up [debug_shared_parent_group](https://docs.google.com/spreadsheets/d/1fSMNZSWxBRoH6WQCrT-xdd5kOJjpM-wW5RQM59uSGgQ/edit?gid=99275306#gid=99275306) somewhat, it now displays the parents' first names (or names from RapidPro) for each parent group. I haven't had a chance to fully flesh out the debug example with features like editing/deleting/adding parents to groups. @ChrisMarsh82 has tried testing this on the relevant production deployment but ran into issues at the end of the working day. 

As we're running out of time this week, @chrismclarke please could you do a code review of this PR now that I've fixed the tests and tidied up the code? I'm not sure how deeply you want to engage with the details of the code as it's so tailored to the use case, but hopefully you can check how things look to you generally.

packages/components/plh/parent-group/plh-parent-group.service.ts
```diff
@@ -461,6 +444,40 @@ export class PlhParentGroupService extends SyncServiceBase {
...
+
+    return parentGroup;
+  }
+
+  /**
+   * Format parent data for upload to shared data by removing RapidPro fields,
+   * since these are managed by RapidPro in shared data
+   */
+  private hackRemoveRapidProFieldsFromParentData(parent: IParent | IParentFromRapidPro): IParent {
+    if (this.isParentFromRapidPro(parent)) {
```
chrismclarke, line 470: nit(non-blocking): Just to check, is the long-term plan to have some parents registered through rapidpro and others from alternate sources? Or is the discrepancy just due to data that's manually been put in the db for testing? Working with one type of data would definitely be easy than multiple, although in this case I think likely the same method can still be ran against any parent data even if not from rapidpro (given you're just filtering out object keys by prefix)

Assuming the service is designed to be closely coupled with rapidpro, I think you can probably remove `hack` from these methods as they are somewhat integral to the service and unlikely to be removed in the future.

packages/components/plh/parent-group/plh-parent-group.service.ts
```diff
@@ -547,6 +571,59 @@ export class PlhParentGroupService extends SyncServiceBase {
...
+    const { rapidpro_uuid, rapidpro_fields, ...rest } = parent;
+    const parsedRapidProFields = Object.fromEntries(
+      Object.entries(rapidpro_fields).map(([key, value]) => [`rp_${key}`, value])
+    );
+    return {
+      ...rest,
+      ...parsedRapidProFields,
+      id: rapidpro_uuid,
+      group_id: parentGroupId,
+      rapidpro_uuid,
```
chrismclarke, line 624: nit(non-blocking): you might want to just keep as `rp_uuid` to save maintaining 2 separate prefixes related to rapidpro. I feel like there's probably not a massive need to distinguish data coming from rapidpro fields vs rapidpro meta give the only non-field is the uuid.

packages/components/plh/parent-group/plh-parent-group.service.ts
```diff
@@ -632,4 +709,86 @@ export class PlhParentGroupService extends SyncServiceBase {
...
+   * Merge two arrays of parents, ensuring that parents representing the same entity are merged,
+   * and any RapidPro fields added to the shared data are preserved.
+   * - For each incoming parent:
+   *   - If an existing parent matches by `id`, or by `rapidpro_uuid` (where `rapidpro_uuid === incoming.id`), merge fields:
+   *     - Use the incoming parent as the base, but add `rapidpro_fields` and `rapidpro_uuid` from the existing parent if present.
+   *   - If no match, use the incoming parent as-is.
+   * - After merging, add any existing parent with a `rapidpro_uuid` (and no `id`) that was not already merged.
+   * This ensures no duplicates and preserves RapidPro data.
+   */
+  private hackMergeParentsArrays(
```
chrismclarke, line 750: nit(non-blocking):
Appreciate the effort you've gone to for tests and documentation. although the method is still a little hard to follow. I think the heavy use of negatives, ternary statements and type-casting makes for a confusing read. I might suggest trying to tidy the logic particularly of
```ts
let incomingParent = (existingParent as IParent).id
        ? incomingById.get((existingParent as IParent).id)
        : undefined;
      if (!incomingParent && existingParent.rapidpro_uuid) {
        incomingParent =
          incomingById.get(existingParent.rapidpro_uuid) ||
          incomingByRapidproUuid.get(existingParent.rapidpro_uuid);
      }
```
I'm guessing maybe splitting into separate lookup function which can return early when an incomingParent is retrieved by the first check. Also guessing you don't need to cast existingParent when both parent types have an id field

chrismclarke (approved):
Thanks @jfmcquade 
I've added a handful of (non-blocking) comments, so happy to just leave to you to decide whether you want to tidy up a bit more as part of follow-up or not.

If wanting to merge first and follow-up later, my main suggestion would be to decide whether you want to maintain both `rp` and `rapidpro` prefixes (or just rename `rp-uuid` for a single prefix), and to confirm you want the level of nesting for the data.

E.g. you could remove the current `parentGroupData` nest and just use `data` (moving the `type` to top-level meta)

```ts
data:{
  text: 'Parent group 2',
  ...
  parents: [ ]
},
type: 'parentGroup'
```

Or, given that you separately manage the `parents` in a separate data_list to the parent_group data, you might want to also separate at the group level, e.g.
```ts
data:{
  parentGroup: { ....},
  parents: [ ]
}
```
Of course this may no longer be a decision you can make if the db is actively being populated with production data (planning for data migrations is painful when considering not all app users will update their app, so need to support both old and new in parallel while migrating) - This is also why good to try to nail data structures as early as posssible.

jfmcquade:
Thanks, @chrismclarke. I've tidied up the code based on your instructions, see [changes since your last review](https://github.com/IDEMSInternational/open-app-builder/pull/3055/files/5625f42d16ebb6b065c6abc637500dc93a40ca5e..bc7fa5b21c12713aafc007b7ad7fa6cb1ad26712). I've also replied to the inline comments with specific new commits, since the changes to the file structure means that the constituent diffs are not clear from all the commits at once. 

The meaningful changes beyond code tidying are:
- Rename local field `rapidpro_uuid` -> `rp_uuid`

Unfortunately, the shared parent group feature is already in use in production, so we're somewhat stuck with the suboptimal data structure.

PR #3058, opened 2025-07-24
chore(plh-parent-group): when importing rapidpro parents, generate unique local ID

jfmcquade:
PR Checklist

## Description

Follow-up to #3055.

When first pulling parent group data from shared data that includes a parent that has been added by RapidPro, a unique ID is generated to be associated with the parent in local data (as the `id` field). This ID is generated by concatenating the parent group ID and the `rapidpro_uuid`, separated with a `+` symbol. Previously, the `rapidpro_uuid` alone was used, but this didn't allow for a parent appearing in multiple groups (which may happen, temporarily, if a parent is moved from one group to another, for example). 

## Git Issues

Closes https://github.com/IDEMSInternational/open-app-builder-functions/issues/8

### DIFF
--- a/packages/components/plh/parent-group/utils/rapidpro.utils.ts
+++ b/packages/components/plh/parent-group/utils/rapidpro.utils.ts
@@ -86,7 +86,8 @@ function transformParentWithRapidProDataToLocalFormat(
   return {
     ...rest,
     ...parsedRapidProFields,
-    id: rapidpro_uuid,
+    // Use a combined id of the parent group id and the rapidpro_uuid to ensure uniqueness
+    id: parentGroupId + "+" + rapidpro_uuid,
     group_id: parentGroupId,
     rp_uuid: rapidpro_uuid,
   } as IParent;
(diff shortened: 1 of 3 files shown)


ChrisMarsh82 (approved):

packages/components/plh/parent-group/utils/rapidpro.utils.ts
```diff
@@ -86,7 +86,8 @@ function transformParentWithRapidProDataToLocalFormat(
   return {
     ...rest,
     ...parsedRapidProFields,
-    id: rapidpro_uuid,
+    // Use a combined id of the parent group id and the rapidpro_uuid to ensure uniqueness
+    id: `${parentGroupId}+${rapidpro_uuid}`,
```
chrismclarke, line 90: nit(non-blocking): haven't seen `+` used to indicate concatenated ids much before, but don't think it would really cause much harm. Given neither parentGroupId no rapidpro_uuid will contain special characters should still be pretty easy to split on. Alternatives could otherwise be `_` (good for files, bad if either ids would include character), or sometimes I see `::` used, e.g. `parentGroup1234::parentUUID1`, but think the `+` should still do the job fine.

PR #3064, opened 2025-07-29
fix: pull shared parent data on ios

### LINKED ISSUE #3063: [BUG] On iOS, pulling shared parent group data does not receive data in parents array
**Describe the bug**
_Include a clear and concise description of what the bug is._

**App version**

**To Reproduce**

**Deployment Sheet(s)**

**Debug Sheet(s)**

**Screenshots and videos**

**Additional context**

jfmcquade:
PR Checklist

## Description

Fixes an issue with shared parent group data on iOS, detail in #3063.

## Dev notes

This was a tricky one to debug. The main breakthrough was realising that the snapshots coming through from firestore were coming from the firestore cache on iOS. It seems that there's a potential race condition in terms of what the `FirebaseFirestore.addCollectionSnapshotListener` returns in terms of whether the data is from the cache or the server. For whatever reason, it seems that on iOS the cache result is reliably returned first in contrast to the other platforms, meaning that taking the `firstValueFrom` in the PLH parent group service was always returning stale data from the cache on iOS.

It still wasn't trivial to resolve after this, since the `@capacitor-firebase/firestore` library doesn't expose a simple method to get a snapshot that can be configured to explicitly query the server over the cache, but I think I've managed to configure our existing methods to handle this.

## Git Issues

Closes #3063
Related to #3015

## Screenshots/Videos

iOS emulator running the `plh_facilitator_my` deployment correctly puling the newly added parent data:

Screenshot 2025-07-29 at 16 53 41

### DIFF
--- a/packages/data-models/fields.ts
+++ b/packages/data-models/fields.ts
@@ -28,6 +28,7 @@ enum PROTECTED_FIELDS {
    * https://capacitorjs.com/docs/plugins/web#aliases
    **/
   NOTIFICATION_PERMISSION_STATUS = "notification_permission_status",
+  OPERATING_SYSTEM = "operating_system",
   PLATFORM = "platform",
   SERVER_SYNC_LATEST = "server_sync_latest",
 }
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -4,6 +4,7 @@ import { Router } from "@angular/router";
 import { Capacitor } from "@capacitor/core";
 import { SplashScreen } from "@capacitor/splash-screen";
 import { App } from "@capacitor/app";
+import { Device } from "@capacitor/device";
 import { DbService } from "./shared/services/db/db.service";
 import { SkinService } from "./shared/services/skin/skin.service";
 import { ThemeService } from "./feature/theme/services/theme.service";
@@ -170,7 +171,8 @@ export class AppComponent {
     this.localStorageService.setProtected("APP_VERSION", _app_builder_version);
     this.localStorageService.setProtected("CONTENT_VERSION", _content_version);
     this.localStorageService.setProtected("PLATFORM", Capacitor.getPlatform());
-
+    const { operatingSystem } = await Device.getInfo();
+    this.localStorageService.setProtected("OPERATING_SYSTEM", operatingSystem);
     const appEnv = environment.production ? "production" : "development";
     this.localStorageService.setProtected("APP_ENVIRONMENT", appEnv);
     this.localStorageService.setProtected("APP_HOSTNAME", location.hostname);
--- a/yarn.lock
+++ b/yarn.lock
@@ -4902,12 +4902,12 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@eslint/core@npm:^0.15.0":
-  version: 0.15.0
-  resolution: "@eslint/core@npm:0.15.0"
+"@eslint/core@npm:^0.15.1":
+  version: 0.15.1
+  resolution: "@eslint/core@npm:0.15.1"
   dependencies:
     "@types/json-schema": ^7.0.15
-  checksum: 74ae0a38678fa1cae44732c5376b800c842993224301e6a4abc57edd123cfcb8c7746f38892adb8747b4f56c0a1c1db71ff619210508628eb9227d08a0d7bb92
+  checksum: 9215f00466d60764453466604443a491b0ea8263c148836fef723354d6ef1d550991e931d3df2780c99cee2cab14c4f41f97d5341ab12a8443236c961bb6f664
   languageName: node
   linkType: hard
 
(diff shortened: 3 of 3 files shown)


chrismclarke (changes requested):
Annoying issue, but thanks for getting to the bottom of the root cause.

From the [Firebase Docs](https://firebase.google.com/docs/firestore/manage-data/enable-offline) it says that offline persistence is set to on by default for Android and IOS platforms, which both explains the issue and why behaviour differs to web. 

From the capacitor-firebase side, I see a discussion related to adding support for configuring persistence (saying not currently available) - an issue was created is still outstanding (although the description isn't very clear). This is consistent with the lack of configuration in the current code
```
https://github.com/capawesome-team/capacitor-firebase/discussions/713
``` 
```
https://github.com/capawesome-team/capacitor-firebase/issues/715
```

I'm somewhat hesistant to us manually managing server/cache layers on firebase for Android and IOS given that we already have our own caching layer (rxdb) and would want to stay consistent with other platforms. The code starts to get quite complicated and may be harder to maintain in the future. The potential race conditions between server and cache is also cause for concern.

So if possible I'd like to first explore fully disabling cache on ios and firestore, as if we can that will make ongoing changes easier to manage. Given this would require change to the capacitor-firebase package, our 3 options would be:

1. Raise a PR to fix
2. Take a fork and fix
3. Patch the package

I think my short term preference would be option (3), and if working as expected we could try to raise a PR to eventually get merged into the main package.

Given that this will also need testing on ios - would you be ok to try this out first @jfmcquade ?
I've written up a separate issue to document steps for patching packages in case useful #3065

jfmcquade:
Thanks, @chrismclarke, I like the suggested approach of patching the package, that's new to me. The instructions in #3065 are really helpful.

However I haven't yet been able to patch the package in a way that resolves the issue. I have added the following to the swift code:
```swift
         super.init()
         if FirebaseApp.app() == nil {
             FirebaseApp.configure()
+            let settings = Firestore.firestore().settings
+            settings.isPersistenceEnabled = false
+            Firestore.firestore().settings = settings
         }
```
But I find that the app still fetches data from the firestore cache. To confirm this (as well as the data not coming through to the UI as expected), I'm logging out the `event.snapshots` returned from the `FirebaseFirestore.addCollectionSnapshotListener` method, the `metadata` property of which includes `fromCache: true` for the stale data.

We're overdue delivering a version of the app for testing, including the feature related to the current PR, and I need to publish one before I finish today. So my plan is to publish a release built from this PR branch temporarily to meet this need, and then work on a more sustainable solution before at real production release. 

If you have capacity this week, would you be able to look into the equivalent Android implementation as suggested in #3065? Don't worry if you don't have the time, but if so it might help me with the issues I'm having working on the iOS solution.

chrismclarke:
> Thanks, @chrismclarke, I like the suggested approach of patching the package, that's new to me. The instructions in #3065 are really helpful.

Yeah for small changes it's pretty convenient, also a good way to test changes to a package before opening a PR without having to fully checkout and build the package locally. 

> However I haven't yet been able to patch the package in a way that resolves the issue. I have added the following to the swift code:
> 
> ```swift
>          super.init()
>          if FirebaseApp.app() == nil {
>              FirebaseApp.configure()
> +            let settings = Firestore.firestore().settings
> +            settings.isPersistenceEnabled = false
> +            Firestore.firestore().settings = settings
>          }
> ```
> 
> But I find that the app still fetches data from the firestore cache. To confirm this (as well as the data not coming through to the UI as expected), I'm logging out the `event.snapshots` returned from the `FirebaseFirestore.addCollectionSnapshotListener` method, the `metadata` property of which includes `fromCache: true` for the stale data. I wonder if there's an issue with the web layer caching even on the native device, but I haven't progressed.

Very strange, looking at the [firestore swift docs](https://firebase.google.com/docs/reference/swift/firebasefirestore/api/reference/Classes/FirestoreSettings?_gl=1*3z2ly8*_up*MQ..*_ga*MzI3Nzg3MzQ0LjE3NTM4OTI3NDQ.*_ga_CW55HF8NVT*czE3NTM4OTI3NDQkbzEkZzAkdDE3NTM4OTI3NDQkajYwJGwwJGgw#ispersistenceenabled) it says `isPersistenceEnabled` is deprecated (although should still work) - but you could also try [cacheSettings](https://firebase.google.com/docs/reference/swift/firebasefirestore/api/reference/Classes/FirestoreSettings.html?_gl=1*18x6p0w*_up*MQ..*_ga*MzI3Nzg3MzQ0LjE3NTM4OTI3NDQ.*_ga_CW55HF8NVT*czE3NTM4OTI3NDQkbzEkZzAkdDE3NTM4OTI3NDQkajYwJGwwJGgw#/c:objc(cs)FIRFirestoreSettings(py)cacheSettings)

```swift
let settings = FirestoreSettings()
settings.cacheSettings = MemoryCacheSettings()
Firestore.firestore().settings = settings
```

It might also just be a case that your changes aren't fully making it into the build. I'd recommending adding some logs in the swift code and seeing if they turn up in xcode

> We're overdue delivering a version of the app for testing, including the feature related to the current PR, and I need to publish one before I finish today. So my plan is to publish a release built from this PR branch temporarily to meet this need, and then work on a more sustainable solution before at real production release.

Sounds sensible for now

> If you have capacity this week, would you be able to look into the equivalent Android implementation as suggested in #3065? Don't worry if you don't have the time, but if so it might help me with the issues I'm having working on the iOS solution.

Yeah I can try take a quick look today

chrismclarke:
@jfmcquade 
Just as a quick update, I was able to make changes to the java code, however looking more closely it seems like Firestore does using caching regardless of configuration, the only different is whether it uses a persistent cache (default) or in-memory cache. The in-memory cache would still prevent server-only requests if the the endpoint has been synced as part of the current app session (clears on app restart)

So actually I think the best way forward will be to enforce server-only requests at the method level as you first attempted. I'll propose a few changes to this PR (mostly to enforce the server-only requests instead of making configurable), but then should hopefully be good to go

chrismclarke:
I've tidied the code and moved into #3064 - would you be happy to do one last test to confirm working @jfmcquade ?

PR #2990, opened 2025-06-10
Feature: Add toast action

markharmon:
PR Checklist

- [x] PR title descriptive (can be used in release notes)

## Description

Adds the toastService which allows the toast action to be used:

`click | toast | message: toast this`

[feature_action_toast](https://docs.google.com/spreadsheets/d/1WGR_pxk-7iBHKX2BDdzdmjiaVjiwe-qKIou47uZFqHo/edit?gid=99275306#gid=99275306)

## Screenshots/Videos

image

PR #3073, opened 2025-08-01
fix: delete account

### LINKED ISSUE #3074: [BUG] Delete account sometimes fails with 'requires recent login' error
**Describe the bug**
When attempting to delete the user's auth account via the `auth: delete_account` action, sometimes the action fails, and error is thrown with the code `auth/requires-recent-login`.

[Google's ios docs](https://firebase.google.com/docs/reference/ios/firebaseauth/api/reference/Enums/FIRAuthErrorCode#:~:text=password%20more%20than-,5%20minutes,-after%20signing%20in) suggest that the amount of time for which the latest login is considered "recent" is 5 minutes, although this may differ by platform.

**App version**
0.20.2

**Debug Sheet(s)**
[example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705) launching [example_delete_account](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=79442240#gid=79442240) as pop up:

https://github.com/user-attachments/assets/75a031b0-b32d-42fe-9416-b6a9c2ea8606

jfmcquade:
PR Checklist

## Description

Fixes an issue where the `auth: delete_account` action (introduced in #3011) would fail if the user had not recently logged in. The new behaviour is that, in such cases, the user is prompted to re-authenticate via their original login method before the operation completes.

Ideally, it would be good to display a user-facing message in such cases, e.g. (from [this thread](https://github.com/capawesome-team/capacitor-firebase/issues/539#issuecomment-1879614601)):
```
This operation is sensitive and requires recent authentication. Sign in again before retrying this request.
```

However, authoring this is difficult without a method of exposing and responding to success/error states of action logic from the authoring level. One option might be to make use of the toast feature (#2990), to display a message that is passed as a string param (therefore translatable) to the `auth: delete_account` action. This could be added to this PR or addressed as a follow-up. However, I think that the behaviour as of this PR, where the user is prompted to log in again when attempting to delete their account, is not particularly jarring to the user.

## Testing

If wanting to test, be aware that the length of time after which a log in stope to be considered "recent" seems to be [5 minutes](https://firebase.google.com/docs/reference/ios/firebaseauth/api/reference/Enums/FIRAuthErrorCode#:~:text=password%20more%20than-,5%20minutes,-after%20signing%20in).

## Git Issues

Closes #3074 

## Screenshots/Videos

Demo showing [example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705) launching [example_delete_account](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=79442240#gid=79442240) as pop up (I had already let 5 minutes elapse since logging in):

https://github.com/user-attachments/assets/65db0c8c-5d90-4843-a9ef-53265d4c4907


jfmcquade:
> Just to check - do we also need to delete local and server data related to the user too? (Is the purpose of account deletion to clear all user data for GDPR/store compliance?)
> 
> If so we should perhaps make a follow-up issue to address

The impetus for adding account deletion was to conform with Apple's hard requirement for approving apps that "apps that support account creation must let users initiate deletion of their account within the app" ([docs](https://developer.apple.com/support/offering-account-deletion-in-your-app)). This basic implementation was sufficient to allow apps to pass Apple's review process, but I agree we should consider whether we need a more thorough implementation of account deletion in order to truly satisfy Apple's requirements and GDPR compliance. I've made an issue here: #3132.

PR #3092, opened 2025-08-14
feat: display group background image tiling

### LINKED ISSUE #3077: [FEATURE] Repeating background image behind progress path
**What?**
Ability to display a pattern / tiling / repeating background image behind the progress path component

**Other details**
Content issue: https://github.com/ParentingForLifelongHealth/plh-kids-teens-app-za-content/issues/47

jfmcquade:
PR Checklist

- [x] Builds on #3083, which should be merged first

## Description

- Enables overriding of the display group component's `background-size` CSS property from an authored style list
- Updates the `plh_kids_kw` theme to set the colour of the path rendered by the `progress_path` component to yellow

This enables authoring of a repeating tiled background image behind the `progress_path_component`, by wrapping it in a display group with the necessary parameters and style overrides. See demo below, achieving the desired use case of #3077.

Originally I had intended to expose options on the `progress_path` component itself, however this seemed like a more general solution. Likewise, at first I tried exposing dedicated parameters on the display group that would encompass a collection of CSS styles (e.g. `background-style: tile`), however to get the flexibility that will likely be required, it made the most sense to access the relevant CSS properties directly in the style list for now.

## Git Issues

Closes #3077 
Addresses some of the points in https://github.com/ParentingForLifelongHealth/plh-kids-teens-app-za-content/issues/183 (pending authoring changes)

## Screenshots/Videos

[demo_progress_path_bg](https://docs.google.com/spreadsheets/d/14BBKG07nqUWDmcscdUMv1AKYSWGBUPCNDT-VcSERHyU/edit?gid=44241983#gid=44241983):

Screenshot 2025-08-14 at 14 38 45

Screenshot 2025-08-28 at 10 58 34

### DIFF
--- a/src/theme/themes/plh_kids_kw/_index.scss
+++ b/src/theme/themes/plh_kids_kw/_index.scss
@@ -91,7 +91,7 @@
     points-item-background-complete: var(--ion-color-primary-200),
     points-item-background: var(--ion-background-color),
     points-item-border: rgba(black, 0.07),
-    progress-path-line-background: white,
+    progress-path-line-background: var(--ion-color-primary-variant),
     radio-group-background-secondary: var(--color-accent-orange-90),
     radio-group-background-selected-secondary: var(--color-accent-orange-80),
     radio-group-background-selected: var(--ion-color-primary-300),


chrismclarke (approved):
I definitely think applying the background image within a display group is a much nicer solution than trying to expose various options for setting at the component level.

From what I can tell, by not explicitly forcing the `background-size: contain` property you can provide static values, e.g. `background-size: 630px` which will prevent the repeated background from scaling in size on smaller devices.
|before|after|
|----|----|
| Repeated background auto-scales on small devices | Fixed background size can be authored for consistent sizing on all devices (default handling will still auto-scale) |
| localhost_4200_template_demo_progress_path_bg(Nexus 5X) (1) | localhost_4200_template_demo_progress_path_bg(Nexus 5X) |




I think this both makes sense to me from a dev point and looks nicer, so overall I think this is a really nice PR

Issue #2865, opened 2025-03-21
[BUG] Full-screen pop-up excessive padding


esmeetewinkel:
**Describe the bug**
Full screen pop ups have slightly excessive padding, causing their content to show narrower than templates accessed through a `go_to` action. This holds both when the `showCloseButton` is true and false.

**App version**
v0.18.0

**Deployment Sheet(s)**
- splash screens of any app.
   ![Image](https://github.com/user-attachments/assets/c70ad793-9adf-40c0-840d-772fb5ea47b4)
- [module](https://docs.google.com/spreadsheets/d/1hIT1V3xGZRFyD-k2heJibYr-UTqkkCc6eCvreiokNMk/edit?disco=AAABgrhIKrI) template in `plh_kids_tz` deployment

**Screenshots and videos**
`padding: 0 2rem` (default)
![Image](https://github.com/user-attachments/assets/87c92851-f2be-4358-a2b0-5e75fd4ddb6e)

~`padding: 0 2rem`~ (manually in dev tools side panel)
![Image](https://github.com/user-attachments/assets/bb6abbe7-4f9e-4057-9000-fef31ce44e42)

esmeetewinkel:
@jfmcquade not essential, but would be nice to get done for `plh_kids_teens_za` so I'm adding it to the milestone just in case. 

I'm wondering if this might relate to https://github.com/IDEMSInternational/open-app-builder/issues/3042

PR #3093, opened 2025-08-15
feat: allow sticky display groups to ignore safe area

### LINKED ISSUE #3042: Footer of full screen pop up should touch bottom of screen
**Details**
A footer on a full screen pop up should touch the border of the screen. By default it seems to have some padding (maybe 24px?). @jfmcquade Could you advise an authoring workaround or code fix? 

**Screenshots and videos**

jfmcquade:
PR Checklist

## Description

Adds a new parameter to the `display_group` component: `ignore_safe_area`. This parameter only has an effect on "sticky" instances of the component, i.e. those with the parameter `sticky: top` or `sticky: bottom`.

Possible values of `ignore_safe_area` and behaviour:
- `false` (default)
  - The sticky display group will attach to the top/bottom of the usable screen, within the "safe" area of the device, e.g. avoiding the native device area covered by a notch and the native bottom nav bar
- `true`
  - The sticky display group will attach to the very top/bottom of the visible screen, potentially overlapping the sections outside of the "safe" are of the device. This will only happen if the app is already displayed as "edge-to-edge" on the device, which is determined by platform and OS version, see examples below.

@esmeetewinkel Note that this can be used to replace the workaround you currently use of applying a negative margin on the sticky header in some instances.

## Naming

I am open to suggestions of better names for this param. "ignore_safe_area" is a bit of a technical name to expose to authors, however it was the best descriptive name I could come up with, and we can consider this behaviour somewhat advanced anyway. 

Additionally, there is a complication that on web platforms, the param has a desired effect even though there is no official "safe area" to speak of. Specifically, if the sticky display group appears within a fullscreen pop-up, then the author may want it to ignore the default padding of that pop-up, or to stay within it. This was the use-case that prompted the #3042 being opened. I think it makes sense that the pop-up padding should be respected by default, as the content within the sticky display group may want to be displayed within those bounds (currently there are many cases of authored navigation buttons in a `sticky: bottom` display group, for example). However, there are also use-cases like #3042, where a purely visual element is being made sticky and would be better suited to sit at the very bottom of the screen, ignoring any pop-up padding. With this in mind, a more accurate name for the param might be `use_screen_edges` or `sticky_absolute` or `sticky_ignore_padding` or something along those lines. But I think that `ignore_safe_area` is probably a more helpful name than any of those.

#2865 is also relevant here – perhaps fullscreen pop-ups shouldn't apply their own additional padding at all? I will investigate that independently.

## Git Issues

Closes #3042

## Screenshots/Videos

Demos of [comp_display_group](https://docs.google.com/spreadsheets/d/1IlM2fbVYzVdwPBa9UEaKZhGGlLFhRHCo--T6uC-EQiE/edit?gid=7532536#gid=7532536), which launches [dg_sticky_safe_area_1](https://docs.google.com/spreadsheets/d/1IlM2fbVYzVdwPBa9UEaKZhGGlLFhRHCo--T6uC-EQiE/edit?gid=1881605374#gid=1881605374) and [dg_sticky_safe_area_2](https://docs.google.com/spreadsheets/d/1IlM2fbVYzVdwPBa9UEaKZhGGlLFhRHCo--T6uC-EQiE/edit?gid=1478732095#gid=1478732095) as fullscreen pop-ups.

Screenshot 2025-08-15 at 15 21 39

Screenshot 2025-08-15 at 15 21 44

### Web

https://github.com/user-attachments/assets/5435d160-57b9-4068-b005-9e13e19061c5

### iOS

https://github.com/user-attachments/assets/aa663ec0-af0c-4862-91c7-f65449424f7c

### Android

On Android, the safe screen area is treated in two different ways depending on OS version. For devices running Android 14 (API 34) and earlier, the app always displays within the safe area. This sometimes means that visible bars are displayed above and below our app content, but the behaviour is consistent. On devices running Android 15 (API 35) and later, the app is displayed "edge-to-edge" by default. This has some visual inconsistencies that should be addressed (open issue #3094). However, these are independent of the current PR and associated issue, which should be addressed directly and will apply in both contexts.

#### Android 14

https://github.com/user-attachments/assets/8f50e114-9af4-415b-afcb-af33d82891da

#### Android 15 (edge-to-edge)

https://github.com/user-attachments/assets/8d99e630-7517-40da-ae77-aea9aac22b93

### DIFF
--- a/src/app/shared/components/template/components/layout/display-group/sticky/display-group-sticky.component.scss
+++ b/src/app/shared/components/template/components/layout/display-group/sticky/display-group-sticky.component.scss
@@ -9,7 +9,7 @@
 
 // Top sticky positioning
 :host[sticky-position="top"] {
-  top: var(--pop-up-top-padding, 0);
+  top: var(--ion-safe-area-top, 0);
 }
 
 :host[sticky-position="top"][ignore-safe-area="true"] {
@@ -18,7 +18,7 @@
 
 // Bottom sticky positioning
 :host[sticky-position="bottom"] {
-  bottom: var(--pop-up-bottom-padding, 0);
+  bottom: var(--ion-safe-area-bottom, 0);
 }
 
 :host[sticky-position="bottom"][ignore-safe-area="true"] {


chrismclarke (changes requested):
Thanks for all the details in this PR, very easy to follow and track examples and cases for what is a pretty tricky combination of factors.

My gut instinct is that we probably don't want to pass the onus of managing safe areas to authors at all - it's quite a niche area of development and hard to evaluate without careful examination of multiple specific android devices (beyond what I would expect a typical author to have access to). Similarly, I think the behaviour could be subject to change as we continue to work on handling edge-to-edge displays.

So I think what I would probably propose would be always using full edge-to-edge display when launching the sticky popup, assuming that typically the sticky components are used to display top/bottom images which are fine to overlap the safe area. If we discover specific cases where this is no longer desirable we could suggest authoring workarounds like positive margins/padding, maybe even exposing variables to apply safe-area padding

Let me know what you think of the suggestion and if there's any potential complications that I haven't thought of. Otherwise I'd hope it would be a fairly simple refactor to remove the safe-area variables and just change default padding to 0

jfmcquade:
> My gut instinct is that we probably don't want to pass the onus of managing safe areas to authors at all - it's quite a niche area of development and hard to evaluate without careful examination of multiple specific android devices (beyond what I would expect a typical author to have access to). Similarly, I think the behaviour could be subject to change as we continue to work on handling edge-to-edge displays.
> 
> So I think what I would probably propose would be always using full edge-to-edge display when launching the sticky popup, assuming that typically the sticky components are used to display top/bottom images which are fine to overlap the safe area. If we discover specific cases where this is no longer desirable we could suggest authoring workarounds like positive margins/padding, maybe even exposing variables to apply safe-area padding
> 
> Let me know what you think of the suggestion and if there's any potential complications that I haven't thought of. Otherwise I'd hope it would be a fairly simple refactor to remove the safe-area variables and just change default padding to 0

I definitely see where you're coming from. Although I am not sure that we can sensibly avoid exposing both options to authoring (respect or ignore the device safe area).

I do think that on reflection it doesn't make a lot of sense for the sticky display groups to sit inside the default padding of the fullscreen pop-up (e.g. on web, where the fullscreen pop-up has non-zero vertical padding even without a "safe area"). I think that even if the sticky display group is used to author interactive elements like buttons, then setting it to sticky should pin it to the very top/bottom of the screen. So I've added https://github.com/IDEMSInternational/open-app-builder/pull/3093/commits/5b18f6ec7ed659b3ac3bae97603a677e5e0317fc which makes the sticky display groups stick to the safe area edge as opposed to the inner padding of the fullscreen pop-up. NB that the `--pop-up-*-padding` vars are defined as:
```scss
--pop-up-bottom-padding: max(var(--ion-safe-area-bottom), 24px);
```

However, I would suggest that it's still desirable to expose the `ignore_safe_area` param to authors. It's not the case that sticky display components are only used to display images that should overlap the safe area, another use case in production (potentially more common although I'm not sure) is to display a title at the top or navigation buttons at the bottom. In both of these cases, it is important that the content not overlap the safe area. I agree that managing safe areas isn't something that authors should generally need to worry about, but I think it's somewhat unavoidable given these two real use cases. We can't rely on the author adding their own margins (or negative margins) because the safe area, characteristically, varies by device.

I did consider wrapping the same functionality in a param with a potentially more friendly name, e.g. `sticky_image_only: true` or `background_only: true`, which would imply some special handling that can remain unknown to authors. But I think that as long as the name of the param and the functionality are not aligned, it's potentially quite confusing for both authors and developers. We could also consider other descriptive names that might be more understandable to authors, e.g. `edge_to_edge` or `overlap_device_ui` or something else along those lines. Perhaps `overlap_device_ui` does the job of describing the effect without assuming any additional knowledge?

PR #3106, opened 2025-08-20
fix: shared data caching fallback

### LINKED ISSUE #3107: [BUG] PLH parent group PULL sometimes hangs
**Describe the bug**
Sometimes, but not always, an attempt to perform the `plh_parent_group: pull` action will hang and not complete. So far I have only recreated this on web, I am not sure if the issue is present on other platforms.

**App version**
2.2.4

**Screenshots and videos**
Demo on `plh_facilitator_my` deployment

https://github.com/user-attachments/assets/556da301-2f22-4057-94df-cce20930e5db

jfmcquade:
PR Checklist

## Description

Addresses an issue with shared data caching. This is a follow-up to #3068, which has explanatory details in the discussion on the PR it builds upon, #3064. Those changes added a system of ignoring cached firestore data to bypass an issue where stale cached data would be emitted first and taken, by not emitting any cached data and waiting for live emits from the server.

However, an additional issue arises: at least on web platforms, and at least some of the time, the firestore plugin will only emit cached data. This means that, for example, an attempt to pull data can be left stalling or incomplete. This PR works around that by providing a fallback: after a given amount of time (currently 2 seconds) the cached data will be emitted if no data from the server has been received in that window. This is a somewhat heavy-handed solution, but following on from the discussion on #3064, it's not clear that we have access to proper configuration of the cached vs server data from firestore.

## Git Issues

Closes #3107 

## Screenshots/Videos

`plh_facilitator_my` user flow, showing successful pull in contrast to behaviour documented in #3107. There is a delay in navigation caused by the explicit 2 second timeout.

https://github.com/user-attachments/assets/610701a9-7de7-4508-b04e-01b18edd3d68

### DIFF
--- a/src/app/feature/shared-data/providers/firebase.ts
+++ b/src/app/feature/shared-data/providers/firebase.ts
@@ -146,8 +146,14 @@ export class FirebaseDataProvider implements SharedDataProviderBase {
   private collectionToObservable(
     options: AddCollectionSnapshotListenerOptions
   ): Observable<ISharedDataCollection[]> {
+    // Timeout for server data to be received before falling back to cache
+    const TIMEOUT = 2000;
+
     return new Observable<ISharedDataCollection[]>((observer) => {
       let callbackId: CallbackId;
+      let hasEmittedServerData = false;
+      let cachedData: ISharedDataCollection[] | null = null;
+      let cacheTimeout: any;
 
       const registerListener = async () => {
         try {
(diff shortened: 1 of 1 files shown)


src/app/feature/shared-data/providers/firebase.ts
```diff
@@ -146,8 +146,14 @@ export class FirebaseDataProvider implements SharedDataProviderBase {
   private collectionToObservable(
     options: AddCollectionSnapshotListenerOptions
   ): Observable<ISharedDataCollection[]> {
+    // Timeout for server data to be received before falling back to cache
+    const TIMEOUT = 2000;
```
chrismclarke, line 150: Blocking
Having an arbitrary timeout to receive server data is a really bad pattern - large data payloads, flaky connection, slow servers, distanct server locations etc. can all play a factor into why a credible request may still take a considerable amount of time. This may limit the ability to ever receive live data from server, breaking functionality

chrismclarke:
I've opened #3108 to see if there might be a simpler solution working with individual queries instead of subscriptions. I don't think the solution here is headed in the right direction, so I'm going to close - although if we do want to keep similar snippets I'd suggest opening a follow-up PR where the workarounds are applied at the plh-parent-group service level instead

PR #3111, opened 2025-08-22
Feat: expose pop-up background colour to theming

jfmcquade:
PR Checklist

- [x] PR title descriptive (can be used in release notes)

## Description

Exposes a theme variable, `--pop-up-background`, for customising the background colour of pop-up modals (including fullscreen). By default, this will take the value of the theme's `page-background` colour. Previously, this was hardcoded as `white`.

As it happens, the only theme that doesn't use a white page background is the `default` theme. In order to preserve current behaviour, where the `default` theme styles page backgrounds as a pale variant of the primary theme colour, and pop-up backgrounds as white, the `default` theme has been updated to set `--pop-up-background` explicitly to white.

Also resolves an issue where any styling that made use of the page background colour would not work as expected within pop-ups, since it the global page background colour may differ from the pop-up background colour. See screenshots below.

## Git Issues

Closes #

## Screenshots/Videos

[debug_pop_up_bg](https://docs.google.com/spreadsheets/d/1ESiY3It66heuxfPl9PMZuYOKM6jjjpk9qIWH48Q_m-s/edit?gid=2079927361#gid=2079927361). Note the subtle box around the title elements previously, which are due to the display group with `style: background_solid` making use of the page background colour in its styling.

Screenshot 2025-08-22 at 13 33 06

| Pop-up type | Before | After |
|---|---|---|
| Not fullscreen | Screenshot 2025-08-22 at 13 20 22 | Screenshot 2025-08-22 at 13 18 48 |
| Fullscreen | Screenshot 2025-08-22 at 13 20 30 | Screenshot 2025-08-22 at 13 18 57 |

PR #3112, opened 2025-08-22
Feat: sticky display group safe area alt

### LINKED ISSUE #3042: Footer of full screen pop up should touch bottom of screen

jfmcquade:
PR Checklist

- [x] Builds on #3109, which should be merged first
- [x] Builds on #3111, which should be merged first
- [x] Builds on #3118, which should be merged first

## Description

Handle interaction between sticky display groups and device safe area, including case of fullscreen pop-ups. There are lots of combinations of configuration and device. I've added [feat_display_group_sticky](https://docs.google.com/spreadsheets/d/1tY8sa8ocDl552gIUAP263zgl3JQVcIeS4foaKk7nlMk/edit?gid=909315452#gid=909315452) which should allow us to test all these combinations.

## Testing

Appetize links:
- [iOS](https://appetize.io/apps/ios/international.idems.debug-app/builds/b_tet7cbvfwaysrukc63u5ofgqvq)
- [Android](https://appetize.io/apps/android/international.idems.debug_app/builds/b_xi6tie4jnrgqts6cxg6pimoxji)

## Notes

Currently, we use the behaviour of the ionic header and footer to handle the device safe area. They include the necessary additional padding when the device has a notch, for example, and we render a header with no additional height even in the case of the header being hidden at the app config level, in order to handle these cases.

Ultimately, our sticky display group top/bottom should include equivalent functionality. However, we don't always want to include this functionality. There are cases where the top sticky display group, for example, does not appear at the very top of the device, for example if it appears underneath the main app header, or within a nav-stack. In such cases, we don't want to add additional padding to the component.

I realised late in this process that I was working under the false assumption that fullscreen popups are the only case where the sticky display groups are at the very top or bottom of the screen. Therefore this PR handles pop-ups as a special case and only applies the relevant safe area padding in the case that the sticky display group is inside a pop-up. In actual fact, the sticky display group can also be at the top/bottom of the screen if the app (or template) is configured to have no header/footer. This case is not currently accounted for, however the changes included in this PR represent progress that can be built upon to handle that case.

See last set of screenshots below for an example of this case. I removed the header and footer by temporarily adding the following to the deployment config:
```ts
config.app_config.APP_HEADER_DEFAULTS.hidden = true;

config.app_config.APP_FOOTER_DEFAULTS.template = ""
```

## Git Issues

Closes #3093
Closes #3042

## Screenshots/Videos

[feat_display_group_sticky](https://docs.google.com/spreadsheets/d/1tY8sa8ocDl552gIUAP263zgl3JQVcIeS4foaKk7nlMk/edit?gid=909315452#gid=909315452) demos. There are more combinations of sticky display group type, but the image and content one is used here as it has the most complex behaviour.

Note that as of #3118, iOS is the only platform that includes true edge-to-edge support, although we would like to expand this support to Android 15+ in future.

| Platform | Within app content | in fullscreen pop-up |
|---|---|---|
| Web | web | web popup |
| Android 14 | android 14 | android 14 popup |
| Android 15 | android 15 | android 15 popup |
| iOS | ios | ios popup |
| iOS no header/footer | Screenshot 2025-08-29 at 17 40 35 | Screenshot 2025-08-29 at 17 40 46 |

### DIFF
--- a/.github/workflows/reusable-app-build.yml
+++ b/.github/workflows/reusable-app-build.yml
@@ -1,8 +1,8 @@
 ##################################################################################
 #         About
 ##################################################################################
-# Reuseable workflow to be called from content repos. 
-# Allows for parent repo. 
+# Reuseable workflow to be called from content repos.
+# Allows for parent repo.
 # Must specify all below secrets and variables - see documentation for details
 #
 #         Version : 1.0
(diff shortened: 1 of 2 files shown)


esmeetewinkel (approved):
Functional test passed on web and Appetize for iOS and Android. 

> In actual fact, the sticky display group can also be at the top/bottom of the screen if the app (or template) is configured to have no header/footer. This case is not currently accounted for, however the changes included in this PR represent progress that can be built upon to handle that case.

> See last set of screenshots below for an example of this case. I removed the header and footer by temporarily adding the following to the deployment config:

I tested this locally, but couldn't easily remove the header and footer on Appetize/Native so didn't test this case.

On iOS, I observed that the splash screen full screen pop up close icon is only half visible: 
image
This was not the case when accessing another full screen pop up on the same device, so maybe it's something specific to splash screens (and out of scope for this PR).

chrismclarke (approved):
Thanks @jfmcquade - the demos are looking waaay better, and great that the overall styling is pretty minimal (much better for long term maintainability)

I really like the edge-to-edge on ios so think definitely worth enabling on Android when we can, but otherwise all looking good to me

jfmcquade:
@esmee:
> On iOS, I observed that the splash screen full screen pop up close icon is only half visible

Thanks, I've pulled this out into an issue: #3133 



@chrismclarke 
> I really like the edge-to-edge on ios so think definitely worth enabling on Android when we can, but otherwise all looking good to me

Yeh, I agree that the edge-to-edge looks really nice and modern. Although interestingly when I look at Android apps I have installed, there are very few that make full use of the edge-to-edge feature – the majority have some kind of header, and of those that don't they mostly block out the unsafe areas as we're doing (although they colour-match the bars tot he background colour of the app, which we could attempt too).

In any case I think the work of this PR should pin down a few of the variables so that we can focus on the remaining factors to handle, but it may be that we want a bit of an overhaul of our fullscreen handling when working on this again, taking into account the points made in the PR description. It also may make sense to wait until the next Capacitor release to see if they include any more tools to handle it more easily. I've made a placeholder issue here: #3134.

PR #3115, opened 2025-08-25
fix: radio-button-grid text wrapping

### LINKED ISSUE #3114: [BUG] Radio button grid flex variant does not handle strings containing long words
**Describe the bug**
See https://github.com/ParentingForLifelongHealth/plh-kids-teens-app-za-content/issues/176.

The component should be updated to ensure that strings which contain individual long words do not break the styling.

**App version**

**To Reproduce**

**Deployment Sheet(s)**

**Debug Sheet(s)**

**Screenshots and videos**

**Additional context**

chrismclarke:
PR Checklist

## Description
Fix text wrapping in radio-button-grid component

## Dev Notes
Not entirely sure why width isn't working with default, I'm guessing something to do with the nested `<p>` inside the `<div>` - could probably just be a single span although I don't want to make any breaking changes to the component so hopefully for now setting the width should fix

## Git Issues

Closes #3114
Closes https://github.com/ParentingForLifelongHealth/plh-kids-teens-app-za-content/issues/176

## Screenshots/Videos
`plh_kids_teens_za` - `/template/module_oneononetime` (a few pages in)

localhost_4200_template_module_oneononetime(Moto G4)

### DIFF
--- a/src/app/shared/components/template/components/radio-button-grid/radio-button-grid.component.scss
+++ b/src/app/shared/components/template/components/radio-button-grid/radio-button-grid.component.scss
@@ -39,6 +39,7 @@ $text-color: var(--radio-button-font-color, var(--ion-color-primary));
     .item-text {
       font-size: $text-size;
       color: $text-color;
+      width: 100%;
     }
   }
 


jfmcquade (approved):
Thanks, this looks good. I think ideally we'd want the flex properties to handle wrapping the buttons onto multiple lines before they get to this height/width ratio. But this is definitely an improvement as it fixes the bug, so I think good to merge.

PR #3116, opened 2025-08-25
Feat: plh bottom nav text hide

### LINKED ISSUE #3113: [FEATURE] Allow plh-bottom-navigation-bar component to only display the text under the active link
**What?**
Update the `plh-bottom-navigation-bar` to expose an option to only show text under the button corresponding to the currently active link.

**Why?**
See bullet point on this content issue: https://github.com/ParentingForLifelongHealth/plh-kids-teens-app-za-content/issues/183

**How?**
This will likely be a new variant with a descriptive name, e.g. `variant: show_text_only_for_active_link`. NB: The component in use is the one at `packages/components/plh/plh-kids-kw/components/bottom-navigation-bar`, NOT `src/app/shared/components/template/components/navigation-bar`.

chrismclarke:
PR Checklist

## Description
- Add a `hide_inactive_text` authoring parameter to hide text from all non-active buttons in nav bar
- Improve text wrapping to avoid resizing buttons depending on text and instead just use ellipsis to indicate more text

## Review Notes
I debugged directly on local `plh_kids_teens_za` instead of any debug sheets, manually changing the parameter default in code to test. Should be tested on authored version

## Dev Notes
I noticed that the default component actually had no styling, but relied on overrides from themes. I migrated some of the more basic styling to the component itself (flex layout), and removed some of the duplicate theme styling. There are still a number of duplicate styles, but I think they are slightly more opinionated so possibly best left as overrides

## Git Issues

Closes #3113

## Screenshots/Videos
**Example** - with `hide_inactive_text` parameter, and artificially long link label
localhost_4200_template_activity_library(Moto G4) (3)


jfmcquade (approved):
Thanks, @chrismclarke. I've added [these commits](https://github.com/IDEMSInternational/open-app-builder/pull/3116/files/79788a31ad36539e0882973a3dd42c6534adb3d5..6452f41331edd143b3983c410efa21f11b9a05da) which include these changes:
- Makes the default value of `hide_inactive_text` `false`
  - This preserves behaviour for deployments already using this component
- Adds new param, `hide_wrapping_text`, to handle the newly introduced behaviour of truncating the text with an ellipsis
  - I think content authors won't generally want the labels to be truncated, and so by default the behaviour is now to wrap over multiple lines, see demo below
  - This isn't ideal as it causes the nav-bar height to increase to accomodate the tallest label, but I think it makes sense to leave it up to authors to balance up the options:
    1. Restrict text strings to a maximum length (best user experience)
    2. Truncate strings with an ellipsis via setting `hide_wrapping_text` to `true`
    3. Allow long strings to wrap over multiple lines, potentially increasing the footer height

https://github.com/user-attachments/assets/de8c9c73-c5aa-4f14-8395-2f6f96fd8920

Another option, which I chose not to expose, would be to allow the footer height to change dynamically based on the selected option. This seems less desirable than having a consistently taller footer, but could be re-exposed to authoring if desired:

https://github.com/user-attachments/assets/03367bd9-0f8d-46a1-bd3d-c064b759edee

@esmeetewinkel, I'll request a functional review from you, taking into account these comments

jfmcquade:
> @jfmcquade - looking at the example @esmeetewinkel shared - do you think we could keep just make the footer horizontal scrollable if required?
> 
> I'm somewhat opposed to options that either make one button much wider than others or dynamically changes height of just one button in the row - to me both look fairly jarring. I think most apps don't have this problem as they are much stricter on limiting the amount of text put into a footer button (guessing we should probably try to do the same), but if we really want to allow for long text then maybe auto overflow scroll on horizontal might be a reasonable solution.
> 
> Or we allow 2 lines of text for all and truncate at 2 lines (although that typically needs a bit of css/js workaround to truncate at 2 lines instead of 1).

I tried the functionality implied by the original issue, and I agree that it's quite jarring (along with the other potential solutions offered), see demo:


https://github.com/user-attachments/assets/3b6f6295-2a24-4f07-a92d-f8a34663c58c


So in the end I've reverted the behaviour I added so that the ellipsis truncation is still there. I think this should be seen as a catch for potentially buggy behaviour: it's not desirable that the label text gets truncated, but it catches the case of the provided text string being too long and encourages the content to be re-thought. See the results of [these commits](https://github.com/IDEMSInternational/open-app-builder/pull/3116/files/0d268a9d6dd74d7dfe86e3b8600c3b206de08ae3..c3ef240db0970551634f720e6ae5af4d18aee5f1) for my updates, I've also moved some additional styling that was previously in the theme overrides to the component level.

For a more thorough solution to the core problem, @FaithDaka has suggested the following designs, based upon @esmeetewinkel's example, which I think should work for handling the case of longer text strings. I plan to implement this design as a variant of the component in a follow-up PR to this one.

nav bar redesign

chrismclarke:
> I tried the functionality implied by the original issue, and I agree that it's quite jarring, see demo:

> I also agree that my other suggestions result in unsatisfactory behaviour. And I don't think I like the idea of a horizontally scrolling nav-bar, although that would still be an option.

> So in the end I've reverted the behaviour I added so that the ellipsis truncation is still there. I think this should be seen as a catch for potentially buggy behaviour: it's not desirable that the label text gets truncated, but it catches the case of the provided text string being too long and encourages the content to be re-thought. See the results of [these commits](https://github.com/IDEMSInternational/open-app-builder/pull/3116/files/0d268a9d6dd74d7dfe86e3b8600c3b206de08ae3..c3ef240db0970551634f720e6ae5af4d18aee5f1) for my updates, I've also moved some additional styling that was previously in the theme overrides to the component level.

For this basic navbar type I really think simply truncating text on smaller screens with longer translations is the best compromise - typically the page displays the name at the top anyway so users should be able to interpret the abbreviation in a reasonable way, and having the text flow onto multiple lines doesn't solve the issue where a single word is long (e.g. `Settings` -> `Izicwangciso`)

localhost_4200_template_home_screen(Moto G4)


> For a more thorough solution to the core problem, @FaithDaka has suggested the following designs, based upon @esmeetewinkel's example, which I think should work for handling the case of longer text strings. I plan to implement this design as a variant of the component in a follow-up PR to this one (UPDATE: WIP draft PR here: #3130)
> 
> nav bar redesign

I like this design as an alternate variant, and I think it would be a good way to address longer nav items. I'd likely vote in favour of merging this PR for default equal spacing with ellipsis for standard length text, and addressing longer text use case in draft pr referenced if that seems reasonable to you @esmeetewinkel ?

esmeetewinkel (approved):
> I like this design as an alternate variant, and I think it would be a good way to address longer nav items. I'd likely vote in favour of merging this PR for default equal spacing with ellipsis for standard length text, and addressing longer text use case in draft pr referenced if that seems reasonable to you @esmeetewinkel ?

Sounds good to me, thanks!

PR #2928, opened 2025-05-06
[FEATURE] Audio Player Styling - TZ theme

FaithDaka:
PR Checklist

- [x] PR title descriptive (can be used in release notes)

## Description

- TZ theme styling refactor for the audio component, to match KW theme style

## Git Issues

Closes #2861 

## Screenshots/Videos
Default:
 || 

Refactor:
 ||

PR #3125, opened 2025-08-29
Fix: audio component styling

### LINKED ISSUE #3124: [BUG] Audio player margin too large when title is missing
**Describe the bug**
When no title is specified, the large top margin of the audio player should disappear. 

**App version**
v0.22.0

**Debug Sheet(s)**
comp_audio

**Screenshots and videos**

Image Image

jfmcquade:
PR Checklist

## Description

Fixes some issues with spacing around the audio component that led to inconsistencies.

- `default` theme:
  - removes excess whitespace at top of audio component (https://github.com/IDEMSInternational/open-app-builder/pull/3125/commits/71d8a1be12b54002fdb570391f63d2590c61e7ff)
- `plh_kids_kw` theme and its descendants:
  - fixes issues with display of audio component title
  - removes excessive vertical spacing around audio component
  - refactors styling overrides
  - applies consistency in colours across variants and themes, see screenshots below

## Dev notes

The KW-based themes override the core structure of the component, moving the title from inside the demarcated area to outside. Previously, this was achieved with a workaround that absolutely positioned the title, which was brittle. This is refactored so that the same result is achieved in the theme overrides by moving the visible bounding box from the top-level container to the functional audio player element of the component, excluding the title. @FaithDaka if/when we generally refactor the styling of components to use the KW-based overrides by default, then the HTML and SCSS of the base component can be tidied up.

## Git Issues

Closes #3124

## Screenshots/Videos

[comp_audio](https://docs.google.com/spreadsheets/d/1utD4Kr0WVjvDwnMbg2wYupaPQ0VY0H2csb86fyi4bPI/edit?gid=603128459#gid=603128459) template. Note the changes to the background colour of the `large` variant for some themes – I don't believe this variant is actually used in production currently.

Also note the more significant colour changes for the `plh_kids_tz` deployment – this brings it in line with the other themes, but the colours could be overridden to preserve the current look if desirable (which was updated in #2928). Indeed there's a general question for @FaithDaka of whether the colour of the controls of the audio player should use the new secondary colour post theme generator refactor, which is taken to be a contrasting colour, or use the primary-variant colour, which is taken to be a similar variant of the primary colour.

| theme | before | after |
|---|---|---|
| default | default before | default after |
| plh_kids_kw | kw before | kw after |
| plh_facilitator_mx | mx before | mx after |
| plh_facilitator_my | my before | my after |
| plh_facilitator_ph | ph before | ph after |
| plh_kids_tz | tz before | tz after |

### DIFF
--- a/src/app/shared/components/template/components/audio/audio.component.html
+++ b/src/app/shared/components/template/components/audio/audio.component.html
@@ -1,4 +1,4 @@
-<div class="container-player margin-t-large" [attr.data-variant]="params.variant">
+<div class="container-player" [attr.data-variant]="params.variant">
   @if (params.title) {
     <div class="top-row">
       <h3>{{ params.title }}</h3>
--- a/src/theme/themes/plh_facilitator_mx/_overrides.scss
+++ b/src/theme/themes/plh_facilitator_mx/_overrides.scss
@@ -347,35 +347,54 @@
 
   // audio
   plh-audio {
-    border: 1px solid transparent;
-    .container-player[data-variant~="compact"] {
-      width: var(--buttons-full-width);
-
-      border: 0.9px solid var(--ion-color-primary-100);
-      border-radius: var(--ion-border-radius-rounded);
+    .container-player[data-variant~="compact"],
+    .container-player[data-variant~="large"] {
+      // Remove visible bounding box from container level and apply to second-row instead
+      border: unset;
+      border-radius: unset;
       box-shadow: none;
-
-      background-color: var(--ion-color-primary-50);
-      padding: 8px 16px;
-      padding-right: 24px;
-      margin-top: 30px;
+      background-color: unset;
+      padding: 0;
 
       .top-row {
-        min-height: 0 !important;
-        position: absolute;
-        top: -52px;
-        margin: 0 0 0 -14px;
-        // padding-top: 30px;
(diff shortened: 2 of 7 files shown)


esmeetewinkel (approved):
This looks great, thanks @jfmcquade for all the screenshots. Functional test passed.

I'm merging this since it's needed for an imminent release, hopefully code review can happen as a follow up.

chrismclarke:
@jfmcquade - thanks for updating this. I think it makes sense to disable for now.
Given the improvements you've made to layouts in #3112 I think we may want to revisit edge-to-edge display on Android, but possibly wait until Capacitor has support for using different colors for status/navbar backgrounds (default behaviour would want to match header background when header visible, but likely change to transparent when hiding, and similarly only apply a footer background if using a footer)

There's an open PR to apply more configuration via
```
https://github.com/capawesome-team/capacitor-plugins/pull/601
```

For now might just be worth creating a small follow-up issue for future dev

chrismclarke:
Thanks @jfmcquade - nice to see a decent-sized trim of the css overrides. There still is a fair bit of duplication which hurts long term maintainability, as you've suggested I'd probably recommend updating the default component and styles to align more closely with the kw variant and just exposing a handful of variables as required, but appreciate this PR is still a good first step in the right direction.

PR #3127, opened 2025-08-29
Chore: refactor theme definition code alt

### LINKED ISSUE #3086: [CHORE] Refactor theme definition code
**What?**
Following #3084, some further refactoring could be done to remove duplicated code.

From https://github.com/IDEMSInternational/open-app-builder/pull/3084#pullrequestreview-3108526654:

> I'm guessing you could probably go even a bit further to also incorporate the duplicate code that I think is included in the bottom of all themes
> ```scss
> @include utils.generateTheme(
>     $p: $color-primary,
>     $s: $color-secondary,
>     $bg: $page-background,
>     $pv: $color-primary-variant,
>     $sv: $color-secondary-variant
>   );
>   @each $name, $value in $variable-overrides {
>     --#{$name}: #{$value};
>   }
> ```
> This would likely mean passing the list of variables and overrides to the theme-selector, and calling the other mixins from there (possibly renaming the current generateTheme to generateColorPalletes or similar, and calling the selector mixin generateTheme). But also happy to approve as-is if it helps fixes the regression introduced

jfmcquade:
PR Checklist

## Description

Builds on #3117. Goes further in refactoring theme generation code by abstracting more logic to general SCSS mixins/functions, allowing us to reduce duplicated code in theme definition files. This should make adding new themes slightly easier in future.

## Testing 

We should confirm that themes are generally still functioning as expected, including within fullscreen pop-ups (resolved by #3081 and #3084). From my testing, that does appear to be the case.

## Git Issues

Closes #3086 
Closes #3117

## Screenshots/Videos

### DIFF
--- a/.github/workflows/reusable-app-build.yml
+++ b/.github/workflows/reusable-app-build.yml
@@ -1,8 +1,8 @@
 ##################################################################################
 #         About
 ##################################################################################
-# Reuseable workflow to be called from content repos. 
-# Allows for parent repo. 
+# Reuseable workflow to be called from content repos.
+# Allows for parent repo.
 # Must specify all below secrets and variables - see documentation for details
 #
 #         Version : 1.0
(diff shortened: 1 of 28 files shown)


jfmcquade:
> Thanks for the further abstraction @jfmcquade
> 
> When I run the branch I got a `config` error which I thought was due to a naming collision with global Sass files. I changed the mixin name from `theme` to `theme-generator` and the error was resolved. I don't know if this error appeared on your end.
> 
> 

That's interesting, I can't replicate that error. It does sound like some kind of naming collision issue with SASS, but I can't figure out what could be causing it. I wouldn't be averse to renaming the mixin if we can't find another solution (maybe back to `generate-theme`), but would be better to figure out the root cause.

Could you try running
```sh
yarn cache clean
```
and see if that resolves the issue on your end?
If that doesn't work, it could also be worth trying:
```sh
rm -rf node_modules
yarn install
yarn build 
```
to make sure your SASS related packages are all installed correctly.

src/theme/themes/utils/generate-theme.scss
```diff
@@ -2,14 +2,68 @@
...
+  $theme-name,
+  $color-primary,
+  $color-secondary,
+  $page-background: null,
+  $color-primary-variant: null,
+  $color-secondary-variant: null,
+  $color-red: null,
+  $color-green: null,
+  $variable-overrides: ()
+) {
```
chrismclarke, line 19: nit(blocking) - this is a long list of positional args which will be difficult to manage if we ever want to make changes. I think it would be better to support a map which gets merged with defaults, e.g.
```
$theme-defaults: (
  theme-name: "default",
  color-primary: hsl(220, 90%, 50%),
  color-secondary: hsl(280, 80%, 50%),
  page-background: null,
  color-primary-variant: null,
  color-secondary-variant: null,
  color-red: null,
  color-green: null,
  variable-overrides: (),
);

// Mixin that accepts a config map
@mixin theme($config) {
  // Merge user config with defaults
  $cfg: map.merge($theme-defaults, $config);
```

Then if referencing particular colors from the merged config you can extract via something like `$theme-name: map.get($cfg, theme-name);` - although you may find for may of the child mixins you could still just pass the full map config.

It might also be a little cleaner to keep 2 args, the first being the `theme-name` and second color variables.

chrismclarke (changes requested):
Thanks @jfmcquade - I think this is definitely moving in the right direction (any reduction in total scss feels like a win!)

It does still feel like there's quite a lot of duplication and redundancy, e.g. multiple themes explicitly setting the `color-red` variable as `null` feels off, but the general way themes are generated feels nicer. See inline comment for one potential improvement I noticed.

Issue #3139, opened 2025-09-11
[FEATURE] Create `plh_kids_teens_za` theme


FaithDaka:
**What?**
Create a dedicated theme file for the `plh_kids_teens_za` theme following Kuwait overrides

**How?**
_Do you have an example template that demonstrates the desired behaviour or syntax? If so, specify its name and provide a link._

**Other details**
_Add any extra details that you think may be useful._

Issue #3099, opened 2025-08-20
[CHORE] Refactor `plh_kids_kw` theme to use new theme colour vars


jfmcquade:
**Details**
See parent issue for more details

PR #3142, opened 2025-09-12
style: plh_kids_teens_za theme updates

jfmcquade:
PR Checklist

- [x] Builds on #3139. After that has been merged, this PR should be updated to target master

## Description

Updates to the `plh_kids_teens_za` theme that were originally included as part of #3140. Having them in a separate PR here lets us test and iterate.

## Git Issues

Closes #

## Screenshots/Videos


jfmcquade:
@FaithDaka See this PR description – it includes the changes that were originally in #3140, which were addressing #3200. 

Please could you either update this PR or create a new one to address #3200 directly. Some points to note:
- #3099 has now been merged, which may have knock-ons for this PR (I believe the ZA theme is base don the KW theme directly)
- See @esmeetewinkel's comment above for a list of colour changes that probably aren't acceptable. We'll likely want to tweak this theme to override its `*-variant` colours, or else make a case for the changes from a design point of view

PR #3147, opened 2025-09-24
Feat: content json sync

chrismclarke:
PR Checklist

## TODO
- [ ] Fix Tests
- [ ] Check if local-first and watch workflows still work
- [ ] Check handling of intermediates if google drive remote files moved or deleted

## Description
Adds support for storing raw json representation of google sheet content, opening up pathway for future conversion back to xlsx.

Includes metadata with additional `lastModifiedBy` to track both when and who made last changes to a sheet

This does not remove the need for the full app_data content synced in git repo (follow-up work required in future), so for now the repos will have somewhat duplicate content (but designed with different purposes)

## Review Notes
Sync a repo
```bash
yarn workflow sync
```
View generated outputs in _intermediates_ content repo folder
```
 .idems_app\deployments\debug\intermediates
```

## Author Notes
**Overview**
The purpose of the intermediates output is to track a more basic representation of data in google sheets. In the future it could be used to recreate a google sheet from content repo at a specific time in history

All sheets have a single entry to describe the google sheet contents, e.g. _component_apple_sign_in_button.json_
```json
{
  "==content_list==": [
    {
      "flow_type": "template",
      "flow_name": "comp_apple_sign_in_button",
      "status": "released",
      "flow_subtype": "component_demo"
    }
  ],
  "comp_apple_sign_in_button": [
    {
      "type": "title",
      "value": "Apple Sign In Button"
    },
    {
      "type": "apple_sign_in_button",
      "value": "Continue with Apple",
      "action_list": "click | emit: force_reprocess;",
      "comments": "Button is styled in line with Apple's guidleines: https://developer.apple.com/documentation/sign_in_with_apple/displaying-sign-in-with-apple-buttons-on-the-web"
    },
    {
      "type": "apple_sign_in_button",
      "value": "Sign in",
      "action_list": "click | emit: force_reprocess;"
    },
    ...
  ]
}
```

There is also a top-level `_metadata.json` file which tracks (by google drive id) a snapshot of sheet data, including latest time modified and who by (if info available from google drive api)
```json
"1YcG0GDJME1Ds1UzwXcEmo97Ckb5tnUWc": [
    {
      "relativePath": "component_sheets/component_apple_sign_in_button.xlsx",
      "size_kb": 53.9,
      "modifiedTime": "2025-09-18T07:03:00.776Z",
      "modifiedBy": "faithux.co",
      "md5Checksum": "82e176f9bb44f491bf9d6943a667cd24",
      "remoteUrl": "https://docs.google.com/spreadsheets/d/1P6EhHmElOJOAoMH4h_vingVL82jyCfSTYPXj8eY2Iek"
    },
    ...
```
**Additional Notes**
For any repos where this is not required (or feels excessive), the entire _intermediates_ folder can just be added to `.gitignore` file

Given the data undergoes a couple transformation (sheet -> plain json -> merged with contents -> processed json) I wasn't entirely sure what the best point to generate this intermediate data would be. I opted for the plain json representation as that would be easiest to convert back into xlsx in the future, although that also means all sheets marked as draft will be included. Open to input about what would be the most useful from a content side. There is also separate consideration about what would be best from a dev side to separate out the various layers of processing.

## Dev Notes

Current the intermediates are mostly generated and output as an after-thought, and not integral to the processing itself (we keep all jsons in memory and further process to do things like filter out draft sheets, merge contents etc.) before fully parsing. A case could be made for using this latter json representation instead (harder to convert back to xlsx, but with slightly more useful info), as could further refactors to separate up these processes even more. However for now I'm hoping this achieves short term goal of pathway to xlsx <---> json conversions, and some additional useful info for authors.

If we wanted to continue to build on this and _only_ store the plain json representations, workflows, and gh actions would need a little bit more work to ensure content fully processed if syncing from a content repo as part of CI action (syncing from content repo is no longer sufficient to have runnable app content). In the long term I think this would be desirable, but beyond the scope of initial PR.

Additional work was needed on the gdrive downloader to track the modifiedBy user, and I also implemented better type-safety through zod. There were a few small knock-ons to the main parser as the gdrive downloader content is now slightly less tightly coupled with the app data (additional mapping functions added). 

I also refactored to remove some of the api metadata that we weren't actually using, and dropped usage of server `md5` checksum comparison as it's somewhat unreliable (google drive only populates md5 if the sheet was originally uploaded as an xlsx file, and not if was created or converted to gsheet format). So for now simply using the modified timestamp to compare local and server files.

## Future Tasks
- [ ] Consider allowing author to define config google drive folder ids with custom `label` to make outputs more human-readable. e.g.
```
config.google_drive = {
  sheets_folder_ids: [{id:"1YcG0GDJME1Ds1UzwXcEmo97Ckb5tnUWc",label:"debug_sheets"}]
};
```
This would allow for outputs grouped with correct google drive folder names, while still avoiding issue if syncing from multiple sources that have same name (e.g. multiple folders with a folder named `app_sheets` or similar)\

- [ ] Further refactor to remove need to store app_data sheets in content repos, and rely just on synced google sheet data (would require updates to workflows and github actions)

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/scripts/src/commands/app-data/convert/processors/xlsxWorkbook.ts
+++ b/packages/scripts/src/commands/app-data/convert/processors/xlsxWorkbook.ts
@@ -7,7 +7,7 @@ import BaseProcessor from "./base";
 import { existsSync } from "fs-extra";
 import { IContentsEntry, parseAppDataCollectionString } from "../utils";
 
-const cacheVersion = 20250923.0;
+const cacheVersion = 20251001.0;
 const sheetsFolderBaseUrl = "https://drive.google.com/drive/u/0/folders";
 
 export class XLSXWorkbookProcessor extends BaseProcessor<


jfmcquade (commented):
Thanks @chrismclarke, I think this is definitely moving things in the right direction.

I think that this "plain json" level is the right one to save. At least for now – we can decide if we'd want to support a system of excluding drafts or grouping by content sheet, but I think it makes sense to start with the most direct representation of the sheets as possible.

I think we do want to be working towards clearly separating the steps so that the first round of parsing spits out these JSONs that are then the input for the next round of parsing, as you've mentioned in the dev notes, but I'm happy for that to be handled as a follow-up. So this should be good to merge pending the immediate TODOs you've highlighted (and addressing the minor comments from the Gemini review).

@ChrisMarsh82 I'm tagging you for review to be aware of the changes described in the PR description

PR #3156, opened 2025-09-30
refactor: translated assets signal

### LINKED ISSUE #3155: [BUG] force reload sometimes has no effect
**Describe the bug**
The force reload action (`emit: force_reload`) sometimes does not cause the page to reload as expected.

**App version**
0.23.0

**To Reproduce**

**Deployment Sheet(s)**

**Debug Sheet(s)**
[debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit?gid=1422144003#gid=1422144003)

**Screenshots and videos**
See demo of [debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit?gid=1422144003#gid=1422144003). The asset with a language override does not change when the language is changed, nor when the force reload is called. It does update when the page is refreshed via the browser.

https://github.com/user-attachments/assets/620f70e6-003c-4504-b158-4d08eb301ad5

**Additional context**
This is a regression issue as the behaviour previously was to always reload the template from scratch. See, for example, the first video demo on #2408.

Using `git bisect`, I have identified that the problematic commit is [this one](https://github.com/IDEMSInternational/open-app-builder/pull/2427/commits/f357220ba119e1642bee6257ad8fb05f001634f6), from #2427. Prior to that commit, the assets on [debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit?gid=1422144003#gid=1422144003) would change when shanging the language via the buttons on that template. I suspect that the issue is to do with the removal of the `ChangeDetectorRef` and explicit `detectChanges()` call that was removed in that commit, although I haven't verified this.

chrismclarke:
PR Checklist

## Description

_What this PR does_

## Review Notes

```
changed | emit: set_language: @local.language_select;
```

## Git Issues

Closes #3155

## Screenshots/Videos

### DIFF
--- a/src/app/feature/theme/services/theme.service.ts
+++ b/src/app/feature/theme/services/theme.service.ts
@@ -4,12 +4,20 @@ import { LocalStorageService } from "src/app/shared/services/local-storage/local
 import { AppConfigService } from "src/app/shared/services/app-config/app-config.service";
 import { IAppConfig } from "data-models";
 import { SyncServiceBase } from "src/app/shared/services/syncService.base";
+import { toSignal } from "@angular/core/rxjs-interop";
 
 @Injectable({
   providedIn: "root",
 })
 export class ThemeService extends SyncServiceBase {
+  /**
+   * @deprecated 2025-09
+   * Prefer to use `currentTheme` signal instead
+   */
   currentTheme$ = new BehaviorSubject<string>(null);
+
+  public currentTheme = toSignal(this.currentTheme$);
+
   availableThemes: IAppConfig["APP_THEMES"]["available"];
   defaultThemeName: string;
 
--- a/src/app/shared/components/template/components/image.ts
+++ b/src/app/shared/components/template/components/image.ts
@@ -1,12 +1,22 @@
-import { Component, OnInit } from "@angular/core";
+import { Component, computed, inject } from "@angular/core";
 import { TemplateBaseComponent } from "./base";
-import { getStringParamFromTemplateRow } from "../../../utils";
+import { TemplateAssetService } from "../services/template-asset.service";
+
+interface IAuthorParams {
+  /** TEMPLATE_PARAMETER: "style". Default null */
+  style?: string;
+}
+
+interface IComponentParams {
+  style: "no-padding";
+}
 
 @Component({
   selector: "plh-tmpl-image",
   template: `
-    <div class="tmpl-image-container" [class]="style" [attr.data-param-style]="style">
-      <img *ngIf="_row.value" [src]="_row.value | plhAsset" />
+    <div class="tmpl-image-container" [attr.data-param-style]="params().style">
+      @if (assetSrc()) {}
+      <img [src]="assetSrc()" />
     </div>
   `,
   styleUrls: ["./tmpl-components-common.scss"],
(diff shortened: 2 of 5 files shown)


chrismclarke:
> Thanks @chrismclarke, I appreciate you getting to this promptly.
> 
> Generally the implementation looks good, your explanation of why the pipe approach wasn't working makes sense to me, so good to understand what the issue was, and nice to see the signal based implementation.
> 
> I don't love the new pattern of needing to wrap potentially every asset reference string in `this.assetService.translatedAssetSignal()` and expose as a new public property before use in the angular template, but I can't think of a better alternative. I suppose this can be incorporated into our parsing of parameters in the case that the value in question is a parameter. I'm just thinking of a dev coming across this down the line and finding it confusing – perhaps it could be resolved with a different name, e.g. `this.assetService.assetSignal()` or `this.assetService.assetSource()` or `this.assetService.assetSrc()` or `this.assetService.assetUrl()`? I might favour `assetSignal`, I think what we lose in descriptiveness we make up in general applicability. Let me know what you think, I'm happy to leave as it is if you disagree.

It does feel somewhat clunky for sure, particularly passing the a signal as an input (e.g. `row.value` instead of `row.value()`). 

I've taken another quick look and updated to code to also include a new `translatedAsset` pipe which can abstract the use of computed signals and essentially work as a drop-in replacement for the `plhAsset` pipe. The drawback to using a pipe instead of computed signal is that the pipe will re-evaluate on every change detection cycle (needs to be defined as non-pure to allow pipes to update from external service signals), however I've attempted to minimise the impact of this by caching created signals (so will just return the same signal subscription following change detection). I feel like this is a pretty reasonable balance, but have tagged you for re-review to confirm if happy. 

> I've also added an update that extends the signal based approach to cover `TemplateAssetService.assetsContentsList`, which was previously an rxjs observable. This ensures that the assets update as expected when a relevant asset pack is downloaded, which will update the asset paths without changing the app's theme or language (so previously the new computed signal would not evaluate). This means that the demo of the remote assets feature in `debug_remote_assets` should now work as expected without the need for a force refresh (although until #3154 is merged, only the first asset will update as the second one throws an error).
> 
> Finally, I've made an issue to track the follow-ups you've highlighted: #3158. I'm hoping I'll be able to get to to it and tidy up before too long, but need to focus on the core remote asset functionality next.

Sounds good, I'll review that PR following this one

jfmcquade:
> I've taken another quick look and updated to code to also include a new `translatedAsset` pipe which calls the service code directly to remove the need of component-level computed signals, essentially working as a drop-in replacement for the `plhAsset` pipe.
> 
> The drawback to using a pipe instead of computed signal is that the pipe will re-evaluate on every change detection cycle (pipe needs to be defined as non-pure to allow pipes to update from external service signals, and this falls back to somewhat more verbose angular change detection). I've attempted to minimise the impact of this by caching created signals so that even if triggering recalculation it can just return the already-generated signal (with added benefit of reusing the same signal if the same asset used in multiple places). I haven't included explicit methods for removing signals from cache when no longer required (garbage collection), however I don't think it will grow very large (maybe 1MB memory per 1,000 unique asset paths to store the original path, signal code and current value - super rough calculations).
> 
> I feel like this is a pretty reasonable balance overall, but have tagged you for re-review to confirm if happy. Code updates in [2c0720d](https://github.com/IDEMSInternational/open-app-builder/pull/3156/commits/2c0720dd8ede5840336a17cb7f848c87ec4b15d9)

Nice, this does feel like a nicer pattern, even with the caveats you've mentioned. However I think that the cache might be implemented in a way that means it only exists for each separate instance of the pipe? So the benefit of reusing the same signal for the same asset globally is lost. I've added https://github.com/IDEMSInternational/open-app-builder/pull/3156/commits/1dc4deabffc96c55be628f3e42d8bf6f74079367, which moves the cache from the pipe to the service level to ensure it's shared globally. If you agree with those changes then I think this PR should be good to merge.

PR #3122, opened 2025-08-28
Style: reduce text component margins (default theme)

jfmcquade:
PR Checklist

- [x] PR title descriptive (can be used in release notes)
- [x] Builds on #3121, which should be merged first

## Description

Currently the text component has quite large vertical margins. This can make spacing look inconsistent when a list of components includes text instances, and effectively reduces the amount of usable space on the screen.

This PR reduces the default margin from `15px` to `5px` (reading from the existing `--tiny-margin` var). This will have an effect on the `default` theme and its descendants (`professional` etc.), but not on the `plh_kids_kw` theme and its descendants, which already override this value.

A new theme-level variable, `--paragraph-spacing`, is exposed to override this default value.

## Dev notes

FAO @FaithDaka: the new theme variable `--paragraph-spacing` can now be used to override the default spacing, and should be implemented instead of the custom `--paragraph-margin-small` introduced for the `plh_kids_kw` deployment when refactoring that theme.

## Git Issues

Closes #

## Screenshots/Videos

| template | Before | After |
|---|---|---|
| [comp_text](https://docs.google.com/spreadsheets/d/1iZHeAAZAQ8q-XNKrW_ww3yvPUVGwc407uhAF7vu1TtM/edit?gid=569531329#gid=569531329) | Screenshot 2025-08-28 at 15 17 13 | Screenshot 2025-08-28 at 15 09 36 |
| [debug_form_layout](https://docs.google.com/spreadsheets/d/1-SX-9_ns4DJXHZL7fe653cchnaj8cXKQoBL5gVF-J2w/edit?gid=99275306#gid=99275306) | Screenshot 2025-08-28 at 15 22 39 | Screenshot 2025-08-28 at 15 09 57 |

PR #3175, opened 2025-10-23
style: text component margins alt

chrismclarke:
PR Checklist

## Description

- Replaces #3122 
- Fixes broken merge
- Restores default 0.75em (12px) paragraph margin instead of pr 5px

## Author Notes
As mentioned in previous PR, new style variable `paragraph-spacing` can be defined at the theme level to control default text margins. This PR makes the fallback value 12px (for cases where the variable is not explicitly set) - previously pr had set as 5px which I suggested was too narrow

## Git Issues

Closes #3122

## Screenshots/Videos
Comparing 5px fallback vs 12px fallback spacing

| Previous PR (5px default) | Current PR (12px default) |
| --- | --- |
| image | image | 

Alternatively, theme could choose to use a value like 16px to space out further (although would likely need additional line-height variable exposed in similar way to improve spacing between)

| 16px spacing (no line height change) | 16px spacing (custom line height 1.5) |
| --- | --- |
| image | image |


jfmcquade (approved):
Thanks, looks good, I approve of the changes to the default spacing for the reasons given.

jfmcquade:
Thanks @chrismclarke, I've broken out into #3179

PR #3176, opened 2025-10-23
fix: translation bug

jfmcquade:
PR Checklist

## Description

Fixes an issue where a string containing a dynamic field would not be translated as expected.

The exact conditions are not clear, but the bug was observed on the `plh_kids_teens_za` deployment, specifically on the [intro_loop](https://docs.google.com/spreadsheets/d/1m0TIn3wJO2Y3W6dwWwh695e3q3lWTJe0y-PBHcWBpzc/edit?gid=2022338829#gid=2022338829) template and the string in E235 of the [intros](https://docs.google.com/spreadsheets/d/1pl75hk-zfJVDY22auU2Kf9l0dPe8KsGnXCSSQCU1LQU/edit?gid=185224490#gid=185224490) data list.

## Dev notes

We haven't been able to fully isolate the conditions that create the bug, but there was some strange behaviour that differed depending on levels of nesting of template. Additionally, both the translation logic and the expression evaluation logic are very hard to follow, and take place across multiple files, with various handlings of specific cases. This makes it hard to understand exactly when the translations and evaluations take place.

The included solution is therefore a belt-and-braces approach, that adds an additional check to ensure that any expression is translated as part of evaluation. Whilst this may result in unnecessary checks, there is already a lot of un-optimised translation and evaluation code, which would likely be better addressed with a rewrite of the translation and evaluation systems (as part of the new reactive templates renderer) rather than tracking another specific edge case to handle.

## Git Issues

Closes https://github.com/ParentingForLifelongHealth/plh-kids-teens-app-za-content/issues/249 (@esmeetewinkel to confirm)

## Screenshots/Videos

`article_praise_t_praise_t3` template from `plh_kids_teens_za`

| Before | After |
|---|---|
| Screenshot 2025-10-23 at 15 04 15 | Screenshot 2025-10-23 at 14 31 29 |


chrismclarke (changes requested):
Thanks @jfmcquade 
Looking at the code for the value translation it's mostly just a quick object lookup, so I see no harm in quickly checking on all strings like you've done, and assume should catch case where full expression needs translating, e.g. 
```
"Hi {@fields.userName}, how are you feeling today?"
```
Which I'm guessing may have previously depended on patterns like
```
"@global.greeting @fields.userName how are you feeling today"
```
and would have easily gotten mistranslated if word order changes in translation, or languages like spanish where you would prepend `¿` to the beginning

My only request would be to add at least 1 actual test to capture this edge-case/scenario, as I imagine it being something easily lost if migrating dynamic content runtime processing in the future.

Also as a sidenote would be good to remember for future general preferred processing order at runtime would be along the lines of

```mermaid
  graph TD;
      A["Base Text"]-->B["Runtime Translation"];
      B-->C["Runtime Dynamic Text Parse"];
```
And probably setting it up more like a pipeline instead of including so many inline methods would probably make for easier code maintenance in the future (but not a job for now)

src/app/shared/components/template/services/template-variables.service.ts
```diff
@@ -175,6 +175,17 @@ export class TemplateVariablesService extends AsyncServiceBase {
     context: FlowTypes.TemplateRowEvalContext
   ) {
     const fullExpression = evaluators[0].fullExpression;
+
+    // HACK: Ensure expression is translated (evaluators may be created from untranslated strings in some cases)
```
chrismclarke: Not sure this is actually a hack. Can probably drop from the comment prefix assuming we're able to add a test case that validates the use

chrismclarke (approved):
Thanks @jfmcquade - the added tests look good to me. I haven't tried running them against master to check how default behaviour would have been otherwise, but I think makes sense to have either way

esmeetewinkel:
@jfmcquade I'll merge here. I still have creating a proper debug example on my to-do list, but I think it shouldn't hold up this PR.

PR #3181, opened 2025-10-28
Feature: Add reactive query component

markharmon:
Replace by: #3215 

PR Checklist

## Description

Adds a 'query' component which allows execution of raw Mango queries.

image

image

[example sheet](https://docs.google.com/spreadsheets/d/1cDndPmELXtmx65GfoBKVnp-JxD21wxMMVWGeJUryuL0/edit?gid=1343090831#gid=1343090831)

### JSEvaluator chages

I noticed an issue with javascript evaluation which I have fixed as part of this PR, specifically: "options_value: name" parameter value was compiling as valid javascript unintentionally.

image

The fixes mean that the following single words will longer compile as valid javascript and will return their string values. I couldn't find a way to remove these from the evaluation context entirely

    "window",
    "name",
    "globalThis",
    "self",
    "document",
    "parent",
    "top",
    "frames",
    "global",

### json5

Introduces [json5](https://json5.org/) so we can parse 

```
selector: {active: true},
sort: [ {name: 'desc'} ],
limit: 3
```
instead of having to do this this: 
```
"selector": {"active": true},
"sort": [ {"name": "desc"} ],
"limit": 3
```

### Content PR

https://github.com/IDEMSInternational/app-test-content/pull/7

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/shared/src/models/jsEvaluator/jsEvaluator.ts
+++ b/packages/shared/src/models/jsEvaluator/jsEvaluator.ts
@@ -23,16 +23,19 @@
 export class JSEvaluator {
   private evaluationContextBase: string;
 
+  // Identifiers that should be restricted from access within evaluated expressions
+  // Some global identifiers may conflict with common expression / data column names e.g. name and parent
+  // The commented items are potential candidates for restriction but are left accessible for now
   private readonly restrictedIdentifiers = [
-    "window",
+    // "window",
     "name",
-    "globalThis",
-    "self",
-    "document",
+    // "globalThis",
+    // "self",
+    // "document",
     "parent",
-    "top",
-    "frames",
-    "global",
+    // "top",
+    // "frames",
+    // "global",
   ];
 
   constructor() {


packages/shared/src/models/jsEvaluator/jsEvaluator.ts
```diff
@@ -23,6 +23,18 @@
...
+    "window",
+    "name",
+    "globalThis",
+    "self",
+    "document",
+    "parent",
+    "top",
+    "frames",
+    "global",
+  ];
```
jfmcquade: nit(blocking): I think that we currently potentially make use of some of these (specifically "window" and "document"). So probably safest to just restrict "name" and "parent" for now (as these are likely to be the ones that could conflict with variable values)? Perhaps it would make sense to leave the other strings here in the list but comment them out, as it could be useful reference in future.

jfmcquade (changes requested):
Very nice how minimal these changes are. I think it's all making sense to me.

As discussed, we likely want to migrate away from relying on indexedDB to store the dynamic data ([Capacitor docs](https://capacitorjs.com/docs/guides/storage)), but that should be handled as a follow-up.

Also, as discussed, I think the approach of exposing the raw complex syntax (in this case mango queries), and then deciding later down the line whether or not to offer a simplified syntax, is a good idea going forward. 

See one blocking comment inline regarding the protected names in JS evaluation.

PR #3184, opened 2025-10-28
Feat: gdrive remote folder metadata

chrismclarke:
PR Checklist

## TODO
- [x] Builds on top of #3166, should merge after

## Description
- Add support for providing custom names alongside google drive folder ids
- Update flow metadata to include full path to file (including parent folder), and sheet url

## Review Notes
Compare outputs generated when syncing content repo. See author notes below for example repo config update required.

Should also test whether `--content-watch` and local-first workflows still working or not

## Author Notes
This PR deprecates the `config.google_drive.sheets_folder_ids` property and instead uses `config.google_drive.sheets_folders` which takes an array of both google drive `id` and a local `name` (can be anything, as long as unique with list of folders). 

E.g on `plh_kids_teens_za` using the names previously commented next to the IDs

```ts
config.google_drive.sheets_folders = [
  { id: "19wSspWYMbRc75een-kS0q0aq24--75u8", name: "library_app_menu" },
  { id: "1UXVz71HniwdtklFnGUEBzzj8ZHI9oQVo", name: "library_field_values" },
  { id: "1Y8uC9-rqQtsjQgUfeX9qp-vNzsFDUQFU", name: "kids_global" },
  { id: "1GnKk8luhnYcWobeeEfbR23ZSoZakcNF9", name: "kids_teens_global" },
  { id: "1XBq4iGIZHEwzwPk3xbHDAm9WCesjR7kR", name: "library PLH onboarding" },
  { id: "1Bzlnwts9mkoLRhDy-SN5O1A3bUlBOynI", name: "kids_teens_za" },
];

config.google_drive.assets_folders = [
  {id: "1abaL1QGd33NqqLoKuo2t9fVWKmh5ouM9", name: "kids_teens_global"},
  {id: "1T93qsaSBbYa-lCF6ChPkfoX85PLugJCZ", name: "kids_teens_za"},
];
```
Recompile deployment after setting
```bash
yarn workflow deployment set plh_kids_teens_za
```

## Dev Notes
I've chosen to use a single `_source` metdata property instead of previous suggested `_remoteUrl` and  `_remotePath` as a more generic catch-all that could in the future include any input-source specific metadata. 

I haven't included any additional checks to throw error if author uses same name for multiple drive folders (would cause issue as one folder would overwrite other, however they'd have to be defined in same config so assume not likely/easy authoring fix).
I did however add an additional layer of namespacing in the google drive download provider to split sheets and assets into separate subfolders, so that an asset folder can be named identically to a sheet folder (currently the case for `plh_kids_teens_za` which both refer to sheet/asset folders with name `kids_teens_za`)

## Git Issues

Closes #

## Screenshots/Videos
**Sync download now includes named folders in progress updates**
image

**Content repos now include nested path with parent folder prefix as well as direct URL**
Previously `_xlsxPath` would only be relative path and would need to follow `_sheetsFolderUrl` to identify which parent folder the flow originated from. Now can read everything inline and follow direct link to sheet
image

Before
```json
  "_xlsxPath": "data/ZA relax.xlsx",
  "_sheetsFolderUrl": "https://drive.google.com/drive/u/0/folders/1Bzlnwts9mkoLRhDy-SN5O1A3bUlBOynI"
```
After
```json
  "_source": {
    "path": "kids_teens_za/data/ZA relax.xlsx",
    "url": "https://docs.google.com/spreadsheets/d/1GJDojW4H5hLw8wO-EZmHrPcbrKvQ8WX9vDnJQk5Za2I"
  }
```

**Tasks now use intermediate folders**
Mostly only used for debugging, but still can be moderately helpful. Note that these folders are not automatically purged, so legacy folders will remain unless manually deleted all calling `yarn workflow clear_workflow_cache`
image

### DIFF
--- a/android/app/capacitor.build.gradle
+++ b/android/app/capacitor.build.gradle
@@ -14,6 +14,7 @@ dependencies {
     implementation project(':capacitor-firebase-crashlytics')
     implementation project(':capacitor-firebase-firestore')
     implementation project(':capacitor-firebase-performance')
+    implementation project(':capacitor-firebase-storage')
     implementation project(':capacitor-app')
     implementation project(':capacitor-clipboard')
     implementation project(':capacitor-device')
--- a/android/app/src/main/assets/capacitor.plugins.json
+++ b/android/app/src/main/assets/capacitor.plugins.json
@@ -20,6 +20,10 @@
 		"classpath": "io.capawesome.capacitorjs.plugins.firebase.performance.FirebasePerformancePlugin"
 	},
 	{
+		"pkg": "@capacitor-firebase/storage",
+		"classpath": "io.capawesome.capacitorjs.plugins.firebase.storage.FirebaseStoragePlugin"
+	},
+	{
 		"pkg": "@capacitor/app",
 		"classpath": "com.capacitorjs.plugins.app.AppPlugin"
 	},
--- a/android/capacitor.settings.gradle
+++ b/android/capacitor.settings.gradle
@@ -17,6 +17,9 @@ project(':capacitor-firebase-firestore').projectDir = new File('../node_modules/
 include ':capacitor-firebase-performance'
 project(':capacitor-firebase-performance').projectDir = new File('../node_modules/@capacitor-firebase/performance/android')
 
+include ':capacitor-firebase-storage'
+project(':capacitor-firebase-storage').projectDir = new File('../node_modules/@capacitor-firebase/storage/android')
+
 include ':capacitor-app'
 project(':capacitor-app').projectDir = new File('../node_modules/@capacitor/app/android')
 
--- a/ios/App/Podfile
+++ b/ios/App/Podfile
@@ -16,6 +16,7 @@ def capacitor_pods
   pod 'CapacitorFirebaseCrashlytics', :path => '../../node_modules/@capacitor-firebase/crashlytics'
   pod 'CapacitorFirebaseFirestore', :path => '../../node_modules/@capacitor-firebase/firestore'
   pod 'CapacitorFirebasePerformance', :path => '../../node_modules/@capacitor-firebase/performance'
+  pod 'CapacitorFirebaseStorage', :path => '../../node_modules/@capacitor-firebase/storage'
   pod 'CapacitorApp', :path => '../../node_modules/@capacitor/app'
   pod 'CapacitorClipboard', :path => '../../node_modules/@capacitor/clipboard'
   pod 'CapacitorDevice', :path => '../../node_modules/@capacitor/device'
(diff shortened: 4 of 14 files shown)


esmeetewinkel (commented):
Thanks for working on this, I think this will be helpful in debugging content issues.

**Sync_watch workflow** 
I've made some sheet changes while running `yarn workflow sync_watch` but it doesn't seem to be picking up on them:
image

_UPDATED: I also tested this on `master` just in case, turns out also there `yarn workflow sync_watch` doesn't pick up on sheet changes. So this issue doesn't seem to be caused by this PR._

`yarn workflow sync_local` seems to work okay (but I've never tested/used this intensively). 

**Separate folder from path**
I see that when no names are specified (in particular when the old config format is used) the folder id is used instead - which is fine.
image

It would make more sense to me to (independent of whether or not folder names are specified) separate out the path inside the folder from the folder itself. Having the folder id resp. name appear in a single variable alongside the rest of the path suggests that 
1. the Google Drive folders have the exact names as the folder names specified in the config (not true)
2. there exists a top level folder that contains all sheet/asset folders as subfolders (not true)

To prevent confusion I'd prefer to separate these out, e.g. 
```json
  "_source": {
    "folder": "1Bzlnwts9mkoLRhDy-SN5O1A3bUlBOynI", // show folder name if exists
    "path": "data/ZA activities.xlsx",
    "url": "https://docs.google.com/spreadsheets/d/1-bu6vC4Po5n6GdHPRHOACEJ3ltI07UyTNynN4VN6Pc8"
  }
```

chrismclarke:
> Thanks for working on this, I think this will be helpful in debugging content issues.
> 
> **Sync_watch workflow** I've made some sheet changes while running `yarn workflow sync_watch` but it doesn't seem to be picking up on them: image
> 
> _UPDATED: I also tested this on `master` just in case, turns out also there `yarn workflow sync_watch` doesn't pick up on sheet changes. So this issue doesn't seem to be caused by this PR._

Good to know - I'll try take a quick look at this as a follow-up

esmeetewinkel (commented):
I'm happy with the new format: 

1. If config has deprecated format (no folder names):
   image

2. If config has updated format, with folder names: 
   image

**Asset folder name cannot have spaces**
I discovered an additional issue: is it intentional that asset folder names cannot contain spaces? Folder names with spaces seem to work fine for sheet folders.

E.g. I used the format 
```
config.google_drive.assets_folders = [
  { id: "1T93qsaSBbYa-lCF6ChPkfoX85PLugJCZ", name: "kids teens za" },
];
```
which caused an error in the `assets_post_process` step of the content sync:
image

**Folder name in console ≠ folder name in jsons**
The name logged in the console is the actual name of the folder on GDrive ("ParentApp Kids & Teens ZA assets"), whereas I would have expected that to be the name specified in the config ("kids teens za"). Although maybe this makes sense when the config file doesn't specify folder names.

esmeetewinkel (approved):

PR #3196, opened 2025-10-31
feat: combo box component - expose params to override default key and value names in answer list

jfmcquade:
PR Checklist

## Description

Allows for authoring a combo box component with an answer list that uses columns other than `name` and `text` to act as the identifier and user-facing value for each option in the answer list. Those are still the default values, but each can be overridden by two newly exposed params:
- `options_key`
- `options_value`

## Testing

Each version of the combo box, as demonstrated in the [comp_combo_box](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=569531329#gid=569531329) template, should be tested to ensure it still works as expected.

## Dev notes

This pattern is borrowed from the reactive dropdown component introduced in #3150. The code for the combo box component is quite messy and this addition only makes it messier. However the existing code is quite fragile so I've haven't refactoring anything to avoid breaking production deployments. I think it's acceptable to add this complexity, and building a new component based around the new reactive dropdown should allow us to keep things simpler.

## Git Issues

Closes #

## Screenshots/Videos

New demo on [comp_combo_box](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=569531329#gid=569531329) template:

Screenshot 2025-10-31 at 17 10 22

[comp_combo_box_items](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=818486196#gid=818486196):

Screenshot 2025-10-31 at 17 11 06

Screenshot 2025-10-31 at 17 11 33

### DIFF
--- a/packages/@idemsInternational/env-replace/package.json
+++ b/packages/@idemsInternational/env-replace/package.json
@@ -8,7 +8,7 @@
     "test": "ts-mocha -p tsconfig.spec.json src/**/*.spec.ts"
   },
   "dependencies": {
-    "glob": "^9.0.0"
+    "glob": "^10.5.0"
   },
   "devDependencies": {
     "@types/chai": "^4.2.22",
--- a/src/app/shared/components/template/components/navigation-bar/navigation-bar.component.html
+++ b/src/app/shared/components/template/components/navigation-bar/navigation-bar.component.html
@@ -1,9 +1,9 @@
 <div class="navigation-bar-wrapper" [attr.data-variant]="params().variant">
-  @for (button of params().buttonList; track button.target_template) {
+  @for (button of params().buttonList; track button.targetTemplate) {
     <a
       class="button-wrapper"
       routerLinkActive="active-link"
-      [routerLink]="'template/' + button.target_template"
+      [routerLink]="'template/' + button.targetTemplate"
       [attr.data-name]="button.name"
     >
       <div class="image-wrapper">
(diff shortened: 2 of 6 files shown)


chrismclarke (changes requested):
The changes to the modal and search make sense to me - I think having the hardcoded fields was always a bit of a fragile implementation so good to at least have some means of providing a variable.

The combo-box component itself is indeed quite messy already and this doesn't make things any better, but can appreciate the challenge in trying to unpick and clean up while addressing this specific issue. 

I think short term I've added on small nit inline which can hopefully help tidy up a bit, and in the future just try to migrate to the reactive alternative which I'm guessing has a much cleaner implementation

esmeetewinkel (approved):
Functional test passed. @MichelePancera as part of handover could you functionally test too? 

I wanted to comment here that there _is_ an authoring workaround to this issue _as long as the data list exists at compile time_: you can use a `data_pipe` `map` operation to create a new data lists that has columns named as the combo box expects. I've added an example of this to 
[comp_combo_box_items](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=818486196#gid=818486196)
[comp_combo_box_pipe](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=1755915540#gid=1755915540)

https://idems-debug.web.app/template/comp_combo_box
image

chrismclarke (approved):
Thanks @jfmcquade - much easier to follow now, looks good to me

PR #3201, opened 2025-11-04
Refactor: app-data script cli

chrismclarke:
PR Checklist

## TODO 
- [x] Builds on top of #3184 - merge after

## Description
Following issue identified in #3184 where passing app-data assets post-process command would break if folder name has spaces in it, refactor the legacy `app-data` cli command to directly call underlying method in workflows.

## Review Notes
I've manually checked for any code or documentation that includes `app-data ` to catch cases where any app-data subcommands are directly referenced/invocated, e.g.
```
let cmd = `app-data post-process assets ${args}`;
await parseCommand(`${cmd}`);
```

The main thing to check is that the sync workflow still works as intended (which previously called the app-data commands). Can also confirm if now supports asset folders with spaces in names.

## Dev Notes
The original scripts command-line system was designed originally to allow triggering any script in isolation from the command line (e.g. setting a deployment, authenticating gdrive), usually with a limited number of configurable args designed around a single deployment workspace. 

Over time, however, we also implemented the tasks system to allow easier chaining of multiple commands, and handling multiple workspaces, which would still mostly trigger these commands but would opt to directly call underlying methods if more advanced arg configuration required (e.g. passing a function for sheet/asset filter, which would otherwise need to be serialised as string)

Currently we're in a slightly odd halfway-state where pretty much all interaction with the cli comes from workflow system (only exception I'm currently aware of is `yarn scripts version` - check code/docs for `yarn scripts` or `yarn workspace scripts start`), and pretty much all of the workflow tasks directly call underlying scripts instead of invocating via cli (check code/docs for `parseCommand`)

Moving to a single system would not only make the code easier to maintain, but will also improve type-safety when invocating scripts (only minimal validation available when using commander), and performance with reduced overhead spawning multiple child processes. There is a minor downside where the multiple processes could have been used to allow for parallel processing, however given the linear nature of our workflows this wasn't every really part of the design and could always be brought back in at the workflow-invocation level if desirable in the future.

## Follow-up
It would be good to eventually remove all scripts cli commands in favour of calling from workflows (I expect most has already been migrated), although best to do one command at a time to make for easier testing and review.

On quick inspection I think it's only actually the versioning script and workflow runner that are called via command line, but as a rough checklist should go through:

- [ ] Refences to cli invocation via `parseCommand`
- [ ] References to `child_process` and `execa` to track spawned processes
- [ ] References to `yarn workspace scripts start` to track spawned processes
- [ ] References to `commander` (once full cli no longer needed)

Where possible 

## Git Issues

Closes #

## Screenshots/Videos


chrismclarke:
@jfmcquade - tagging you for review as this should be good to open after #3184 is merged (can temp target that branch to see simpler diff, or just update post-merge)

PR #3202, opened 2025-11-05
[STYLING] Update: Combo Box styling

FaithDaka:
PR Checklist

## Description
- With this PR, the `combo-box` styling has been updated across themes to improve legibility and colour cohesiveness. In instances where the `combo-box` is combined with other components, its style shouldn't appear overpowering/combative.
- Facilitator apps should have more neutral/gray coloured dropdown states (unless designed otherwise) while the kids/teen apps have use of colour in all states.

### Other themes
| Theme | Image |
|--------|-------|
| facilitator-mx |    |
| default |    |
| kids-kw |   | 
| early-math |   |

### DIFF
--- a/packages/scripts/src/lib/app-data/convert/convert.spec.ts
+++ b/packages/scripts/src/lib/app-data/convert/convert.spec.ts
@@ -92,18 +92,17 @@ describe("App Data Converter", () => {
       paths.outputFolder,
       "../",
       "sheet_json",
-      "sheets",
-      "test_input.json"
+      "template",
+      "test_template.json"
     );
     expect(pathExistsSync(sheetJsonPath)).toEqual(true);
     const sheetJson = readJSONSync(sheetJsonPath);
     expect(Object.keys(sheetJson)).toEqual([
-      "==content_list==",
-      "test_data_list",
-      "test_template",
-      "test_draft",
-      "test_data_pipe",
-      "test_data_pipe_list",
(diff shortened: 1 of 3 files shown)


esmeetewinkel (commented):
Somehow what I see doesn't agree exactly with the screenshots in the PR description. I tested this on the PR preview link [here](https://plh-teens-app1--pr3202-3197-bug-cw-cbox-tex-yh9k4hhn.web.app/template/comp_combo_box). 

**Across themes**
I like the new look of the default (pop up) variant and the dropdown variant with search filter. 

The dropdown variant without search filter has a thick dark line (I think on all themes, but I tested `plh_facilitator_cw`, `default`, `plh_facilitator_mx`) which looks a little strange. Also the box with the possible responses no longer appears to "touch" the box, so it now is more of a pop up than a dropdown. And sometimes (for sure when taking a screenshot) the selected item appears in blue rather than a colour from the theme. 
   
| Before | After |
|:--:|:--:|
|image | image |

**`default` theme**
Since the `default` theme generally has a light blue background, it shows clearly that some components have a white background, inconsistently: 
- dropdown variant without search filter has white background
- dropdown variant with search filter doesn't have white background
- search filter pop up has white background behind listed items
image   image

**`plh_kids_kw` theme**
I don't like the inconsistency between the dropdown variant with/without search filter. 
image

PR #3206, opened 2025-11-13
[REFACTOR] Tidy up line height theme variables

FaithDaka:
PR Checklist

### DIFF
--- a/src/app/shared/components/template/components/combo-box/combo-box-modal/combo-box-modal.component.html
+++ b/src/app/shared/components/template/components/combo-box/combo-box-modal/combo-box-modal.component.html
@@ -7,7 +7,7 @@
       <div class="checkboxes">
         <div
           class="container-radio"
-          [class.checked-radio]="form.get('answer').value === radio.name"
+          [class.checked-radio]="form.get('answer').value === radio[optionsKey]"
           *ngFor="let radio of answerOptions()"
         >
           <label>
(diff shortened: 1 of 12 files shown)


src/app/shared/components/template/components/accordion/accordion.component.scss
```diff
@@ -1,11 +1,11 @@
 // Create overlapping effect
 ion-accordion {
-  background: white;
-  border: var(--border-standard);
-  border-radius: 10px;
-  margin-top: -12px;
-  padding-top: 8px;
-  box-shadow: var(--ion-default-box-shadow);
+  background: var(--accordion-background-color, white);
```
jfmcquade: nit(blocking): when introducing a new component-level theme var such as this, we should use the pattern of extracting that to a SASS `$` variable and defining that at the top of the file, including the fallback value. This makes it easy to see at a glance what values within a component-level stylesheet are exposed to theming. See an example in `packages/components/demo/basic/basic.component.scss`

jfmcquade (changes requested):
Thanks @FaithDaka, I appreciate that this is fiddly work, but great to keep moving in the right direction with this stuff and really nice to see the reduction in override style rules. 

I think you'll need to make a new PR to bypass the 10,000 commits issue I'm afraid. I'd also request that you update the title and description of the new PR to be explicit about what changes are included (it now goes beyond the line height theme variables). Would also be good to be clear about the ways in which it addresses https://github.com/ParentingForLifelongHealth/plh-facilitator-app-cw-content/issues/124.

See also a couple of review comments that I've added inline.

Then I imagine it would be good for us to do quite extensive functional testing on all themes, as there will be many expected visual changes (e.g. around margin/padding sizes), although they should be minor.

Issue #3163, opened 2025-10-02
[FEATURE] Workflow to create (and upload) an asset pack to remote asset storage


jfmcquade:
**What?**
Authors should be able to run a workflow to create an asset pack from a specified set of assets. Ultimately this should include the upload of the pack to the chosen remote assets provider (firebase/supabase), however this can be considered a stretch goal as uploading the files manually for the time being wouldn't be a big issue.

The major piece of functionality required is to generate the asset pack manifest json that must accompany the files. in order for them to constitute an asset pack.

PR #3213, opened 2025-11-17
feat: support remote asset pack source folders in deployments' synced assets

jfmcquade:
Replaces #3203 (new PR to avoid excessive commits post- repo git history updates)

PR Checklist

## Description

Adds the option for source asset folders provided in the deployment config to be flagged as `remote`, in which case the assets for these folders are not included in the main assets app data and instead get bundled into standalone asset packs, ready for manual upload.

Currently, the asset pack is exported but not uploaded anywhere. It will be committed to the content repo for versioning etc., but will not be available to download to users until the manual step of uploading the asset pack to a hosting provider (firebase or supabase) has been performed. In future this can be an automated step: the appropriate place for this is likely as a workflow that gets run when content PRs are merged to master.

Likewise, this workflow currently must be run manually, but should be incorporated into the general `sync_assets` workflow, so that both core and remote assets are synced when running a sync.

NB: The current thinking is that asset pack/remote asset functionality when testing locally will work the same as it would for end users, in that asset pack assets will not be available in the running app until they've been downloaded via the corresponding action.

Additionally, this PR refactors the remote asset code to remove the need for authors to include the asset pack name when referencing an asset in a template. Previously, the `debug_asset_pack_1` asset pack contained assets which were intended to override assets in a corresponding folder of the core assets, also called `debug_asset_pack_1`. This meant that when referencing these assets, authors would need to include `debug_asset_pack_1/*` in the path. This pattern is undesirable since we want assets from asset packs to be able to override arbitrary assets in the core, that may not be nested in a specific subfolder in this way. This required updating the remote asset service code to handle appending this asset pack name where necessary itself.

## Notes

Example config (for `debug` deployment):
```ts
config.google_drive = {
  sheets_folders: [{ id: "1YcG0GDJME1Ds1UzwXcEmo97Ckb5tnUWc", name: "Debug Sheets" }],
  assets_folders: [
    { id: "1bnGPB70wPp9hzVOpT4LkPX9h3FwIrqfs", name: "Debug Assets" },
    { id: "19Kbwidpb_nOKpJAuRCkVLpAzWbKPzNm1", name: "debug_asset_pack_1", remote: true }
  ],
};
```

This links to this [Google Drive](https://drive.google.com/drive/u/1/folders/19Kbwidpb_nOKpJAuRCkVLpAzWbKPzNm1) folder.

Note that this is not quite the same as the pattern suggested in https://github.com/IDEMSInternational/open-app-builder/pull/3203#pullrequestreview-3450654397. Namely, `remote` is just a boolean flag rather than specifying a platform that will host the assets. The configuration for the remote asset host remains in a separate config param for now:
```ts
config.remote_assets = {
  provider: "supabase",
  bucketName: "debug",
  folderName: "asset_packs",
};
```

Functionally, this means that a given app can only have remote assets served from a single provider. In future, this information could be included in a general `data_sources` property, as suggested in the comment linked above, but I think that decision can be made later, along with handling of other types of data source.

One point of discussion around naming: I've used the terms "asset pack" and "remote assets" throughout the code. It might make sense to try and rationalise this. I think that potentially "remote assets" could expand to cover more than just asset packs (e.g. remotely hosted assets that are delivered by a different means). But for the time being, the only remote assets functionality concerns asset packs, and I think "asset pack" is a more easily understandable phrase than "remote assets", so I have used the former for most author-facing functionality. However the fact that both terms are used in the code could cause a bit confusion.

## Testing

Experiment with creating a gdrive folder, referencing it in the deployment config, then running `yarn workflow sync` and inspect the output at `deploymeent_folder/app_data/remote_assets`.

Example content changes in debug PR: https://github.com/IDEMSInternational/app-debug-content/pull/217

## TODO/future follow-ups

- [ ] Remote assets upload workflow
- [ ]  Asset pack versioning
  - Versioning remote assets has the potential to get quite complicated, as the assets hosted remotely have the potential to be out of sync with a given user's app version. For now, it should be sufficient to keep a single version of a remote asset pack, but this will need discussion is future (also relates to @ChrisMarsh82's thoughts around storing and versioning core assets in future).

## Git Issues

First step towards addressing #3163

## Screenshots/Videos

Generated asset packs sit in a new `remote_assets` folder inside deployment's app_data directory. These will be committed to the content repo, but will not be copied to the running app folder and will not be included within the app bundle.

Screenshot 2025-11-19 at 09 35 58

Remote assets still successfully populated from an asset pack on all platforms:

| web | android | ios |
|-|-|-|
| Screenshot 2025-11-19 at 12 08 11 | Screenshot 2025-11-19 at 12 11 20 | Screenshot 2025-11-19 at 12 18 46 |

### DIFF
--- a/packages/test-e2e/README.md
+++ b/packages/test-e2e/README.md
@@ -29,13 +29,15 @@ https://drive.google.com/drive/u/0/folders/1EpRmiR2zTpU17EndofUHFUyca85IDyLV
 **Open the Cypress UI:**
 
 ```sh
-yarn cy:open
+cd ./packages/test-e2e
+yarn start
 ```
 
 **Run all tests in headless mode:**
 
 ```sh
-yarn cy:run
+cd ./packages/test-e2e
+yarn run test
 ```
 
 > These scripts are defined in this package's `package.json`.
--- a/packages/test-e2e/cypress/e2e/components/loop.cy.ts
+++ b/packages/test-e2e/cypress/e2e/components/loop.cy.ts
@@ -1,0 +1,217 @@
+describe("Loop Component", () => {
+  beforeEach(() => {
+    cy.visit("/template/test_loop");
+  });
+
+  it("Loops over a list", () => {
+    cy.getDataTest("listLoop.key_1")
+      .should("exist")
+      .within(() => {
+        cy.getDataTest("listLoop.key_1.loopTitle")
+          .should("exist")
+          .should("contain", "List item key_1");
+
+        cy.getDataTest("listLoop.key_1.loopText")
+          .should("exist")
+          .should("contain", "This is value 1");
+
+        cy.getDataTest("listLoop.key_1.plainText").should("exist").should("contain", "Plain text");
+      });
+
+    cy.getDataTest("listLoop.key_2")
+      .should("exist")
+      .within(() => {
+        cy.getDataTest("listLoop.key_2.loopTitle")
+          .should("exist")
+          .should("contain", "List item key_2");
+
+        cy.getDataTest("listLoop.key_2.loopText")
+          .should("exist")
+          .should("contain", "This is value 2");
+
+        cy.getDataTest("listLoop.key_2.plainText").should("exist").should("contain", "Plain text");
+      });
+
+    cy.getDataTest("listLoop.key_3")
+      .should("exist")
+      .within(() => {
+        cy.getDataTest("listLoop.key_3.loopTitle")
+          .should("exist")
+          .should("contain", "List item key_3");
+
+        cy.getDataTest("listLoop.key_3.loopText")
+          .should("exist")
+          .should("contain", "This is value 3");
+
+        cy.getDataTest("listLoop.key_3.plainText").should("exist").should("contain", "Plain text");
+      });
+  });
+
+  it("Loops over a query result", () => {
+    cy.getDataTest("queryLoop.id5")
+      .should("exist")
+      .within(() => {
+        cy.getDataTest("queryLoop.id5.loopTitle")
+          .should("exist")
+          .should("contain", "Query item id5");
+
+        cy.getDataTest("queryLoop.id5.loopText")
+          .should("exist")
+          .should("contain", "Test 5 Value (B)");
+
+        cy.getDataTest("queryLoop.id5.plainText").should("exist").should("contain", "Plain text");
+      });
+
+    cy.getDataTest("queryLoop.id4")
+      .should("exist")
+      .within(() => {
+        cy.getDataTest("queryLoop.id4.loopTitle")
+          .should("exist")
+          .should("contain", "Query item id4");
+
+        cy.getDataTest("queryLoop.id4.loopText")
+          .should("exist")
+          .should("contain", "Test 4 Value (B)");
+
+        cy.getDataTest("queryLoop.id4.plainText").should("exist").should("contain", "Plain text");
+      });
+
+    cy.getDataTest("queryLoop.id3")
+      .should("exist")
+      .within(() => {
+        cy.getDataTest("queryLoop.id3.loopTitle")
+          .should("exist")
+          .should("contain", "Query item id3");
+
+        cy.getDataTest("queryLoop.id3.loopText")
+          .should("exist")
+          .should("contain", "Test 3 Value (A)");
+
+        cy.getDataTest("queryLoop.id3.plainText").should("exist").should("contain", "Plain text");
+      });
+  });
+
+  it("Loop includes a nested template", () => {
+    cy.getDataTest("nestedLoop.0")
+      .should("exist")
+      .within(() => {
+        cy.getDataTest("nestedLoop.0.loopKey").should("exist").should("contain", "key: key_1");
+
+        cy.getDataTest("nestedLoop.0.loopTemplate.templateText")
+          .should("exist")
+          .should("contain", "value: This is value 1");
+
+        cy.getDataTest("nestedLoop.0.countText").should("exist").should("contain", "0 of 3");
+      });
+
+    cy.getDataTest("nestedLoop.1")
+      .should("exist")
+      .within(() => {
+        cy.getDataTest("nestedLoop.1.loopKey").should("exist").should("contain", "key: key_2");
+
+        cy.getDataTest("nestedLoop.1.loopTemplate.templateText")
+          .should("exist")
+          .should("contain", "value: This is value 2");
+
+        cy.getDataTest("nestedLoop.1.countText").should("exist").should("contain", "1 of 3");
+      });
+
+    cy.getDataTest("nestedLoop.2")
+      .should("exist")
+      .within(() => {
+        cy.getDataTest("nestedLoop.2.loopKey").should("exist").should("contain", "key: key_3");
+
+        cy.getDataTest("nestedLoop.2.loopTemplate.templateText")
+          .should("exist")
+          .should("contain", "value: This is value 3");
+
+        cy.getDataTest("nestedLoop.2.countText").should("exist").should("contain", "2 of 3");
+      });
+  });
+
+  it("Loop includes @first, @last, @item, @count, @is_first, @is_last", () => {
+    cy.getDataTest("firstLastLoop.key_1")
+      .should("exist")
+      .within(() => {
+        cy.getDataTest("firstLastLoop.key_1.firstText")
+          .should("exist")
+          .should("contain", "This is value 1");
+
+        cy.getDataTest("firstLastLoop.key_1.thisText")
+          .should("exist")
+          .should("contain", "This is value 1");
+
+        cy.getDataTest("firstLastLoop.key_1.lastText")
+          .should("exist")
+          .should("contain", "This is value 3");
+
+        cy.getDataTest("firstLastLoop.key_1.countText")
+          .should("exist")
+          .should("contain", "key_1 of 3");
+
+        cy.getDataTest("firstLastLoop.key_1.isFirstText")
+          .should("exist")
+          .should("contain", "Is first: true");
+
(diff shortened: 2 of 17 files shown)


packages/scripts/src/lib/app-data/postProcess/assets.ts
```diff
@@ -34,27 +35,32 @@ const APP_CORE_SIZE_KB = {
   total: APP_CORE_BUILD_KB + APP_CORE_ASSETS_KB,
 };
 
+interface IAssetFolderMetadata {
+  /** Whether this folder contains remote assets (asset pack) */
+  remote?: boolean;
+  /** Name of the folder (used for remote asset pack output directory) */
+  folderName?: string;
+}
```
chrismclarke: nit(blocking): the deployment config already defines
```
 /** gdrive id from end of url */
      id: string;
      /** local name for download */
      name: string;
      /** if true, the assets from this folder are treated as an asset pack to be uploaded to remote storage */
      remote?: boolean;
```
Adding another type here is confusing, as really anything that needs to be definied should be defined in the deployment config either within the data source config (currently nested in google drive), or within the remote_assets config. This should be passed to the processor directly from the workflow

packages/scripts/src/lib/app-data/postProcess/assets.ts
```diff
@@ -34,27 +35,32 @@ const APP_CORE_SIZE_KB = {
...
+
 interface IAssetPostProcessorOptions {
   sourceAssetsFolders: string[];
+  /** Map of source folder path to metadata (remote flag and folder name) */
+  folderMetadata?: Map<string, IAssetFolderMetadata>;
```
chrismclarke: nit(blocking): I think the list of folders is the same as the keys for the metadata no? If so then let's just pass a single list to iterate over.

I'd also probably recommend sticking with arrays over maps in case we ever need funcationality that merges multiple input sources into a single remote asset pack (otherwise remote folder names would override each other)

packages/scripts/src/lib/app-data/postProcess/assets.ts
```diff
@@ -65,52 +71,197 @@ export class AssetsPostProcessor {
...
         }
-        mergedAssetsHashmap[relativePath] = entry;
-      });
+        const remoteAssets = remoteAssetsBySourceFolder.get(sourceAssetsFolder)!;
+        Object.assign(remoteAssets, sourceAssetsFiltered);
+      } else {
+        // Core assets - merge into coreAssetsHashmap (later folders can overwrite earlier ones)
+        Object.assign(coreAssetsHashmap, sourceAssetsFiltered);
+      }
+    }
```
chrismclarke: nit(blocking): The processing here feels messy, jumping in and out of custom logic for remote assets vs non. I'm trying to wrap my head around just how much of the processing is the same and how much is different? The code makes it imply like a lot of the processing is different, but in reality don't both core and remote assets get processed in the exact same way, just populating to different output folders? 

Once populated to folders a final `remote_sync` workflow could be used to keep the local remote_assets folder in sync with target remote server (just like local script copies deployment app_data/assets folder to app assets folder)

So I'd expect all things going smoothly this PR to actually only add a small handful of lines to this script and a small number to a new workflow step

packages/scripts/src/tasks/providers/appData.ts
```diff
@@ -13,9 +13,12 @@ import {
...
 /** Prepare sourcely cached assets for population to app */
-const postProcessAssets = async (options: { sourceAssetsFolders: string[] }) => {
-  const { sourceAssetsFolders } = options;
-  return new AssetsPostProcessor({ sourceAssetsFolders }).run();
+const postProcessAssets = async (options: {
+  sourceAssetsFolders: string[];
+  folderMetadata?: Map<string, { remote?: boolean; folderName?: string }>;
+}) => {
+  const { sourceAssetsFolders, folderMetadata } = options;
+  return new AssetsPostProcessor({ sourceAssetsFolders, folderMetadata }).run();
```
chrismclarke: We're using `folderMetadata` as an additional arg that only applies to remote assets. It would be better to define a unified structure that can just be used for both core and remote, likely dropping the sourceAssetsFolder string and using the structure from the deployment config instead (which should include the metadata fields anyways) - essentially using 1 data structure instead of 3

packages/workflows/src/sync.workflows.ts
```diff
@@ -132,34 +132,51 @@ const workflows: IDeploymentWorkflows = {
...
+                folderName: folderConfig.name,
+              });
+            }
+          }
+
+          return tasks.appData.postProcessAssets({
+            sourceAssetsFolders,
+            folderMetadata,
+          });
+        },
```
chrismclarke: nit(blocking): I would still expect to see one additional processing step, which runs conditionally when the deployment has enable remote asset deployment config. Something like `sync_remote_assets` which handles the final step of pushing new assets to the server and deleting outdated (TBC how best to manage given legacy app users may still want access to the packs not required in latest release...)

Short term this is likely just copying from one local folder to another for visibility (if not already done so in previous processing step)

jfmcquade:
Thanks, @chrismclarke, very helpful comments. I've implemented your suggestions in [these commits](https://github.com/IDEMSInternational/open-app-builder/pull/3213/files/92b7b5413e1b5d9948a126f29e8897437a591e61..dcfb6099047de7d464fbf430bb2c741200c9ca61#diff-31ab5ac110845895d658007e00132444acb41cea92099c9b320d07a973981240). I've left your comments as unresolved for reference.

To summarise the updates:
- Refactored `AssetsPostProcessor` to use a single data structure matching the deployment config.
  - This isn't actually the same as that in the config, since the config source includes the google drive folder ID and the post processor source includes the local folder downloaded folder location instead, but I think these structures mirroring each other is sufficient
- Unified processing for all assets in a single loop, treating core and remote assets identically except for output location. And the additional difference that has now been made clearer is that the remote assets get a manifest file in the format of the `asset_pack` flow type.
  - I wouldn't have thought of using the `CORE_ASSETS_PACK` Symbol without Gemini's help, but it feels like a rare sensible use of Symbol so I quite like it.
- Added a placeholder (logic not yet implemented) `sync_remote_assets` workflow step that runs conditionally when remote assets are configured.

chrismclarke (approved):
>Refactored AssetsPostProcessor to use a single data structure matching the deployment config.
This isn't actually the same as that in the config, since the config source includes the google drive folder ID and the post processor source includes the local folder downloaded folder location instead, but I think these structures mirroring each other is sufficient

_Mirroring_ is a bit of a misnomer/dangerous concept here as there isn't anything really in the code that forces one definition to keep aligned with the other. It's usually better to extend an existing type interface, and if needing to make minor changes you can both extend and omit properties at the same time (but would recommend only doing so if the omit has significance, e.g. if a property isn't strictly required by a processor but is required by the parent just keep as required instead of adding omit)

I've added fe6fbc4be609cbf400d8c20ab1db4c03bdc48538 to at least export the type definition in case we want to reuse 

> Unified processing for all assets in a single loop, treating core and remote assets identically except for output location. And the additional difference that has now been made clearer is that the remote assets get a manifest file in the format of the asset_pack flow type.

Definitely following a more logical order now - agreed the reduced branching makes it a lot easier to actually understand the difference between core and remote assets (which isn't very much, beyond a few minor method changes and the manifest file)

> I wouldn't have thought of using the CORE_ASSETS_PACK Symbol without Gemini's help, but it feels like a rare sensible use of Symbol so I quite like it.
Added a placeholder (logic not yet implemented) sync_remote_assets workflow step that runs conditionally when remote assets are configured.

Sensible enough for now (seems to be a clean way to ensure a remote asset pack isn't named the same as core and accidentally overrides it), although I think in the future we can probably use asset pack type to avoid conflict and not worry about the name. 

I think we can probably go a little bit further still in terms of code tidying (likely using base and extended AssetPack classes to make it easier for remote asset pack to extend core asset pack with methods and names), but I don't want to hold up this PR any more and think in it's current shape it's more than reasonable. 

I have opened one minor PR targetting this branch just to reduce the length of the main file and separate out some of the longer methods to dedicated files in #3238 - I think in the future this will make it easier to add more focused tests, and organise the asset-pack post-processor to have more of a processing pipeline pattern with individual sub-processors, so if happy with that please go ahead and merge (and quickly check still working as I had to rush the PR a bit), and I'd say good to go

chrismclarke:
> Good point. I've added [1670cf2](https://github.com/IDEMSInternational/open-app-builder/commit/1670cf26cb222f427c6fadf8bd791eaa5b3c3b82) to make the interface used in the post-processor extend the one used in the deployment config. I have omitted the `id` field, since it is required in the deployment config interface but not used in the post-processor, which seemed sensible to me but I might be missing a clearer approach.

Yeah I think that's cleaner. Dropping the id is fine, whilst both extending and omitting and the same time always feels a bit messy (and is usually a good indicator that something isn't quite as clean as it could be further down the line), the reason for this is due to id being a field used to couple with google drive ids - so really there should be a cleaner base that both google drive downloaded asset packs and generated remote asset packs can extend, but at this point not really worth the refactor.

> I also noticed that we were saving an `untracked-assets.json` file, which logs assets that exist as overrides with no core version, for remote assets as well as the core assets. Remote asset packs, by design, can contain assets that are override versions only, so this file isn't relevant. I resolved this and then was able to improve the control flow slightly to distinguish between core/remote more cleanly with [38584c1](https://github.com/IDEMSInternational/open-app-builder/commit/38584c1b4a0026abfa1a71ca006cf5483a88425f)

Yeah that check doesn't make sense in both contexts. Probably what would make the most sense is actually building separate core and core-override asset packs, and then applying the same checks on core-override as well as remote assets (i.e. core asset exists fore override), but again probably overkill for now. 

> If you're happy with these latest changes then I'll go ahead and merge

Sound good to me

PR #3218, opened 2025-11-20
Chore: refactor nav-bar component to use new params pattern

jfmcquade:
Replaces #3171

PR Checklist

## Description

Refactors the `navigation_bar` component to use the new pattern for defining component params introduced in #3096. This includes adding a new coerce method to the params utils to handle the case of a parameter of the type of an array of objects.

There is no immediate need for updating this component in particular, but I chose to do so as a representative component as part of testing the new logic for defining params. This was originally done in #3171, but is broken out to a new PR (avoiding excessive commits after the repo history rewrite). For full context, see [review comments on that PR](https://github.com/IDEMSInternational/open-app-builder/pull/3171#issuecomment-3399924196) as well as https://github.com/IDEMSInternational/open-app-builder/pull/3096#issuecomment-3487735923.

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -21249,7 +21249,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"js-yaml@npm:4.1.0, js-yaml@npm:^4.1.0":
+"js-yaml@npm:4.1.0":
   version: 4.1.0
   resolution: "js-yaml@npm:4.1.0"
   dependencies:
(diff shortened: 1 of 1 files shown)


chrismclarke (approved):
I still think the `|` should probably be a `,` and would then follow more conventional object notation (but I lose track of these things) - but ultimately having the parsing utility to at least make the output more type-safe makes sense both with or without wider refactor

PR #3225, opened 2025-11-25
feat: tab variant of button

### LINKED ISSUE #3212: [FEATURE] Tab Bar
**What?**
Tab bar / tabbed navigation: a horizontal group of mutually exclusive buttons (tabs).
Selecting a tab usually shows the corresponding content panel below, but the component should also support cases where tab changes trigger other behaviours (e.g. filters, view modes).

Image

**Why?**
The functionality is already possible (for example in CW deployment):

Image

from the template: https://docs.google.com/spreadsheets/d/1J_fEngLyEzP64X7PE1XZyYGf3kDo_EpXx1fSzgY0sdo/edit?gid=867304914#gid=867304914

with the code:

Image

**How?**

- The tab bar is a single-select group: only one tab can be active at a time.
- Selecting a tab:
    - visually highlights the active tab, and
    - optionally updates the associated content panel or triggers a defined action.
- The component should support:
    - a state with no active tab (if needed), and
    - an initial active tab (default selection) for typical use.

Behaviour and configuration can likely mirror the single-select pattern described in #3211 (radio button groups), with a different visual treatment.

**Other details**

While we are mostly missing aesthetics, a dedicated component could be a good idea? ie Should we just have a Theaming option to have buttons in this format or should we have a new component?

jfmcquade:
PR Checklist

## Description

Enables authoring "tab bar"/"tab group" functionality. This is primarily achieved through two updates to the `button` component:
- Adds a new variant, `tab`
- Adds a new param, `highlighted`, that takes a boolean value
  - Currently this only has significance for the tab variant of the button, but the implementation could be extended to support a "highlighted" state on other variants

## Git Issues

Closes #3212

## Screenshots/Videos

Example on [comp_button](https://docs.google.com/spreadsheets/d/1OmgZICjM5EMT1KgLOU_ovDljRF_SPlpQAgkKWPrNX0s/edit?gid=569531329#gid=569531329) template:
Screenshot 2025-11-25 at 16 33 01

[example_tab_group](https://docs.google.com/spreadsheets/d/1HTqpAEkwxFJPwh1ulKmAEuH1jcMiBZQU3rfSVTlVT5Y/edit?gid=569531329#gid=569531329) template:

Screenshot 2025-11-25 at 16 37 41

| theme | [example_tab_group](https://docs.google.com/spreadsheets/d/1HTqpAEkwxFJPwh1ulKmAEuH1jcMiBZQU3rfSVTlVT5Y/edit?gid=569531329#gid=569531329) |
|-|-|
| default | Screenshot 2025-11-25 at 16 20 09 |
| professional | Screenshot 2025-11-25 at 16 21 37 |
| pfr | Screenshot 2025-11-25 at 16 21 50 |
| plh_facilitator_mx | Screenshot 2025-11-25 at 16 22 38 |
| plh_facilitator_my | Screenshot 2025-11-25 at 16 24 53 |
| plh_facilitator_ph | Screenshot 2025-11-25 at 16 25 41 |
| plh_facilitator_cw | Screenshot 2025-11-25 at 16 25 58 |
| early_family_math | Screenshot 2025-11-25 at 16 26 09 |
| plh_kids_kw | Screenshot 2025-11-25 at 16 26 42 |
| plh_kids_tz | Screenshot 2025-11-25 at 16 26 57 |
| plh_kids_teens_za | Screenshot 2025-11-25 at 16 27 44 |

### DIFF
--- a/src/theme/themes/plh_facilitator_cw/_overrides.scss
+++ b/src/theme/themes/plh_facilitator_cw/_overrides.scss
@@ -336,6 +336,14 @@
       }
     }
 
+    ion-button[data-variant~="tab"] {
+      --background: none;
+      color: var(--ion-color-primary);
+      &:hover {
+        --background: unset;
+      }
+    }
+
     ion-button[data-language-direction~="rtl"] .children {
       position: unset !important;
       align-self: unset !important;
--- a/src/app/shared/components/template/components/button/button.component.html
+++ b/src/app/shared/components/template/components/button/button.component.html
@@ -7,6 +7,7 @@
     (click)="triggerActions('click')"
     [attr.data-param-style]="params().style"
     [attr.data-variant]="params().variant"
+    [attr.data-highlighted]="params().highlighted ? 'true' : null"
     [attr.data-has-children]="_row.rows ? true : null"
     [attr.data-language-direction]="templateTranslateService.languageDirection()"
     [attr.data-icon-align]="params().iconAlign"
--- a/src/app/shared/components/template/components/button/button.component.scss
+++ b/src/app/shared/components/template/components/button/button.component.scss
@@ -224,3 +224,24 @@ ion-button[data-variant~="card"] {
     max-width: 10rem;
   }
 }
+
+ion-button[data-variant~="tab"] {
+  --background: none;
+  color: var(--ion-color-primary);
+  font-weight: var(--font-weight-bold);
+  --border-radius: 8px 8px 0 0;
+  --padding-bottom: 8px;
+  --padding-top: 8px;
+  --padding-end: 8px;
+  --padding-start: 8px;
+  min-height: 48px;
+  --box-shadow: none;
+  border-bottom: 4px solid transparent;
+  --background-activated: none;
+
+  opacity: 0.5;
+  &[data-highlighted="true"] {
+    opacity: 1;
+    border-bottom: 4px solid var(--ion-color-primary);
+  }
+}
--- a/src/app/shared/components/template/components/button/button.component.ts
+++ b/src/app/shared/components/template/components/button/button.component.ts
@@ -18,11 +18,14 @@ interface IButtonParams {
     | "navigation"
     | "short"
     | "standard"
+    | "tab"
     | "tall";
   /** TEMPLATE PARAMETER: "style". Legacy, use "variant" instead. */
   style: string;
   /** TEMPLATE PARAMETER: "disabled". If true, button is disabled and greyed out */
   disabled: boolean;
+  /** TEMPLATE PARAMETER: "highlighted". Currently only used for the "tab" variant, but implementation could be extended */
+  highlighted: boolean;
   /** TEMPLATE PARAMETER: "text_align" */
   textAlign: "left" | "centre" | "right";
   /** TEMPLATE PARAMETER: "button_align" */
@@ -73,6 +76,7 @@ export class TmplButtonComponent extends TemplateBaseComponent {
   private getParams(authorParams: FlowTypes.TemplateRow["parameter_list"]): IButtonParams {
     return {
       disabled: parseBoolean(this.parameterList().disabled),
+      highlighted: parseBoolean(this.parameterList().highlighted),
       style: `${getStringParamFromTemplateRow(this._row, "style", "information")} ${
         this.isTwoColumns() ? "two_columns" : ""
       }` as any,
--- a/src/theme/themes/plh_facilitator_mx/_overrides.scss
+++ b/src/theme/themes/plh_facilitator_mx/_overrides.scss
@@ -312,6 +312,14 @@
       }
     }
 
+    ion-button[data-variant~="tab"] {
+      --background: none;
+      color: var(--ion-color-primary);
+      &:hover {
+        --background: unset;
+      }
+    }
+
     ion-button[data-language-direction~="rtl"] .children {
       position: unset !important;
       align-self: unset !important;
--- a/src/theme/themes/plh_facilitator_my/_overrides.scss
+++ b/src/theme/themes/plh_facilitator_my/_overrides.scss
@@ -336,6 +336,14 @@
       }
     }
 
+    ion-button[data-variant~="tab"] {
+      --background: none;
+      color: var(--ion-color-primary);
+      &:hover {
+        --background: unset;
+      }
+    }
+
     ion-button[data-language-direction~="rtl"] .children {
       position: unset !important;
       align-self: unset !important;
--- a/src/theme/themes/plh_facilitator_ph/_overrides.scss
+++ b/src/theme/themes/plh_facilitator_ph/_overrides.scss
@@ -334,6 +334,14 @@
       }
     }
 
+    ion-button[data-variant~="tab"] {
+      --background: none;
+      color: var(--ion-color-primary);
+      &:hover {
+        --background: unset;
+      }
+    }
+
     ion-button[data-language-direction~="rtl"] .children {
       position: unset !important;
       align-self: unset !important;
--- a/src/theme/themes/plh_kids_kw/_overrides.scss
+++ b/src/theme/themes/plh_kids_kw/_overrides.scss
@@ -101,6 +101,15 @@
         }
       }
     }
+
+    ion-button[data-variant~="tab"] {
+      --background: none;
+      color: var(--ion-color-primary);
+      &:hover {
+        --background: unset;
+      }
+    }
+
     ion-button[data-language-direction~="rtl"] .children {
       position: unset !important;
       align-self: unset !important;
--- a/src/theme/themes/plh_kids_teens_za/_overrides.scss
+++ b/src/theme/themes/plh_kids_teens_za/_overrides.scss
@@ -101,6 +101,15 @@
         }
       }
     }
+
+    ion-button[data-variant~="tab"] {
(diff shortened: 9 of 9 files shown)


chrismclarke (changes requested):
I really think this should just be it's own component. There's no real reason an author should have to manually keep the tab pressed and content rendered in sync, and in it's current form doesn't really give nice scope for animating between tabs. 

Angular material has a good example of a tab group interface:
https://material.angular.dev/components/tabs/overview

It looks to reasonably align with the material spec:
https://m3.material.io/components/tabs/guidelines

If styling slightly different for IOS then would probably recommend adding an outside border, drop shadown and using background color to emphasis active instead of underline, e.g.: https://developer.apple.com/design/human-interface-guidelines/tab-views

From our side I think we could have a pretty minimal equivalent behind a `begin_tabs` parent group and `begin_tab` children.

It would also avoid further bloating the button component which is already doing quite a lot

jfmcquade:
> I really think this should just be it's own component. There's no real reason an author should have to manually keep the tab pressed and content rendered in sync, and in it's current form doesn't really give nice scope for animating between tabs.

We did discuss a variety of options ranging from the PR's current implementation where the logic lives at the authoring level to a set of custom components that would map onto [ionic's tab implementation](https://ionicframework.com/docs/api/tabs) (e.g. `begin_tab_group`, `begin_tab_bar`, `tab_button`, `begin_tab_content`). The reason for favouring the author-level approach was that, long term, we'd like the ability to create components at an authoring level. This can almost be achieved already by creating a reusable template with a considered I/O that can function effectively as a component, but we'd like to eventually be able to author components in sheets that can be instantiated in the same way as a component that has been written in the code. This would expose more of our logic at the "code as data" level and potentially allow us more flexibility be applying data tools to this component logic as well as template logic. So the thought is that we should offer a variety of atomic components in the code that can be configured in combination in authoring to then be re-exposed as more complex components. So in this case, whilst I agree that an author shouldn't have to manually keep the tabs and content in sync, the idea would be that most authors would never have to be aware of this as the authored component would be encapsulated and behave like any other component, but that there would be a level of author who _could_ customise this behaviour from the authoring level if desired.

However you raise a very good point about the possibility of animating between tabs. What that makes me think is that when we decide to build up some component at the authoring level as opposed to creating a new custom component, there is a always potentially a trade-off between having that logic/functionality/customisability exposed to authoring vs having a good, slick UI experience. We don't want our apps to have a compromised user experience for the benefit of an unseen structure, and I think in this particular case, that probably is a compromise not worth making. So I agree that this would be better as a dedicated (set of) component(s).

I think I would propose still merging the current PR and then implementing a dedicated component as a follow-up. I take your point about bloating the button component, but I can see a use for the included variant, and I think the `highlighted` param could have general applications. And then we offer authors the flexibility to author tabs in this way if they wish.

> Angular material has a good example of a tab group interface: https://material.angular.dev/components/tabs/overview
> 
> It looks to reasonably align with the material spec: https://m3.material.io/components/tabs/guidelines
> 
> If styling slightly different for IOS then would probably recommend adding an outside border, drop shadown and using background color to emphasis active instead of underline, e.g.: https://developer.apple.com/design/human-interface-guidelines/tab-views
> 
> From our side I think we could have a pretty minimal equivalent behind a `begin_tabs` parent group and `begin_tab` children.

Yeh I think the Angular Material pattern of authoring content blocks that also implicitly define a tab bar of buttons is probably a good approach, and nicer than explicitly defining both as in the [ion-tabs HTML](https://ionicframework.com/docs/api/tabs). Although we could always still use the ionic components and just expose our own components to authoring with our API of choice. When it comes to implementation, do you have a preference between the following:
- Using [ion-tabs](https://ionicframework.com/docs/api/tabs)
- Using [Angular Material tabs](https://material.angular.dev/components/tabs/overview)
- Building custom components from scratch
I think I'd favour `mat-tab`s – as you say, they're reasonably aligned with Material 3, and I don't really like the look of the ionic tab bar. I'd also probably see the iOS adaptation as a stretch/follow-up.

For the authoring API, I'm imagining it would look like the following ([debug sheet here](https://docs.google.com/spreadsheets/d/1HTqpAEkwxFJPwh1ulKmAEuH1jcMiBZQU3rfSVTlVT5Y/edit?gid=1905914480#gid=1905914480)):
Screenshot 2025-11-26 at 13 16 49

chrismclarke:
> Yeh I think this relates to the ongoing question of when things should be dedicated components vs when they should be built up in authoring. @ChrisMarsh82 and I have been discussing this recently and I think it's hard to come up with general rules so remains best approached on a case-by-case basis.

Case-by-case is best, but I would say as a general rule of thumb that if you need to add more than 10-20 lines of code to a core component to implement you're probably dealing with either a new component, or something that should remain just in templates.

For this current example, I would argue that the core button changes don't make sense because they don't add any real value when used in isolation. The tab button only really makes sense when included in another component that also handles tab change and content display logic. It would be like adding a nav-button style without any support for actual link navigation (forcing the author to always implement an additional logic layer when using).

> We did discuss a variety of options ranging from the PR's current implementation where the logic lives at the authoring level to a set of custom components that would map onto [ionic's tab implementation](https://ionicframework.com/docs/api/tabs) (e.g. `begin_tab_group`, `begin_tab_bar`, `tab_button`, `begin_tab_content`). The reason for favouring the author-level approach was that, long term, we'd like the ability to create components at an authoring level. This can almost be achieved already by creating a reusable template with a considered I/O that can function effectively as a component, but we'd like to eventually be able to author components in sheets that can be instantiated in the same way as a component that has been written in the code. This would expose more of our logic at the "code as data" level and potentially allow us more flexibility by applying data tools to this component logic as well as template logic. So the thought is that we should offer a variety of atomic components in the code that can be configured in combination in authoring to then be re-exposed as more complex components. So in this case, whilst I agree that an author shouldn't have to manually keep the tabs and content in sync, the idea would be that most authors would never have to be aware of this as the authored component would be encapsulated and behave like any other component, but that there would be a level of author who _could_ customise this behaviour from the authoring level if desired.

That actually sounds like you're justifying the case for separate components. Given that these will eventually move into an authored component repository, we should be focusing on keeping separate the current implementation. E.g. if in the future we're planning on authoring a tab system that shows different content depending on tab, then for now we should author a component that handles that functionality and later consider removal. Although to me tabbed content sections is a fundamental UI element that makes sense to keep in core repo. Material have a dedicated component, as does [shadcn](https://ui.shadcn.com/docs/components/tabs), [ionic](https://ionicframework.com/docs/api/segment) (via segment view) amongst others 

> However you raise a very good point about the possibility of animating between tabs. What that makes me think is that when we decide to build up some component at the authoring level as opposed to creating a new custom component, there is always potentially a trade-off between having that logic/functionality/customisability exposed to authoring vs having a good, slick UI experience. We don't want our apps to have a compromised user experience for the benefit of an unseen structure, and I think in this particular case, that probably is a compromise not worth making – animating between tab buttons or tab content would not be possible with the authored component approach. So I agree that this would be better as a dedicated (set of) component(s).
> 
> I think I would propose still merging the current PR and then implementing a dedicated component as a follow-up. I take your point about bloating the button component, but I can see a use for the included variant, and I think the `highlighted` param could have general applications. And then we offer authors the flexibility to author tabs in this way if they wish.

If planning a follow-up component I'd say just do it now, the code would be pretty minimal and have immediate benefit to authors and better future extensibility

> Yeh I think the Angular Material pattern of authoring content blocks that also implicitly define a tab bar of buttons is probably a good approach, and nicer than explicitly defining both as in the [ion-tabs HTML](https://ionicframework.com/docs/api/tabs). Although we could always still use the ionic components and just expose our own components to authoring with our API of choice. When it comes to underlying implementation, do you have a preference between the following:
> 
> * Using [ion-tabs](https://ionicframework.com/docs/api/tabs)
> * Using [Angular Material tabs](https://material.angular.dev/components/tabs/overview)
> * Building custom components from scratch
>   I think I'd favour `mat-tab`s – as you say, they're reasonably aligned with Material 3, and I don't really like the look of the ionic tab bar. I'd also probably see the iOS adaptation as a stretch/follow-up that wouldn't be required initially.
> 
> For the authoring API, I'm imagining it would look like the following ([debug sheet here](https://docs.google.com/spreadsheets/d/1HTqpAEkwxFJPwh1ulKmAEuH1jcMiBZQU3rfSVTlVT5Y/edit?gid=1905914480#gid=1905914480)): Screenshot 2025-11-26 at 13 16 49

src/theme/themes/plh_facilitator_cw/_overrides.scss
```diff
@@ -336,6 +336,14 @@
...
     }
 
+    ion-button[data-variant~="tab"] {
+      --background: none;
+      color: var(--ion-color-primary);
+      &:hover {
+        --background: unset;
+      }
+    }
+
```
chrismclarke: question(non-blocking):
It seems like most deployment are modifying the base styles. Is it a case that the base style still makes more sense for the rest of the deployments that do not include override (so aren't shown up here), or would it make more sense to update the base style to avoid the need for overrides (if these happen to be the only deployments currently using?)

chrismclarke (changes requested):

jfmcquade:
Thanks @chrismclarke, on reflection I agree that it makes sense to treat tabbed content sections as a fundamental UI element that belongs in the code. I've opened an alternative PR, #3230, and will close this one.

PR #3226, opened 2025-11-26
Feat: radio list component

### LINKED ISSUE #3211: [FEATURE] Radio Buttons
**What?**
Radio Buttons (functionality and aesthetics): mutually exclusive single choiches as in:

Image

**Why?**
This is a feature required by PLH, for example here: https://github.com/ParentingForLifelongHealth/plh-facilitator-app-cw-content/issues/15

**How?**
- Only one option in the group can be selected at any time.
- Selecting a new option automatically unselects the previously selected one.
- By default, the group should support having no option selected.
- It should also be possible to configure an initial selected value when needed (e.g. for pre-filled forms or “edit” flows).

**Other details**
Should this be in essence a different visualization of combo boxes or should we allow the option of more than one choice to be selected?

jfmcquade:
PR Checklist

## Description

Adds a new component, `radio_list`, to display a list of single-selectable options.

This component exposes the following parameters:
- `answer_list`
- `options_key`
- `options_value`
The latter two options can be used to override the default names for the key/value properties in the provided answer list, as was introduced for the combo box component in #3196.

## Dev Notes

We already have two components with similar functionality: `radio_group` and `radio_button_grid`. An argument could be made that this new component should be exposed as a variant of `radio_group` instead of a standalone component (or indeed that the existing components should also be handled in this way). I think that as the HTML template is completely different for this component, it made sense to create it independently, plus I wanted a clean TS file to keep the logic as simply as possible. We could expose the `radio_list` as a variant of the `radio_group` component at the authoring level, while still keeping the code files separate (e.g. as a subcomponent in the code), but I think that separate template components is acceptable.

Whilst I was able to keep the logic reasonably minimal, the majority of lines were added to enable the customisable key/value names and to allow authoring of the answer list via a nested data-items row ([these commits](https://github.com/IDEMSInternational/open-app-builder/pull/3226/files/24414e63a98e90182d040d265af1477b2106c626..febef6230fb31b3e2a21b2660ad454b8b1c1c703)). The latter functionality was copied over from the combo-box component, and whilst I was able to simplify the logic significantly, it's still a bit of a hack tied to how we handle data-items within a dedicated component (I will add a follow-up to refactor the combo-box component to use the zod params and simplify this equivalent logic). Potentially this could be simplified significantly under the core rewrite, with components that expect a list of options (including the radio components and combo-box) reading from any child rows they may have in a more general way.

I've kept the `answer_list` parameter so named to retain interoperability, but in the core rewrite we'll likely want to rename this parameter and make it clear that its properties are not fixed and depend upon component implementation and author param config. 

## Git Issues

Closes #3211

## Screenshots/Videos

[comp_radio_list](https://docs.google.com/spreadsheets/d/1ykG_OU7bZJhNyoVKzOGjjGJ2F0OHbdJC0sh_7d67fTY/edit?gid=569531329#gid=569531329):

Screenshot 2025-11-26 at 16 29 54

Screenshot 2025-11-26 at 16 30 16

### DIFF
--- a/src/app/shared/components/template/components/radio-list/radio-list.component.ts
+++ b/src/app/shared/components/template/components/radio-list/radio-list.component.ts
@@ -36,7 +36,12 @@ export class TmplRadioListComponent extends TemplateBaseComponentWithParams(Auth
     toObservable(this.rows).pipe(
       map((rows) => rows.find((r) => r.type === "data_items")),
       filter((row) => row !== undefined),
-      switchMap((row) => this.dataItemsService.getItemsObservable(row, this.parent.templateRowMap))
+      switchMap((row) =>
+        this.dataItemsService.getItemsObservable(
+          row,
+          this.parentContainerComponentRef.templateRowMap
+        )
+      )
     )
   );
 }
--- a/src/app/shared/components/template/components/radio-list/radio-list.component.ts
+++ b/src/app/shared/components/template/components/radio-list/radio-list.component.ts
@@ -1,11 +1,16 @@
-import { Component } from "@angular/core";
+import { Component, computed } from "@angular/core";
+import { toObservable, toSignal } from "@angular/core/rxjs-interop";
+import { filter, map, switchMap } from "rxjs/operators";
 import { defineAuthorParameterSchema, TemplateBaseComponentWithParams } from "../base";
 import { IAnswerListItem } from "../../../../utils";
+import { DataItemsService } from "../data-items/data-items.service";
 
 const AuthorSchema = defineAuthorParameterSchema((coerce) => ({
   answer_list: coerce.objectArray<IAnswerListItem>([
     { name: null, text: null, image: null, image_checked: null },
   ]),
+  options_key: coerce.string("name"),
+  options_value: coerce.string("text"),
 }));
 
 @Component({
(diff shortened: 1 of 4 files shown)


src/app/shared/components/template/components/radio-list/radio-list.component.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  // Allow radio_list to include data_items child row to define answer list
+  private dataItemRows = toSignal(
+    toObservable(this.rows).pipe(
+      map((rows) => rows.find((r) => r.type === "data_items")),
+      filter((row) => row !== undefined),
+      switchMap((row) =>
+        this.dataItemsService.getItemsObservable(
+          row,
+          this.parentContainerComponentRef.templateRowMap
```
chrismclarke: nit(non-blocking): a bit annoying we still need this workaround but appreciate nothing can really be done for now (fixed with reactive templating).

chrismclarke (approved):
Looks good to me.  Agreed current radio group is already doing quite a lot and not really setup to handle what the fairly primitive behaviour required here. I could imagine a case in the future where we make a minor refactor to the radio group to extend the radio list, or perhaps just author as a templated component (either using button and state primatives, or exposing a slot in this component to make it easier to control rendered child item component to allow for a button instead of a radio)

MichelePancera (approved):

PR #3227, opened 2025-11-26
Chore: refactor combo box

jfmcquade:
PR Checklist

## Description

Refactor combo box component to simplify the logic, following #3226 and #3196.

Includes renaming a parameter, which I don't think is used in production deployments, from `text` to `modal_title`, to more accurately describe its meaning.

## Testing

Ensure that every example on [comp_combo_box](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=569531329#gid=569531329) works as expected. There should be no functional changes (besides the param name change).

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/src/app/shared/components/template/components/index.ts
+++ b/src/app/shared/components/template/components/index.ts
@@ -48,6 +48,7 @@ import { TmplProgressPathComponent } from "./progress-path/progress-path.compone
 import { TmplQRCodeComponent } from "./qr-code/qr-code.component";
 import { TmplRadioButtonGridComponent } from "./radio-button-grid/radio-button-grid.component";
 import { TmplRadioGroupComponent } from "./radio-group/radio-group.component";
+import { TmplRadioListComponent } from "./radio-list/radio-list.component";
 import { TmplSharedDataComponent } from "./shared-data/shared-data.component";
 import { TmplSimpleCheckboxComponent } from "./simple-checkbox/simple-checkbox.component";
 import { TmplSliderComponent } from "./slider/slider.component";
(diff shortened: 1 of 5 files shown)


chrismclarke (approved):
Looks good to me, nice to see net reduction in the code on a lot of these components that you've started to tidy up. 

If concerned about breaking changes for authoring, you could start to consider exposing a common some of these validators to check at compile time when syncing a deployment (possibly behind an optional flag). 

The general structure to do this would likely be a single top-level `TEMPLATE_PARAMETER_VALIDATORS` that stores key-value pairs of row type against validators, e.g. something along the lines of

```ts
import {ComboBoxAuthorSchema } from './path/to/combobox'

type ITemplateValidators = Record<ICommonComponentName, z.ZodObject<any, any>>

export const TEMPLATE_PARAMETER_VALIDATORS:Partial<ITemplateValidators>={
  combo_box: ComboBoxAuthorSchema 
}
```

and then adding a parser step during template parsing (or later, depending on what is most useful/convenient) to validate rows against the schema. 

I don't think it would be a large feature to introduce, and potentially a pretty useful one longer term as more deployment use components that we want to refactor, so having some basic scaffolding in place to add validation could be nice.

MichelePancera (commented):
I'm having a surprising issue: In main branch (as well as in the tabs branch for example) all seems to work correctly, including the last 3 combo boxes (from line 75 in the corresponding template):
image
image

When this branch is selected the last 3 combo boxes appear disabled:
image

PR #3230, opened 2025-11-27
Feat: tabbed content

### LINKED ISSUE #3212: [FEATURE] Tab Bar

jfmcquade:
PR Checklist

## Description

Allows for authoring of tabbed content blocks via two new components: `tabs` and `tab`. The authoring pattern is similar to how accordions are authored via the `accordion` and `accordion_section` components, see demo below for examples.

## Dev notes

The `tab` component is very minimal. It could be considered redundant: we could allow any components to be direct children of the `tabs` component, e.g. `display_group`, and interpret their value as the tab label to be displayed on the associated button. However the current approach of a dedicated component allows for the possibility of adding tab-level params in future, and arguably keeps the authoring more readable.

I considered making the `tabs` component standalone so it could import the `MatTabsModule` module itself rather than doing so in the general template module, like the table component does with the `MatTableModule`. However, because the `tabs` component also needs the `TemplateComponent` as an import, making it standalone would require further refactoring elsewhere (to avoid circular dependencies, for example), so for now it's much simpler to just import the `MatTabsModule` at the module level.

## Git Issues

Closes #3212

## Screenshots/Videos

[comp_tabs](https://docs.google.com/spreadsheets/d/1raJlddyTZlGxQyv78Nm5PHJxMftisYuCWFWgIVvGxq8/edit?gid=332666066#gid=332666066)

Screenshot 2025-11-27 at 15 59 47

| theme | component demo |
|-|-|
| default | Screenshot 2025-11-27 at 16 00 33 |
| plh_kids_kw | Screenshot 2025-11-27 at 16 01 01 |
| plh_facilitator_cw | Screenshot 2025-11-27 at 16 01 22 |

### DIFF
--- a/packages/data-models/deployment.model.ts
+++ b/packages/data-models/deployment.model.ts
@@ -102,6 +102,15 @@ export interface IDeploymentRuntimeConfig {
   useReactiveTemplates?: boolean;
 }
 
+export interface IAssetSource {
+  /** gdrive id from end of url */
+  id: string;
+  /** local name for download */
+  name: string;
+  /** if true, the assets from this folder are treated as an asset pack to be uploaded to remote storage */
+  remote?: boolean;
+}
+
 /** Deployment settings not available at runtime  */
 interface IDeploymentCoreConfig {
   google_drive: {
@@ -116,7 +125,7 @@ interface IDeploymentCoreConfig {
     /** @deprecated Use `assets_folders` array instead */
     assets_folder_ids?: string[];
     /** gdrive id from end of url and local name for download */
-    assets_folders?: { id: string; name: string }[];
+    assets_folders?: IAssetSource[];
     /** generated gdrive access token. Default `packages/scripts/config/token.json` */
     auth_token_path?: string;
     /** filter function applied to sheets download that receives basic file info such as folder and id. Default `(gdriveEntry)=>true` */
--- a/packages/scripts/src/lib/app-data/postProcess/asset-processors/asset-filters.spec.ts
+++ b/packages/scripts/src/lib/app-data/postProcess/asset-processors/asset-filters.spec.ts
@@ -1,0 +1,104 @@
+import { filterAppAssets } from "./asset-filters";
+import type { IDeploymentConfigJson } from "data-models";
+import type { IContentsEntry } from "shared";
+
+const mockFileEntry: IContentsEntry = {
+  size_kb: 100,
+  md5Checksum: "mock",
+  relativePath: "mock",
+  modifiedTime: "2023-01-01T00:00:00.000Z",
+};
+
+function createMockAssets(paths: string[]) {
+  const assets: { [relativePath: string]: IContentsEntry } = {};
+  paths.forEach((p) => {
+    assets[p] = { ...mockFileEntry, relativePath: p };
+  });
+  return assets;
+}
+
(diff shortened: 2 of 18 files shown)


src/app/shared/components/template/components/tabs/tabs.component.html
```diff
@@ -0,0 +1,13 @@
+<mat-tab-group
+  [mat-stretch-tabs]="!params().compactButtons"
+  [animationDuration]="animationDuration()"
+>
+  @for (childRow of rows() | filterDisplayComponent; track trackByRow($index, childRow)) {
+    <mat-tab [label]="getTabLabel(childRow)" [attr.data-rowname]="childRow.name">
```
chrismclarke: nit(non-blocking): alternatively, you could use a `computed` statement on the component level to automatically map the `rows()` to a new data structure that filters to be tab components and also populates the label. This keeps the display component a bit cleaner (doesn't have to call back to the template logic on render).

You could also fully extract the child tab rows during that step and render directly in this component, removing the need for the separate tab component. Something like
```ts
public tabRows = computed(()=>rows()
  .filter(row=>{...})
  .map(tabRow=>({
    label:tabRow.params.label || row.name,
    childRows: tabRow.rows
  })
))
```

chrismclarke (approved):
Thanks for the updates @jfmcquade - I appreciate is was more work on top of #3212 but I think the end result is a lot cleaner and extensible.

I've added a few minor (non-blocking) comments inline, but overall I think this is a pretty sensible component to add

jfmcquade:
Thanks @chrismclarke, I've implemented your suggestions.

I was initially unsure about removing the `tab` subcomponent, as I had reservations about a pattern that requires authoring a template row that looks like a component row but is not (as is the case for a row with type `tab` now. But on reflection I don't think this is too much of an issue: the `tab` component already only made sense when instantiated within a `begin_tabs` block anyway. I'm aware that there could be complications if we decided we wanted to expose additional parameters on `tab` rows/components – currently we can read `tabRow.parameter_list?.label` from the parent `tabs` row, but if we wanted to work with a parameter that was more complex than a string then that parameter parsing would be best suited to being defined at the level of a standalone template component. but we can deal with that situation in the unlikely event that it arises.

PR #3240, opened 2025-12-04
Feat/asset pack download button component

### LINKED ISSUE #3164: [FEATURE] Remote assets: user-facing feedback for asset pack download
**What?**
There should be some user-facing feedback when downloading an asset pack.

A suggested first version of this might be:

- A dedicated asset pack download component, that, on click, becomes an unclickable loading bar that indicates progress. Then on success/failure of the asset pack download, the button is replaced by a message that reports this status to the user
  - Whilst a true progress bar might be ideal, many UXs now opt for a simple loading animation, as it can be frustrating when the progress bar doesn't advance predictably. This might be a better option, and easier to implement

jfmcquade:
PR Checklist

- [x] Builds on #3213 (wait for merge)

## Description

Adds a new template component, `asset_pack_download_button`.

This component is a dedicated button for downloading a named asset pack. The motivation for this implementation pattern was that we don't currently have a way for actions to return values to the authoring level. Therefore, an approach like this is necessary within our current constraints in order to have the UI reflect a "success" or "error" state of this operation. This component has four distinct states: "initial", "downloading", "success" and "error". The component itself displays differently for each of these states, but it also updates its own row value to reflect its state, so this can be used  elsewhere in the template to react to this state, e.g. within a condition.

## Notes

In the future, it would be good to be able to display continuous download progress (e.g. as a percentage).

If desirable, we could expose the status of an attempted asset pack download to authoring as a dedicated field, available from any template. E.g., a field called `asset_pack_download_status_${asset_pack_name}`, that could have the values "success" or "error" (or numbered error codes), or would be undefined in the event that the download had not been attempted. My instinct is that we want to move away from this kind of dedicated app field, but it may be that we want some version of this.

## Git Issues

Closes #3164 

## Screenshots/Videos

[comp_asset_pack_dl_button](https://docs.google.com/spreadsheets/d/1kV8Bd2wvP3swyrYPYDAJf0px3Gd9PLeDIF4wdgcF-9w/edit?gid=569531329#gid=569531329)

Screenshot 2025-12-05 at 15 05 23

| initial | downloading <br> (progress_display: count) | downloading <br> (progress_display: spinner) | success | error |
|-|-|-|-|-|
| Screenshot 2025-12-05 at 15 00 55 | Screenshot 2026-01-07 at 16 09 18 | Screenshot 2025-12-05 at 15 04 10 | Screenshot 2025-12-05 at 15 01 06 | Screenshot 2025-12-05 at 15 03 38 |

[debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit?gid=1060209527#gid=1060209527) is also updated to include the component implementation of the asset pack download logic alongside the pre-existing template action implementation.

Screenshot 2025-12-05 at 15 08 41


chrismclarke (changes requested):
I think a dedicated component is fine (for now) as at least it simplifies the ui and download status binding, although I really think this needs some level of feedback to users to at least feedback downloaded / total numbers (an indefinite spinner makes it pretty hard to distinguish stalled vs slow downloads, and will likely encourage users to navigate a way and retry, wasting data). Given it's the service that handles downloads I think it should be pretty simple to ask ai to refactor and share some sort of progress signal

I'm also wondering what will happen if an asset pack is partially downloaded, but then interrupted. Will a user have to download everything again? Or is the manifest checked against local files to see what is still remaining? I think this would also be a minor improvement worth prioritising if not already done.

ChrisMarsh82 (approved):

jfmcquade:
> I think a dedicated component is fine (for now) as at least it simplifies the ui and download status binding, although I really think this needs some level of feedback to users to at least feedback downloaded / total numbers (an indefinite spinner makes it pretty hard to distinguish stalled vs slow downloads, and will likely encourage users to navigate a way and retry, wasting data). Given it's the service that handles downloads I think it should be pretty simple to ask ai to refactor and share some sort of progress signal

Good suggestion, I've added a `progress_display` param which defaults to `count`, in which case a progress count is displayed in the format `(X/Y)`. The old spinner is still possible by setting `progress_display: spinner`, and we could expose `progress_display: percentage` in future, although implementation would be tricky without firebase exposing byte-level download progress. See the commits since your last review for the changes.

> I'm also wondering what will happen if an asset pack is partially downloaded, but then interrupted. Will a user have to download everything again? Or is the manifest checked against local files to see what is still remaining? I think this would also be a minor improvement worth prioritising if not already done.

This is indeed an oversight and I agree that it should be prioritised, but I think probably makes sense as a separate follow-up PR. We're not planning on using this feature in production imminently. I've made a new issue: #3268.

chrismclarke (approved):
Code looks good to me, if working functionally I'd say should be good to go

Issue #3199, opened 2025-11-04
[STYLE] Text spacing issues


jfmcquade:
There are style issues around paragraph spacing of text elements in combination with other components. E.g. there is probably now too much space between a text and title. Consider updating paragraph spacing behaviour so that it only applies to consecutive paragraphs within a text component, not at beginning/end of text component itself.

From https://github.com/IDEMSInternational/open-app-builder/pull/3144:

Functional test passed. I can see slight colour difference on a few screens, all of which I think are acceptable. Another difference visible in the comparison screenshots below is that the paragraph spacing has increased (following PR https://github.com/IDEMSInternational/open-app-builder/pull/3193) which I think is an improvement on screens with big blocks of text. 

After / Before:

<img width="991" height="834" alt="image" src="https://github.com/user-attachments/assets/217e2a74-8a79-4149-92cf-8d119da3a68d" />

<img width="863" height="830" alt="image" src="https://github.com/user-attachments/assets/87fa2ba8-68e5-41de-866f-83c6481a4c78" />

<img width="851" height="833" alt="image" src="https://github.com/user-attachments/assets/5c31308a-244c-4cae-ac91-0494eaebb23b" />

<img width="858" height="829" alt="image" src="https://github.com/user-attachments/assets/df93669f-7c17-4635-b34c-8b114bc320ae" />

<img width="846" height="832" alt="image" src="https://github.com/user-attachments/assets/518db723-401f-4410-9830-60a802a5b6e2" />

_Originally posted by @esmeetewinkel in https://github.com/IDEMSInternational/open-app-builder/pull/3144#pullrequestreview-3410752815_

PR #3243, opened 2025-12-10
[CHORE] Tidy up line height theme variables

### LINKED ISSUE #3179: [CHORE] Refactor line-height theme vars
Copied from https://github.com/IDEMSInternational/open-app-builder/pull/3175#issuecomment-3437981085

> Perfect, also just as a quick FYI I did think about adding an extra variable to allow setting default `line-height` within a theme and apply to the `p` tags to improve readability on text-heavy deployment themes (like I did in the screenshot), but quickly realised it's a bit of a mess to unpick the large number of components that use legacy `--line-height-text-small`, `--line-height-text-tiny` alongside hardcoded values like `14px`, `22px`, `27px`, and multipliers `1,25`, `1.05 * $line-height-base` etc.
> 
> **Codebase search for `line-height` - 113 results across 45 files** image
> 
> So I think this would be a good candidate for future tidying. I'd probably recommend focusing on a single default multiplier value (e.g. `1.6`) that should handle small text in a reasonable way
> 
> Looking at the current themes, it looks like every plh theme changes the default variables in the same way, so would probably recommend starting by making that the default and removing need for overrides image
> 
> From there I'd probably guess that given the medium text variable is 28px, that probably a scale factor around 28/16 = 1.75 (nice for reading legibility), and would expect if tiny/large text where also scaled in the same way we probably would still get spacing similar to the hardcoded values. e.g. the `-text-large: 32px` would be the same as scaling a base font 18px by 1.75 - even if the text uses font larger than 18px (which most seem to not, still inheriting default font and just adding a bit more spacing), I don't think the spacing would look significantly different.
> 
> Also to note that there's no code that actually use the -extra-large variant so can just scrap (code uses `--line-height-text-largest` which no themes override but use despite the default value of `25px` being roughly the same as the theme `--line-height-text-small` value of 24px (so no longer the largest at all, and smaller than just leaving unchanged and applying default 16px * 1.75 = 28px). So here the use of extra-large actually has opposite effect on content line-height, which is pretty counter-intuitive.
> 
> Maybe worth a follow-up issue? I can already see these variables starting to be used in the new reactive components, so would probably trying to tidy up asap to prevent too much more spill into new code

FaithDaka:
PR Checklist

## Description
- In this PR, the `line-height` attribute within the theme override files has been tidied up such that it is defined within the component-level and typography styling files.
- Where necessary, certain components' styling overrides were refactored and transferred to their individual styling files as a way to tidy up.
- The global `line-height` var points toward `line-height-text` and set for the `p` tag styling in `_typography.scss`. This variable has been defined with `1.3` to match the current line height.
- Variables within the `variables.scss` have been updated such that they are primarily multiples of 2. These include margins, padding, font-size and line-height.
- 3 additional component-level variables have been created to make styling more dynamic. These were added in the theme definition files i.e `title-text-color` to define title's text colour, `accordion-background-color` to set an accordion's bg and `accordion-border-color` to set the accordion's border colour.

## Git Issues

Closes #3179 

#### Components
| Before | After |
|--------|------|
|      |     |
|     |     |
|    |    |


jfmcquade (changes requested):
Apologies for the very slow review, this slipped off the bottom of my to-do list.

Thanks for this, it's great to see a reduction in the amount of theme overrides code and a tidying up of the themes. I think generally that what has been moved from theme overrides to component styling is sensible.

There are a few visual changes with each updated component that I'm not sure about, see below. It's important to include screenshots like this in the PR description so that we can quickly see what the expected visual changes are. I've tested exclusively on the `default` theme to look for unintended knock-ons, so it would be good if you could test on other themes after updating this PR. Additionally, you'll need to update any themes that have been created since this PR was opened, as they will likely have copy-pasted code from the old versions of affected themes.

## Accordion

comp_accordion

The accordion component has lost its default border. This should be restored, and the `transparent` override should be handled at the theme level via the `--accordion-border-color` var you've introduced

| before | after |
|--|--|
| before | after |

## Task card

comp_task_card

I think the line height changes for this component could be an issue, as it will reduce the amount of content available on the screen which may impact some existing deployments. These changes should be made as minimal as sensibly possible.

| before | after |
|-|-|
| before 1 | after 1 | before 2 | after 2 |

## Title

comp_title

The new spacing between lines within a single paragraph looks much too big to me. I don't think we want to change this from the old version. (see example in comp_task_card above too)

| before | after |
|-|-|
| before 1 | after 1 |
| before 2 | after 2 |

## Text

comp_text

The spacing between paragraphs has increased, which is not desirable. It ahs already been observed that consecutive paragraphs get spaced too far apart by default (e.g. in #3199), so this is a step in the wrong direction. I would also say that to my eyes the increased spacing between lines within a single paragraph looked better before when it was smaller. It is a frequent request to be able to fit more content on the screen, so I think we need to be very deliberate with such changes if they are desirable

| before | after |
|-|-|
| before | after |

PR #3245, opened 2025-12-10
fix: force_reload action

### LINKED ISSUE #3155: [BUG] force reload sometimes has no effect

jfmcquade:
PR Checklist

## Description

Fixes an issue with the `emit: force_reload` action that meant that not all child elements in the template would fully re-render as expected. 

See https://github.com/IDEMSInternational/pfr-app-content/issues/162 for a manifestation of the issue that prompted a fix being implemented. 

## Dev notes

As documented in #3155, the commit that changed the behaviour of the `emit: force_reload` action seems to be [this one](https://github.com/IDEMSInternational/open-app-builder/pull/2427/commits/f357220ba119e1642bee6257ad8fb05f001634f6), from [#2427](https://github.com/IDEMSInternational/open-app-builder/pull/2427). This would mean that the issue has been present for around a year. I think it's plausible that this has gone unnoticed since our authoring patterns have developed to rely much more on dynamic data rather than fields, meaning that the `force_reload` action is much less relied upon. The PFR deployment is still built around the old static patterns, and so requires this action to initiate a more complete reload as it used to. If we wanted to be safe around avoiding unexpected knock-ons, we could wrap the new logic included in this PR in a new template action, specifically intended to be targeted by deployments like PFR that use the old patterns, but I doubt that we've been authoring apps in the last year in a way that relies upon `force_reload` behaving how it has done in that time. Perhaps @ChrisMarsh82 can confirm.

## Git Issues

Closes #3155
Closes https://github.com/IDEMSInternational/pfr-app-content/issues/162

## Screenshots/Videos

See working demo of the PFR app below (relevant template: [home_screen_modular](https://docs.google.com/spreadsheets/d/1-WrPl1r3S_KKfJl4S-Uv5zMX12BvjHXXckk37BtA0Z8/edit?gid=3374259#gid=3374259)). Note that, in contrast with the video included on https://github.com/IDEMSInternational/pfr-app-content/issues/162, the home page now updates to reflect the progress state without needing a manual browser reload.

https://github.com/user-attachments/assets/5dfde071-60ef-430d-a079-9b320b1b2f6b

### DIFF
--- a/src/app/shared/components/template/components/tabs/tabs.component.html
+++ b/src/app/shared/components/template/components/tabs/tabs.component.html
@@ -1,21 +1,24 @@
-<mat-tab-group
-  [animationDuration]="animationDuration()"
-  [dynamicHeight]="params().animateHeight"
-  [mat-stretch-tabs]="!params().compactButtons"
-  [preserveContent]="params().preserveContent"
-  [selectedIndex]="params().selectedIndex"
->
-  @for (tabRow of tabRows(); track tabRow.name) {
-    <mat-tab [label]="tabRow.label" [attr.data-rowname]="tabRow.name">
-      @for (
-        childRow of tabRow.childRows | filterDisplayComponent;
-        track trackByRow($index, childRow)
-      ) {
-        <plh-template-component
-          [row]="childRow"
-          [parent]="parentContainerComponentRef"
-        ></plh-template-component>
-      }
-    </mat-tab>
-  }
-</mat-tab-group>
+@if (shouldRenderTabs()) {
+  <mat-tab-group
+    [animationDuration]="animationDuration()"
+    [dynamicHeight]="params().animateHeight"
+    [mat-stretch-tabs]="!params().compactButtons"
+    [preserveContent]="params().preserveContent"
+    [selectedIndex]="params().selectedIndex"
+    [disableRipple]="true"
+  >
+    @for (tabRow of tabRows(); track tabRow.name) {
+      <mat-tab [label]="tabRow.label" [attr.data-rowname]="tabRow.name">
+        @for (
+          childRow of tabRow.childRows | filterDisplayComponent;
+          track trackByRow($index, childRow)
+        ) {
+          <plh-template-component
+            [row]="childRow"
+            [parent]="parentContainerComponentRef"
+          ></plh-template-component>
+        }
+      </mat-tab>
+    }
+  </mat-tab-group>
+}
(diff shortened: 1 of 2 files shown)


chrismclarke (commented):
question(non-blocking)

The pr description suggests the issue was due to the move to onPush change detection. Did you test calling `cdr.markForCheck` instead of fully recreating the state variables? Might be a slightly simpler solution (don't have the right deployment setup to test against upstream issue myself).

If not the addition looks pretty minimal and only impacts the somewhat legacy emit call anyway, so I'd say should be fine to proceed in this way

chrismclarke (approved):
> Yes, that was my first thought too. I tried adding cdr.markForCheck() (and this.cdr.detectChanges()) in various places(/combinations) but it didn't seem to have the desired effect

Ok, in that case I think this solution seems reasonable

PR #3246, opened 2025-12-10
Fix: auth fields not populated

### LINKED ISSUE #3244: [BUG]
Describe the bug
In the Curaçao (CW) deployment, the fields `@fields._auth_user_name` and `@fields._auth_user_given_name` sometimes evaluate to null. This makes it impossible to reliably display the user’s name fetched during Google login.

App version
Curaçao (CW) deployment

To Reproduce

In the CW deployment, log in with Google

(Ideally use an account where the facilitator’s name has never been edited in the app.)

Go to the Profile page where the name can be edited.

Do not change the name.

Observe that the name shown should be the one provided by Google during login, but instead it shows null (in the English version).

Sometimes refreshing the page causes the name to be loaded correctly.

Screenshots:

Image

Expected behavior
When a user logs in with Google, the Profile page should consistently display the name returned from Google auth (via @fields._auth_user_name / @fields._auth_user_given_name), without requiring a page refresh or prior manual editing.

Scope / other deployments
This is likely the same on any other deployment that uses Google login and relies on these fields.

Deployment sheet(s)
The field @fields._auth_user_name is used in the Login sheet here, where it is used to show the user’s name:
https://docs.google.com/spreadsheets/d/1sgEFVcnisQMCpDrT7O2_Rb3ATV38cI1Elv3ATfr_RvE/edit?gid=1064839943#gid=1064839943

It does not work as intended here, on the Profile page:
https://docs.google.com/spreadsheets/d/1HFALelixOUDU9zpgsBozYLgooQytT2f17DVq_F5g-B4/edit?gid=827249884#gid=827249884

jfmcquade:
PR Checklist

## Description

Fixes an issue where protected auth fields (e.g. `@fields._auth_user_name`) would not be populated on initial template render in some cases.

## Dev notes

Previously, the auth service could finish initialising before the Firebase auth state had settled and the app did not wait for the service to initialise anyway, meaning that the associated fields were not necessarily set by the time a template was rendered. This PR:
- Makes `AuthService` a blocking service during app initialisation
- Updates Firebase authentication provider to wait for auth state to be ready before finishing init
  - `authStateReady()` exists on the Firebase web SDK, but not in the Capacitor plugin, so we've effectively added a custom implementation here
  - There's a fallback to an arbitrary timeout. This isn't ideal and shouldn't strictly be necessary, but I thought it was safest in case Firebase has an issue with returning an auth state.

## Git Issues

Closes #3244

## Screenshots/Videos

[debug_auth_fields](https://docs.google.com/spreadsheets/d/16ddtNuXBYmYM879yIJRI1Gb4-adCFxIR00WVMO-ie98/edit?gid=99275306#gid=99275306) template demo showing consistent behaviour. Contrast with video of broken state on https://github.com/IDEMSInternational/open-app-builder/issues/3244#issuecomment-3642220909.

https://github.com/user-attachments/assets/86d9de04-3a35-4446-b4c5-6ed7f906aedd

### Native testing

Since there's no equivalent to refreshing the browser window when running natively on iOS and Android, it's not as easy to test the particular functionality in question (by the time we've navigated to a template that displays the auth fields, the auth service has finished initialising even without these changes. But I can confirm that on both Android and iOS, the automated login functionality seems to still work as expected when relaunching the app after previously being signed in.

Screenshot 2025-12-17 at 13 11 57
Screenshot 2025-12-17 at 14 01 32

### DIFF
--- a/src/app/shared/components/template/components/combo-box/combo-box-modal/combo-box-modal.component.scss
+++ b/src/app/shared/components/template/components/combo-box/combo-box-modal/combo-box-modal.component.scss
@@ -1,7 +1,5 @@
 @use "/src/theme/mixins";
 
-$background-with-value: var(--combo-box-background-with-value, var(--ion-color-primary-100));
-
 :host {
   display: contents;
 }
@@ -60,8 +58,8 @@ $background-with-value: var(--combo-box-background-with-value, var(--ion-color-p
       margin-bottom: 14px;
       min-height: 55px;
       width: 100%;
-      background: var(--ion-color-gray-50);
-      border: 1px solid var(--ion-color-gray-200);
+      background: var(--ion-color-primary-50);
+      border: 1px solid var(--ion-color-primary-200);
       border-radius: var(--ion-border-radius-small);
       @include mixins.flex-centered;
       input {
@@ -73,15 +71,18 @@ $background-with-value: var(--combo-box-background-with-value, var(--ion-color-p
         @include mixins.flex-centered;
         font-weight: var(--font-weight-standard);
         font-size: var(--combo-box-font-size-standard, 18px);
-        color: var(--ion-color-gray-700);
+        color: var(--ion-color-gray-800);
         min-height: 56px;
         border-radius: var(--ion-border-radius-small);
       }
     }
     .checked-radio {
-      background: $background-with-value;
-      border: 1px solid var(--ion-color-primary-400);
-      color: var(--ion-color-primary-700);
+      background: var(--ion-color-primary-100);
+      border: 1.5px solid var(--ion-color-primary-400);
+      label {
+        font-weight: var(--font-weight-normal);
+        color: var(--ion-color-primary-800);
+      }
     }
   }
 }
--- a/src/app/shared/components/template/components/combo-box/combo-box.component.scss
+++ b/src/app/shared/components/template/components/combo-box/combo-box.component.scss
@@ -26,14 +26,18 @@ $border-with-value: var(--combo-box-border-with-value, var(--ion-color-primary-3
     opacity: 0.9;
   }
   &.with-value {
-    background: $background-with-value;
-    color: $selected-value-text;
-    border: 1px solid $border-with-value;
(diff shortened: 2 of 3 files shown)


jfmcquade:
> Bit of an awkward issue, as in most cases we would probably want the auth system to load in the background and templates to just auto-update with variables when signed in, but appreciate that's not possible in the current system.
> 
> So I think adding the blocking init is probably the best solution until we fully migrate to reactive templating, at which point we could revisit updating global auth variables and subscribing from templates.

Indeed, I was a bit rushed making these PRs yesterday but I certainly agree that that would be a better pattern when available.

> Overall I think the pattern seems pretty reasonable, although I found some of the timeout and race condition code a little long in places so have proposed some minor tidying in #3250

Thanks I've merged. I've also implemented a couple of follow-up suggestions from a Gemini review with https://github.com/IDEMSInternational/open-app-builder/pull/3246/changes/0ded99deea1cd0f3384f0ef184c8d0d56b95a1d2.

> I haven't tested on Android (would be good to do so given the issues with native vs JS layer auth), but testing on web I do see the console error as expected if putting setting the timeout value to something small (e.g. 50ms), so I'd say working as intended.

Good point. Unfortunately I am not able to test auth functionality via my local Android Studio emulators, as. i get a Firebase error despite having added the relevant SHA fingerprints to Firebase (I've been in this position for a while and just can't seem to find what's going wrong). Additionally I think we're lacking the ability publish PR previews to Appetize at the moment, as the workflow needs some updates. I've made an issue to resolve, #3251, but I'm not going to be able to work on it soon. If you happen to be working this weekend perhaps you could try and resolve? Otherwise, I think I'd be happy to go ahead with merging this PR, and we can test that the auth functionality works as expected on a production Android build before release. Testing this particular issue on Android would be difficult as the broken behaviour involve re-initialising the app on a template that includes reference to the auth fields, but I think a general test that the auth feature works as expected when logging in and then relaunching the app would be sufficient.

chrismclarke (approved):

PR #3248, opened 2025-12-11
fix: tabs component

jfmcquade:
PR Checklist

## Description

Follow up to #3247, attempting to fix a version of the same issue. Unfortunately this issue has only been observed on the firebase-hosted web preview, not when running locally, so testing is a bit more difficult, but the issue does some to be resolved on the generated PR preview.

## Git Issues

Closes #

## Screenshots/Videos

| Published web preview | PR web preview |
|-|-|
| Screenshot 2025-12-11 at 11 44 12 | Screenshot 2025-12-11 at 12 08 08 |


jfmcquade:
> @jfmcquade - just to quickly check - is this issue also related to the authState init issue (#3246) or somewhat independent? I'm not really aware of any bugs in mat-tabs that would mean they can't just be rendered, so a bit cautious this might be masking a different issue and making the code slightly less maintainable.
> 
> I'd also suggest changing the changeDetectionStrategy to `OnPush` and seeing if that helps to resolve lifecycle issues. If not let me know and I can try to figure out exactly what's going on

We were a bit rushed yesterday trying to fix these bugs so i didn't put enough detail into these PR descriptions as I would have liked. This issue is unrelated to #3246 as far as I'm aware.

This bug is a bit of a weird one, partly because we could only replicate it on the Firebase-hosted web preview and not locally. My interpretation of the errors we were seeing was that the Angular Material tabs package was trying to run `setupTriggerEvents()` before the necessary DOM elements were created – potentially demonstrating a lack of support for dynamically creating these child tab components programmatically. So this fix ensures that `MatTabGroup` doesn't attempt to render until after Angular's initial render cycle completes, and the DOM is populated with our templated rows. Let me know if you think it would be worth experimenting with addressing the issue via changing changeDetectionStrategy to `OnPush`, or else adding additional documentation comments to better explain the changes.

525318451-8d1f2c78-f86c-4ebd-9544-05470da930bf

chrismclarke:
> This bug is a bit of a weird one, partly because we could only replicate it on the Firebase-hosted web preview and not locally. 

The production build process and environment is a bit different than local development, so not impossible to have production-only issues (were a lot more common in Angular a few years back but tends to be pretty rare these days). My main hesitation is that if this were a core issue then there would be an issue raised on github given how popular the package is and the fact we're still using a version that's been around for quite some time.

That being said, we are in a slightly unique position where we have Angular 18 installed but are using Material 19 (can't remember the full reason why). So the temp workaround is fine, but I would definitely make explicit that this is a temporary workaround and try to encapsulate as much of the logic as possible either within a single method or naming prefix, with comments that link back to this issue and that it should be reviewed following planned major update of Angular and Material.

> My interpretation of the errors we were seeing was that the Angular Material tabs package was trying to run `setupTriggerEvents()` before the necessary DOM elements were created – potentially demonstrating a lack of support for dynamically creating these child tab components programmatically. So this fix ensures that `MatTabGroup` doesn't attempt to render until after Angular's initial render cycle completes, and the DOM is populated with our templated rows. Let me know if you think it would be worth experimenting with addressing the issue via changing changeDetectionStrategy to `OnPush`, or else adding additional documentation comments to better explain the changes.

Given our general move towards signals, all new components should use OnPush change detection. I'm not sure if you create components using cli/tooling or manually, but if possible we should either be configuring the tooling or adding lint rules to try and force (or at least make explicit that a component is not using onPush change detection, for easier switchover later).

So yes, try that first and see if the workaround still required. You can test a production environment locally either running `yarn build` and then serving from the build folder using a tool like `npx serve`, or depending on cli version might support
```sh
ng serve --configuration production
```

PR #3253, opened 2025-12-15
Chore: refactor button component

jfmcquade:
PR Checklist

## Description

- Refactors button component to use zod-based parameter handling
- Adds a reusable `allowedValuesList` coerce method for params
- Minor tidying of button component logic
  - E.g. removing unnecessary aliased params. The logic of the component could likely be simplified further, but I'm wary of making changes that could have unintended knock-ons for such a fundamental component
- Refactors button component template to use modern Angular control flow syntax and signal values of the row data

## Git Issues

Closes #

## Screenshots/Videos

[comp_button](https://docs.google.com/spreadsheets/d/1OmgZICjM5EMT1KgLOU_ovDljRF_SPlpQAgkKWPrNX0s/edit?gid=569531329#gid=569531329) should be functionally unchanged


src/app/shared/components/template/utils/parameter-list.utils.ts
```diff
@@ -68,6 +68,28 @@ const coerceMethods = {
...
+      .transform((v) => {
+        // Normalize to array of strings
+        const parts = Array.isArray(v) ? v : v.includes(",") ? v.split(",") : v.split(/\s+/);
+
+        // Trim, filter empty, and join with space
+        return (
+          parts
+            .map((s) => s.trim())
+            .filter((s) => s.length > 0)
+            .join(" ") || fallback
```
chrismclarke: nit(non-blocking): will the fallback ever be reached? I'm guessing the `parts.map` will return an empty array if empty string (and will have parse error if undefined), so don't think the fallback will be reached

We could consider using the fallback if the array length is zero

chrismclarke (changes requested):
Definitely appreciate the overall reduction in boilerplate code and explicit conversion of inline string list params. See a handful of comments for tidying up

src/app/shared/components/template/utils/parameter-list.utils.ts
jfmcquade: Tidied up in https://github.com/IDEMSInternational/open-app-builder/pull/3253/changes/12fc2140bf361167ac8ad2999b10b7107ea01ace (now outdated)

PR #3257, opened 2025-12-15
chore: tabs component bug workaround documentation

### LINKED ISSUE #3256: [BUG] Error thrown by tabs component
**Describe the bug**
When rendering the tabs component, an error is thrown in the console that can block navigation from functioning correctly from then on.

**App version**
v0.24.1

**To Reproduce**

This bug has been addressed with a workaround by #3248 so is not currently present on `master`.

The branch `spike/tabs-fix` can be used to reproduce the bug.

This bug was first observed on the web preview, and is not present when running the app locally via `yarn start`. To simulate a production build locally, run `ng serve --configuration production`. Then navigate to the [comp_tabs](https://docs.google.com/spreadsheets/d/1raJlddyTZlGxQyv78Nm5PHJxMftisYuCWFWgIVvGxq8/edit?gid=332666066#gid=332666066) template and observe the errors in the console:

```
main.js:1 ERROR TypeError: Cannot read properties of undefined (reading 'addHandler')
  ...
```

As noted in https://github.com/IDEMSInternational/open-app-builder/pull/3248#issuecomment-3647780178, it may be that this bug is due to our mismatched versions of Angular Material/core Angular packages. It would be worth testing whether the issue is resolved on `spike/tabs-fix` (which attempted an alt fix by simply settign the component's change detection strategy to `OnPush`) after we've updated the Angular version in #3234. And if not, then whether it is resolved by the subsequent Angulat (and Angular Material) updates in #3233.

**Debug Sheet(s)**
[comp_tabs](https://docs.google.com/spreadsheets/d/1raJlddyTZlGxQyv78Nm5PHJxMftisYuCWFWgIVvGxq8/edit?gid=332666066#gid=332666066)

**Screenshots and videos**

Image

**More details**

Setting the component's change detection strategy to `OnPush` is not sufficient to resolve the issue

jfmcquade:
PR Checklist

## Description

Follow-up to #3256, implementing a simpler workaround with clear documentation.

## Git Issues

Closes #3256

## Screenshots/Videos

Screenshot 2025-12-16 at 12 50 30

### DIFF
--- a/src/app/shared/components/template/services/instance/template-row.service.ts
+++ b/src/app/shared/components/template/services/instance/template-row.service.ts
@@ -425,4 +425,15 @@ export class TemplateRowService extends SyncServiceBase {
     log_group = SHOW_DEBUG_LOGS ? console.group : () => null;
     log_groupEnd = SHOW_DEBUG_LOGS ? console.groupEnd : () => null;
   }
+
+  /**
+   * Clear the rendered state to force component destruction and enable a clean rebuild.
+   * Used for force reload scenarios to simulate a full template re-initialization.
+   */
+  public clearRenderedState() {
+    this.templateRowMap = {};
+    this.templateRowMapValues = {};
+    // clear rendered rows to force change detection and component destruction
+    this.renderedRows.set([]);
+  }
 }
--- a/src/app/shared/components/template/template-container.component.ts
+++ b/src/app/shared/components/template/template-container.component.ts
@@ -140,7 +140,9 @@ export class TemplateContainerComponent implements OnInit, OnDestroy {
     if (shouldProcess) {
       if (full) {
         console.log("[Force Reload]", this.name);
-        // allow time for other pending ops to finish
+        // Tear down service state to simulate full reload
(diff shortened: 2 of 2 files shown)


chrismclarke:
> > As mentioned in #3256 - I think just disabling the ripple should be enough to workaround the issue (but workaround notes and link to source issue still super useful)
> 
> You're right! I had tried to test this last week and thought I had confirmed a negative result, but it was before you'd showed me how to test the production build locally so perhaps I was interpreting an outdated PR branch web preview or similar.

Perfect, yeah can always be hard to keep track of what actually fixed the issue when trying multiple fixes, but having proper debugging setup and access to stack traces helps a bunch

> In any case, this seems to have resolved the issue, thanks (confirmed on PR preview branch too). I've updated this PR with the much simpler workaround, and have documented the workaround in a comment that references the relevant issues. For review, you can just look at the latest changes from all commits.

Perfect thanks, yeah looks a lot more minimal now, thanks for the update

chrismclarke (approved):

PR #3261, opened 2025-12-18
!chore: native assets generation workflows - refactor Android, add iOS

### LINKED ISSUE #2072: [FEATURE] Migrate package used for generating Android assets cordova-res->@capacitor/assets

jfmcquade:
## PR Checklist

## Description

Refactors native asset (app launcher icon and splash/launch screen) generation to use [@capacitor/assets](https://github.com/ionic-team/capacitor-assets) instead of the legacy [cordova-res](https://www.npmjs.com/package/cordova-res) package.

**Canonical approach (Android & iOS):** One logo image plus one background colour. The logo is centred on the background to generate app icon and splash/launch assets on both platforms. Config: `logo_asset_path` and optional `logo_background_color` under `android` and/or `ios`; if the colour is omitted, white is used and a warning is logged.

**Legacy support:** Deployments that still use the old asset-based config (`splash_asset_path`, `icon_asset_path`, etc.) continue to work: the Android workflow uses the new generator when `logo_asset_path` is set, and falls back to the previous flow otherwise. The four legacy properties are deprecated but supported.

**iOS:** Full iOS support is added. The `ios generate_assets` workflow generates app icon and launch screen from the same logo + background. It uses `ios` config when present, otherwise falls back to the config supplied in `android` (with a warning). 

**Documentation:** A single **Native assets (Android & iOS)** page at `developers/native-assets.md` replaces the old Android-only doc. Dark mode is not supported in the new flow for now; a note in the doc leaves room for future support.

Included in the changes are updates to the default Android and iOS assets. These should always be overridden on production deployments, but the previous plh-specific ones have been replaced by examples with IDEMS branding.

## Breaking changes

This PR is marked as including breaking changes as the old config pattern is considered deprecated. However the deprecated pattern is still supported.

## Testing

1. **Legacy config:** Deployments that only use the pre-existing asset paths should still produce the same Android assets when running `yarn workflow android` with no config changes.
2. **New logo-based config:**
   - **Android:** `yarn workflow android generate_assets` with `logo_asset_path` (and optional `logo_background_color`) under `android`.
   - **iOS:** `yarn workflow ios generate_assets` with `ios.logo_asset_path` or, if unset, `android.logo_asset_path`.

## Git Issues

Closes #2072
Closes #2364

## Screenshots/Videos

Updated docs (single **Native assets** page for Android & iOS):

127 0 0 1_8000_developers_native-assets_

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -6,16 +6,11 @@ name: Docker
 # documentation.
 
 on:
-  schedule:
-    - cron: '38 11 * * *'
   push:
     branches: 
       - "master"
-      - "staging"
     # Publish semver tags as releases.
     tags: [ 'v*.*.*' ]
-  pull_request:
-    branches: [ "master" ]
 
 env:
   # Use docker.io for Docker Hub if empty
@@ -42,7 +37,6 @@ jobs:
       # Install the cosign tool except on PR
       # https://github.com/sigstore/cosign-installer
       - name: Install cosign
-        if: github.event_name != 'pull_request'
         uses: sigstore/cosign-installer@59acb6260d9c0ba8f4a2f9d9b48431a222b68e20 #v3.5.0
         with:
           cosign-release: 'v2.2.4'
@@ -56,7 +50,6 @@ jobs:
       # Login against a Docker registry except on PR
       # https://github.com/docker/login-action
       - name: Log into registry ${{ env.REGISTRY }}
-        if: github.event_name != 'pull_request'
         uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d # v3.0.0
         with:
           registry: ${{ env.REGISTRY }}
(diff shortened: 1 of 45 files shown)


chrismclarke (approved):
Thank for the updates @jfmcquade - the changes and config feel much simpler to me, which I think is a good thing (both for consistency, and to have a clean base for richer functionality in the future as required)

The new docs are also a good addition, thanks for adding (see minor comment in-line).

I can confirm on Android I was able to use the new deployment config to correctly set a custom android splash and icon from assest icon with background color. Not tested on IOS.

```ts
config.android = {
  app_id: 'international.idems.debug_app',
  app_name: 'Debug App',
  logo_asset_path:'./app_data/assets/images/module-1.png',
  logo_background_color:'yellow',
  zoom_enabled: true
}
```

Image

Image

Image

PR #3285, opened 2026-01-13
refactor: slider component

chrismclarke:
PR Checklist

## Description

As part of the angular core updates (#3275), the slider component imports a legacy ng2-slider package which appears to be abandoned. It officially only supports angular 14, and whilst the current code is still compatible with modern angular it would still be preferable to deprecate and remove.

This PR takes a simple copy-pasted of the code used by ng2-slider so that we can drop the npm dependency and modify locally as required moving forwards. It also converts the slider component to standalone to make it easier to import the base nouislider component

## Dev Notes
I'm guessing long term we will probably want to drop the nouislider dep (which ng2-slider wraps), possibly just using the slider provided by angular-material or ion-range. But for now this felt like the simplest solution to deal with the outdated dep. 

## Review Notes
Have given a quick look at the comp-slider template which seems to appear fine to me, but would be good to get a second look

## Git Issues

Closes #

## Screenshots/Videos

image


chrismclarke:
> The changes generally look good to me, I've added #3290 as a follow-up issue to refactor further.
> 
> However, I noticed when testing the component demo template that the "help tooltip" example is broken: the tooltip button is not visible
> This raises questions more generally, as I don't understand why ngIf isn't working. It's considered deprecated, but isn't planned to be removed until Angular v22 (see [docs](https://angular.dev/api/common/NgIf)). What's more, the timer component implements displaying a tooltip icon in exactly the same way, but this does work as expected with `*ngIf`.

> But I don't think that should block the current PR, so I will approve.

Thanks for catching this. It will because I converted to standalone component but did not import the `CommonModule` that angular uses to include specific ngIf/ngFor directive bindings (whereas @if and @for are handled by compiler instead). I'm slightly surprised this wasn't detected by the angular language service or during build, but I guess can be difficult to track when imported/exported across various workspaces and modules.

At some point it would probably be worth just running the automatic migration to convert all legacy ngIf/ngFor to modern syntax - it's highly unlikely we'll manually do this over time so maybe worth just ripping the bandaid off whilst doing other core upgrades
https://angular.dev/reference/migrations/control-flow

There is a similar migration to move everything to standalone components (auto-importing common module parts), although I think I'm slightly less convinced this will work given the dependence on non-angular imports in many of the components (first we would likely want to prioritise removing ngModule imports for 3rd party modules, e.g. `NgxExtendedPdfViewerModule` would need to be manually migrated to the pdf component that uses it - not a bad thing to do just not something auto migrate would likely pick up on)

PR #3288, opened 2026-01-14
Feat: pdf viewer fullscreen option

### LINKED ISSUE #3284: Full Screen option for pdf-viewer
**What?**
The pdf-viewer could have the option to visualize the content in full screen mode.

**Why?**
The Full Screen option functionality would solve [this ](https://github.com/ParentingForLifelongHealth/plh-facilitator-app-ph-content/issues/230) CW issue.
Especially on mobile phones, the combination of the margins of the pdf documents and the template margins can make the actual text small and difficult to read, moreover a full screen visualization makes the pdf viewer more usable.

Image

**How?**
On the side of the current + and - buttons we can add a button for full screen.

jfmcquade:
PR Checklist

## Description

WIP 

Adds an optional button to the pdf viewer component toolbar to enter fullscreen "presentation mode". The visibility of this button can be toggled for a given `pdf` component instance via the new param `allow_fullscreen`.

#### Exiting fullscreen on mobile

The presentation mode is optimised for desktop browsers, where the mode can be exited via the `esc` key. On Mobile, we rely on the native functionality for exiting fullscreen mode.

##### iOS
On iOS, the OS overlays a close button:

Screenshot 2026-01-14 at 15 57 03

##### Android
On Android, there is no UI element to close the presentation. 

Screenshot 2026-01-14 at 16 04 33

Testing on both emulator and physical device, the only way out seems to be via the native device "back" functionality, however this navigates back within the app as well as exiting fullscreen. We could consider trying to add our own UI element to exit fullscreen, however this is not trivial, as there seems to be no way to do so programmatically.

#### Zoom functionality

Zooming is not possible when in fullscreen mode. This seems to be a hard limit of the package we are using, so cannot be overridden unless we switched to a different one.

From [extended-pdf-viewer docs](https://pdfviewer.net/extended-pdf-viewer/presentations):
> The full-screen mode of pdf.js (and, by extension, ngx-extended-pdf-viewer) is optimized for presentations. There's little else you can do with it: you can't scroll, nor can you zoom in or out.

## Git Issues

Closes #3284

## Screenshots/Videos

[comp_pdf](https://docs.google.com/spreadsheets/d/1VXM9zYgrsZIB4h8slC9mg3P9C-jsXuPScNaml-1qaOk/edit?gid=569531329#gid=569531329)

Screenshot 2026-01-14 at 16 29 10

### DIFF
--- a/src/app/shared/components/template/components/pdf/pdf.component.ts
+++ b/src/app/shared/components/template/components/pdf/pdf.component.ts
@@ -13,8 +13,8 @@ const AuthorSchema = defineAuthorParameterSchema((coerce) => ({
   ),
   /** Text for the "Open with..." button. Default "Open with...". */
   open_external_text: coerce.string("Open with..."),
-  /** Show full screen button. Default true. */
-  allow_fullscreen: coerce.boolean(true),
+  /** Show full screen button. Default false. */
+  allow_fullscreen: coerce.boolean(false),
 }));
 
 @Component({


chrismclarke:
@jfmcquade - I think you'll likely need a custom solution as the pdf.js fullscreen mode is quite limited (as you mentioned, prevents scroll and zoom which renders the functionality pretty useless on mobile). I think you could possibly approach in two ways:

1. Use the [Fullscreen API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API) to put the pdf element in fullscreen mode

I haven't tested myself but it's a pretty common use-case, and I think should have good support on web and android (tbc ios). I see the Capacitor team has done some work to get the functionality working in `https://github.com/ionic-team/capacitor/pull/7909`. There might be some extra work required to handle safe area insets, back button press interception and system bars (looks like Capacitor has worked on this). May also need to check what version of Capacitor we are on and when the changes landed

2. Use CSS

You could try using absolute positioning to force the component to render on top of other content. Not sure how well this will work in deeplly nested templates, and in this case would definitely need to add manual workarounds for safe area insets, back button press and system bars. So I'd say this should be less preferable than the fullscreen api if working on all platforms, and could just be a fallback option on specific platforms where not working.

PR #3293, opened 2026-01-16
Chore: refactor to Angular control flow syntax

### LINKED ISSUE #3291: [CHORE] Refactor Angular templates to use modern control flow syntax
**What?**
The Angular directives [`*ngIf`](https://angular.dev/api/common/NgIf), [`*ngFor`](https://angular.dev/api/common/NgFor) etc. are deprecated in favour of the [control flow](https://angular.dev/guide/templates/control-flow) syntax. All uses of the deprecated patterns should be refactored to use the modern syntax.

**Why?**
Whilst the directives aren't due to be removed until Angular v22, some inconsistent behaviour was identified with the existing syntax (https://github.com/IDEMSInternational/open-app-builder/pull/3285#issuecomment-3755437701), which would motivate refactoring all components. This is something we'd have to do at some point anyway, in order to support Anfular v22.

**How?**

Search the codebase for examples of each deprecated directive and refactor to use the new equivalent. AI should be helpful.

| Old directive         | New syntax |
|-----------------------|------------|
| `*ngIf`               | `@if`      |
| `*ngFor`              | `@for`     |
| `*ngSwitch`           | `@switch`  |
| `*ngSwitchCase`       | `@case`    |
| `*ngSwitchDefault`    | `@default` |

jfmcquade:
PR Checklist

## Description

Refactors all cases of the deprecated Angular directives [*ngIf](https://angular.dev/api/common/NgIf), [*ngFor](https://angular.dev/api/common/NgFor) etc. in favour of the [control flow](https://angular.dev/guide/templates/control-flow) syntax.

## Dev notes

The vast majority of changes were made by running [Angular's migration schematic](https://angular.dev/reference/migrations/control-flow):
```sh
ng generate @angular/core:control-flow
```

Manually searching for remaining uses revealed one file that had been missed (https://github.com/IDEMSInternational/open-app-builder/pull/3293/changes/595da388acbf718171e904af7e80929d582d1d42), plus some references to the old syntax in comments (https://github.com/IDEMSInternational/open-app-builder/pull/3293/changes/1a0c2d60101e7c3187edcb9cd1d2eee7115856d1)

## Git Issues

Closes #3291

## Screenshots/Videos

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -95,7 +95,7 @@ module.exports = tseslint.config(
     extends: [...angular.configs.templateRecommended, ...angular.configs.templateAccessibility],
     rules: {
       "@angular-eslint/template/eqeqeq": "warn",
-      "@angular-eslint/template/prefer-control-flow": "off",
+      "@angular-eslint/template/prefer-control-flow": "error",
       "@angular-eslint/template/click-events-have-key-events": "warn",
       "@angular-eslint/template/interactive-supports-focus": "warn",
       "@angular-eslint/template/alt-text": "warn",


chrismclarke (approved):
Thanks for updating and giving a manual check. I've pushed a6d3ab30197dad2420b6b170e2baa1f854eb24ea to enable the lint rule to prevent use in the future but otherwise think this should be good to go.

Image

I'll also update the latest screenshot tests #3292 in case it picks up any issues

I can see gemini review has picked up a few places where we could remove divs due to control syntax not requiring parent div to sit on like ngIf, although I think to avoid any possible layout knock-ons I'll suggest leaving as-is (better to tidy up when manually migrating component to standalone or similar)

There are also a handful of cases where we've used ngIf for inverse conditions (e.g. `ngIf="condition"`...`ngIf="!condition"`), which could be refactored to `@if` and `@else` - but I'm probably more keen to get this merged in and avoid too many conflicts on follow-up work than scrutinizing too closely

PR #3300, opened 2026-01-23
Feat: migration system

chrismclarke:
PR Checklist

## Description
Add a lightweight migration service used for tracking one-time app data migration tasks

## Review Notes
As there are no migrations currently specified, functional review will have to wait for follow-up pr on dynamic data persistence. Tests have been added and a markdown file to provide overview to the service however, for review and feedback.

## Developer Notes
This is a precursor for proposed data persistence changes, so that existing data stored in indexeddb can be populated to json file. A dedicated service is used for this to simplify the logic held in the dynamic data service, and to allow for more legacy code cleaning in the future.

The PR opts for a decentralised approach to migrations (where each service can register their own migrations) instead of a single source of all app migrations to help avoid circular dependencies, and to avoid cases where migrations only apply to apps that have specific feature modules available (migration would break on builds where module omitted).

I had considered trying to implement a system where each migration tracks either timestamp or app version code of when created, and using that to determine which migrations can/can't be skipped for new users compared to legacy, however the tracking logic became overly complex and instead now rely on optional per-migration `preconditions` to decide whether a given migration needs to be run or not. 

## Git Issues

A much older draft for a similar service was created in 2023, but can now be closed in favour of this pr.
Closes #1814

## Screenshots/Videos


jfmcquade (approved):
Looks good to me.

I can see how keeping the feature in this general form and able to run arbitrary migration logic could make it useful in a wide variety of contexts. As documented in the readme, it will be important to keep the number and complexity of these migrations as minimal as possible for performance reasons. Another motivation for this might be to avoid having to keep track of various migrations and possible states of user apps (although I appreciate with idempotency we shouldn't have to be aware of those states explicitly).

PR #3304, opened 2026-01-25
Feat/nav bar persist highlight alt

chrismclarke:
PR Checklist

## Description
Alt implementation to #3296 
Makes template section persist default (and only) behaviour

## Dev Notes
I don't see much downside to making the section highlight default (and only) behaviour - if opening a standalone template the nav bar will likely get covered anyway, and not aware of many other cases where user is expecting navigation to somewhere entirely outside of all sections. If that were the case we could always add an action to remove highlight.

## Git Issues

Closes #

## Screenshots/Videos

(screen recording extension currently broken)

### DIFF
--- a/src/app/shared/components/template/components/pdf/pdf.component.html
+++ b/src/app/shared/components/template/components/pdf/pdf.component.html
@@ -8,7 +8,7 @@
         [showBorders]="false"
         [mobileFriendlyZoom]="'1.3'"
         [height]="'80%'"
-        [minHeight]="'400px'"
+        [minHeight]="'540px'"
         [textLayer]="false"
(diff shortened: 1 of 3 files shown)


MichelePancera (approved):

jfmcquade:
> When I refresh the highlight dissapears. I wonder if this could lead to more real bugs.

This is good to know. I can't recreate on Chrome on Mac, see video below, so I guess it is platform dependent.

In any case, I think we're agreed that this can be addressed as a follow-up issue, I've opened #3340 to track.

https://github.com/user-attachments/assets/4169d2f4-d92e-4d4b-a59c-eff4080d3470

PR #3320, opened 2026-02-04
Feat: pdf component toolbar buttons - open_external and download

jfmcquade:
PR Checklist

## Description

Adds two new optional buttons to the toolbar of the pdf component: for opening the file in a chosen native reader; and for downloading the file to the device. Both of these buttons will not be enabled by default, to preserve existing functionality, but can be enabled through the following params:
- `show_open_external_button`
- `show_download_button`

Additionally, increases the min height of the pdf viewer component, and adds a bottom margin.

## TODO

Consider exposing the icons to authoring. This would allow, for example, an instance of the pdf viewer to use a "fullscreen" icon for the `open_external` button.

## Git Issues

Related to #3284 and https://github.com/ParentingForLifelongHealth/plh-facilitator-app-ph-content/issues/230. Whilst it doesn't address the specific requests of these issues, it addresses the underlying problem by providing an option to enable users to view the PDF in a more convenient way (open external). If desirable, we could add additional customisability options to specify a custom icon for the two new buttons.

## Screenshots/Videos

[comp_pdf](https://docs.google.com/spreadsheets/d/1VXM9zYgrsZIB4h8slC9mg3P9C-jsXuPScNaml-1qaOk/edit?gid=569531329#gid=569531329):

Screenshot 2026-02-05 at 11 32 05

Screenshot 2026-02-05 at 11 31 40


src/app/shared/components/template/components/pdf/pdf.component.html
```diff
@@ -35,5 +35,17 @@
       <pdf-zoom-out id="zoomOut"></pdf-zoom-out>
       <pdf-zoom-in id="zoomIn"></pdf-zoom-in>
     </div>
+    <div [class.invisible]="false" id="toolbarViewerRight">
```
chrismclarke: nit(non-blocking): I'm not sure I understand this class here - I don't see in the scss and surely it would be omitted by default anyway?

MichelePancera (approved):

PR #3328, opened 2026-02-12
fix(auth): force_reprocess sees updated values after sign-in/sign-out

### LINKED ISSUE #3329: [BUG] Auth fields do not update automatically on sign-in/-out
**Describe the bug**
When signing in or out via any of the methods on [example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705), the page does not update to show the latest values of the auth related fields.

**App version**
This is a new bug present in v0.26.0, i.e. post the upgrade to Angular v21

**Debug Sheet(s)**
[example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705)

**Screenshots and videos**

https://github.com/user-attachments/assets/7075b264-65c6-4057-856c-5a0b406e3b28

**Additional context**
Reloading the page causes the rows to display the expected updated values. 

UPDATE:

Triggering an additional `force_reprocess` works – [example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705) updated with a button for triggering this manually. So the issue is with `force_reprocess` when it appears in the action list directly after the sign-in/-out action. Likely a race condition issue

Image

jfmcquade:
PR Checklist

## Description

Fixes an issue with a race related to auth fields so `force_reprocess` now sees updated values when called immediately after sign-in/sign-out.

### Explanation of issue

There was a race between the auth action and the effect that writes auth fields to storage. For flows like:
- `click | auth: sign_in_google; click | emit: force_reprocess`
- Or the Google/Apple sign-in button (which then runs the row’s click actions)

the next action in the list (e.g. `force_reprocess`) could run before the `AuthService` effect had run. That effect reacts to `provider.authUser()` and calls `addStorageEntry` / `clearUserData` to update the storage that `@field`/`@fields._auth_user_id` read from. Because effects run in a later Angular phase, the first `force_reprocess` apparently saw stale or empty auth fields. A second reprocess would then show the correct values.

This became consistent after the Angular v18→v21 upgrade. It's not completely clear to me why this is, but we can probably safely assume it is due to changes in effect scheduling relative to the action queue.

### Solution

- AuthService wrappers
  `signIn(providerId)`, `signOut()`, and `deleteAccount()` now wrap the provider and call `syncStorageToAuthState()` before returning. So any following action in the same turn (e.g. `emit: force_reprocess`) sees up-to-date auth fields in storage.

- Single entry point for auth
  All auth flows now go through these methods: template actions (`auth: sign_in_google`, etc.), the Google/Apple sign-in button components, and debug pages. No direct `provider.signIn` / `provider.signOut` / `provider.deleteAccount` calls from the app.

### Result

The first `force_reprocess` (or any immediate follow-up action) after sign-in or sign-out now sees the correct `@fields._auth_user_id` (and other auth fields) without needing a second reprocess or a full reload.

### Additional considerations

I've updated the auth logic specifically, but it's possible that we have other cases of similar race conditions between actions and signals in the app. However, none come to mind. In any case, I don't think it's worth seeking amore general solution at this moment in time – issues like this are precisely what the new reactive templating system is designed to avoid.

## Git Issues

Closes #3329

## Screenshots/Videos

[example_auth](https://docs.google.com/spreadsheets/d/12xo0ewDN0bbeVrs3klWd9J8DkFQLTCYX1zhGWnovqIo/edit?gid=579616705#gid=579616705) working as expected (contrast with recording on linked issue)

https://github.com/user-attachments/assets/a446049f-7d44-4a20-a91a-3b897c4856d0


src/app/shared/services/auth/auth.service.ts
```diff
@@ -123,10 +144,10 @@ export class AuthService extends AsyncServiceBase {
...
         const [actionId] = args;
         const childActions = {
-          sign_in_google: async () => await this.provider.signIn("google.com"),
-          sign_in_apple: async () => await this.provider.signIn("apple.com"),
-          sign_out: async () => await this.provider.signOut(),
-          delete_account: async () => await this.provider.deleteAccount(),
+          sign_in_google: () => this.signIn("google.com"),
+          sign_in_apple: () => this.signIn("apple.com"),
+          sign_out: () => this.signOut(),
+          delete_account: () => this.deleteAccount(),
```
chrismclarke: nit(non-blocking): I'm not sure what we really gain from losing the async/await here. I'd be inclined to revert back to the more explicit notation. 

If you want more concise list of child actions you could define the consts elsewhere, e.g.
```ts
const sign_in_google = async ()=> this.signIn("google.com")
const sign_in_apple = async () => this.signIn("apple.com")
// ...
const childActions = {sign_in_google, sign_in_apple  ...}
```
You could also keep the actions in a separate file like some others (using a factory pattern to automatically keep reference to the service that creates)... but I'd say both that and the `const` form is more down to personal preference, and not really necessary when the action logic is so minimal

chrismclarke: great, yeah that feels easier to follow

MichelePancera (approved):

PR #3350, opened 2026-02-26
perf: angular dev server

chrismclarke:
PR Checklist

## TODO
- [x] Add scripts to help benchmark performance changes
- [x] Identify and prioritise potential improvements 
- [x] Implement vite-based angular builder
- [x] Fix app breaking changes
- [x] Implement dev/prod browserslist separation
- [x] Optimise tsconfig
- [x] Fix script breaking changes

## Description
Various changes to try and improve dev server bundling time

## Review Notes
Install updated deps
```bash
yarn install
```
Serve the application
```bash
yarn start
```
You should notice initial startup time roughly half of what it previously was. You should also notice when editing `.html` and `.scss` files, that the first file edit will be slow to recompile (on par with initial startup), but subsequent edits much faster (dynamically loaded into page without refresh). Changes to `.ts` files will still reload.

There is also a benchmarking script that can be run via
```bash
node dev-benchmark.mjs
```
Although if wanting to test on master the script would need to be manually copied over and run there. Script outputs to root folder as `benchmark-[timestamp].log`. Example screenshots from script included below 

## Dev Notes
**Benchmark Scripts**
To help understand what does/doesn't a difference I generated a rough script to try and time how long the `ng serve` command takes to run both with and without angular cache present, and how long it takes to recompile after file changes. I'd say probably worth keeping the benchmark script in the repo for now in case we want to test any further optimisation down the road.

**New Angular Builder**
As mentioned on the call, angular released a [new build system](https://angular.dev/tools/cli/build-system-migration) built on top of vite/esbuild that is significantly faster than the legacy webpack-based one. I've migrated the code to use this and have noticed roughly 50-60% speed improvements for initial bundling.

**Breaking Changes**
Vite handles monorepos and linked modules quite differently to webpack. We had a lot of hacks and workarounds to allow relevant code to be imported and read everywhere it was needed, which immediately broke within the new build system. As such I had to do a bit of tidying and refactoring, mostly focused on dropping use of yarn workspace modules to import code into the main app (more reliable with tsconfig paths), and converting packages that are designed for use in the frontend to more simple structures (removing package.json). 

More work could probably be done in the future to reduce some of the circular imports and structures, but for now things seem to at least be holding together and a few of the more confusing abstractions have been removed.

I also found the `character-parser` import used in scripts seemed to break and couldn't really figure out why, so opted to just directly inline the methods and drop the package altogether which seemed to fix things.

There is also a console warning :

>The above dynamic import cannot be analyzed by Vite

This is a known issue with ionic pwa-elements, and will hopefully be resolved in a future update (tried most recent version but still present). It doesn't break the build however. `https://github.com/ionic-team/pwa-elements/issues/129`

**No Difference**
A few other things I investigated but found they made no difference:

 _browserslist_
This isn't used by vite/esbuild at runtime (code is shipped to browser directly). I did make some minor tidying improvements to align the ios version to the same as capacitor (and supress angular warning), and removed some of the legacy configuration that isn't relevant to apps or web previews we produce.

_tsconfig incremental build_
I tried setting `"incremental":true` in tsconfig following guide [here](https://freedium-mirror.cfd/https:/medium.com/codetodeploy/typescript-performance-tips-that-speed-up-angular-builds-2adfb9a6f2bf) but quickly realised the new angular build system has it's own tsconfig overrides build in which already include the setting and populate the corresponding `.tsbuildinfo` file within the `.angular/cache` directory.

I also marked `"isolatedModules": true` following the same guidance, but not sure if it actually takes effect or not (didn't appear to have noticable difference in benchmark, but decided to keep enabled in any case).

**Other Possible Fixes**
I saw noted in a few places recommendation to configure antivirus to not run in app folders. This is something that I think vscode has previously prompted directly so not something we can really configure at the repo level, but may be worth considering a brief note on how to handle on platforms if not the case. E.g. on windows
```ps1
# Run as Administrator - windows
Add-MpExclusion -Path "C:\path\to\project"
Add-MpExclusion -Process "node.exe"
```

## Additional Knock-ons
- Introduce namespace `@components` to import directly from components workspace as a lib, instead of relying on yarn workspaces to pseudo-load via node modules. In turn this helps with some cross-compatibility issues between esm and commonjs modules. In the future it would probably be good to separate `packages` (have external deps and run standalone) from `libs` (shared code to import into packages). 

Similarly I've refactored the `workflows` package to function as a lib because it is ultimately managed by the scripts package that calls it. Again, this is a slightly confusing pattern, but arguably better structured as a lib as the workflows are instruction files that the scripts import, instead of their own standalone scripts.

- Migrate scripts cli runner from `ts-node` to `tsx` which generally has better support for different module types. This had a few minor knock-ons around imports and filepaths which have been resolved.

- Various issues correctly resolving and stubbing code within test suites, resolved by dropping external fs-extra dep in shared (would conflict with stubbed script fs-extra instance), and providing alt path-based import for shared lib in scripts

- Remove legacy code from shared file-utils

- Refactor scripts deployment config stub to correctly apply to all test runs (not entirely sure why previous pattern stopped working, but should be slightly cleaner overall now)

- Migrate eslint config to v9 compatible

## Future Follow-up

**TS Compilation**
During first start and any changes to `.ts` files, angular has to re-process the large dependency tree. This is quite time consuming given the scale and many layers of imports/co-dependencies. Could consider ideas such as:

 Could specifically consider:

- [ ] Reduce the number of `index.ts` barrel files that export in batch, making it easier for angular to cache and separate bundles
- [ ] Tidy up larger common packages such as `shared` and `data-models`, particularly if either call any 3rd party modules. `shared` top-level export is now gone, but still large exports like `shared/utils`.
- [ ] Continue to refactor as many components to standalone as possible (easier for angular to track)
- [ ] Try to break up largest ngModules (would happen alongside standalone component migration)
- [ ] Enforce type-only import (`import type {...} from ...`) for types (omits module compilation)

Longer term suggestions:
- [ ] Consider refactoring larger components like odk and pdf viewer to webcomponents and build externally
- [ ] Reduce external 3rd party dependencies where possible (inlining more as required)
- [ ] Add [NX](https://nx.dev) for more fine-grained caching of build and serve operations
- [ ] Consider including angular build cache within CI pipelines (ci build speed improvements, not local dev)
- [ ] Consider migrating from yarn to bun (would likely reduce tool calling times like prepare script, but minimal ongoing improvements)

Non-recommendations
- [ ] Bypass angular type checking (can disable and rely solely on vscode/eslint, but personally find console and brower errors quite useful in dev)

There will also likely be incremental improvements ongoing with new versions of angular and node (have read the angular team are also looking into porting the compiler to go which could help).

**SCSS Compilation**
This also takes considerable time during startup, and first change to any scss file. We have a large number of internal and external style sheets that could likely be cleaned up, as well as various warnings that need to be written to the console for deprecated features. Could specifically consider:

- [ ] Replace `@import` with `@use` where possible
- [ ] Replace deprecated functions, e.g. `saturation(...)`
- [ ] Review 3rd party style imports which may not be required (e.g. some ionic and angular ones)
- [ ] Investigate ways to streamline theming

## Git Issues

Closes #

## Screenshots/Videos
**Benchmark (Baseline)** - time taken on my local machine to serve application without any `.angular/cache` dir (cold start), with the cache (warm start), and following file change (incremental rebuild)
image

**Benchmark (New Build System)** - shows 50-60% reduction for cold and warm starts, similar incremental rebuild 
image

**Character Parser** - imported script seems incompatible with vite bundler, so have instead just inlined methods required
image

**@ionic/pwa-elements** - warning shown, known issue but doesn't break build (could patch package in future)
`https://github.com/ionic-team/pwa-elements/issues/129`
image


ChrisMarsh82 (commented):
Builds fine although didn't seem any quicker. Had a problem when running `yarn workflow sync sheets` on plh-kids-teens-za. This gave the following error. (It worked fine when back on master)

```
TSError: ⨯ Unable to compile TypeScript:
src/commands/compile.ts:4:27 - error TS2307: Cannot find module 'data-models' or its corresponding type declarations.

4 import { FlowTypes } from "data-models";
```

chrismclarke:
(note to self - recent breakages likely due to incoming code that also now needs refactor to change shared namespace. will have similar issues with #3465)

chrismclarke:
> Builds fine although didn't seem any quicker. Had a problem when running `yarn workflow sync sheets` on plh-kids-teens-za. This gave the following error. (It worked fine when back on master)
> 
> ```
> TSError: ⨯ Unable to compile TypeScript:
> src/commands/compile.ts:4:27 - error TS2307: Cannot find module 'data-models' or its corresponding type declarations.
> 
> 4 import { FlowTypes } from "data-models";
> ```

Thanks for checking this. I've been able to recreate and fix the issue, so should now be working. There's still some potential knock-ons for tools we use (e.g. screenshot tests), but any error would be immediately obvious (will just fail to run). I've checked the main app and sync/deployment set scripts are working as expected so hopefully this should be in a better state for merging. Will flag @jfmcquade to verify

markharmon (commented):
I didn't notice a significant performance improvement on the build (particular on a rebuild when changes are detected). It looks like a good change though. I tried out reactive templates and everything is building ok + all reactive template cypress tests are still passing.

markharmon:
> I didn't notice a significant performance improvement on the build (particular on a rebuild when changes are detected). It looks like a good change though. I tried out reactive templates and everything is building ok + all reactive template cypress tests are still pass

actually after just switching back to master this change does seem to have noticeably improved build time

ChrisMarsh82:
@chrismclarke used this for Panama and it worked fine but had problems when I moved to plh_kids_teens_my. Server started but just got a blank screen. 

This was in the console:
```
Watch mode enabled. Watching for file changes...
NOTE: Raw file sizes do not reflect development server per-request transformations.
  ➜  Local:   http://localhost:4200/
  ➜  press h + enter to show help
11:11:41 [vite] (client) warning: 
C:/Source/CM82/parenting-app-ui/.angular/cache/21.2.2/app/vite/deps/chunk-P5VUEMBS.js
1150 |      }
1151 |    }
1152 |    return import("./".concat(a, ".entry.js").concat("")).then((function(e2) {
     |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
1153 |      {
1154 |        cmpModules.set(a, e2);
The above dynamic import cannot be analyzed by Vite.
See https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#limitations for supported dynamic import formats. If this is intended to be left as-is, you can use the /* @vite-ignore */ comment inside the import() call to suppress this warning.

  Plugin: vite:import-analysis
  File: C:/Source/CM82/parenting-app-ui/.angular/cache/21.2.2/app/vite/deps/chunk-P5VUEMBS.js?v=1cf2db4c
```

PR #3351, opened 2026-02-27
chore: docker ci disable

chrismclarke:
PR Checklist

## Description
- Disable docker ci script on prs

## Dev Notes
I noticed this was running on all pushes to pull-requests targeting main branch. This would include any draft or incomplete PRs, which is both a potential security issue (external contributor could push malicious code into image) and cause a lot of issue for anyone trying to use the image if the code is constantly changing (with potentially broken functionality). 

It will also populate a very large number of images to GHCR which might be awkward for decluttering (fortunately GHCR is still currently free for public projects, so no billing implications, although does eat into action minutes quota). 

I'd strongly recommend linking docker images to published releases to allow for more consistent use, and maybe considering a single _nightly_ build which just links to master branch merge (not actually run on a cron basis, but whenever new code merged into master).  

I also wasn't too sure what purpose the cron schedule serves, given it will essentially be rebuilding the same image that would have been deployed following push to master branch, but haven't removed in case I'm missing something (?).

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/documentation/docs/data/user_disaggregation.md
+++ b/documentation/docs/data/user_disaggregation.md
@@ -1,0 +1,34 @@
+# User Disaggregation
+
+In the App Users tables, several variables across the device information and contact fields columns describe the environment in which the app is being accessed. These variables help identify the type of user. The tables below provide a comprehensive overview of each variable, their possible values, and descriptions of what they represent.
+
+## Device Info
+
+| Variable          | Potential Values       | Description                                                                                     |
+|------------------|----------------------|-------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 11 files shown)


chrismclarke:
I also took a quick scan of the repo and don't see a `staging` branch active anywhere, so I'm assuming the original docker file was either generated or adapted from somewhere else that did. 

I've pushed 437f331 to remove the additional staging and pull_request logic

GabeBolton (approved):
Those are good points, I'm convinced. To improve efficiency and not build intermediary or docs pushes, we should move just to:
```yml
on:
  push:
    tags: [ 'v*.*.*' ]
    workflow_dispatch:
```
I'll update our other Docker CI actions in line with this decision.

PR #3366, opened 2026-03-04
feat: allow special chars in ios app name

### LINKED ISSUE #3091: [FEATURE] Handle special characters in app name in iOS automated build
**What?**
Ensure that app names that include special characters (e.g. "+") are supported in the iOS workflows. I.e. such an app name provided as `config.ios.app_name` in the deployment config should be populated as expected when running the `ios configure` workflow – currently the app cannot be built from the results of running this workflow.

**Why?**
There are some special characters that are supported as the user-facing app names of apps on both Android and iOS that currently cannot be passed through via the `config.ios.app_name` property, meaning they must be handled manually. Specifically, the "+" symbol at the end of "Formando Conciencia+" (the name of the `plh_facilitator_mx` deployment app) has caused issues.

**How?**
When building an iOS app, the `ios configure` workflow handles copying over deployment-specific details into their relevant build files. This includes copying the `config.ios.app_name` property into various locations within the file `ios/App/App.xcodeproj/project.pbxproj`. One of the places that the app name gets populated in the `ios/App/App.xcodeproj/project.pbxproj` file is in the `PRODUCT_NAME` variable in `buildSettings`. This variable is then read into `ios/App/App/Info.plist` as the `CFBundleDisplayName` at build time. It is this `CFBundleDisplayName` that ultimately specifies the user-facing name of the app. The `CFBundleDisplayName` can support a "+" symbol (adding it in here manually has been my workaround), but the `PRODUCT_NAME` variable cannot*. Therefore, we may need to handle a string replace in `ios/App/App/Info.plist` directly as part of the `envReplace` in `packages\scripts\src\tasks\providers\ios.ts`. It may also be worth considering storing the user facing app display name in a separate build settings variable that gets passed through to `inf`o.plist`, as it's only the `PRODUCT_NAME` that cannot contain special chars as far as I'm aware.

(*Note – ensure that the `PRODUCT_NAME` variable is indeed the cause of the error, and not some other value that gets string replaced in `ios/App/App.xcodeproj/project.pbxproj`).

jfmcquade:
PR Checklist

## Description

Adds support for user-facing iOS app names that contain special characters. We currently have two such cases in production: `Formando Conciencia+` and `Mayor Konektá`. I have been managing these special cases manually, but the `+` and `á` chars in the app names mean that these could not have been released through the existing workflows and automated actions. This is handled via a new deployment config property:
- `ios.app_display_name`. This value will appear in user-facing contexts, such as the app store listing and app icon label.

The extant `ios.app_name` property remains unchanged and cannot include special characters. This must be provided and will be used for the official app bundle name, for example.

It is worth noting that whilst the `app_display_name` can differ somewhat from the `app_name`, it should closely resemble it, or risk rejection from Apple ([source](https://stackoverflow.com/a/5449517)). 

## Testing

I have updated the debug deployment to use an `ios.app_display_name` with a special char in this PR: https://github.com/IDEMSInternational/app-debug-content/pull/226

See this successful action run to publish a build to appetize, which was run from that content repo PR branch with the repo's `APP_CODE_BRANCH` var temporarily set to refer to the feature branch of the current PR: https://github.com/IDEMSInternational/app-debug-content/actions/runs/22680858967

## Git Issues

Closes #3091

## Screenshots/Videos

### DIFF
--- a/packages/components/plh/plh-kids-kw/components/bottom-navigation-bar/bottom-navigation-bar.component.html
+++ b/packages/components/plh/plh-kids-kw/components/bottom-navigation-bar/bottom-navigation-bar.component.html
@@ -1,23 +1,25 @@
 <div class="plh-bottom-nav-wrapper" [attr.data-variant]="params.variant">
-  @for (button of params.buttonList; track button.name) {
-    <a
-      class="nav-item"
-      routerLinkActive="active-link"
-      [routerLink]="'template/' + button.target_template"
-      #rla="routerLinkActive"
-      [attr.data-name]="button.name"
-      [attr.data-variant]="params.variant"
-      [attr.data-isActive]="rla.isActive"
-      [attr.data-hideInactiveText]="params.hideInactiveText"
-      [attr.data-dim-when-inactive]="!button.active_icon || null"
-    >
-      <img
-        class="icon"
-        [src]="(button.active_icon && rla.isActive ? button.active_icon : button.icon) | plhAsset"
-      />
-      <div class="label" [attr.data-hideInactiveText]="params.hideInactiveText">
-        {{ button.label }}
-      </div>
-    </a>
-  }
+  <div class="nav-items">
+    @for (button of params.buttonList; track button.name) {
+      <a
+        class="nav-item"
+        routerLinkActive="active-link"
+        [routerLink]="'template/' + button.target_template"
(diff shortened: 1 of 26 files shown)


chrismclarke (approved):
Sounds sensible and implementation looks solid to me.

PR #3385, opened 2026-03-17
chore: combo-box code tidy

jfmcquade:
PR Checklist

## Description

Follow-up to #3361. Minor code tidy to remove need for `$any()` and improve var naming.

## Git Issues

Closes #

## Screenshots/Videos


chrismclarke (approved):
Ah yeah, I can see why it could be a challenge to strongly type when authors can define their own lists. I think this is a fair enough compromise for now

PR #3387, opened 2026-03-17
feat(text-box component): add update_on_input param

jfmcquade:
PR Checklist

## Description

Adds a new param to the text-box component: `update_on_input`. If `true`, the component instance will update its value with each keystroke. The default behaviour is unchanged: only to update on blur (deselecting the element).

Since updating the component's value potentially triggers actions, I added debounce logic to avoid excessive calls (see [these commits](https://github.com/IDEMSInternational/open-app-builder/pull/3387/changes/d2b9e09fa1376e7f8a527f8edf47533262235a26..9411db999f361f193d333c07960e3079b6a3cf5f)). I think that this parameter should still only be enabled in cases where changes to the value aren't authored to trigger additional actions.

## Git Issues

Allows authoring resolution for https://github.com/ParentingForLifelongHealth/plh-facilitator-app-content/issues/92

## Screenshots/Videos

[comp_text_box](https://docs.google.com/spreadsheets/d/10RX7I92ZaLMRDGHuNVdl7d6inTfS7aRP1jVXUs7oqwg/edit?gid=569531329#gid=569531329):

Screenshot 2026-03-17 at 15 22 39

https://github.com/user-attachments/assets/2a7c5e78-a4b5-41ae-af99-e1fd2b9dd133


chrismclarke (changes requested):
I'm not a huge fan of the pattern we're introducing here, feels like the type of decision-making we shouldn't be pushing onto authors (it's a subtle implementation detail, not really an authoring issue).

What exactly is the underlying issue here? That the `ionBlur` event doesn't fire in time to invalidate button that depends on it? I feel like the real thing missing here is a proper `form` component that can wrap all children and ensure validation always live, or possibly just using the reactive templating system instead which I understand should be setup much better for this.

Alternatively, I could possibly suggest a short-term/hacky workaround where the default behaviour is to trigger the corresponding `setValue` if (and only if) we know the input does not have any actions attached.

If determined to proceed with the current author-led pattern let me know and I can try to review the code as-is, although I honestly think the edge-case mentioned doesn't really justify the changes to the text-box component required to workaround (would also need to repeat pattern across all other input components with similar potential issues)

jfmcquade:
Closing in favour of #3391

PR #3391, opened 2026-03-18
feat: text box set value behaviour

chrismclarke:
PR Checklist

## Description

Alt to #3387
Change how text_box triggers changes as a user types and when box blurs

## Author Notes
`text_box` inputs essentially have 2 different side-effect triggers on input:
1. A `set_self` action which informs other UI components of its value, allowing sibling components to update accordingly
2. A `changed` action which triggers any side-effects from the row `action_list`, e.g. `changed | set_field : some_field : some_value`

Previously both of the actions would trigger only when a user taps out of the input box, meaning that UI wouldn't appear to keep in sync very well with the user as they type, and could also allow for edge-cases like clearing an input box but still being able to tap a button that depends on it having a value (https://github.com/ParentingForLifelongHealth/plh-facilitator-app-content/issues/92)

The new behaviour now treats these two side-effects slightly differently. The `set_self` action is automatically triggered as the user types (with a minimal 500ms delay to avoid too frequent ui updates - could be exposed as authoring option in future if required), and any `changed` action_list is still triggered on blur. There is also improved logic to ensure the `set_self` is always processed before any other actions (skips the delay timer if tapping anything outside the input before the delay has completed)

This is now the default handling for all `text_box` components (so no authoring changes required), and the pattern could be extended to other components as required.

## Review Notes
See example in [debug_text_box_value](https://docs.google.com/spreadsheets/d/14q7yu_WhVjew6xGzlqP589Dc9RAswrdv33ZTLTw9Tlc/edit?gid=2063929093#gid=2063929093)
https://github.com/user-attachments/assets/39bb612e-366a-48e9-96b8-b1103a0a9bf5

## Dev Notes
I've tried to add better separation between the template base row component setValue and action side-effects by refactoring the methods and providing an optional `triggerActions` arg (default true). In doing so I'm hoping it should allow for slightly more fine-grained updates from the dev side, like in this component where the parent bound value is kept in sync.

I'm hopeful that in doing so we can find a more sensible default pattern to avoid authors having to configure any sort of value-based action triggers for themselves (ideally things should just work).

## Git Issues

Closes #3387

## Screenshots/Videos

[debug_text_box_value](https://docs.google.com/spreadsheets/d/14q7yu_WhVjew6xGzlqP589Dc9RAswrdv33ZTLTw9Tlc/edit?gid=2063929093#gid=2063929093)

_Example_ - UI stays reasonably in sync with text box updates while typing (small delay), and immediately updates when trying to click outside of the text box
 
https://github.com/user-attachments/assets/39bb612e-366a-48e9-96b8-b1103a0a9bf5


jfmcquade (approved):
Thanks @chrismclarke, I think this is a better approach.

I've made a debug sheet here: [debug_text_box_value](https://github.com/IDEMSInternational/open-app-builder/pull/3391/changes/f8b1eed059082a50bdb97eace93e2e42226c87e7)

Image

Here's a demo of the "broken" state:

https://github.com/user-attachments/assets/e79f0e08-680d-467f-89eb-44b86dff496f

In order to resolve the issue, we need the `set_self` action to run on input too, since this triggers the `set_local` that handles the row's value. So I've added https://github.com/IDEMSInternational/open-app-builder/pull/3391/changes/f8b1eed059082a50bdb97eace93e2e42226c87e7 to only exclude the `changed` actions when calling `setValue()` with `triggerActions` as false. Let me know what you think.

Demo of the new behaviour where the value updates on keystroke:

https://github.com/user-attachments/assets/40cbd4a1-2291-439c-a281-97bffc2a81b2

@ChrisMarsh82 Could you test this new behaviour with the in-production onboarding flow? No authoring changes are needed since this PR changes the default behaviour of the text box. There could potentially be knock-ons by making this default behaviour – e.g. for "forms" with non-trivial validation on a text field, this will now be evaluated on keystroke (so the user might see a a validation error message as they're typing). I think the behaviour is still an improvement on the previous behaviour of having to manually unblur the element, but worth being aware of the change.

jfmcquade (approved):
Looks good, should be the best of both worlds

jfmcquade:
@MichelePancera Please could you test this? It should resolve some of our issues around form validation

MichelePancera (approved):

Issue #3400, opened 2026-03-24
[FEATURE] Combo-box: optional 'label'/'badge'/icon for each option


jfmcquade:
**What?**

When viewing options for a combo-box, the user should be able to see a badge or label on some options. 

<img width="500" height="214" alt="Image" src="https://github.com/user-attachments/assets/9bbe1040-e5df-41a0-ae77-0f2a788f5bda" />

<img width="500" height="440" alt="Image" src="https://github.com/user-attachments/assets/e7f697f4-f723-4094-9807-183ee995e15b" />

<img width="600" height="620" alt="Image" src="https://github.com/user-attachments/assets/9e8b0192-751b-4b49-a03e-0253280b5b23" />

**Why?**
See https://github.com/ParentingForLifelongHealth/plh-facilitator-app-my-content/issues/257

**How?**

PR #3411, opened 2026-03-30
Chore/combo box dropdown refactor

jfmcquade:
PR Checklist

## Description

For authors, there should be no functional changes, besides some minor styling changes to the combo-box `dropdown` component.

Refactors the combo-box component in the following ways:
- The dropdown variant code is moved to a dedicated sub-component
  - This include both the default version and the "search" version that is used in the case of there being a large number of options
- Refactors the dropdown variant to use a custom overlay rather than ionic's own `ion-select`
  - This uses an `ion-popover` element directly with a list of `ion-item`s within it

## Testing

Thoroughly test the `dropdown` variants of the combo-box component on [comp_combo_box](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=569531329#gid=569531329). This includes both the default `dropdown` variant and the `dropdown` variant with search filter. There should be no changes in behaviour vs master.

## Git Issues

A step towards #3400 

## Screenshots/Videos

Minor visual changes to combo-box dropdown:

| theme | before | after |
|--|--|--|
| default | Screenshot 2026-03-30 at 13 02 09 | Screenshot 2026-03-30 at 13 01 58 |
| plh_kids_teens_za | Screenshot 2026-03-30 at 13 30 26 | Screenshot 2026-03-30 at 13 30 13 |

### DIFF
--- a/src/app/shared/components/template/components/combo-box/combo-box-dropdown/combo-box-dropdown.component.ts
+++ b/src/app/shared/components/template/components/combo-box/combo-box-dropdown/combo-box-dropdown.component.ts
@@ -1,4 +1,4 @@
-import { Component, computed, EventEmitter, Output, signal, input } from "@angular/core";
+import { Component, computed, EventEmitter, Output, signal, input, output } from "@angular/core";
 import { ModalController } from "@ionic/angular";
 import { ComboBoxSearchComponent } from "../combo-box-search/combo-box-search.component";
 import { IAnswerOption } from "src/app/shared/utils";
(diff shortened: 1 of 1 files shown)


chrismclarke (approved):
I think the custom dropdown gives more flexibility which is good. The underlying combo-box component still feels quite messy, and now the combination of combo components adds quite a bit of code duplication that I'm pretty sure ai should be able to more cleanly refactor, but happy to approve for now if it unblocks things upstream.

I triggered a couple more gemini reviews which flagged issue passing some signals instead of values (fixed in [0fb5d96](https://github.com/IDEMSInternational/open-app-builder/pull/3411/commits/0fb5d96cb66db312fd32b80e2058cc359af3d869)), I also went to remove the `prioritisePlaceholder` logic from the dropdown component, but reverted as it does appear to be vaguely in use by the child search component (although still not really sure to what extent it actually is or isn't used)

I haven't addressed the most recent review (I find gemini typically will only flag 3-4 issues at a time these days for whatever reason, so can take subsequent calls), so you may want to take a quick look prior to merge.

PR #3404, opened 2026-03-26
Feat/combo box badge

jfmcquade:
PR Checklist

- [x] PR title descriptive (can be used in release notes)

## Description

Adds support for "badges" on options of a combo-box. This is achieved with a new `options_badge` author param to specify a the name of the badge field on the source options data. If provided, then any option which has a value for this named field will have it displayed as a badge next to the option's display text in the dropdown.

Note: the combo-box component has three distinct versions: the `modal` variant (default); the `dropdown` variant; and the `search` version, which is active when the `dropdown` variant is used and there are more than a certain number of options to select from. Due to some peculiarities of the underlying ionic components we use, implementing this feature is much more straightforward for the `modal` and `search` versions than it is for the `dropdown` version. Therefore, currently this PR implements an MVP workaround for the case of the `dropdown` variant. See screenshots below for examples.

## Dev notes

The limitation of the `dropdown` variant is due to the [ion-select-option](https://ionicframework.com/docs/api/select-option) component not exposing "slots" like other ionic components do. This seems to leave us with two choices for implementing this feature in this case:
1. Rebuild a custom "popover" select component to use as the `dropdown` variant
2. Use DOM manipulation to insert the required element within the existing component
  - I have tried this implementation in #3406

Ionic may implement this functionality in the future, indeed there is an [open issue](https://github.com/ionic-team/ionic-framework/issues/29890), but there is not a lot of activity around the issue.

## Git Issues

Closes #3400

## Screenshots/Videos

[comp_combo_box](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=569531329#gid=569531329)

Screenshot 2026-03-26 at 11 52 14

[comp_combo_box_items](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=818486196#gid=818486196):

Screenshot 2026-03-26 at 11 50 51

#### Modal version

Screenshot 2026-03-26 at 17 23 41

#### Search version

Screenshot 2026-03-26 at 17 23 48

#### Dropdown version 

Screenshot 2026-03-26 at 17 27 02

PR #3412, opened 2026-03-30
Feat: combo box badge (alt)

### LINKED ISSUE #3400: [FEATURE] Combo-box: optional 'label'/'badge'/icon for each option

jfmcquade:
PR Checklist

- [x] Builds on #3411 (merge that PR first)

## Description

Adds support for "badges" on options of a combo-box. This is achieved through new params on the combo-box:
- `options_meta_badge_text`
  - The name of a property key on each answer option (e.g. a column of the corresponding data list). If a given option has a value for this key, a badge will display for that option using the value as text to display
- `options_meta_badge_color`
  - The name of a property key on each answer option (e.g. a column of the corresponding data list), to change the colour of the badge for that option. Default is "primary". Currently limited to the [colour options exposed by ionic](https://ionicframework.com/docs/api/chip#color), but should be expanded in a follow-up to support our theme colours generally.

This params suggest an extendible syntax for future use. For example, if we wanted to support an icon instead of (or as well as) a badge, we could add a new `options_meta_icon` param.

## Git Issues

Closes #3400
Closes #3406
Closes #3404

## Screenshots/Videos

Screenshot 2026-03-30 at 16 25 14

| modal (default) variant | dropdown variant | dropdown variant with search |
|--|--|--|
| Screenshot 2026-03-30 at 16 23 43 | Screenshot 2026-03-30 at 16 23 48 | Screenshot 2026-03-30 at 16 23 54 |

### DIFF
--- a/packages/components/plh/certificate/plh-certificate.actions.ts
+++ b/packages/components/plh/certificate/plh-certificate.actions.ts
@@ -1,0 +1,92 @@
+import type {
+  IActionHandler,
+  ITemplateActionServiceHandle,
+} from "src/app/shared/components/template/services/instance/template-action.registry";
+import { PlhCertificateService } from "./plh-certificate.service";
+import type {
+  IPlhCertificateGenerateParams,
+  IPlhCertificateGenerateResponse,
+  IPlhCertificateResponse,
+} from "./plh-certificate.types";
+import { isPlhCertificateSuccessResponse } from "./plh-certificate.types";
+
+export type { IPlhCertificateGenerateParams };
+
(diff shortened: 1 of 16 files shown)


chrismclarke (changes requested):
Just putting a block on this until after #3411 merged in. Please re-request when updated for easier review

PR #3438, opened 2026-04-10
spike: youtube player native

chrismclarke:
PR Checklist

## TODO
- [ ] Determine if wanting to use on Android/Web (or just IOS)
- [ ] Improve UI for native play (add thumbnail and/or better play button or placeholder container)
- [ ] Explore inline proxy option (if required)

## Dev Notes
- Currently enabled for all native platforms for testing, may want to consider only ios if prefer to keep existing android functionality
- Does not include changes from #3431 - to be confirmed whether still desirable to have the web player updates or not. Could also consider using this plugin on web as it does appear to have a similar iframe approach on web - [sourcecode - web.ts](https://github.com/Cap-go/capacitor-youtube-player/blob/main/src/web.ts)
- Current player accepts url, which could generically be used for any video player platform. This implementation requires extraction of youtube id, so if using youtube component for other platforms some functionality may be lost (would recommend migrating to more generic video-player component or similar).

## Review Notes
- Will need to build/sync native code and test with example template
- Requires testing on IOS
- Check whether closes relevant deployment issues (and link here)

## Git Issues

Hack/workaround for #3432

## Screenshots/Videos


jfmcquade:
Thanks @chrismclarke, appreciate you looking into potential solutions.

This does provide the expected user experience on iOS (see demo below). However I'm not sure this approach is better than a more heavy-handed workaround of just including a button that links to the youtube video in the iOS case. This could either by done within the component via a platform check (and passing in a param like "open_external_button_text" to ensure that the text displayed on the button is exposed to the authoring and translation system); or just through an author-level platform check and custom button.

The potential advantage of the current approach is that the user stays within our app. But personally this would be my least preferred option as a user clicking on a youtube link, after opening in the youtube app and opening in the external browser. The fact that the full youtube UI is accessed through the in-app browser means it doesn't feel like you're still within our app, and so I think the UX advantage is minimal. Also, I'm aware that we may have to acknowledge additional permissions when submitting the app to Apple, as they ask about whether we allow "unrestricted web access", which arguably we would via the inline browser (links within youtube will open external sites).

https://github.com/user-attachments/assets/653040e5-4da0-475d-b48a-16233a145c4b

chrismclarke:
> Thanks @chrismclarke, appreciate you looking into potential solutions.
> 
> This does provide the expected user experience on iOS (see demo below). However I'm not sure this approach is better than a more heavy-handed workaround of just including a button that links to the youtube video in the iOS case. This could either by done within the component via a platform check (and passing in a param like "open_external_button_text" to ensure that the text displayed on the button is exposed to the authoring and translation system); or just through an author-level platform check and custom button.
> 
> The potential advantage of the current approach is that the user stays within our app. But personally this would be my least preferred option as a user clicking on a youtube link, after opening in the youtube app and opening in the external browser. The fact that the full youtube UI is accessed through the in-app browser means it doesn't feel like you're still within our app, and so I think the UX advantage is minimal. Also, I'm aware that we may have to acknowledge additional permissions when submitting the app to Apple, as they ask about whether we allow "unrestricted web access", which arguably we would via the inline browser (links within youtube will open external sites).

Yeah I'm inclined to agree, I'll move ahead with proxy-based iframe in #3439 and close here

PR #3443, opened 2026-04-13
feat: initialise remote-function service on app launch

jfmcquade:
PR Checklist

## Description

Configures the remote-function service to initialise eagerly on app launch. This addresses a performance issue where the service would only init when a remote function was actually invoked for the first time via a template action – this would cause a noticeable delay in the action functionality.

## Dev notes

It's a bit untidy that, when working on deployments with remote functions enabled, devs and authors will now see these errors in the console on every app load, unless they configure App Check in firebase using their debug token. Any quick ideas on how to tidy this up @chrismclarke?

Screenshot 2026-04-13 at 15 28 36

## Git Issues

Should address https://github.com/ParentingForLifelongHealth/plh-kids-teens-app-my-content/issues/162

## Screenshots/Videos


ChrisMarsh82 (approved):
This looks like it works from a functional point of view

PR #3463, opened 2026-04-16
feat!: disabled state handling

chrismclarke:
PR Checklist

## Breaking Changes
Removes the `row.disabled` property. I couldn't find any examples in the code where this was being managed/set, so assume this was an authored property passed directly from sheet columns. I'm not sure how common an authoring pattern this was, in the debug sheets the only non-parameter disabled property I could find was in `example_workshops_accordion`.

Maintaining both row-level disabled state and parameter-level disabled state is quite confusing, so this PR proposes to move to parameter-level pr state which will be made available to all components by default.

The parser will throw an error if it detects authored content that explicitly uses a `disabled` column (see screenshots)

## TODO
- [x] Update parser to either migrate or throw error if detected
- [ ] improve example templates for specific components to show parameter-based disable state (where it is used)
- [ ] provide example where you can see the action list automatically disabling during a long-running action (either as part of button demo, or a specific actions demo)
- [ ] either update reactive components or create follow-up issue
- [ ] functional test components that are impacted by change

## Description
1. Adds a top-level `disabled` signal to all template components, that can be programatically triggered by the component

2. Add a default handler to set the component disabled state to `true` when triggering actions and `false` after trigger completes/fails

3. Refactor parameter list handling to allow all components to set a `disabled` parameter, and reflect to top-level disabled state. Remove duplicate parameters within components.

## Dev Notes
This will have a knock-on for reactive templates, which inherited much of the parameter-based disabled setting.

## Git Issues

Closes #

## Screenshots/Videos
**Example Sync Errors** - processing fails if content sets `disabled` column in a sheet

image

```ts
{
  source: 'converter',
  message: 'Template parse error',
  details: {
    error: "Cannot create property 'message' on string '[\n" +
      '  "Do not set \\"disabled\\" property on row. Use parameter_list disabled property instead"\n' +
      "]'",
    flow: {
      flow_type: 'template',
      flow_name: 'example_workshops_accordion',
      status: 'released',
      flow_subtype: 'debug',
      rows: [
        {
          type: 'begin_workshops_accordion',
          name: 'workshops_accordion'
        },
        { type: 'end_workshops_accordion' }
      ],
      _source: {
        name: 'Debug Sheets',
        path: 'example_sheets/to_be_sorted/example_accordion.xlsx',
        url: 'https://docs.google.com/spreadsheets/d/1gfnWmr59DF0ZsuSuBh0K1o9fMqEg-z6Qa7eJNsiB-rQ'
      }
    }
  },
  level: 'error'
}

```


jfmcquade (commented):
Thanks @chrismclarke, I hadn't foreseen these issues and may have overlooked them if I'd tried to implement this, so I appreciate you looking into it. 

I'm happy for you to continue working on this. In terms of the to-dos, I think it would make sense for me to handle the update to reactive templates, as I'm planning on porting over components and component functionality in the coming months.

I'm not aware of how extensively the legacy `row.disabled` was used, but I think we should definitely be able to move away from it. My instinct is that the parser throwing the error may make more sense than migration, just in terms of adding complication to the code (although long-term being able to define parser migrations independently of the parser code would be a valuable feature, but probably not a priority)

chrismclarke:
@jfmcquade - I've updated to always report if sheets are using the `disabled` column property, however found the reporting inconsistent so have opened #3470 to resolve.

Once that is tidied up I would be keen if you are able to take over this PR to work on examples and confirming things behave as expected as my time will be somewhat limited over the next couple weeks and would be good to avoid this going to stale.

PR #3464, opened 2026-04-17
[CHORE] Expose new title component variable

### LINKED ISSUE #3460: Expose new title component variable
**Details**
Update text component to expose `--title-text-color` to define title's text colour. Update all themes to add the override values where possible

FaithDaka:
PR Checklist

## Description
- This PR is a follow-up to #3462. A new `title-text-color` variable has been created to ease theme overrides. The default colour remains `ion-color-primary` but can now be defined in different theme designs without overrides.

## Git Issues

Closes #3460
Closes #3462 

## Screenshots/Videos

| Master Branch | This Branch |
|----------------|-------------|
|  |  |
|  | |
|  | |
|  |  |

### DIFF
--- a/src/app/shared/components/template/components/title/title.component.scss
+++ b/src/app/shared/components/template/components/title/title.component.scss
@@ -3,10 +3,18 @@
 .title-wrapper {
   display: flex;
   align-items: center;
-  h1 {
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6 {
     margin: 0;
+  }
+  h1 {
     line-height: var(--line-height-text-largest);
     font-weight: var(--font-weight-bold);
+    color: var(--title-text-color, var(--ion-color-primary));
     @media (max-width: 375px) {
       font-size: var(--font-size-title-tiny);
     }
--- a/src/app/shared/components/template/components/title/title.component.scss
+++ b/src/app/shared/components/template/components/title/title.component.scss
@@ -53,4 +53,19 @@
       text-align: center !important;
     }
   }
+
+  &[data-variant~="section_banner"] {
+    padding: var(--regular-padding);
+    background-color: color-mix(in srgb, var(--ion-color-primary-50), white 2%);
+    border-bottom: 1.5px solid var(--ion-color-primary-200);
+    h1 {
+      font-size: var(--font-size-title-tiny);
+      font-weight: var(--font-weight-bold);
+    }
+  }
+
+  &[data-variant~="page_banner"] h1 {
+    font-size: var(--font-size-title-medium);
+    margin: var(--regular-margin) 0;
+  }
 }
--- a/src/app/shared/components/template/components/title/title.component.scss
+++ b/src/app/shared/components/template/components/title/title.component.scss
@@ -55,7 +55,7 @@
   }
 
   &[data-variant~="section_banner"] {
-    padding: var(--regular-padding);
+    padding: 18px;
     background-color: color-mix(in srgb, var(--ion-color-primary-50), white 2%);
     border-bottom: 1.5px solid var(--ion-color-primary-200);
     h1 {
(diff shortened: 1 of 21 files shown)


src/app/shared/components/template/components/title/title.component.scss
```diff
@@ -3,10 +3,18 @@
 .title-wrapper {
   display: flex;
   align-items: center;
-  h1 {
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6 {
```
jfmcquade, line 11: comment: I think at some point we'll want to update the title component to expose these options to authors, e.g. via a param like `size: 1;`, which would implement the corresponding `h*` element in the component. This has been requested in some form by PLH, see my comment [here](https://github.com/ParentingForLifelongHealth/plh-facilitator-app-content/issues/105#issuecomment-4074701651). Something for us to think about at some point.

src/app/shared/components/template/components/title/title.component.scss
```diff
@@ -3,10 +3,18 @@
...
     margin: 0;
+  }
+  h1 {
     line-height: var(--line-height-text-largest);
     font-weight: var(--font-weight-bold);
+    color: var(--title-text-color, var(--ion-color-primary));
```
jfmcquade, line 17: nit(blocking): To keep to our conventions, we should wrap this in a SCSS variable that we declare at the top of the file, to make clear what vars are exposed to themes for every component, i.e.
```scss
$text-color: var(--title-text-color, var(--ion-color-primary));
```
at the top of the file, then here use:
```scss
color: $text-color;
```

src/theme/themes/plh_facilitator_cw/_index.scss
```diff
@@ -99,6 +99,7 @@
       plh-module-list-item-background-1: hsl(334, 43%, 70%),
       plh-module-list-item-background-2: hsl(29, 77%, 70%),
       plh-module-list-item-background-3: hsl(123, 34%, 70%),
+      title-text-color: var(--ion-color-primary-shade),
```
jfmcquade, line 102: question(blocking): Would it make more sense to use a specific colour here, like `var(--ion-color-primary-900)`? Same applies to all themes. That was my initial thought on seeing this, as there's no way of knowing how dark `-shade` is. But I suppose as this is being set within a theme, we do know that this is the colour we want.

I think in fact I'd suggest that we leave this as it is, but update the default/fallback value of `--title-text-color`, defined in the title component SCSS, to be `var(--ion-color-primary-900)` or `var(--ion-color-primary-900)` (instead of `var(--ion-color-primary)` as it is currently). Ironically, the `default` theme is used by the fewest deployments, and none is sensitive to changes, so it's OK if we change the default values in a way that will change the default theme when we're doing clean-ups like this. And in this case, I think it makes sense to set a dark text colour as the default, apps don't tend to use the primary brand colour for text.

jfmcquade (changes requested):
Thanks @FaithDaka, generally looks good.

I'm going to suggest we include #3462 in this PR (i.e. mark the current PR as closing #3462, as it already builds on the changes), just because we need the title colour variable in order to preserve the title colour across themes, so I don't want to merge #3462 independently.

I've also added a couple of minor requests inline

FaithDaka:
@jfmcquade This PR will close #3462. It was necessary to separate them to avoid bulk changes. 
I've decided to keep `plh-kids-kw` theme with cleaned up title overrides as its styling changes might be more sensitive. 

The typography styling for `plh-facilitator-cw` will have to be noted for later issues.

Also, the heading tags (h1 ... h6) definitely need cleaning up so that there is variety of title sizes as html intends [ `h1` is the main tag and sized as `--font-size-title-tiny` yet it's meant to be the largest heading tag]. 
It will require knowledge on the full scope of the affected components that use titles in their inner html. I see `<p>` tags wrapped by `<h1>`s

jfmcquade (approved):
Thanks for the updates, I'm happy to merge now

PR #3468, opened 2026-04-22
fix: crashlytics init config handling

jfmcquade:
PR Checklist

## Description

Fixes an issue where the Crashlytics service would throw an error on init on native devices in the case where the deployment config does not specify a value for `firebase.crashlytics.enabled`. This is set to a `default` of `true`, i.e. Crashlytics will be enabled unless explicitly disabled.

## Git Issues

See, for example https://app.glitchtip.com/idems/issues/3220862?project=4993 (this isn't definitely the cause of this error, but it seems highly likely).

## Screenshots/Videos


chrismclarke (approved):
sounds like a reasonable solution to check, can see how setting `enabled: undefined` could be invalid

PR #3474, opened 2026-04-29
Refactor `lesson-cta` styling for accessibility

### LINKED ISSUE #3475: Refactor `lesson-cta` styling for accessibility

FaithDaka:
PR Checklist

## Description

- The Course CTA reflows under extreme scaling such that the inner elements remain seen and accessible on devices with large displays as shown here: 

   - Now, the button width doesn't break the layout, content wraps instead of remaining in a row, text adapts to the changing width.

## Git Issues

Closes #3475 

## Screenshots/Videos
The app was tested in an emulator under the following constraints
| Constraints | Screen |
|-------------|--------|
| Large Display Size, Large Font Size |  |
| Medium Display Size, Large Font Size |  |
| Small Display Size, Medium Font Size |  |


jfmcquade (dismissed):
Thanks @FaithDaka, appreciate you getting the emulator workflow set up to work on this, good to know we do now have a way of testing for these accessibility settings. 

See gemini review, feel free to resolve/dismiss those comments as appropriate.

Generally the changes look good, I haven't used the `@container` selector before, but this seems like a perfect use case. My only concern is browser compatibility – it [looks like](https://caniuse.com/css-container-queries) the feature should be supported on Chrome versions 105+.

We try to support down to WebView version 89 (versions roughly match between WebView and Chrome), see: https://github.com/IDEMSInternational/open-app-builder/blob/3427d1d1f0bca61ef841fe750db326db3b52f08d/android/app/src/main/java/utils/WebViewChecker.java#L18

This means that there would be some users on older devices who wouldn't get these style changes when their text is scaled in this way. I think we probably want to still get these changes merged and move forward. From my research, it doesn't seem like there's an alternative CSS query we can use for capturing native text scaling in order to apply layout changes like this. And we don't want to compromise the default version of the designs that the majority of users will see. So the only reliable option is to try and handle these cases within the built-in responsiveness of our styling, which is going to be quite difficult. I think we should see this as a process, where we should be able to get better at styling for these edge cases, but I think changes like the ones in this PR are still a useful tool even without full compatibility.

One additional request: I think it would be good to add some top margin to the component as a whole in order to prevent it overlapping with components above like this:

Image

I'll tag @ChrisMarsh82 for functional review, and @chrismclarke in case you have any thoughts about styling for these cases more generally

FaithDaka:
The changes suggested by in the gemini review have been addressed. 
Also, the margin has been added to the top of the `CTA` container however, this change will have to be noted by content authors who have added margins in the page sheets so their designs don't look off.



I share the same sentiments about support for devices running older versions of software. Accessibility styling will be done in phases according to the different edge cases as long as we create a good template to follow.

ChrisMarsh82:
@FaithDaka The text in the button is now slightly higher
current:
image
proposed:
image

I assume this is due to the text needing to be higher if the continue button is underneath. Is it possible to set it back to slightly lower text if the continue button is on the side?

FaithDaka:
Thanks for noting that @ChrisMarsh82. It's been addressed.

Before:
 

After:

ChrisMarsh82 (approved):

PR #3488, opened 2026-05-11
fix!: preserve snake_case keys in nested parameter values

jfmcquade:
PR Checklist

## Description

For components that use the modern zod-based params parsing, the logic for parsing parameter lists was recursively converting all object keys from snake_case to camelCase, including keys inside nested values such as `answer_list` option fields (an answer list may be defined in a data list, for example, and use custom column names for the fields that contain each options identifier and display text). This required a HACK in the combo-box component logic that had to look up each key under both its original and camelCase form.

These changes are marked as breaking since in theory it may have previously been possible for authors to access nested snake_case keys on objects they pass to parameters using camelCase – in the context of components that take an answer_list object param. However, there should be no such cases of this.

## Dev Notes
- Changed `snakeToCamelKeys` to a shallow `snakeToCamelTopLevelKeys` that only converts the top-level param keys (e.g. `button_list` → `buttonList`), leaving nested object/array contents untouched. Updated the `ISnakeToCamelKeys` type to match.
- Added a test asserting that nested keys are preserved

Knock-on changes:

- Removed the `rawOptionFieldForMetaBadgeKey` HACK from `combo-box-meta-badge.config.ts`
- Updated navigation-bar to access `button.target_template` instead of `button.targetTemplate` (nested keys are no longer rewritten)
  - This is the only case in the code of directly referring to these nested param keys

## Git Issues

Closes #

## Screenshots/Videos

[comp_combo_box](https://docs.google.com/spreadsheets/d/1uIkaMlDjoDN7uTpHkSeEQ6Yp-4ehX9IrBQMrolpfjQc/edit?gid=569531329#gid=569531329) updated with example that uses a snake_case key identifier, which should be unaffected by the changes

Screenshot 2026-06-01 at 15 43 53

### DIFF
--- a/src/app/shared/components/template/components/combo-box/combo-box-meta-badge.config.ts
+++ b/src/app/shared/components/template/components/combo-box/combo-box-meta-badge.config.ts
@@ -1,5 +1,4 @@
 import { IAnswerOption } from "src/app/shared/utils";
-import { snakeToCamel } from "../../utils";
 
 /** Fallback values when a field is missing on an answer option (not default key names). */
 export const OPTION_META_BADGE_VALUE_DEFAULTS = {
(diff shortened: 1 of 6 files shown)


ChrisMarsh82 (approved):
checked in parent app and facilitator app and appears to be ok

PR #3518, opened 2026-06-08
Feat: asset packs data list

### LINKED ISSUE #3519: core_assets dynamic data list
**Details**
I noticed that after downloading remote assets, a dynamic data list `core_assets` appears on the `/user` page. 

Is that intended? What does it represent (if it truly represents only assets that are part of the core app download I don't understand why it only exists after downloading an asset pack)? Should it be called `_core_assets` (name starting with underscore)?

**Screenshots and videos**

Image

Image

jfmcquade:
PR Checklist

## Description

Adds a protected data list, `_asset_packs`, to track metadata of downloaded asset packs and expose it to authoring (see #2993 for the first instance of a protected data list, `_notifications`).

Currently this tracks:
- `id`
- `name`
  - Currently same as ID, but could be an author assigned friendly name? Or could be removed?
- `download_status`: `"in_progress" | "completed" | "error"`

Additional fields could be exposed, e.g. `assets_total_count`, `assets_downloaded_count`

## Git Issues

Closes #3519

## Screenshots/Videos

Updated [debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit?gid=1422144003#gid=1422144003):

Screenshot 2026-06-08 at 10 46 29

Screenshot 2026-06-08 at 10 38 36


esmeetewinkel:
I was wondering if "status" of an asset pack sounds too general when we're only talking about status of the download. 
Alternatively, we could call it e.g. `download_status` with states `in_progress`, `completed`, `error/failed`

esmeetewinkel (changes requested):
It looks like the action 
```
click | asset_pack: reset;
```
doesn't reset the dynamic data list. See PR preview link. 

Image

jfmcquade:
Thanks @esmeetewinkel. I've acted on your comments (including #3519) and updated the PR description

esmeetewinkel (approved):
Thanks @jfmcquade for updating, this works well. I've updated the debug template to reflect the revised syntax accordingly. 

> - `name`
>    - Currently same as ID, but could be an author assigned friendly name? Or could be removed?

I can see how this might be useful in the future. I think my preference would be to remove as long as we don't have a purpose for it (to avoid clutter and since documentation is still lagging behind), but not feeling strongly.

PR #3525, opened 2026-06-11
feat: Improve asset pack download state handling; new Network service

### LINKED ISSUE #3521: In progress vs error status for asset pack download
**Details**
Further testing to PR #3518 (post merge). I was trying to sample the `error` download status. I thought I would reach this status by running the app locally without internet connection and attempting to download an asset pack. However, doing this I just end up in `in_progress` status. Is that the expected behaviour? Should this eventually time out and become `error`? I could imagine poor or lack of connectivity would be the main download issue end users will experience. 

**Screenshots and videos**

Image

jfmcquade:
PR Checklist

## Description

Added more complete asset pack download lifecycle handling.

- Added a shared `NetworkService` using `@capacitor/network`, exposing connection status as Angular signals and supporting abortable waits for reconnection.
- Synced the Capacitor Network plugin into both Android and iOS native projects.
- Extended `_asset_packs` metadata with `waiting_for_connection` and `cancelled` statuses, plus `download_started_at`, `download_completed_at`, and `download_status_updated_at`.
- Updated asset pack downloads to enter `waiting_for_connection` when offline, resume when connectivity returns, and avoid arbitrary timeout-based failures.
- Added `asset_pack: cancel_download` template action to explicitly cancel an active download attempt, clean up network listeners, clear progress, and persist `cancelled`.
- Added tests covering success/error transitions, offline wait and resume, stale manifest handling, and cancellation while waiting for connection.

### Current limitations / follow-ups:

- Asset packs can remain `waiting_for_connection` indefinitely unless `asset_pack: cancel_download` is explicitly triggered.
- There is no automatic retry limit, timeout, or stale-download cleanup policy yet.
- Downloads are tracked in memory only while the app session is active; a restart will preserve `_asset_packs` status but will not automatically resume an active attempt.
- Only one active asset pack download attempt is supported; broader queueing/concurrent download management is still a future improvement.
- Download progress is still global on `RemoteAssetService`, so concurrent asset pack progress is not independently tracked.

## Git Issues

Closes #3521

## Screenshots/Videos


esmeetewinkel:
Qs:
- Would it be feasible to have specific error statuses for "failed because lack of phone storage" and "server issue"? Besides a connectivity issue, these are the two other specific errors that Deepali identified [here](https://github.com/ParentingForLifelongHealth/plh-kids-teens-app-pa-content/issues/177#issuecomment-4620744247).
- Are there easy ways to simulate the different statuses for testing?

jfmcquade:
Thanks for looking into this, @esmeetewinkel 

> * Are there easy ways to simulate the different statuses for testing?

I've added https://github.com/IDEMSInternational/open-app-builder/pull/3525/changes/7b0427f09b901147ca76a9f873ecda09ebada42e which exposes a variable at the top of the file that will apply a time delay, in ms, to each asset being downloaded within the asset pack. This should help us when testing the feature and the various states, since it gives us time to change things and check the statuses.

To test `waiting_for_connection`/`in_progress`, we can test by starting a download, then disabling the network connection, then eventually re-enabling it. This can be done from within the "Network" tab in Chrome devtools, see screenshot below, or just toggling off the wifi on your computer, for example, if that is easier.

Screenshot 2026-06-15 at 15 21 31

I've also updated [debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit?gid=1422144003#gid=1422144003) to include some features at the top, including a button for the new `asset_pack: cancel_download` action. Currently, this explicit action is the only way to cancel the download, but eventually this logic should be triggered by specific conditions, e.g. a certain amount of time elapsing, to be discussed.

 
> * Would it be feasible to have specific error statuses for "failed because lack of phone storage" and "server issue"? Besides a connectivity issue, these are the two other specific errors that Deepali identified [here](https://github.com/ParentingForLifelongHealth/plh-kids-teens-app-pa-content/issues/177#issuecomment-4620744247).

I think those should be possible, although I haven't looked into it yet. I think it would make sense to get the current PR merged in first and address these as follow-ups, as this PR adds quite a lot of structure that those features should hopefully be able to build on top of.

esmeetewinkel (approved):
Description looks good, I'll test this in detail on a deployment post merge.

PR #3548, opened 2026-06-23
feat: implement caching for generators and data pipes as part of sync

jfmcquade:
PR Checklist

## Description

Experimental PR – code generated by AI without much oversight.

Speeds up repeat syncs by caching `generator` and `data_pipe` flows together with their generated children, instead of always re-parsing deferred child flows when the parent and its inputs are unchanged.

Previously, `shouldUseCachedEntry` always returned `false` for these flow types, so every sync re-ran generators/pipes and re-queued all children (~97% of FlowParser time on large deployments).

## Dev notes

- Cache keys include a dependency checksum from referenced `data_list` rows, so cache hits invalidate when inputs change.
- On a parent cache hit, stored `_generated_children` are restored into the hashmap and outputs without re-deferring.
- After processing, `finalizeParentCaches()` attaches parsed child flows to the parent cache entry.
- Cache keys use stable payloads (normalized `parameter_list`, row fields, parsed `args_list`) so deferred retries and fresh inputs hash consistently.
- Cache entries are deep-cloned on write; dependency rows are cloned in the hashmap to avoid drift from `postProcess` mutations.
- `JsonFileCache` batches index writes via `flush()` at end of processing.

Cache version bumped to `20260622` — first sync after deploy does one full parse to populate `_generated_children`; subsequent syncs benefit.

## Test plan

- [x] `flowParserCache.spec.ts` — generator/data_pipe cache + restore, invalidation on data list change
- [x] Full `flowParser/` suite (89 tests)
- [ ] Run sync twice on a large deployment; confirm second run skips generator/pipe reprocessing and FlowParser time drops significantly

## Git Issues

Closes #

## Screenshots/Videos


esmeetewinkel:
@jfmcquade when using this on `plh_kids_teens_za` deployment I'm seeing a lot of deleted generated data lists

image

PR #3561, opened 2026-06-30
Created plh_course_lesson_accordion component

ChrisMarsh82:
## Description

This pull request introduces a new `plh-course-lesson-accordian` component, which provides an interactive, data-driven accordion UI for course lessons. The implementation includes the component class, template, styles, a basic test, and full integration into the shared PLH components library for reusability and mapping.

**New Component: Course Lesson Accordion**


ChrisMarsh82:
@jfmcquade compare [module_list_test](https://docs.google.com/spreadsheets/d/1SkxfFFUzSjkUaxGKvYxiec-mx0szbeR6TaqBwfbBJyQ/edit?gid=281794208#gid=281794208) to [module_list](https://docs.google.com/spreadsheets/d/1SkxfFFUzSjkUaxGKvYxiec-mx0szbeR6TaqBwfbBJyQ/edit?gid=450485313#gid=450485313)

jfmcquade (commented):
I've made a few changes to tidy up the code, based on copilot's review and my own opinions. 

I've also added some minor updates to [module_list_test](https://docs.google.com/spreadsheets/d/1SkxfFFUzSjkUaxGKvYxiec-mx0szbeR6TaqBwfbBJyQ/edit?gid=281794208#gid=281794208) to handle the `completed` state.

From my testing, it does seem to function exactly the same as the pre-existing component, I haven't been able to break it. And testing performance through the Chrome devtools "Largest Contentful Paint" property seems a bit random, but I do think I notice a general trend of the page with the new component loading faster on average.

I'd still be wary of moving over to this component. There are some quite painful bits of code, in particular the `toCourseSubItemRow` method returning a whole template-like row object, with hardcoded actions. I feel like potentially with some refactoring we could make it a bit nicer, but I'm inclined to think that the performance gains aren't worth sacrificing the authoring editabiliity that we currently have. Although if we can show that the performance gains are very significant, that decision might be different, at least for the immediate term of getting a release out that is more performant.

PR #3566, opened 2026-07-02
chore: expose 'year' type param on datetime picker

### LINKED ISSUE #3551: [FEATURE] Select only the year from the calendar
**What?**`
A ‘year’ type for the date_time_picker component, allowing users to select only the year from the calendar. 
This ensures that only the year is retained, which is necessary to avoid unnecessary complexity when selecting the year 

**Why?**
It enables farmers with limited education and understanding to select the year for their farming season.

**How?**
Here is an example that perfectly illustrates the desired feature [year_selection](https://docs.google.com/spreadsheets/d/1ZkkT4Kqe6qzxeo0ifM-K44M3xoDjImUATbYHmwAdvrs/edit?gid=1411621432#gid=1411621432&range=A9:E12)

**Other details**
Here is an illustration 
Image

jfmcquade:
PR Checklist

## Description

Adds `year` as an allowed `type` parameter on the date-time-picker component, enabling Ionic's year-only presentation mode via `parameter_list.type: year`.

## Git Issues

Closes #3551

## Screenshots/Videos

[comp_date_time_picker](https://docs.google.com/spreadsheets/d/1Ff1QbrAKWpENp6QNeHVAw6_7KtkXz6zngRTSKqHMssw/edit?gid=569531329#gid=569531329)

Screenshot 2026-07-02 at 14 57 52

### DIFF
--- a/src/app/shared/components/template/components/date-time-picker/date-time-picker.component.md
+++ b/src/app/shared/components/template/components/date-time-picker/date-time-picker.component.md
@@ -20,7 +20,8 @@
 | disabled	            | true                | Disables the button and greys it out |
 | type                  | date-time (default) | Allows date and time to be selected  |
 | type                  | date                | Allows date only to be selected      |  
-| type                  | time                | Allows time only to be selected      |  
+| type                  | time                | Allows time only to be selected      |
+| type                  | year                | Allows year only to be selected      |
 
 ## Events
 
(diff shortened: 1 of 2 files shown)


fagiothree (approved):

fagiothree:
@jfmcquade I think we can merge this. I have tested and it is working as expected, one minor point is that, with some themes, the selection is not very visible 
image

jfmcquade:
> @jfmcquade I think we can merge this. I have tested and it is working as expected, one minor point is that, with some themes, the selection is not very visible image

Thanks for flagging, I've added https://github.com/IDEMSInternational/open-app-builder/pull/3566/changes/b245e3e745f5571ac58bab572ada67dabda2e799 which a highlighted background to the selected option. This will affect all themes, but is a positive change and it's not as if we'd specifically designed the previous look


Screenshot 2026-09-18 at 15 53 26

Screenshot 2026-09-18 at 15 54 29

PR #3582, opened 2026-07-20
fix: notification sent_recorded_timestamp; notification metadata in DB

### LINKED ISSUE #3581: [BUG] App notification interaction table – notification shows as "sent" after it has been interacted with
**Describe the bug**
@esmeetewinkel Spotted that data in the DB showed cases where the `sent_recorded_timestamp` value would be _after_ the value of `action_recorded_timestamp` (e.g. for a "tap" action), e.g. by 30 minutes. This doesn't make sense logically as the user cannot have tapped on a notification before it was sent.

**Explanation**

For each notification interaction we store two timestamps:

- **Action time** — when the user taps (or otherwise interacts with) the notification  
- **Sent time** — when the app first records that the notification has become due / is processed in that app session  

Action time is recorded once and left alone. Sent time is refreshed repeatedly while the app stays open (on the regular notification sync), even when nothing new has happened for that notification.

**What this looks like in data**

A user can tap at 9:10, keep the app open, and later syncs move sent time forward (e.g. to 9:40). The row then shows a tap *before* “sent,” which is misleading: the tap time is correct; sent no longer means “first recorded as due,” only “last refreshed.”

**Expected**

Sent time should be write-once: set the first time we record it, and not updated on later refreshes.

jfmcquade:
PR Checklist

## Description

See #3581 for description of the bug.

- Stop overwriting `sent_recorded_timestamp` on session notification refreshes
- Keep the first recorded sent time so it no longer drifts forward while the app stays open (which could make action time appear earlier than sent time)

Also improve what we store in `notification_meta` for campaign notifications:

- Include authored `row_id` and `row_name` in notification `extra` when scheduling
- Persist `title` and `text` alongside existing meta (`campaign_id`, `action_list`, etc.)

The changes are entirely client-side so nothing is required to be updated on the server. They will not fix existing bad rows in the DB, and already-scheduled notifications keep the old meta until replaced; both only apply from future app versions / newly scheduled notifications.

## Testing

In theory, the testing procedure might be:

- [ ] Schedule a notification, open the app so it is recorded as sent, then tap it
- [ ] Keep the app open through at least one notification sync cycle
- [ ] Confirm `sent_recorded_timestamp` stays unchanged in the DB and remains at or before `action_recorded_timestamp`
- [ ] Confirm new interaction rows include `row_id`, `row_name`, `title`, and `text` in `notification_meta`

However the changes are minimal and this procedure could be time-consuming, so it may be sensible to merge with minimal testing.

## Git Issues

Closes #3581

### DIFF
--- a/src/app/feature/campaign/campaign.service.ts
+++ b/src/app/feature/campaign/campaign.service.ts
@@ -35,6 +35,10 @@ type IScheduledNotificationsHashmap = {
 
 interface ICampaignNotificationExtra {
   campaign_id: string;
+  /** Authored campaign row id */
+  row_id: string;
+  /** Authored campaign row name, when provided */
+  row_name?: string;
   action_list?: FlowTypes.TemplateRowAction[];
 }
 
@@ -278,7 +282,12 @@ export class CampaignService extends AsyncServiceBase {
     // HACK - remove markdown form title and text as not currently supported in capacitor notifications
     row.text = this.hackStripNotificationMarkdown(row.text);
     row.title = this.hackStripNotificationMarkdown(row.title);
-    const extra: ICampaignNotificationExtra = { campaign_id, action_list: row.action_list };
+    const extra: ICampaignNotificationExtra = {
+      campaign_id,
+      row_id: row.id,
+      row_name: row.name,
+      action_list: row.action_list,
+    };
 
     const notificationSchedule: ILocalNotification = {
       schedule: { at: _schedule_at },
--- a/src/app/shared/services/notification/local-notification-persist.adapter.ts
+++ b/src/app/shared/services/notification/local-notification-persist.adapter.ts
@@ -91,6 +91,13 @@ export class LocalNotificationPersistAdapter {
         notification_id,
       } as any;
     }
-    await this.db.put({ ...entry, ...update, _sync_status: "pending" });
+    await this.db.put({
+      ...entry,
+      ...update,
+      ...(entry.sent_recorded_timestamp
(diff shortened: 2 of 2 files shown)


esmeetewinkel (approved):
I've tried testing in the `debug` app, starting by trying to reproduce the issue in the current Google Play store version of the app. However, I'm seeing so many issues with campaign-notifications there that it is really difficult to identify whether the issue is present. So I think I'll stop here (meaning I haven't tested on this PR branch!). 

One example I got - where Sent Recorded is before Action Recorded - in screenshot below and [here](https://apps-server.idems.international/dashboard/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7InF1ZXJ5Ijp7InNvdXJjZS10YWJsZSI6MTUsImZpbHRlciI6WyJhbmQiLFsiPSIsWyJmaWVsZCIsMzgzOCxudWxsXSwiMzQwIl1dfSwidHlwZSI6InF1ZXJ5IiwiZGF0YWJhc2UiOjV9LCJkaXNwbGF5IjoidGFibGUiLCJ2aXN1YWxpemF0aW9uX3NldHRpbmdzIjp7fX0=). However, I am mainly confused why this appears under an old ID (with CreatedAt June 20, 2025, 6:23 PM) rather than now. Maybe that is because we're not keeping track of notification IDs properly and so my newly sent + tapped notification appears under a row in the database corresponding to something that happened a year ago? I am not sure this row even belonged to "my" user ID previously, or that my new interaction even replaced someone else's row...

Image
 
@jfmcquade please let me know in case you would like me to continue testing, otherwise I'm happy for you to merge.

PR #3584, opened 2026-07-21
chore: add system of retrying failing downloads as part of sync from Canto

jfmcquade:
PR Checklist

## Description

- Add retry with exponential backoff for Canto CloudFront asset downloads so transient network failures (e.g. `FetchError: … read ECONNRESET`) no longer abort `canto_assets_dl` on the first blip
- Still fail the sync after retries are exhausted; non-retryable HTTP errors like 403 fail immediately
- Cover retry success, non-retryable 403, and exhausted retries in unit tests

## Test plan
- [x] Run `yarn workflow deployment sync` on a deployment with Canto assets and confirm downloads complete (or only fail after retries with a clear file-named error)

## Git Issues

Closes #

## Screenshots/Videos

### DIFF
--- a/packages/scripts/src/tasks/providers/canto/download.spec.ts
+++ b/packages/scripts/src/tasks/providers/canto/download.spec.ts
@@ -1,0 +1,152 @@
+jest.mock("../../../commands/workflow/run", () => ({
+  WorkflowRunner: { config: {} },
+}));
+
+jest.mock("node-fetch");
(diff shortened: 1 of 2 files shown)


esmeetewinkel (approved):
Functional test passed, on this branch for me `yarn workflow sync_assets` goes through without error (and it doesn't on `master`)

Image

PR #3591, opened 2026-07-24
feat: resume interrupted remote asset pack downloads

jfmcquade:
PR Checklist

### Background

Remote asset packs download file-by-file in the WebView JS runtime, so if the app is killed or backgrounded mid-download the loop dies. Previously nothing resumed on next launch, and re-triggering a pack re-downloaded **every** file from scratch.

This is **Phase 1** of hardening asset-pack downloads: making an interrupted download resumable and self-restarting. Surviving a user force-quit is explicitly *not* a goal — a killed download is treated as "interrupted → resume on next open." (Phase 2, continuing downloads while the app is backgrounded via native transfers, is a separate follow-up — see below.)

### What changed

- **Skip already-downloaded files on resume** — before fetching each file (base asset and each theme/language override), the network download is skipped when there is *positive* evidence the local copy is good: the file exists on disk, its size matches the manifest, **and** a prior run integrated this exact slot (the `_assets_contents` entry recorded for it carries the manifest checksum **and** has had its `filePath` rewritten away from the manifest value). Anything unverified — including a file saved but never integrated (interrupted mid-write) — is re-downloaded.
- **Auto-resume on launch** — on startup, packs left `in_progress` / `waiting_for_connection` (process death skips cleanup) are automatically restarted without blocking app init. `error` packs wait for an explicit re-trigger; `cancelled` packs are left alone.
- **Correct completion accounting (bug fix)** — `downloadAndIntegrateAssetPack` now reports how many files failed, and a pack only reaches `completed` when every file is present. Previously per-file download errors were swallowed and the pack was still marked `completed` with missing files.
- **Concurrent same-pack requests join** — a download request for a pack that is already downloading now joins the in-flight attempt instead of returning failure (auto-resume can race a template-triggered `ensure_downloaded` for the same pack).
- **Queued packs survive a refusal** — when a background `ensure_downloaded` is refused because a *different* pack holds the single download slot, the remaining packs are retried once that download settles rather than being silently dropped. The awaited path still fails fast by design, so a pack parked in `waiting_for_connection` can't wedge the template action queue behind unrelated work.
- **`debug_download_delay_ms` action param** — `download` and `ensure_downloaded` accept an optional per-call pause before each asset file, e.g. `asset_pack | download: my_asset_pack | debug_download_delay_ms: 3000`, to open a reliable window for interrupting a download. This replaces the `REMOTE_ASSET_DOWNLOAD_DEBUG_DELAY_MS` module constant, which had to be edited in source and reset before merge — there is no longer a committed debug value to forget about.
- **New `FileManagerService.getSavedFileInfo`** — reports a previously saved file's existence/size/`src` using the same path rule as `saveFile`.
- **Feature README** — `src/app/shared/services/remote-asset/README.md` documents the feature end to end: the `_assets_contents` model, slots, download lifecycle, the resume gate, concurrency, the template API, and where packs come from at sync time.

### Notable behaviour / decisions

- The resume unit is the file "slot": a row can be partially done (base saved, overrides not), so the base and each override are checked independently.
- The `filePath` half of the resume gate is load-bearing. Integrating a base asset upserts the whole manifest entry, so the row also gains every *override's* checksum before those files exist — the checksum alone would wrongly vouch for overrides that were never downloaded. It is deliberately compared against the manifest's own value rather than against the expected local path: recorded paths come from `saveFile` while resume only has a `Filesystem.stat` uri, and a platform path-spelling difference (e.g. iOS `/private/var` vs `/var`) would otherwise silently disable resume and re-download every pack.
- On-disk content hashing (MD5) is intentionally **not** done here. `saveFile` resolves only after a complete write, so size + recorded-checksum give positive proof for integrated files at negligible cost. True byte verification (e.g. external corruption of an already-integrated file) is deferred to a future `asset_pack: verify`/repair action, which is where the MD5 dependency belongs.
- On resume, `_asset_packs.assets_downloaded_count` resets to 0 and climbs again (skipped files increment quickly) — this is expected; there is no separate pre-scan pass.

## Testing

I have created a debug sheet here for testing: [debug_remote_assets](https://docs.google.com/spreadsheets/d/1QPtxcCZEgUpXA26niqoXD0OQni2WAa3kgTPExonkCKU/edit?gid=1422144003#gid=1422144003)

**What to watch**: the **`_asset_packs`** table (`download_status`, `assets_downloaded_count` / `assets_total_count`, `download_started_at`, `download_completed_at`). You can also inspect the **`_assets_contents`** table on the `/user` page (each downloaded file's row gains a local `filePath` once integrated).

| # | Action | Expected in the tables |
|---|--------|------------------------|
| 1 | **Happy path** — trigger a download, let it finish | `_asset_packs`: `in_progress` → `completed`, `assets_downloaded_count` reaches `assets_total_count`, `download_completed_at` set. `_assets_contents`: the pack's file rows now have local `filePath`s |
| 2 | **Auto-resume after force-kill** — force-quit while `in_progress`, relaunch, do **not** re-trigger | `_asset_packs`: the row (left `in_progress`) gets a **new `download_started_at`** on its own and proceeds `in_progress` → `completed`, `downloaded == total` |
| 3 | **Prior work isn't lost** — during the resume, watch `assets_downloaded_count` | Resets to 0 then climbs (counter reset is expected); with no debug delay set, it jumps quickly past the already-downloaded count — finished files are re-integrated, not re-fetched |
| 4 | **Offline handling** — start a download, airplane mode on, then restore | `_asset_packs`: `waiting_for_connection` while offline → `in_progress` → `completed` after reconnect |
| 5 | **Cancel is not auto-resumed** — `asset_pack: cancel_download` mid-download, force-kill, relaunch | `_asset_packs`: `cancelled`, and after relaunch **stays `cancelled`** — no new `download_started_at`, count does not advance |

> The debug delay applies to *skipped* files too, so with it set a resume won't look *faster* — verify resume by status/count reaching completion (row 2), not by speed. Drop the param for a realistic timing check.

## Out of scope / follow-ups

- **Phase 2 — background downloads:** continue transfers while the app is backgrounded (native per-file transfers, e.g. `@capgo/capacitor-downloader`, gated on a spike confirming the iOS background-session setup). This PR's resume/reconcile is the groundwork it hands back to on relaunch.
- **On-disk content verification:** a future `asset_pack: verify`/repair action (would add an MD5 dependency).
- **Updated completed packs:** a pack already `completed` is not re-checked when its content changes server-side (no pack versioning yet); a changed file is only picked up if a pack is re-downloaded while incomplete.
- **Download queue:** downloads remain strictly serial — one pack at a time, one file at a time. Requesting a second pack is refused rather than queued (the retry above is a narrow fix, not a general queue).

## Git Issues

Relates to https://github.com/ParentingForLifelongHealth/plh-kids-teens-app-mx-content/issues/14

### DIFF
--- a/src/app/shared/services/file-manager/file-manager.service.ts
+++ b/src/app/shared/services/file-manager/file-manager.service.ts
@@ -11,6 +11,14 @@ import { ErrorHandlerService } from "../error-handler/error-handler.service";
 import { DeploymentService } from "../deployment/deployment.service";
 import { basenameFromExternalUrl, isExternalHttpUrl } from "shared/src/utils/string-utils";
 
+/** Result of statting a file previously written via `FileManagerService.saveFile` */
+export interface ISavedFileInfo {
+  exists: boolean;
+  sizeBytes?: number;
+  /** Webview-usable path to the file, matching the `src` returned by `saveFile` */
+  src?: string;
+}
+
 @Injectable({
   providedIn: "root",
 })
@@ -198,21 +206,22 @@ export class FileManagerService extends SyncServiceBase {
 
   /**
    * Native only: check whether a file previously written via `saveFile` exists on disk, and if so
-   * return its size in bytes. Uses the same path rule as `saveFile` so callers do not re-derive it.
+   * return its size in bytes and the same `src` that `saveFile` would have returned for it. Uses the
+   * same path rule as `saveFile` so callers do not re-derive it.
    * Returns `{ exists: false }` when the file is not present (stat throws for a missing path).
    */
   public async getSavedFileInfo(
     targetPath: string,
     options: { directory?: keyof typeof Directory; subdirectory?: string } = {}
-  ): Promise<{ exists: boolean; sizeBytes?: number }> {
+  ): Promise<ISavedFileInfo> {
     if (!Capacitor.isNativePlatform()) {
       throw new Error("getSavedFileInfo() is only supported on native platforms");
     }
     const { directory = "Data", subdirectory = "" } = options;
     const path = (subdirectory ? subdirectory + "/" : "") + `${this.cacheName}/${targetPath}`;
     try {
-      const { size } = await Filesystem.stat({ path, directory: Directory[directory] });
-      return { exists: true, sizeBytes: size };
+      const { size, uri } = await Filesystem.stat({ path, directory: Directory[directory] });
+      return { exists: true, sizeBytes: size, src: Capacitor.convertFileSrc(uri) };
     } catch {
       return { exists: false };
     }
--- a/src/app/shared/services/file-manager/file-manager.service.ts
+++ b/src/app/shared/services/file-manager/file-manager.service.ts
@@ -197,6 +197,28 @@ export class FileManagerService extends SyncServiceBase {
   }
 
   /**
+   * Native only: check whether a file previously written via `saveFile` exists on disk, and if so
+   * return its size in bytes. Uses the same path rule as `saveFile` so callers do not re-derive it.
+   * Returns `{ exists: false }` when the file is not present (stat throws for a missing path).
+   */
+  public async getSavedFileInfo(
+    targetPath: string,
+    options: { directory?: keyof typeof Directory; subdirectory?: string } = {}
+  ): Promise<{ exists: boolean; sizeBytes?: number }> {
+    if (!Capacitor.isNativePlatform()) {
+      throw new Error("getSavedFileInfo() is only supported on native platforms");
+    }
+    const { directory = "Data", subdirectory = "" } = options;
+    const path = (subdirectory ? subdirectory + "/" : "") + `${this.cacheName}/${targetPath}`;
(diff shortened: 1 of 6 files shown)


jfmcquade:
Thanks for testing, @ChrisMarsh82 

The multiple downloads was an authoring mistake of mine, I've updated the debug sheet and it now downloads both asset packs in sequence as expected (this PR doesn't actually change the logic for downloading multiple packs with a single action call):

Screenshot 2026-08-13 at 12 34 03

There was an issue with the `cancel_download` that was a bit more involved, fix in https://github.com/IDEMSInternational/open-app-builder/pull/3591/changes/9d9eaf22db3ac99a5eee469b9be5acf8a54fdb0c. As `cancel_download` is only used for debugging at the moment, I don't think we necessarily need to test again on native devices, so I would say we're probably good to merge this PR.

Screenshot 2026-08-13 at 14 32 05

ChrisMarsh82 (approved):

